PR #15, opened 2015-09-19
TST: fix py35 on Travis

jakevdp:
I think this will fix the Python 3.5 issue – found the fix here: https://github.com/pytest-dev/pytest/issues/744


jakevdp:
Whoo! All passed!

PR #36, opened 2015-09-20
Html rendering

freeman-lab:
This PR adds simple rendering of the specification's generated by the API via html. Example usage looks like this (here we're targeting display in an IPython notebook):

```
from altair import api, html
from IPython.display import HTML, display

data = dict(x=[1, 2, 3], y=[1, 2, 3])
spec = api.Viz(data).encode(x='x:Q', y='y:Q').mark_line()
out = html.render(s)
display(HTML(out))
```

I included a pretty bare bones test `html`. Also the size handling needs some testing. But otherwise should be functional.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
   - 3.5
 
 env:
-  - DEPS="pandas pytest traitlets pip"
+  - DEPS="pandas pytest traitlets pip jinja2"
 
 install:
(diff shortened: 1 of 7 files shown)


jakevdp:
One thing: Brian has an open PR #39 to change ``mark_points()`` to ``points()``. I think we're going to merge that, which will break these tests. Maybe wait for that to be merged, and then update this to make tests pass

freeman-lab:
Sounds good!

PR #59, opened 2015-09-22
Added dependencies to setup

alope107:
Added the dependencies for the install_requires portion of setup.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,24 @@ This approach enables users to peform exploratory visualizations with a much sim
 
 We realize that a declarative API will necessarily be limited compared to the full programatic APIs of Matplotlib, Bokeh, etc. That is a deliberate design choice we feel is needed to simplify the user experience of exploratory visualization.
 
+## Installation
+
+Altair requires [cycler](http://github.com/matplotlib/cycler), [matplotlib](http://matplotlib.org/), [numpy](http://www.numpy.org/), [pandas](http://pandas.pydata.org/) and [py.test](http://pytest.org/latest).
+
+Once the dependencies have been installed, Altair can be installed to the started Python path using:
+
+```
+python setup.py install
+```
+
+To install to another location use:
+
+```
+python setup.py install --prefix=/path/to/location
+```
+
+Then make sure your PYTHONPATH environment variable points to this location.
+
 ## Testing
 
 We are writing tests using [py.test](http://pytest.org/latest/). The run the Altair test suite, run:
--- a/README.md
+++ b/README.md
@@ -46,27 +46,32 @@ We realize that a declarative API will necessarily be limited compared to the fu
 
 ## Installation
 
-Altair requires [cycler](http://github.com/matplotlib/cycler), [matplotlib](http://matplotlib.org/), [numpy](http://www.numpy.org/), [pandas](http://pandas.pydata.org/) and [py.test](http://pytest.org/latest).
+Altair requires 
+* [cycler](http://github.com/matplotlib/cycler)
+* [matplotlib](http://matplotlib.org/) (1.5.0rc1 or better)
+* [numpy](http://www.numpy.org/)
+* [pandas](http://pandas.pydata.org/)
+* [py.test](http://pytest.org/latest)
 
-Once the dependencies have been installed, Altair can be installed to the started Python path using:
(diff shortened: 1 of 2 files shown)


jakevdp:
I'll chime in: I would avoid using ``install_requires`` for things like numpy and matplotlib. Installing matplotlib via pip is **never** the right thing to do. Before install, you need to do some system-dependent changes to the setup file to specify e.g. graphics backends. Pip doesn't know how to do this, so the installation will pretty much always fail.

On top of that, when working in virtual environments, pip in Python 2.x may not recognize your numpy install, and in trying to rectify that will destroy your Python stack. I've had it happen, and it's not pretty.

For those reasons, I avoid ``install_requires`` in all my packages, and I'd encourage everyone else to do so as well – especially if the requirements include matplotlib or numpy.

ellisonbg:
Do you think we should list any deps here?

jakevdp:
> Do you think we should list any deps here?

That's tough – I think it's confusing if we list some deps but not all. I tend to just avoid it altogether in favor of a detailed installation section in the README. I know lots of others disagree with me, though.

alope107:
Would it be reasonable to check in setup whether the dependencies are installed and display a warning for those that are not? The warning could direct users to the installation instructions in the README.

ellisonbg:
I am also wondering if we can isolate the dependencies in the code so that
the main API will run with only traitlets (and not pandas or mpl).

Then just requiring traitlets would be a good options (it is pure python
and trivial to install).

On Wed, Sep 23, 2015 at 12:24 PM, alope107 <notifications@github.com> wrote:

> Would it be reasonable to check in setup whether the dependencies are
> installed and display a warning for those that are not? The warning could
> direct users to the installation instructions in the README.
>
> —
> <https://github.com/ellisonbg/altair/pull/59#issuecomment-142703463>.
>



-- 
Brian E. Granger

jakevdp:
One difficulty of that is that the type inference depends on Pandas, and it would be difficult to make it work apart from that.

ellisonbg:
Hmm, yeah that is right in the middle of it all. So maybe let's leave it empty for now. Also, most people who will use this will already have all that installed - and we can add an Installation section to the README. @alope107 can you do that instead of this one?

tacaswell:
We could also provide a conda recipe / binaries (might be another good place to use conda-forge).

ellisonbg:
+1 on that - Thomas if you have done that before, do you want to have a go
at that?

On Thu, Sep 24, 2015 at 10:01 AM, Thomas A Caswell <notifications@github.com
> wrote:

> We could also provide a conda recipe / binaries (might be another good
> place to use conda-forge).
>
> —
> <https://github.com/ellisonbg/altair/pull/59#issuecomment-142989193>.
>



-- 
Brian E. Granger

alope107:
Added installation instructions.  Do we need to specify which version of each dependency to use, and if so, what are they?

tacaswell:
mpl needs to be 1.5.0rc1 or better.

README.md
```diff
@@ -44,6 +44,24 @@ This approach enables users to peform exploratory visualizations with a much sim
 
 We realize that a declarative API will necessarily be limited compared to the full programatic APIs of Matplotlib, Bokeh, etc. That is a deliberate design choice we feel is needed to simplify the user experience of exploratory visualization.
 
+## Installation
+
+Altair requires [cycler](http://github.com/matplotlib/cycler), [matplotlib](http://matplotlib.org/), [numpy](http://www.numpy.org/), [pandas](http://pandas.pydata.org/) and [py.test](http://pytest.org/latest).
```
ellisonbg: Let's like the dependencies in a bulleted list as it makes it easier for users to scan visually.

README.md
```diff
@@ -44,6 +44,24 @@ This approach enables users to peform exploratory visualizations with a much sim
...
+
+Once the dependencies have been installed, Altair can be installed to the started Python path using:
+
+```
+python setup.py install
+```
+
+To install to another location use:
```
ellisonbg: I think we can leave out the "install to another location" part, but let's include a version that uses `pip install` directly from the our GitHub repo.

alope107:
Sorry for the slow response. I've added the suggested changes.

PR #84, opened 2015-11-03
Create AltairWidget

aggFTW:
I'll add my own questions in line

ping @ellisonbg


altair/view.py
```diff
@@ -1,116 +1,239 @@
...
+        'col': Col,
+        'shape': Shape,
+        'color': Color,
+        'detail': Detail,
+        'text': Text,
+        'size': Size
+    }
+
+    def __init__(self, viz, shelves=_shelves, renderer='lightning'):
+        use_renderer(renderer)
```
aggFTW: Do you agree that not asking the user to remember to initialize the renderer could be useful or should I remove it?

altair/view.py
```diff
@@ -1,116 +1,239 @@
...
+        # Create output area
+        self.output = w.HBox()
+        self.output.children = []
+
+        # Create widget
+        self.controls = self._create_left_pane()
+
+        self.widget.children = [self.controls, self.output]
+
+    def render(self):
```
aggFTW: I didn't know whether to make AltairWidget inherit from HBox and remove the render method or not. This was the only way I found of having an auto-refreshable widget.

altair/view.py
```diff
@@ -1,116 +1,239 @@
...
+
+    def on_render_viz(self, *args):
+        self._clear_output()
+
+        to_display = w.Textarea(value=str(self.viz))
+
+        # TODO: how to display properly?
+        # to_display = w.Output()
+        # with to_display:
+        #     self.viz.render()
```
aggFTW: ping @mathisonian @freeman-lab 

If you were to comment out line 65 and uncomment 61-63, you should see the lightning rendered graph, but instead you only get blank space. I see no errors, though, so I don't know why there's no display of anything. I wonder if this is the same thing @ellisonbg saw.

altair/view.py
```diff
@@ -1,116 +1,239 @@
...
+        encoding = self._create_encoding_view(columns, types)
+        vbox.children = [encoding]
+        return vbox
+
+    def _create_viz_types_buttons(self):
+        hbox = w.HBox()
+
+        heading = w.HTML('Mark:', width='50px', height='32px')
+
+        # TODO: Could not express this method succinctly as on_press always ends up calling viz.text()
```
aggFTW: This really bothered me, not being able to do without all the manual on_render methods below and whatnot, but as the comment states, I couldn't get on_press to have different viz_mark_method. It always ends up calling viz.text()

A better python programmer could probably tell me why that is.

aggFTW:
As it stands, widget looks like this:

screen shot 2015-11-03 at 1 02 21 am

ellisonbg:
I think I favor the dropdown for mark types if we are going to expose that
many. If there were only a few, and we had good icons for each, then a
button selector like this would be fine.

On Tue, Nov 3, 2015 at 4:02 AM, Alejandro Guerrero Gonzalez <

> As it stands, widget looks like this:
>
> [image: screen shot 2015-11-03 at 1 02 21 am]
> <https://cloud.githubusercontent.com/assets/654204/10904244/9c432bde-81c6-11e5-80ec-62e835bc6c11.png>
>
> —
> <https://github.com/ellisonbg/altair/pull/84#issuecomment-153291886>.
>



-- 
Brian E. Granger

aggFTW:
Yes, I don't like the text either. We do plan to add icons for each mark type though. With 8 icons, we should be able to have the buttons not extend beyond the encoding section. Are you OK with that?

domoritz:
Interestingly this looks a lot like our first version of Polestar :-) Right now Polestar doesn't work with Vega-Lite 1.0 but will be updated soon.

ellisonbg:
Yes, we may end up throwing a simple viz UI on top of altair but my main
focus is on the programmatic API. Then again, maybe just getting the
polestar integration into the notebook really working well is a better
approach. I will likely be in Seattle the last week of April - would be
great to stop by and work some if you and others are around.

On Fri, Feb 26, 2016 at 4:02 PM, Dominik Moritz <notifications@github.com>
wrote:

> Interestingly this looks a lot like our first version of Polestar :-)
> Right now Polestar doesn't work with Vega-Lite 1.0 but will be updated
> soon.
>
> —
> <https://github.com/ellisonbg/altair/pull/84#issuecomment-189531669>.
>



-- 
Brian E. Granger

domoritz:
> I will likely be in Seattle the last week of April - would be
great to stop by and work some if you and others are around.

Definitely!

ellisonbg:
Closing for now...

kynan:
Is this still on the roadmap? It would be great to have `Chart` work with `ipywidgets.interact` (which it currently does not afaict).

The following example doesn't render the chart (running in a notebook):
```python
from altair import Chart, load_dataset
from ipywidgets import interact
cars = load_dataset('cars')
cols = tuple(c for c in cars.columns if c != 'Name')
@interact(x=cols, y=cols, color=cols)
def plot(x, y, color):
    Chart(cars).mark_point().encode(x=x, y=y, color=color)
```
Am I just missing a trick here?

I'm happy to create a separate issue for this if you prefer.

aggFTW:
Hi @kynan! Don't know if altair wants to include this as part of its repo, but I think you might be interested in this project: https://github.com/jupyter-incubator/sparkmagic/blob/master/autovizwidget/examples/autoviz.ipynb

It might be a good idea to decouple a UI from this.

What it does is it provides a widget interface that renders pandas dataframes based on plotly, but I think with some work we could easily extend it to use Altair as a renderer instead. Would that be interesting to you?

Another interesting route would be to see if there's a way to host the polestar UI in a notebook and integrate it with Altair: http://vega.github.io/polestar/
This might be a cleaner solution...

kynan:
@aggFTW thanks for the pointers, will take a look!

PR #105, opened 2016-05-21
ENH: add traitlet tab completion hack

jakevdp:
This is a hack that enables tab completion for traits of any ``HasTraits`` class. Operationally, this means that, for example, if you type
```
In [1]: from altair import *
In [2]: Layer(des<TAB>
```
it will be auto-completed to
```
In [2]: Layer(description=
```
because ``Layer`` has a ``description`` trait.

I wasn't able to figure out how to do this through the normal completer hooks, so I used a hack that works but will likely be very brittle. I don't think this solution should be merged, but I'd love some feedback on what a better solution might be.


jakevdp:
Closing in favor of #113, which is a saner solution :smile:

PR #131, opened 2016-06-14
ENH: add to_html() methods

jakevdp:
I started comparing our docs to those of Vega-Lite, and noticed that they emphasize how to publish visualizations online: https://vega.github.io/vega-lite/tutorials/getting_started.html#embed

This PR uses that example as a template, and creates a ``to_html()`` method for top-level objects which export charts to stand-alone HTML.


jakevdp:
Merging so I can finish docs

PR #134, opened 2016-06-17
[MRG] delay type inference & allow it to handle parent data

jakevdp:
Addresses the statefulness issue brought up in https://github.com/ellisonbg/altair/issues/90#issuecomment-226656429

### DIFF
--- a/altair/api.py
+++ b/altair/api.py
@@ -109,8 +109,9 @@ class TopLevelMixin(object):
 
     def _to_code(self, data=None):
         """Emit the CodeGen object used to export this chart to Python code."""
-        self._finalize()
-        return visitors.ToCode().visit(self, data)
+        # do not call _finalize(), as we want the output code
+        # to reflect the exact input
+        return visitors.ToCode().visit(self, data=data)
 
     def to_altair(self, data=None):
         """Emit the Python code as a string required to created this Chart."""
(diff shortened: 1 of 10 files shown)


jakevdp:
Basically, now instead of having observers that watch for changes in objects that need types inferred, I now instead have a ``_finalize()`` method in the baseobject that triggers type inference just before the JSON or Python code is written. This is constructed so that data from wrapper objects is referenced appropriately, and possibly overwritten by data in child objects (I did some experiments with the Vega-Lite renderer to make sure this matches how it handles data in multiple places).

jakevdp:
This can be streamlined a bit... let me do one more commit.

jakevdp:
@ellisonbg: one question came up as I was going through this: do you recall why we originally added the ability for ``type`` to be arbitrary unicode? https://github.com/ellisonbg/altair/blob/177d92ab523534f49cf3284f3c0cced5299fc4fd/altair/schema/_wrappers/channel_wrappers.py#L44

Looking at it now, I can't imagine any situation in which this would be helpful. Wouldn't we always either want ``type`` to be a valid value, or ``None``?

jakevdp:
Figured it out: it's to allow for abbreviations. I just added a commit to make that more explicit.

jakevdp:
Tests pass; merging after rebase.

PR #149, opened 2016-07-10
Explicitly assign chart in fist example. Don't use Out[1]

Carreau:
screen shot 2016-07-09 at 17 43 34


Carreau:
Well, too late, already done :-)

PR #160, opened 2016-07-16
missing a dependency for git in python

chendaniely:
was getting a `no module named git` error when I ran `make all`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,5 @@ traitlets
 ipython
 pandas
 vega>=0.4.1
+gitpython
+


jakevdp:
The makefile has nothing to do with package installation; it's used to parse vega-lite and auto-generate the source code for package developers. To install the package, follow the instructions in the README.

It might be that the existence of this make file is potentially confusing - we might want to remove it or move it to a subdirectory.

jakevdp:
I changed the Makefile in https://github.com/ellisonbg/altair/commit/3fe2183730db5ddcd32fa55a8cd1104a4d07097c

Thanks for the feedback!

chendaniely:
I'm slowly translating altair for R and needed to read the source code.

does this mean if I want to rebuild everything I'd have to run the old `all` command: `make sync-schema sync-datasets sync-examples generate-schema generate-examples test`?

jakevdp:
No, you'll not need those commands unless you're re-generating the source code from a different vega-lite tag.

For details on the developer scripts, see [NOTES_FOR_MAINTAINERS](https://github.com/ellisonbg/altair/blob/master/NOTES_FOR_MAINTAINERS.md)

PR #166, opened 2016-07-26
Forward kwargs from to_html

idmit:
`to_html` function handles kwargs, but `to_html` method doesn't forward any.


jakevdp:
Thanks!

PR #174, opened 2016-08-21
DOC: adds documentation for heat maps

stsievert:
I added a new notebook that documents heat maps. It shows example heat maps for

* the default heat map
* after applying colors to the background, not just the text
* applying a custom colormap and getting rid of the text. This is the heat map I wanted to generate.


jakevdp:
This is great – thanks!

PR #194, opened 2016-09-06
ENH: add ability to return dataset URL only

jakevdp:


jakevdp:
Another couple commits snuck in; I meant to push these directly to master so I'll just keep them as part of this PR.

jakevdp:
Tests passed; merging. Planning to use this in the new docs.

PR #201, opened 2016-09-12
MAINT: clean-up derived classes in API

jakevdp:
This simplifies the top-level classes (``Chart``, ``LayeredChart``, ``FacetedChart``) and makes sure their traits inherit documentation from base classes.


jakevdp:
I mainly did this because I noticed that the trait ``help`` strings weren't propagating through to the top-level object. After doing some changes by hand, I realized that inheritance should work just fine... the only change would be that ``class_own_traits()`` would no longer include these traits. I think that's OK.

PR #203, opened 2016-09-14
ENH: improve Chart.to_dict() to remove need for load_spec()

jakevdp:


jakevdp:
I believe this has mostly been used internally, so there shouldn't be any problem merging.

PR #214, opened 2016-09-26
MAINT: build docs on travis

jakevdp:
Attempt to build docs on Travis


jakevdp:
I rebased and squashed because there was a lot of trouble-shooting in there. I think having the doc build as part of the tests is a good idea (especially since it takes only ~a minute longer than the normal test suite). I'm going to merge

Issue #224, opened 2016-10-12
CalledProcessError usage is not python 2 compliant


gramhagen:
utils/node.py:83 is doing this:
`raise CalledProcessError(p.returncode, cmd, output=out, stderr=err)`
but the stderr input was not added until python 3.5

jakevdp:
CLosed by #225

PR #225, opened 2016-10-12
making calledprocesserror python2 compatible

gramhagen:
for #224 
combining output and error to pass into CalledProcessError output


jakevdp:
Thanks!

PR #237, opened 2016-10-18
DOC: correct case in GitHub

willingc:
@ellisonbg I noticed this when we looked at the docs earlier today. I've gone through all the files in the `docs` directory. These are the only files that needed edits.


jakevdp:
Thanks!

PR #239, opened 2016-10-20
MAINT: add-in missing examples

jakevdp:
Bug in the generation code was ignoring a number of examples. Also had to slightly tweak things to work correctly with explicitly specified data.


jakevdp:
Tests look good; merging

PR #242, opened 2016-10-21
Require traitlets 4.3.1

SylvainCorlay:
Traitlets 4.3 fixes the issues that you reported wrt Union trait types. Error messages should now make sense.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-traitlets>=4.1
+traitlets>=4.3.1
 ipython
 pandas
 vega>=0.4.4
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,7 @@ AUTHOR_EMAIL        = "ellisonbg@gmail.com / jakevdp@cs.washington.edu"
 URL                 = 'http://altair-viz.github.io'
 DOWNLOAD_URL        = 'http://github.com/altair-viz/altair/'
 LICENSE             = 'BSD 3-clause'
-INSTALL_REQUIRES    = ['traitlets','ipython','pandas','vega>=0.4.4']
+INSTALL_REQUIRES    = ['traitlets>=4.3.1','ipython','pandas','vega>=0.4.4']
 
 
 import io


SylvainCorlay:
I don't think you need to pin the version with `==`. The 4.3 version fixed two issues with unions:

 - we have more meaningful error messages
 - we can set metadata for the union types (besides the metadata for the individual traits in the union). But you are not using metadata in altair.

jakevdp:
OK – so it sounds like the main impact on users is better error messages in some cases.

Regarding metadata: do you see any places we should be using it where we aren't?

SylvainCorlay:
> Regarding metadata: do you see any places we should be using it where we aren't?

In ipywidgets, we specify custom serializers and deserializers via metadata. (specified with the `tag` method). It seems that in altair, you are only using basic types that can be converted to json natively.

However, if you were to adopt custom trait types like [those](http://traittypes.readthedocs.io/en/latest/) you would need to specify the serialization.

I think that it would be interesting to converge in terms of how these things are serialized between the widgets and altair.

jakevdp:
Good thoughts. Our handling of dataframes at the moment bypasses traitlets, and it would be nice to implement a cleaner solution.

SylvainCorlay:
> Good thoughts. Our handling of dataframes at the moment bypasses traitlets, and it would be nice to implement a cleaner solution.

Then you can already use `traittypes.DataFrame` and specify a serializer that does the conversion you are doing manually, and a custom validation that only accepts the ones of the form that you accept (no hierarchical index).

```python
data = DataFrame().valid(no_hierachical)
```

where `no_hierarchical` is a callable that raises a trait error if the dataframe has a hierarchical index.

Example validating the shape of an array:

```python
from traitlets import HasTraits, TraitError
from traittypes import Array

def shape(*dimensions):
    def validator(trait, value):
        if value.shape != dimensions:
            raise TraitError('Expected an of shape %s and got and array with shape %s' % (dimensions, value.shape))
        else:
            return value
    return validator

class Foo(HasTraits):
    bar = Array(np.identity(2)).valid(shape(2, 2))
foo = Foo()

foo.bar = [1, 2]  # Should raise a TraitError
```

SylvainCorlay:
The test failure is a network issue.

jakevdp:
Could you make the same change in ``requirements.txt``?

SylvainCorlay:
Done. Sorry for the oversight.

jakevdp:
Will ``traittypes.DataFrame`` work within a union?

SylvainCorlay:
Sure. However, if you use serializers in a union, they must be specified for each member of the union (since they are only valid for one or the other).

```python
foo = DataFrame().tag(to_json=my_df_to_json) | Dict(trait=List())
```

jakevdp:
I actually didn't know you could use ``|`` to make a union :smile:

jakevdp:
Thanks!

jakevdp:
I'll chat with @ellisonbg at some point about moving to using traittypes.

Issue #238, opened 2016-10-18
Run CI tests with pip?


jakevdp:
We currently run CI tests with conda. Should we run with pip as well, to ensure our installation instructions work correctly?

jakevdp:
Pros: more comprehensive testing. Cons: doubles the CI load.

stonebig:
pip is the Python community standard, so you should test with pip.

jakevdp:
conda is far more of a standard in the Python data science community. For altair, it also has the advantage that it can fully set-up the package, where pip can't (we've already had users report issues that boiled down to them ignoring the install instructions and assuming ``pip install altair`` would work). I would prefer to keep conda as the primary recommended installation path.

stonebig:
keeping conda as your "best advice" is fine, but the community standard is still "pip".

jakevdp:
We support both routes of installation, as well as Python 2 and 3, and we provide conda builds on OSX, Linux, and Windows. I'm not choosing sides here. The question in this Issue is whether it's worth the CI resources to test multiple means of installation.

jakevdp:
Closed by #243

PR #243, opened 2016-10-21
MAINT: add pip install to CI

jakevdp:
addresses #238


jakevdp:
This doesn't significantly impact the overall CI load: a pip build is about ~2 minutes faster than a conda build. I added three pip builds and changed the conda test build to pip, meaning the overall clock time is roughly constant.

PR #250, opened 2016-10-28
ENH: interactive widget to plot different columns (WIP)

stsievert:
Adds `altair.utils.interact_with` that allows functionality to plot different columns:

untitled-3

This depends on ipywidgets.interact. I see there's some work on using the branch altair#widget (not my branch). In my implementation, I have dropdown menus for the type of mark (line, point, area, etc) and the different encodings (x, y, color, etc).

There's a ton of work to do still:

* tests, documentation
* support for different stats function (median, max, mean, etc)
* style dropdown menus
* ensure works for heat maps (for my use case this will require `applyColorToBackground`). This will likely require styling the drop downs.

### DIFF
--- a/altair/utils/__init__.py
+++ b/altair/utils/__init__.py
@@ -1,1 +1,2 @@
 from .core import *
+from .widget import interact_with
--- a/altair/utils/widget.py
+++ b/altair/utils/widget.py
@@ -1,0 +1,27 @@
+# from .. import altair
+import altair
+
+
+def interact_with(df):
+    from ipywidgets import interact, interactive, fixed
+    import ipywidgets as widgets
+
+    def plot(x=None, y=None, color=None, text=None,
+         mark='mark_point', shape=None, size=None):
+        if df is None:
+            raise ValueError('pandas.DataFrame needs to be passed in')
+        kwargs = {'x': x, 'y': y, 'color': color, 'text': text,
+                'shape': shape, 'size': size}
+        for key, value in list(kwargs.items()):
+            if value in {None, False, 'None'}:
+                del kwargs[key]
+
+        c = getattr(altair.Chart(df), mark)().encode(**kwargs)
+        return c
+
+    cols = [None] + list(df.columns)
+    marks = ['mark_' + f for f in ['line', 'point', 'bar', 'tick', 'text',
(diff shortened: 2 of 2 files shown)


jakevdp:
This is pretty cool! One note: what you're doing here is quite similar to the voyager 2 JS frontend that the Vega-Lite team are developing; see a demo at https://uwdata.github.io/voyager2/

Long-term I'd love to enable seamless two-way interaction between Altair and Voyager2 within JupyterLab. With that in mind, we'll have to think carefully about whether we want to duplicate that type of functionality at the ipywidget level as well. An enhancement like the one you show here would lead to a non-negligible maintenance cost in the long-term, which is an important consideration.

stsievert:
Interesting -- thanks for showing! Voyager 2 has a better user interface  -- has more clear how Vega works and better dashboard interface to select functions.

However, right now I do prefer the ipywidgets implementation slightly. I like the immediate showing of results by selecting from a drop-down menu, instead of clicking and dragging (which feels like more work).

> a non-negligible maintenance cost in the long-term

I had the same thought. Maybe we should add to the docs instead? We can mention Voyager 2 as the upcoming software then mention that if a basic and immediate implementation needed, use the function implemented here.

jakevdp:
> Maybe we should add to the docs instead?

How would you feel about putting this widget interface under a separate package in the altair-viz org? perhaps altair-viz/altair_widgets? I'd love to see this developed more, and a separate package would allow more flexibility in that.

stsievert:
Perfect! I've published this with the same license as altair at https://github.com/stsievert/altair_widgets. I've added you as a collaborator so you can transfer the repo into altair-viz.

I've also added support for row and column.

stsievert:
I've added a better interfacing option.

car_example

jakevdp:
I think only the repo owner can transfer... I've added you as a member of altair-viz, so I think you can do it now.

stsievert:
Repo transferred. It's only possible to transfer repos to organizations you have admin rights to :/. But I'm comfortable closing this PR now.

I have put more work into better styling the menu (the current toolbar is below). I have run into a fairly small issue, and can resolve it by putting more effort into tests and simplifying the code.

screen shot 2016-10-29 at 3 01 33 pm

PR #252, opened 2016-11-01
'mean(temp_range):Q' had a bug in Jupyter Notebook

muxuezi:
There was a bug on [Exploring Data: Seattle Weather](https://altair-viz.github.io/tutorials/exploring-weather.html) when we try the code in Jupyter Notebook:

```python
Chart(df).mark_line().encode(
    X('date:T', timeUnit='month'),
    y='mean(temp_range):Q'
).transform_data(
    calculate=[temp_range]
)
```
However, it was ok if we change `'mean(temp_range):Q'` to `'mean(temp_range)'`.

```python
Chart(df).mark_line().encode(
    X('date:T', timeUnit='month'),
    y='mean(temp_range)'
).transform_data(
    calculate=[temp_range]
)
```

### DIFF
--- a/doc/tutorials/exploring-weather.rst
+++ b/doc/tutorials/exploring-weather.rst
@@ -139,7 +139,7 @@ be calculated by the renderer.
 
     Chart(df).mark_line().encode(
         X('date:T', timeUnit='month'),
-        y='mean(temp_range):Q'
+        y='mean(temp_range)'
     )
 
 Of course, the same calculation could be done by using Pandas manipulations to


jakevdp:
The ``Q`` specifies that the temperature range is a quantitative type. While this can be inferred if ``temp_range`` is a column in the dataframe, it is better to specify it explicitly in the case of a computed column. It may be that more recent versions of Vega-Lite can handle this, but as of a month ago this would break if the ``Q`` is not specified.

The Vega-Lite docs do not make clear when we can expect such type inference to happen automatically, and so until they do I would prefer to specify the data type explicitly here.

Thanks for the report!

kanitw:
We currently don't infer yet so i think if it passes in any case, it's
passing by accident!

PR #257, opened 2016-11-01
MAINT: run sphinx doctests in travis

jakevdp:


jakevdp:
Running into Python 2.7 unicode issues, similar to [this](http://stackoverflow.com/questions/31243623/doctest-fails-due-to-unicode-leading-u). After poking around online, I can't find any clean way to fix this within sphinx-doctest, so I'm going to leave this out.

PR #308, opened 2017-03-15
Fix Javascript URLs

### LINKED ISSUE #306: Docs are broken
Plots on the [Altair docs](https://altair-viz.github.io/) no longer render. This is because the script at http://vega.github.io/vega/vega.js has been updated to version 3.0, and it's not backward compatible.

Are there persistent URLs for older versions of vega & vega-lite anywhere?

jakevdp:
This adds an option to the altairplot extension to specify vega javascript paths. Fixes #306

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -301,6 +301,12 @@ texinfo_documents = [
 # If true, do not generate a @detailmenu in the "Top" node's menu.
 #texinfo_no_detailmenu = False
 
+# Hide extra class members
+numpydoc_show_class_members = False
 
-# XXX: just for testing
-todo_include_todos = True
+# For the altairplot extension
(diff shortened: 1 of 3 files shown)


jakevdp:
We could think about populating the URLs using version info within the altair library itself, but that may be overkill.

jakevdp:
Seems to work, but I'll still need to push a doc build.

Unfortunately, that requires installing vega via npm (to generate example thumbnails), which depends on canvas, which seems to be nigh impossible to install on OSX Sierra, which I just upgraded to. *Grumble grumble...*

jakevdp:
OK – I hacked the fix on github pages. I'm going to merge this because it will fix the issue long-term.

PR #319, opened 2017-04-20
MAINT: fix travis issues; allow doc build to fail

jakevdp:


jakevdp:
Failures in the second commit are really strange... it looks like it's using ``.travis.yml`` from the first commit.

jakevdp:
Seems to be working in the third commit. Strange.

PR #334, opened 2017-05-23
Migrate to pandas.api from pandas.lib where available

### LINKED ISSUE #332: FutureWarning in Pandas 0.20.1
Since upgrading to Pandas 0.20.1 I get this warning when first using altair in a notebook.

```
site-packages\altair\utils\core.py:110: FutureWarning: pandas.lib is deprecated and will be removed in a future version.
You can access infer_dtype as pandas.api.types.infer_dtype
  typ = pd.lib.infer_dtype(data)
```

ElliotJH:
Fixes #332 by using the newer API.

Let me know if we'd rather break compat with older pandas and avoid the import check.

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -7,6 +7,10 @@ import warnings
 import pandas as pd
 import numpy as np
 
+try:
+    from pandas.api.types import infer_dtype
+except ImportError: # Pandas before 0.20.0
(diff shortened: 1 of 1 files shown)


rgbkrk:
Looking forward to this shipping, I ran into this today for a presentation. :smile:

PR #347, opened 2017-06-30
correct default url for version 1 of vega-lite.js

eocarragain:
The default url for vega-lite.js no longer seems to work which means it has be overwritten in chart.to_html() every time.


jakevdp:
It was recently suggested that we use the cdnjs url instead. See https://github.com/altair-viz/altair/issues/342#issuecomment-310876271

Would you mind updating the PR?

eocarragain:
Sure. Updated to use minified versions from cdnjs. Thanks

jakevdp:
Thanks!

PR #357, opened 2017-07-19
MAINT: make update_subtraits a helper rather than a baseobject method

jakevdp:
This PR rearranges some code to streamline the transition to Altair 2.0.


jakevdp:
Forgot to commit some files

PR #358, opened 2017-07-19
MAINT: make update_subtraits a helper rather than a baseobject method

jakevdp:
This rearranges some code in a way that will streamline the transition to Altair 2.0


jakevdp:
Tests passing, I'm going to merge.

PR #409, opened 2017-10-17
ENH to_dict(): clone chart object before mutation

jakevdp:
As discussed today with @ellisonbg.

This makes it so that exporting and/or visualizing the chart doesn't silently modify the object, but rather creates a clone of the chart before modification.


jakevdp:
Thanks! I'll merge

PR #414, opened 2017-10-18
DOC: change website to ReadTheDocs theme

jakevdp:
These are the changes required to use the basic ReadTheDocs theme. Only one nontrivial thing in here: because the theme loads javascript at the bottom rather than the top of each page, we need to use a ``DOMContentLoaded`` listener to build the charts once the rest of the content is loaded.

This is supported by all modern browsers, but not by IE8 or before, which according to [this site](http://caniuse.com/#feat=domcontentloaded) only account for about 0.2% of web traffic.

Can we live with that?

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -124,7 +124,7 @@ todo_include_todos = False
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'haiku'
+html_theme = 'sphinx_rtd_theme'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -2,7 +2,7 @@ Declarative Visualization in Python
 ===================================
 
 .. altair-minigallery::
-   :size: 6
+   :size: 5
    :width: 125px
    :shuffle:
    :seed: 0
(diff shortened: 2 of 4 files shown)


jakevdp:
CC/ @mroswell @ellisonbg

ellisonbg:
I am fine with that!

jakevdp:
OK – I'll wait for tests to run and then merge & tomorrow morning will publish the new website. FYI, the conda-forge PR for Altair 1.2.1 has been merged, so once that's done we'll be officially released :smile:

jakevdp:
I'm restarting the stalled job.

mroswell:
Lots of improvements here; Left sidebar, search, much more readable characters per line, responsive! One thing we should fix is cell wrap within a table for instance at: 
https://altair-viz.github.io/API.html#encoding-channels

Probably an answer somewhere here: 
https://stackoverflow.com/questions/1057574/html-td-wrap-text 
though I haven't quite nailed it yet.

mroswell:
I also wasn't successful with this stack answer: 
https://stackoverflow.com/questions/6666532/how-to-force-table-cell-td-content-to-wrap

But I found this offered solution, that's specific to the ReadTheDocs issue. Will see if I can get it to work.
http://rackerlabs.github.io/docs-rackspace/tools/rtd-tables.html

jakevdp:
Thanks @mroswell – would you mind opening a new issue so we can keep track of these doc readability/usability observations?

mroswell:
Will do.

PR #442, opened 2018-02-03
flake8 finds three undefined names

cclauss:
https://travis-ci.org/altair-viz/altair/jobs/336795398#L1426-L1434
* parse_shorthand() needs to be imported from https://github.com/altair-viz/altair/blob/master/altair/v1/traitlet_utils.py#L165
* Change __path__ --> __schema_path__ on https://github.com/altair-viz/altair/blob/master/tools/altair_tools/generate_interface_v1.py#L274 ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,13 @@ before_install:
 install:
   - python setup.py install
 
+before_script:
+    - pip install flake8  # pytest  # add another testing frameworks later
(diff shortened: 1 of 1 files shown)


jakevdp:
Hi – thanks for this! We ended up crossing commits, and now the repo is entirely re-organized, so this PR is no longer valid. I'll keep this in mind as we continue to develop the v2 code. Thanks!

PR #461, opened 2018-02-23
Added a multi series line example

eitanlees:
I added a multi series line example using the `stocks` dataset.


jakevdp:
Awesome, thanks!

PR #462, opened 2018-02-23
added simple line chart with percent axis

afonit:
Trying to do my first commit.
I saw the github request for examples.
I tried to keep the format I saw in the provided example of naming the item as 'chart' to aid in testing.

Please let me know if I have done anything wrong - as this is my first pull request.

### DIFF
--- a/altair/vegalite/v2/examples/line_percent.py
+++ b/altair/vegalite/v2/examples/line_percent.py
@@ -1,0 +1,11 @@
+import altair as alt
+from vega_datasets import data
+
+source = data.jobs.dataframe()
+
+base = source[source['job'] == 'Accountant / Auditor']
+    
+chart = alt.Chart(base).mark_line().encode(
+    x='year:T',
+    y=alt.Y('perc', axis=alt.Axis(format='%')),
+    color='sex')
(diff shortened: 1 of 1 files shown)


jakevdp:
Looks good, thanks!

A couple things:

- it looks like you're using an old version of ``vega_datasets``: the ``.dataframe()`` attribute is no longer valid
- please add a description at the top of the file
- when I run this, the x-axis labels look a bit off. Are you seeing the same thing?

vega 4

afonit:
Thanks for the feedback @jakevdp 
I will add that to the title, and investigate how to update my vega_datasets package.
I deleted the environment and reloaded everything - I do see that date issue on the x-axis, I will work on figuring out the cause for that.

jakevdp:
FYI, you can leave this PR open and push new commits to the same branch, and they will show up here.

afonit:
@jakevdp I closed it because it looks like it may be a bug in 2.0?
I rand the same example in v1 of vega lite and the x axis looks fine, but in v2 the axis is indeed all scrunched up so it looked like a regression and I was not sure how to handle that.  I though I shouldn't leave this open since it may take some time for me to figure out why there is a difference, but I will go ahead and reopen it then.

afonit:
closed to fix beginner branching issues.

PR #463, opened 2018-02-23
Interval selection

jakevdp:
This PR is an initial attempt at streamlining the creation and use of selections in Altair. I've added a ``SelectionMapping`` object to the API, which users should not have to directly use. I've also added an ``alt.selection`` function which allows users to create and manipulate named selection.

The example file shows how I envision this working, and it creates an interactive example similar to the one [here](https://vega.github.io/vega-lite/examples/interactive_overview_detail.html).

Any thoughts on this Python API for selections? cc/ @ellisonbg @domoritz @kanitw @jheer

### DIFF
--- a/altair/vegalite/v2/examples/interval_selection.py
+++ b/altair/vegalite/v2/examples/interval_selection.py
@@ -8,13 +8,13 @@ top chart can be selected by interacting with the bottom chart.
 
 import altair as alt
 from vega_datasets import data
-sp500 = data.sp500()
+sp500 = data.sp500.url
 
 brush = alt.selection(type='interval', encodings=['x'])
 
 upper = alt.Chart().mark_area().encode(
-    x=alt.X('date', scale={'domain': brush.ref()}),
-    y='price'
+    x=alt.X('date:T', scale={'domain': brush.ref()}),
+    y='price:Q'
 ).properties(
     width=600,
     height=200
--- a/altair/vegalite/v2/examples/interval_selection.py
+++ b/altair/vegalite/v2/examples/interval_selection.py
@@ -10,7 +10,7 @@ import altair as alt
 from vega_datasets import data
 sp500 = data.sp500()
 
-brush = alt.selection(type=interval, encodings=['x'])
+brush = alt.selection(type='interval', encodings=['x'])
 
 upper = alt.Chart().mark_area().encode(
     x=alt.X('date', scale={'domain': brush.ref()}),
--- a/altair/vegalite/v2/examples/interval_selection.py
+++ b/altair/vegalite/v2/examples/interval_selection.py
@@ -10,7 +10,7 @@ import altair as alt
 from vega_datasets import data
 sp500 = data.sp500()
 
-brush = alt.selection('brush', {"type": "interval", "encodings": ['x']})
+brush = alt.selection(type=interval, encodings=['x'])
 
 upper = alt.Chart().mark_area().encode(
     x=alt.X('date', scale={'domain': brush.ref()}),
--- a/altair/vegalite/v2/examples/interval_selection.py
+++ b/altair/vegalite/v2/examples/interval_selection.py
@@ -1,0 +1,28 @@
+"""
+Interval Selection Example
+==========================
+
+This is an example of creating a stacked chart for which the domain of the
+top chart can be selected by interacting with the bottom chart.
+"""
+
+import altair as alt
+from vega_datasets import data
+sp500 = data.sp500()
+
+brush = alt.selection('brush', {"type": "interval", "encodings": ['x']})
+
+upper = alt.Chart().mark_area().encode(
+    x=alt.X('date', scale={'domain': brush.ref()}),
+    y='price'
+).properties(
+    width=600,
+    height=200
+)
+
+lower = upper.properties(
+    selection=brush,
+    height=60
+)
+
+chart = alt.VConcatChart([upper, lower], data=sp500)
--- a/altair/vegalite/v2/examples/scatter_linked_brush.py
+++ b/altair/vegalite/v2/examples/scatter_linked_brush.py
@@ -1,0 +1,25 @@
+"""
+Faceted Scatter Plot with Linked Brushing
+-----------------------------------------
+This is an example of using an interval selection to control the color of
+points across multiple facets.
+"""
+
+import altair as alt
+from vega_datasets import data
+
+cars = data.cars()
+
+brush = alt.selection(type='interval', resolve='global')
+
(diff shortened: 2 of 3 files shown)


jakevdp:
Sorry, a stray commit from master ended up in the diff here. Focus on the diff in the second commit.

jakevdp:
FYI, I added one more feature of the selection: looking at the examples you'll see how it can be used. In short, once you've defined a selection object, you can do something like the following:

```python
selection = alt.selection(type='interval')

alt.Chart().encode(
    color=selection.condition(value_if_in_selection, value_if_not_in_selection)
)
```

altair/vegalite/v2/examples/interval_selection.py
```diff
@@ -0,0 +1,28 @@
...
+    width=600,
+    height=200
+)
+
+lower = upper.properties(
+    selection=brush,
+    height=60
+)
+
+chart = alt.VConcatChart([upper, lower], data=sp500)
```
kanitw: `VConcatChart` is a bit verbose -- I guess you would have `VConcat` or even `V` (that @ellisonbg proposed) as aliases?

altair/vegalite/v2/examples/interval_selection.py
```diff
@@ -0,0 +1,28 @@
...
+upper = alt.Chart().mark_area().encode(
+    x=alt.X('date', scale={'domain': brush.ref()}),
+    y='price'
+).properties(
+    width=600,
+    height=200
+)
+
+lower = upper.properties(
+    selection=brush,
```
kanitw: I wonder how would this `selection = brush` thing works when there are multiple selections on the same unit specs.

I wonder if there should be a `.add_selection(brush)` instead.

altair/vegalite/v2/examples/scatter_linked_brush.py
```diff
@@ -0,0 +1,25 @@
...
+import altair as alt
+from vega_datasets import data
+
+cars = data.cars()
+
+brush = alt.selection(type='interval', resolve='global')
+
+base = alt.Chart(cars).mark_point().encode(
+    y='Miles_per_Gallon',
+    color=brush.condition('Origin',
```
kanitw: Seeing this syntax for condition with selection as predicate, I wonder how would condition works for general `test` predicate (see bottom of https://vega.github.io/vega-lite/docs/condition.html). 

I wonder if we should make a general construct that works for both `test` and `selection`? 

For example:

```
color=alt.condition(brush, "Origin", alt.ColorValue('gray'))
```
where `brush` is a selection could work

or we can put a test predicate (i'm not sure if there is a specific Altair syntax for it, but let's say it can support Vega expression string as well). 
```
color=alt.condition("datum.x > 5", "Origin", alt.ColorValue('gray'))
```

altair/vegalite/v2/examples/scatter_linked_brush.py
```diff
@@ -0,0 +1,25 @@
...
+                          alt.ColorValue('gray'))
```
kanitw: This is not the point of this PR, but I think I should ask anyway.

I wonder why this has to be `alt.ColorValue('gray')` -- wouldn't `alt.Value('gray')` or something similar sufficient?  

(At the end of the day, it's just `{value: "gray"}` in JSON.)

altair/vegalite/v2/examples/interval_selection.py
jakevdp: Well, in the schema it's ``TopLeve<lVConcatSpec>``, so I thought ``VConcatChart`` was a nice concise alias 😀

jakevdp: And yes, ``&`` works as well, but it's less convenient to specify data if you use ``&``

altair/vegalite/v2/examples/interval_selection.py
jakevdp: If there are multiple selections, you can add them together, e.g. ``selection = brush + pan``.

Yes, an ``add_selection`` method is a good idea.

altair/vegalite/v2/examples/scatter_linked_brush.py
jakevdp: The way the schema is set up, the encodings all have slightly different definitions (i.e. ``color`` is a ``MarkPropValueDefWithCondition``, ``text`` is a ``TextValueDefWithCondition``, ``x`` is a ``ValueDef``, etc.) Given that every channel has a specific value type that it can accept, I thought the easiest approach for usability would be to alias the particular value type to the channel name.

That way, the user doesn't need a lookup table to remember that ``x`` should accept a ``ValueDef``. They just need to remember that ``x`` takes an ``XValue()``.

altair/vegalite/v2/examples/scatter_linked_brush.py
jakevdp: I really like this.

altair/vegalite/v2/examples/interval_selection.py
jakevdp: I think in the end I'd probably keep the object name something like ``VConcatChart``, but provide an ``alt.vconcat`` function to make things more concise.

domoritz:
cc @arvind

altair/vegalite/v2/examples/interval_selection.py
kanitw: This sounds good.  Also, thanks for bringing up about `TopLevel<VConcatSpec>`. 

Now I'm also curious how Altair deals with the fact that [some properties such as `autosize` and `config`   are only supported at the `TopLevel`](https://vega.github.io/vega-lite/docs/spec.html#top-level-specifications).

(That's why we have `VConcatSpec` and `TopLevel<VConcatSpec>` -- the latter is for the top-level spec only -- when nested, only the former can be used.)

jakevdp: Yeah, I haven't really figured that out yet. Any ideas? :smile:

jakevdp: I added ``vconcat()``, ``hconcat()``, and ``layer()`` functions in https://github.com/altair-viz/altair/pull/476

altair/vegalite/v2/examples/scatter_linked_brush.py
jakevdp: OK, changed!

jakevdp:
After discussing with @ellisonbg this morning, I'm going to merge this and we can continue discussing from master. That doesn't mean this API is necessarily finalized, but we think it's a good start.

Please feel free to weigh-in on this PR or open another issue if you have thoughts on this!

PR #465, opened 2018-02-23
MAINT: add  property to top-level objects on to_dict().

jakevdp:
This will make them automatically render correctly when exported to the vega editor.


jakevdp:
Sounds good. Just FYI, the reason I wanted to add this is so that when I do ``Chart.to_dict()`` and convert the output to JSON, it displays correctly in the vega editor. Without the ``$schema`` argument, the editor assumes the spec is vega rather than vega-lite.

Relatedly, can we use the rendering layer to implement ``to_json()`` for the top-level objects?

PR #466, opened 2018-02-23
added layered chart, bar and mark

afonit:
This could be prettier, I am playing around with the mark attributes but size does not seem to being doing anything.

This is a chart that I have been really waiting for to use in altair - I have been doing it in bokeh since I could not do it in altair.

It is a visual way to see how the goals (in red) are being met by current activity (the bars).

### DIFF
--- a/altair/vegalite/v2/examples/line_percent.py
+++ b/altair/vegalite/v2/examples/line_percent.py
@@ -1,0 +1,11 @@
+import altair as alt
+from vega_datasets import data
+
+source = data.jobs.dataframe()
+
+base = source[source['job'] == 'Accountant / Auditor']
+    
+chart = alt.Chart(base).mark_line().encode(
+    x='year:T',
+    y=alt.Y('perc', axis=alt.Axis(format='%')),
+    color='sex')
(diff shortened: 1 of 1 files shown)


jakevdp:
Very cool. One comment: we have a shortcut to creating layered charts:
```
chart = a + b
```

Also, it looks like the tick configuration has to happen via the top-level config setting. Something like this works:
```python
chart = a + b
chart.config = {'tick': {'thickness': 2, 'bandSize': 50}}
```

Altair 1 had ``configure_xxx()`` methods, which we should implement in Altair 2, so we could do something like 
```
chart = (a + b).configure_tick(thickness=2, bandSize=50)
```

afonit:
Thanks for the recommendations.  I have added those.
I am looking through vegalite to see if there is a way to force a legend as that would be helpful in making the chart more understandable to have the labels 'score' and 'goal' show up.

afonit:
I asked the vega-lite team since I don't see it in the docs:
https://github.com/vega/vega-lite/issues/3411

afonit:
closing to fix some beginner mistakes with branching

PR #470, opened 2018-02-23
Anscombe’s Quartet Example

eitanlees:
An example of a Trellis scatter plot


jakevdp:
Nice. I'd add an explicit call to 
```
Chart(...
     ...
).properties(
    width=150, height=150
)
```
to make the panels a convenient size. Currently Altair sets the default width a bit too large for this kind of faceted plot... we may want to visit that

jakevdp:
Very nice – thanks!

PR #471, opened 2018-02-23
adding simple line example with percent axis

afonit:
trying this again with clean copies.

### DIFF
--- a/altair/vegalite/v2/examples/line_percent.py
+++ b/altair/vegalite/v2/examples/line_percent.py
@@ -1,0 +1,17 @@
+"""
+Line Chart with Percent axis
+-----------------------
+This example shows how to set an axis as a percent.
+"""
+
+import altair as alt
+from vega_datasets import data
+
+source = data.jobs()
+
+base = source[source['job'] == 'Welder']
+    
+chart = alt.Chart(base).mark_line().encode(
+    x='year:T',
+    y=alt.Y('perc', axis=alt.Axis(format='%')),
+    color='sex')
--- a/altair/vegalite/v2/examples/line_percent.py
+++ b/altair/vegalite/v2/examples/line_percent.py
@@ -9,6 +9,8 @@ from vega_datasets import data
 
 source = data.jobs()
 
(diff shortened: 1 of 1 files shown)


jakevdp:
OK, so I took a look at this. It turns out the issue is that because the year is encoded as an integer, it's not correctly interpreted as a date. If you convert the column to string, vega-lite will parse the dates correctly:
```python
import altair as alt

source = data.jobs()

# The year here is interpreted as an integer. When treating columns as dates,
# it is best to use either a string representation or a datetime representation.
source.year = source.year.astype(str)

welders = source[source.job == 'Welder']
    
chart = alt.Chart(welders).mark_line().encode(
    x=alt.X('year', timeUnit='year'),
    y=alt.Y('perc', axis=alt.Axis(format='%')),
    color='sex',
    detail='job'
).properties(
    title='Percent of work-force working as Welders'
)
```
vega 6

altair/vegalite/v2/examples/line_percent.py
```diff
@@ -0,0 +1,23 @@
...
+-----------------------
+This example shows how to set an axis as a percent.
+"""
+
+import altair as alt
+from vega_datasets import data
+
+source = data.jobs()
+
+source.year = source.year.astype(str)
```
jakevdp: Could you add a comment about why we convert the column to a string? Since this will be an example in the documentation, it will help to err toward didacticism.

altair/vegalite/v2/examples/line_percent.py
```diff
@@ -0,0 +1,23 @@
...
+
+welders = source[source.job == 'Welder']
+    
+chart = alt.Chart(welders).mark_line().encode(
+    x=alt.X('year', timeUnit='year'),
+    y=alt.Y('perc', axis=alt.Axis(format='%')),
+    color='sex',
+    detail='job'
```
jakevdp: This line should be removed – since the data contains only one job, it doesn't do anything.

jakevdp:
Looks good, thanks!

PR #473, opened 2018-02-23
adding layered chart bar and mark

afonit:
This is useful to quickly see how performance is against some preset targets.


jakevdp:
Looks good!

Eventually I'd like to add ``config_*`` methods so that the chart can be made using the normal method-chaining style, but we can merge this for now!

PR #474, opened 2018-02-24
add in vega-lite example of line layer rule

afonit:
Jake, if I notice things not working while building examples - please let me know if you would rather me file a github issue, or do a pull request of the example noting the difficulties.

I saw this example:
https://github.com/altair-viz/altair/blob/master/altair/vegalite/v2/examples/mutli_series_line.py
and then looked at the vega-lite example:
https://vega.github.io/editor/#/examples/vega-lite/layer_line_color_rule

I noticed that I could not get the .config to modify an attribute, in this case the line:
 chart.config = {'rule': {'size': 20}}
does not seem to be doing anything, so I commented it out for the moment.  I have looked in the vega-lite documentation and I seem to be using the correct terms so I am not sure what is causing it.

I know you showed me how .config can work in this example:
https://github.com/altair-viz/altair/blob/master/altair/vegalite/v2/examples/layered_chart_bar_mark.py

I can't get chart.interactive() to work on the layered plot.
If I un-comment that line I get:
AttributeError: 'super' object has no attribute '__getattr__'

If I only put rule.interactive() on rule I get zooming on the y axis only (undoubtedly because I am only calling y for the rule).
If I only put line.interactive() on line, it works fine.

Is the expectation to not but interactive() on the layered chart - and just do it on one of the component charts?

If that is the case then line.interactive() and add a comment for others.

### DIFF
--- a/altair/vegalite/v2/examples/layer_line_color_rule.py
+++ b/altair/vegalite/v2/examples/layer_line_color_rule.py
@@ -18,6 +18,6 @@ line = alt.Chart(stocks).mark_line().encode(
 rule = alt.Chart(stocks).mark_rule().encode(
     y = alt.Y('average(price)', ),
     color='symbol'
-)
+).encode(size=alt.SizeValue(2))
 
 chart = (line + rule)
--- a/altair/vegalite/v2/examples/layer_line_color_rule.py
+++ b/altair/vegalite/v2/examples/layer_line_color_rule.py
@@ -13,12 +13,13 @@ line = alt.Chart(stocks).mark_line().encode(
     x = 'date', 
     y = 'price', 
     color = 'symbol'
-).properties(width=800,
-            title="Daily closing prices with their aggregate prices").interactive()
+).properties(width = 800,
+            title = "Daily closing prices with their aggregate prices").interactive()
 
 rule = alt.Chart(stocks).mark_rule().encode(
     y = alt.Y('average(price)', ),
-    color='symbol'
-).encode(size=alt.SizeValue(2))
+    color = 'symbol',
+    size = alt.SizeValue(2)
+)
 
 chart = (line + rule)
--- a/altair/vegalite/v2/examples/layer_line_color_rule.py
+++ b/altair/vegalite/v2/examples/layer_line_color_rule.py
@@ -13,7 +13,7 @@ line = alt.Chart(stocks).mark_line().encode(
     x = 'date', 
     y = 'price', 
     color = 'symbol'
-).interactive()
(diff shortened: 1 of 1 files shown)


jakevdp:
Yes, the ``interactive`` can't be called at the top level; it has to be called on the component charts.

The reason you get the error ``AttributeError: 'super' object has no attribute '__getattr__'`` is becuase the traceback is trying to tell you that ``LayeredChart.interactive`` is not a callable function, but because of the ``__getattr__`` magic used in ``SchemaBase`` this part of the traceback fails. That's something I need to figure out how to rectify.

jakevdp:
OK, the attribute error should be fixed in b33af2cbe3a9bf97d35c2. That should make things a bit clearer: it will now tell you that the ``LayeredChart`` does not have an ``interactive`` attribute.

jakevdp:
As for ``chart.config = {'rule': {'size': 20}}``, the reason this does nothing is that the ``size`` config value is not valid for ``rule``, but only for ``point``, ``circle``, and ``square`` marks. See the [vega-lite docs](https://vega.github.io/vega-lite/docs/mark.html#point-size-config-for-point-circle-and-square-marks).

afonit:
humm, on this page:


It shows 'size' being used for rule
https://vega.github.io/vega-lite/docs/rule.html

{
  "data": {"url": "data/stocks.csv"},
  "mark": "rule",
  "encoding": {
    "y": {
      "field": "price",
      "type": "quantitative",
      "aggregate": "mean"
    },
    "size": {"value": 2},
    "color": {"field": "symbol","type": "nominal"}
  }
}

but there must be something I am missing still, so I will take some time tomorrow are read more through the docs as you have suggested.

jakevdp:
Oh, that's not a size ``config``, that's a size ``encoding``. They're two different things.

You can set that by calling ``chart.encode(size=SizeValue(2))``

altair/vegalite/v2/examples/layer_line_color_rule.py
```diff
@@ -0,0 +1,24 @@
...
+    x = 'date', 
+    y = 'price', 
+    color = 'symbol'
+).properties(width=800,
+            title="Daily closing prices with their aggregate prices").interactive()
+
+rule = alt.Chart(stocks).mark_rule().encode(
+    y = alt.Y('average(price)', ),
+    color='symbol'
+).encode(size=alt.SizeValue(2))
```
jakevdp: You only need to call ``encode`` once, and put all the encodings in there.

jakevdp:
Looks good, thanks!

PR #477, opened 2018-02-24
Horizontal stacked bar chart example

eitanlees:
A horizontal stacked bar chart using the barley dataset.


jakevdp:
Very nice, thanks!

PR #479, opened 2018-02-24
adding vega-lite example of table binned heatmap

afonit:
only issue is I tried many permutations to get the color scheme defined without using the config dictionary method.  If that is implemented I would love to change it - I just could not figure it out if it is there.


jakevdp:
I would modify the encoding's scale directly rather than adjusting the global config (you can think of the config as a chart theme):
```
import altair as alt

source = data.movies()

chart = alt.Chart(source).mark_rect().encode(
    x = alt.X('IMDB_Rating', bin=alt.BinParams(maxbins=60)), 
    y = alt.Y('Rotten_Tomatoes_Rating', bin = alt.BinParams(maxbins=40)),
    color = alt.Color('count(IMDB_Rating)',
                      scale=alt.Scale(scheme='greenblue'))
)
```

jakevdp:
Nice!

PR #481, opened 2018-02-24
adding trellis plot from vega-lite

afonit:
All works - but for some reason the rangeStep does not seem to be having the affect of reducing the spaces between the y axis items.

It was throwing errors until I did the capital 'S', but now that there aren't any errors - it also does not seem to be changing the chart, even when using extreme values.


jakevdp:
I think the reason ``rangeStep`` is not working is because the ``Chart`` sets a width and height by default, and the width and height setting override the ``rangeStep``.

You can see this by explicitly setting width and height to ``alt.Undefined``; then the range step will work.

I've been thinking we should get rid of this width/height default, because it causes issues for faceted plots among other things. This issue might tip the scales.

PR #484, opened 2018-02-24
added vega-lite example of falkensee

afonit:
This has one issue:
When building the final chart
chart = line + recta + point

recta should be in the first position so it does not cover up the lines and points, but if I put recta there errors will result.  line + point + recta work fine as well, just (at this point) can't have recta in first position where it needs to be for the visual composition to work.

The errors are: KeyError: 'population'

I am assuming vega-lite is expecting the full axis with the item in first position - or it is building components before they are loaded(since population is only in the line/points data and not recta's data)?

The vega-lite example is here:
https://vega.github.io/vega-lite/examples/layer_falkensee.html

### DIFF
--- a/altair/vegalite/v2/examples/falkensee.py
+++ b/altair/vegalite/v2/examples/falkensee.py
@@ -80,4 +80,4 @@ recta = alt.Chart(source2).mark_rect().encode(
 
 
 
-chart = line + recta + point
+chart = recta + line + point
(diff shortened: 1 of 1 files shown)


jakevdp:
Ah, I think I know why this is happening. It has to do with how the ``context`` variable is handled within the recursive call of ``to_dict``. Let me see if I can fix it.

jakevdp:
OK, I think I got it. Once #486 is merged, you'll be able to do ``recta + line + point`` and there won't be an error.

afonit:
Great once I see it merged I will reinstall and rerun the chart. Thank you.

jakevdp:
Looks good!

PR #485, opened 2018-02-25
adding vega-lite example of layered plot with dual-axis

afonit:
Here is one I have been struggling with.  In the vega-lite example:
https://vega.github.io/vega-lite/examples/layer_bar_dual_axis.html

It has this part:
"resolve": {"scale": {"y": "independent"}}
that sits at the same level as:
"$schema": "https://vega.github.io/schema/vega-lite/v2.json",
  "data": {"url": "data/seattle-weather.csv"},

I have toyed with ideas like:
y = alt.Y('mean(temp_max)', resolveMode = 'independent'))
also tried to put it in config, but I get errors every time.

I do note that I see resolvemode in here:
https://github.com/altair-viz/altair/blob/0fec41674f062c5e37d5af0b6b318702362f0115/altair/vegalite/v2/vega-lite-schema.json

So I am missing how to connect it up.

Note: the plot included works, it just overwrites the line's y axis with the bar's y axis, if you call out line or bar you will see each looks how they should.  Just combining them I am not seeing the place to add the resolvemode.

### DIFF
--- a/altair/vegalite/v2/examples/layered_plot_with_dual_axis.py
+++ b/altair/vegalite/v2/examples/layered_plot_with_dual_axis.py
@@ -22,5 +22,5 @@ line = alt.Chart(source).mark_line(color = 'red').encode(
     y = 'mean(temp_max)')
 
 
-chart = line + bar
+chart = bar + line
 chart.resolve = {'scale': {'y': 'independent'}}
--- a/altair/vegalite/v2/examples/layered_plot_with_dual_axis.py
+++ b/altair/vegalite/v2/examples/layered_plot_with_dual_axis.py
@@ -1,0 +1,25 @@
+"""
+Layered Plot with Dual-Axis
+-----------------------
+This example shows how to combine two plots and keep their axes.
+"""
+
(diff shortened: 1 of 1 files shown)


jakevdp:
Nice! So the ``resolve`` argument has to be at the top level, and the top level is the ``chart`` object created by ``line + bar``. So you can do this (putting bar first so the line appears on top):

```
chart = bar + line
chart.resolve = {"scale": {"y": "independent"}}
```

or alternatively,
```
chart = alt.layer(bar, line, resolve={"scale": {"y": "independent"}})
```

I think eventually adding a ``set_resolve()`` method or something similar would be helpful here.

afonit:
ah, I did not see it tab complete so I did not think it was there, I only saw .resolve_references

I will try what you wrote.

jakevdp:
Oh, I see. The tab-completion for attributes is not quite working yet...

jakevdp:
OK, attribute tab completion is fixed in d450c9963fc0943fe87

jakevdp:
Looks good!

PR #487, opened 2018-02-25
added gantt_chart from vega-lite

afonit:
This one is complete - but seems to suffer from the same issue as:
https://github.com/altair-viz/altair/pull/481

Where I can add
scale = alt.Scale(rangeStep = 20)
But it does nothing to the chart.

In this example it appears comically large, and can't be shortened.

### DIFF
--- a/altair/vegalite/v2/examples/gantt_chart.py
+++ b/altair/vegalite/v2/examples/gantt_chart.py
@@ -1,0 +1,24 @@
+"""
+Gantt Chart
+-----------------
+This example shows how to make a simple gantt chart.
+"""
+
+import altair as alt
+from vega_datasets import data
+
+data = [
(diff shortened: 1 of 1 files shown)


jakevdp:
The default width/height mechanism is changed now, so things look much better:
```python
import altair as alt

data = [
      {"task": "A","start": 1, "end": 3},
      {"task": "B","start": 3, "end": 8},
      {"task": "C","start": 8, "end": 10}
       ]

source = alt.pd.DataFrame(data)


    x = 'start',
    x2 = 'end',
    y = 'task',
)
```
vega 11

afonit:
Wonderful - thank you, that looks so much nicer.

jakevdp:
Great!

PR #488, opened 2018-02-25
added binned scatterplot example from vega-lite

afonit:
I will note that:
size = 'count(*)'
does not work and gives a keyerror of:
KeyError: '*'

I thought this was a capability that was here before.
But for this example I just set it to:
size = 'count(IMDB_Rating)'

And it works fine and matches the vega-lite example.


jakevdp:
Thanks! The ``count(*):Q`` form should work after cc8000bca51cdc

jakevdp:
It works now, FYI
```python
import altair as alt

source = data.movies()

chart = alt.Chart(source).mark_circle().encode(
    x = alt.X('IMDB_Rating',  bin = alt.BinParams(maxbins=10)),
    y = alt.Y('Rotten_Tomatoes_Rating',  bin = alt.BinParams(maxbins=10)),
    size = 'count(*):Q'
)  
```
vega 10

jakevdp:
Thanks!

PR #494, opened 2018-02-25
Stacked Area Chart Example

eitanlees:
An example of a stacked area chart. I reshaped the original dataset using `pd.melt` to get it into a tidy format. I can make a simpler example with a different dataset but I thought it could be useful to see.

### DIFF
--- a/altair/vegalite/v2/examples/stacked_area_chart.py
+++ b/altair/vegalite/v2/examples/stacked_area_chart.py
@@ -1,0 +1,22 @@
+"""
+Stacked Area Chart
+==================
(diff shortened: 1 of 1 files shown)


afonit:
@eitanlees 
I tried to add a comment to the file, but I think I may have did it wrong,
so I will put it here:

I think if you don't want to import pandas, you could use alt as it has
alt.pd.melt

That will save an import if that matters.

eitanlees:
The `bar.py` example has an explicit pandas import so I figured it was ok.

jakevdp:
Looks great! I think in this case the explicit pandas import is more clear than using the alias that is in altair.

PR #495, opened 2018-02-25
added ranged dot plot example from vega-lite

afonit:
One issue,
it looks like I can't pass the transform on the layered chart (well, I can pass it, but it is not reducing the dataset as requested).
I have to do the filters individually for line and point.

Unless I have missed something.

So right now I have the chart.transform commented out with both the line and point transforms in there.

### DIFF
--- a/altair/vegalite/v2/examples/ranged_dot_plot.py
+++ b/altair/vegalite/v2/examples/ranged_dot_plot.py
@@ -1,0 +1,35 @@
+"""
+Ranged Dot Plot
+-----------------
+This example shows a ranged dot plot that uses 'layer' to convey changing life expectancy for the five most populous countries (between 1955 and 2000).
+"""
+
+import altair as alt
+from vega_datasets import data
+
+source = data.countries()
+
+
+line = alt.Chart(source).mark_line(color='#db646f').encode(
+    x = 'life_expect',
+    y = 'country',
+    detail = 'country'
+    ).interactive()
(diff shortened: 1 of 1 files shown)


jakevdp:
Very nice! So what you're running into with the transforms is the fact that data can be specified at every level of the chart. The ``LayeredChart`` can have its own data, and each of the ``Chart`` instances within the layers can have their own data.

Here, since you're using the same data and transform within each layer, you can specify the layer and transform at the top level, and leave the data out in each of the individual charts.

So something like this would work:

```python
import altair as alt

source = data.countries()

line = alt.Chart().mark_line(color='#db646f').encode(
    x = 'life_expect',
    y = 'country',
    detail = 'country'
).interactive()

point = alt.Chart().mark_point(size = 100, opacity = 1, filled = True).encode(
    x = 'life_expect',
    y = 'country',
    color=alt.Color('year:O',
        scale=alt.Scale(
            domain=['1955', '2000'],
            range=['#e6959c', '#911a24']
        )
    )
)

chart = alt.layer(line, point, data=source,
                  transform = [{'filter': {"field": 'country', "oneOf": ["China", "India", "United States", "Indonesia", "Brazil"]}},
                               {'filter': {"field": 'year', "oneOf": [1955, 2000]}}])
```

afonit:
oh, that is fascinating - thanks for the clear explanation.

jakevdp:
Looks good, thanks!

PR #497, opened 2018-02-25
Trellis Stacked Bar Chart Example

eitanlees:
Added the Trellis Stacked Bar Chart example. Let me know if the dimensions are too big.


jakevdp:
Awesome, thanks!

PR #498, opened 2018-02-25
added vega-lite example of diverging stacked bar chart

afonit:
The parameters inside the axis don't seem to be doing anything when I change them, but I kept them in as they were in the vega-lite example and they are not throwing errors in altair

y = alt.Y('question:N', axis = alt.Axis(title = 'Question',
                                           offset = 5,
                                           ticks = False,
                                           minExtent = 60,
                                           domain = False))

### DIFF
--- a/altair/vegalite/v2/examples/diverging_stacked_bar_chart.py
+++ b/altair/vegalite/v2/examples/diverging_stacked_bar_chart.py
@@ -337,23 +337,30 @@ data =     [
       }
     ]
 
-source = alt.pd.DataFrame(data)
 
-chart = alt.Chart(source).mark_bar().encode(
-    x = alt.X('percentage_start:Q'),
-    x2 = 'percentage_end:Q',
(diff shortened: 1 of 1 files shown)


jakevdp:
Those axis options do seem to be doing something, though the change is subtle (the y-axis label is capitalized, shifted slightly leftward, and the y-axis ticks are removed).

For readability sake, it might be worth defining this axis and scale separately, e.g.

```
x_scale = alt.Scale(
            domain=["Strongly disagree",
            "Disagree",
            "Neither agree nor disagree",
            "Agree",
            "Strongly agree"],
            range=["#c30d24", "#f3a583", "#cccccc", "#94c6da", "#1770ab"]
        )

y_axis = alt.Axis(title = 'Question',
                                           offset = 5,
                                           ticks = False,
                                           minExtent = 60,
                                           domain = False)),
```
and then reference those later by name.

jakevdp:
Looks good, thanks!

PR #499, opened 2018-02-25
added the streamgraph example from vega-lite

afonit:
Everything is working except turning off the y-axis.

In the vega-lite json it says:
"axis": null

I have tried:
axis = False
and
axis = 'null

neither have worked.

### DIFF
--- a/altair/vegalite/v2/examples/steamgraph.py
+++ b/altair/vegalite/v2/examples/steamgraph.py
@@ -1,0 +1,18 @@
+"""
+Streamgraph
+-----------------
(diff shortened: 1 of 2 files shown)


jakevdp:
use ``axis=None``. It's the convention used within the ``json`` package that python ``None`` becomes Javascript ``null``.

hydrosquall:
Minor typo in the filename, I think you might mean `streamgraph`

afonit:
@hydrosquall, yes you are correct, thanks for the catch

jakevdp:
Thanks!

Issue #457, opened 2018-02-22
Example Visualizations


jakevdp:
In preparation for the Altair 2.0 release, we need some good example visualizations for the documentation! These could be everything from simple one-panel scatter and line plots, to more complicated layered or stacked plots, to more advanced interactive features.

Note that the v2 API is not finalized yet, and so another purpose of creating these is to find bugs in the current package as we prepare for release. If you find anything, please report it on the issues tracker!

I've started a folder for these examples in [altair/vegalite/v2/examples/](https://github.com/altair-viz/altair/tree/master/altair/vegalite/v2/examples). You can treat [simple_scatter.py](https://github.com/altair-viz/altair/blob/master/altair/vegalite/v2/examples/simple_scatter.py) as a template.

Every example should:

- have a descriptive docstring, which will eventually be extracted for the documentation website.
- define a ``chart`` variable with the main chart object (This will be used both in the unit tests to confirm that the example executes properly, and also eventually used to display the visualization on the documentation website).
- not make any external calls to download data within the script (i.e. don't use ``urllib``). You can define your data directly within the example file, generate your data using ``pandas`` and ``numpy.random``, or you can use data available in the [vega_datasets](https://github.com/jakevdp/vega_datasets) package. 

The easiest way to get started would be to adapt examples from the [Vega-Lite example gallery](https://vega.github.io/vega-lite/examples/). Or you can feel free to be creative and build your own visualizations.

Note that the new display architecture is still being built; for display troubleshooting please see the wiki page: https://github.com/altair-viz/altair/wiki/Display-Troubleshooting

We'll look forward to your pull requests!

bdore:
I would be glad to contribute to this.

jakevdp:
Great! Let me know if you need help getting started! And please feel free to use this issue as a forum for any questions you have about example creation, that way other potential contributors will benefit as well.

kfk:
Hi Jake, this is cool and I also want to help out.

I think some directions on how to setup Altair would help. I cloned the master branch and installed with:

`make`

Then I tried to run your example and produce an html file and I get the below. Maybe I am not supposed to run the latest version of Altair this way though, in that case a bit of direction would help.

`Traceback (most recent call last):
  File "/home/k/env/lib/python3.5/site-packages/altair-2.0.0.dev0-py3.5.egg/altair/utils/schemapi.py", line 81, in __getattr__
    return self._kwds[item]
KeyError: 'to_html'

During handling of the above exception, another exception occurred:

[...]

PR #500, opened 2018-02-26
#457: Add Basic Histogram Example

hydrosquall:
Implements https://vega.github.io/vega-lite/examples/histogram.html .

The primary change compared to Altair 1.3 seems to be that the `Bin` class was renamed `BinTransform`. ( https://altair-viz.github.io/gallery/histogram.html ) . 

I'm not sure if there is a practical difference is between using `:Q` as a suffix, vs explicitly setting `type='quantitative'`. However I felt the second option was more explicit, and so I went with that instead.

### DIFF
--- a/altair/vegalite/v2/examples/histogram.py
+++ b/altair/vegalite/v2/examples/histogram.py
@@ -1,0 +1,18 @@
+"""
+Histogram
+-----------------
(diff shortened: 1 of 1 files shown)


jakevdp:
Looks good. Regarding the ``Bin`` name change: we may be creating some aliases for this and other objects; see #454

PR #501, opened 2018-02-26
#457: Add Basic Trellis Histogram Example

hydrosquall:
Implements the vega-lite example from

https://vega.github.io/vega-lite/examples/trellis_bar_histogram.html

Note that the row labels will show up on the right instead of on the left (as they are in the vega-lite online example).


jakevdp:
Thanks!

PR #502, opened 2018-02-26
#457: Add Box Plot (Max/Min Whiskers) Example

hydrosquall:
Implements https://vega.github.io/vega-lite/examples/box-plot_minmax_2D_vertical_normalized.html .

A few open questions on style:

- Wasn't sure if there's a way to carry over the "create calculated fields and give them names" concept from the vega-lite spec, so I can rewrite this if there's a more idiomatic way to do this.

- The `size` parameter passed into the `mark_rule` and `mark_tick` methods appears to have no effect, as in the rendered chart, the widths of those items stays constant

- The `style` parameter appears to have no effect either (for setting rule style to `boxWhisker`).

### DIFF
--- a/altair/vegalite/v2/examples/boxplot_max_min.py
+++ b/altair/vegalite/v2/examples/boxplot_max_min.py
@@ -1,0 +1,47 @@
+"""
+Box Plot with Min/Max Whiskers
+-----------------
+This example shows how to make a basic box plot using US Population
+data from 2000.
+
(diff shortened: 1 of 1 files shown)


kanitw:
FYI, we will add "box-plot" compound mark the public API in Vega-Lite in the near future (1-3 months).  (Currently it's a hidden / experimental feature.) 

There is nothing wrong about adding this example in Altair using layer for now, but please watch Vega-Lite and be prepare to update / add another example with the compound mark version. :)

jakevdp:
Looks good, thanks!

PR #506, opened 2018-02-26
added the vega-lite example of natural disasters

afonit:


jakevdp:
Very nice! Keep them coming :smile:.

PR #516, opened 2018-02-26
added example from vega-lite: one dot per zipcode

afonit:
Note,
I had to have the row in here:
zipcodes = zipcodes[5000:10000]

to limit the amount of data or I get the error:
MaxRowsError: The number of rows in your dataset is greater than the max of 5000

and searching in the repo I don't see a way to temporarily lift that restriction.

### DIFF
--- a/altair/vegalite/v2/examples/one_dot_per_zipcode.py
+++ b/altair/vegalite/v2/examples/one_dot_per_zipcode.py
@@ -7,9 +7,6 @@ This example shows a geographical plot with one dot per zipcode.
 import altair as alt
 from vega_datasets import data
 
-states = alt.UrlData(data.us_10m.url,
-                     format=alt.TopoDataFormat(type='topojson',
-                                               feature='states'))
 zipcodes = data.zipcodes.url
 
 chart = alt.Chart(zipcodes).mark_circle(size = 3).encode(
--- a/altair/vegalite/v2/examples/one_dot_per_zipcode.py
+++ b/altair/vegalite/v2/examples/one_dot_per_zipcode.py
@@ -14,7 +14,7 @@ zipcodes = data.zipcodes.url
 
 # US states background
 background = alt.Chart(states).mark_geoshape(
-    fill='lightgray',
+    fill='white',
     stroke='white'
 ).properties(
     projection={'type': 'albersUsa'},
@@ -23,7 +23,7 @@ background = alt.Chart(states).mark_geoshape(
 )
 
 # Zip Codes labeled on background
-points = alt.Chart(zipcodes).mark_circle(size = 2).encode(
+points = alt.Chart(zipcodes).mark_circle(size = 3).encode(
     #alt.Text('city', type='nominal'),
     alt.X('longitude', type='longitude'),
     alt.Y('latitude', type='latitude'),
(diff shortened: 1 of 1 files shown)


jakevdp:
You could get around that by passing the data url rather than the dataframe, I think: replace ``zipcodes = data.zipcodes()`` with ``zipcodes = data.zipcodes.url``

jakevdp:
Also, I think it looks a bit better with ``size=5`` instead of ``size=1``, but I'm fine with you making that call if you feel otherwise.

eitanlees:
Also I think the plot is more impressive if you leave off the background states and only plot the points.

jakevdp:
> Also I think the plot is more impressive if you leave off the background states and only plot the points.

Yeah, that does look pretty cool!
vega 8

afonit:
@eitanlees , I agree it does look nice without the background - but in the spirit of replicating the vega-lite example, I have it in there.

@jakevdp I went ahead and put the .url after the data as you suggested and it works, but why?
I went to:
https://github.com/vega/vega-datasets/search?utf8=%E2%9C%93&q=url&type=
and searched on url, it only brought up two items as you can see in the search result - neither of which are methods.  If you get a sec could you explain for educational sake where that .url method is coming from.

afonit:
Is there a way to make this layered plot interactive,
I tired combinations of putting .interactive() on either background, points and both, nothing renders in those cases.

afonit:
@eitanlees , thanks for catching that, 
I was looking at two examples and had missed that the background was not on this example.

jakevdp:
The maximum lines issue is because if you make a chart with a dataframe, it actually embeds the data within the plot spec, which is then included in the notebook. (you can see this by calling ``chart.to_dict()`` on any chart made with a dataframe). When the dataset is too large, this can get very big very quickly, and doesn't perform well in the browser.

By contrast, if you specify the data by url, the dataset itself is not embedded in the spec, only the URL reference to the dataset: the dataset is then downloaded at render time, which is much more efficient in terms of displaying plots in the notebook.

This large dataframe thing presents some real usability issues (it's easy to accidentally make gigantic notebooks, particularly when building multiple plots). There are potential solutions down the line—vega-lite has the concept of datastores with named datasets that we may be able to take advantage of—but for now we've found that raising an error if the dataset is too big is the best way to keep users from accidentally shooting themselves in the foot.

jakevdp:
I've not seen any examples of interactive lat/lon plots. @kanitw, @domoritz, is it possible to have interactive scales here?

domoritz:
Not yet: https://github.com/vega/vega-lite/issues/3306

jakevdp:
Looks good. It would be simpler to put the ``projection`` properties directly on the point chart, and then you can remove the background layer.

altair/vegalite/v2/examples/one_dot_per_zipcode.py
```diff
@@ -0,0 +1,25 @@
...
+-----------------------
+This example shows a geographical plot with one dot per zipcode.
+"""
+
+import altair as alt
+from vega_datasets import data
+
+states = alt.UrlData(data.us_10m.url,
+                     format=alt.TopoDataFormat(type='topojson',
+                                               feature='states'))
```
jakevdp: states data is not used anymore

jakevdp:
Thanks!

PR #518, opened 2018-02-27
Change default width/height to config.view.width/height

### LINKED ISSUE #482: Remove default width/height from Chart?
Having defaults causes issues with faceted plots (which become comically large) and with things like ``rangeStep``, which are ignored if ``width`` and ``height`` are set (see #481).

I would propose removing these defaults.

The disadvantage is that straightforward charts end up being smaller than is perhaps ideal for display in the notebook.

jakevdp:
Closes #482

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -3,6 +3,8 @@ Utility routines
 """
 import re
 import warnings
+import collections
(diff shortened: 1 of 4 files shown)


jakevdp:
I'm going to merge & iterate.

PR #520, opened 2018-02-27
added vega-lite example: bar chart with highlight

afonit:
I have having difficulty moving the text around in the plot, if anyone has ideas.

Here is the json from the vega-lite example:
{
        "mark": {
          "type": "text",
          "align": "left",
          "dx": 215,
          "dy": -5
        }

here is what it should look like:
https://vega.github.io/vega-lite/examples/layer_bar_annotations.html


jakevdp:
You should be able to add these args to the text mark like this:

``mark_text(align='left', dx=215, dy=-5)``

afonit:
@jakevdp , humm, I could have sworn I did that and it threw errors.
I apologize for what ever mistake I made.
I will add that in now.

jakevdp:
Cool, glad it worked!

PR #521, opened 2018-02-27
added vega-lite example: mean overlay..bar line layered chart

afonit:
This is working and looks like:
https://vega.github.io/vega-lite/examples/layer_precipitation_mean.html

But, I can't seem to specify a size.
Obviously size = alt.Size(value = 3)) won't work because it requires a field.
In this example there is no field - we are using a calculated average and so we just need to set a static size.

I tried this: mark_rule(color = 'red', size = 3)
Which does not throw an error - but it also does not do anything (changing it to any number even 300).


jakevdp:
Try ``alt.SizeValue(3)``

jakevdp:
Looks good!

PR #523, opened 2018-02-27
Removed duplicate airports example

eitanlees:
There seemed to be two examples of the same plot `airports.py` and `us_airports.py`


jakevdp:
Oops, I should have caught that. Thanks!

PR #525, opened 2018-02-27
added vega-lite example: grouped bar chart

afonit:
Note, this uses a configuration line:
chart.config = {"view": {"stroke": "transparent"}, "axis": {"domainWidth": 1}}

Which is what was done in:
https://github.com/altair-viz/altair/pull/473/files

Jake, I did not know if you wanted this merged like you did the last one I just mentioned.  I am assuming that when these options show up in the api they will be here:
https://github.com/altair-viz/altair/blob/master/altair/vegalite/v2/api.py#L306
?

Asking as I am trying to dive a bit deeper and learn more.

### DIFF
--- a/altair/vegalite/v2/examples/grouped_bar_chart.py
+++ b/altair/vegalite/v2/examples/grouped_bar_chart.py
@@ -1,0 +1,21 @@
+"""
+Grouped Bar Chart
+-----------------------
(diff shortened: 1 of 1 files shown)


jakevdp:
Yep! I'm working on adding those additional methods today.

I think it's fine to merge this... when those APIs are available I'll do a grep through the examples and replace them as necessary

PR #528, opened 2018-02-27
Transform method

jakevdp:
This adds a ``transform_filter``, ``transform_calculate``, ``transform_bin``, etc. methods to top-level charts, so you can make a spec like this:

```python
import altair as alt

source = data.population()

    x='sum(people):Q',
    y='age:O'
).transform_filter("datum.year == 2000")
```
vega

### DIFF
--- a/altair/vegalite/v2/api.py
+++ b/altair/vegalite/v2/api.py
@@ -197,7 +197,7 @@ class TopLevelMixin(mixins.ConfigMethodMixin):
         """Return a MIME bundle for display in Jupyter frontends."""
         return renderers.get()(self.to_dict())
 
-    def add_transform(self, *transforms):
+    def transform_data(self, *transforms):
         copy = self.copy()
         if copy.transform is Undefined:
             copy.transform = list(transforms)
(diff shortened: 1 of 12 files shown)


jakevdp:
It might be interesting to have specific methods (with tab-completable call signatures) for each transform type. 

Something like:
```
chart.transform_filter("datum.year == 2000")
```

jakevdp:
Thoughts, @ellisonbg?

jakevdp:
``filter_transform`` might make more sense gramatically, but I went with ``transform_*`` because it's more easily discoverable via tab completion.

jakevdp:
Similarly added the ``resolve_axis``, ``resolve_filter``, ``resolve_scale`` methods.

jakevdp:
Regarding ``add_transform`` and ``set_resolve``... I think we could probably make them private. The reason they're different is because the ``transform`` property is a list of transforms, and calling ``add_transform`` appends to that list. On the other hand, calling ``set_resolve`` will overwrite any resolve hook that was there previously.

jakevdp:
Merged by rebase in 94892d3b726251d0

jakevdp:
It's a method that sets the filter attribute of the resolve... so

```python
Chart().resolve_filter('independent')
```
leads to the spec
```
{
  'resolve': {
    'filter': 'independent'
  }
}
```

PR #529, opened 2018-02-28
Interactive Examples

jakevdp:
This adds a few additional examples that make use of interactive features.


jakevdp:
These depend on #528

PR #534, opened 2018-02-28
Layered Histogram Example

eitanlees:
I know technically a layered bar chart is the same thing as a layered histogram but I thought this example was useful.

### DIFF
--- a/altair/vegalite/v2/examples/layered_histogram.py
+++ b/altair/vegalite/v2/examples/layered_histogram.py
@@ -1,0 +1,31 @@
+"""
+Layered Histogram
+=================
+This example shows how to make a layered histogram in Altair.
+"""
+
+import pandas as pd
+import altair as alt
+import numpy as np
+np.random.seed(42)
+
+# Generating Data
+df = pd.DataFrame({'Trial A':np.random.normal(0, 0.8, 1000),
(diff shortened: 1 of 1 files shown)


jakevdp:
Nice!

One thing: I think we should avoid using top-level config parameters in examples. These are a bit tricky, because once you do something like ``config_bar()``, you'll get an error if you try to use the chart within ``vconcat``, ``hconcat``, and ``layer``. I'm still thinking about how to inform the user of that (the current traceback is not all that helpful).

Unfortunately, it doesn't look like ``binSpacing`` can be set anywhere but the top-level config.

But there's an alternative: an area plot with ``interpolate="step"``. How would you feel about creating the plot this way instead?

```
chart = alt.Chart(df).mark_area(opacity=0.3, interpolate='step').encode(
    x=alt.X('Measurement',bin = alt.BinParams(maxbins=100)),
    y=alt.Y('count(*):Q', stack=None), 
    color=alt.Color('Experiment', 
                    scale=alt.Scale(range=['#0000ff', 
                                           '#008000', 
                                           '#ff0000'])
                   )
)
```

eitanlees:
Interesting. I didn't know about interpolate! I guess histograms can be made many ways.

jakevdp:
Looks good, thanks!

PR #548, opened 2018-03-02
Improve auto-generated docs

jakevdp:


jakevdp:
Changed the indent code from 80 to 100 chars, because some lines were really short.

PR #550, opened 2018-03-02
moved line to be behind the circles

afonit:
In reviewing this the line was on top of the dots and looked strange.
Sorry for missing this before.


jakevdp:
Looks good, thanks!

PR #553, opened 2018-03-05
Horizon Graph Example

eitanlees:
I made the Horizon Graph example from the vega-lite website. I had never heard of these plots before.


jakevdp:
Looks good, thanks!

PR #554, opened 2018-03-05
Add programmatic export utility based on selenium and chrome-headless

jakevdp:
This allows you to save altair charts to PNG without going to the browser.

Dependencies can be installed this way:

```
$ conda install selenium PIL
$ conda install -c clinicalgraphics chromedriver
```

Then you can use the utility this way:

```python
import alt
chart = alt.Chart(...)  # make your favorite chart here

from altair.utils import save_spec
save_spec(chart.to_dict(), 'mychart.png')
```

Eventually we can make this more streamlined for the user (e.g. ``chart.save('mychart.png')``) but we'll have to figure out the dependency story... and I wanted to get the initial version of this utility committed so that I can use it to build docs for the release candidate.

### DIFF
--- a/altair/utils/__init__.py
+++ b/altair/utils/__init__.py
@@ -3,3 +3,5 @@ from .core import *
 from .codegen import CodeGen
 
 from .plugin_registry import PluginRegistry
+
+from .headless import save_spec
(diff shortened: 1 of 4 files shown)


jakevdp:
This clearly could use some improvements, but I'm going to merge so that I get incorporate it into #552 and get the doc build working.

PR #556, opened 2018-03-06
Add savechart() method to vegalite/v2

jakevdp:


jakevdp:
This will eventually have to be refactored to be used with other schema versions (#557) but in the interest of fast progress I'm going to merge now 😄

PR #561, opened 2018-03-08
Add stem leaf plot example

eitanlees:
My attempt at a stem and leaf plot.

### DIFF
--- a/altair/vegalite/v2/examples/stem_and_leaf.py
+++ b/altair/vegalite/v2/examples/stem_and_leaf.py
@@ -9,22 +9,25 @@ import pandas as pd
 import numpy as np
 np.random.seed(42)
 
-# Generating Random Data
+# Generating random data
 original_data = pd.DataFrame({'samples':np.array(np.random.normal(50, 15, 100), dtype=np.int)})
 
-# Splitting Steam and Leaf
+# Splitting steam and leaf
 original_data['stem'] = original_data['samples'].apply(lambda x: str(x)[:-1])
 original_data['leaf'] = original_data['samples'].apply(lambda x: str(x)[-1])
 
-# Grouping Leafs for each Stem
-grouped_data = pd.DataFrame(columns=['stem', 'leaf'])
-for key, group in original_data.groupby('stem'):
-    grouped_data = grouped_data.append({'stem':key,
-                                        'leaf': ''.join(group['leaf'].sort_values())},
-                                        ignore_index=True)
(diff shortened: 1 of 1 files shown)


jakevdp:
I think you could do the grouping with a groupby, not sure if this is more clear or more obfuscated :smile:

```python
grouped_data = original_data.groupby('stem')['leaf'].apply(lambda leaves: ''.join(sorted(map(str, leaves))))
```

I wonder if there would be a better way to do this by actually encoding the x positions of the digits?

eitanlees:
haha, nice one liner! 

Obfuscation always creeps in the longer I work on a problem. It's the [curse of knowledge](https://en.wikipedia.org/wiki/Curse_of_knowledge)! 

With these examples it's sometimes tough to strike the balance between "Hey look at this cool thing I did" vs "This is a building block of visualization"

I will go back to the drawing board and think of how to encode the positions.

eitanlees:
I think calculating positions is better than just plotting a concatenated string. This is a first pass at trying to calculate the positions of the leaves.

jakevdp:
This is nice! Though I generally think it's clearer to use ``apply`` rather than looping over groups; something like this:

```python
get_position = lambda x: 1 + pd.Series(range(len(x)))
    
original_data['position'] = = original_data.groupby('stem')\
                                           .apply(get_position)\
                                           .reset_index(level=1, drop=True)
```

eitanlees:
I see, I haven't gotten comfortable with `apply` yet and often resort to loops.

altair/vegalite/v2/examples/stem_and_leaf.py
```diff
@@ -0,0 +1,34 @@
...
+
+# Determining leaf position
+get_position = lambda x: 1 + pd.Series(range(len(x)))
+    
+original_data['position'] = original_data.groupby('stem')\
+                                         .apply(get_position)\
+                                         .reset_index(drop=True)
+
+# Creating stem and leaf plot
+chart = alt.Chart(original_data).mark_text(align='left', baseline='middle', dx=-5).encode(
```
jakevdp: Formatting-wise, I'd break up the ``mark_text`` args:
```python
chart = alt.Chart(original_data).mark_text(
    align='left',
    baseline='middle',
    dx=-5
).encode(
    ...
```

jakevdp:
Looks pretty good!

jakevdp:
Looks good to me. If you're happy with it, I'll merge!

eitanlees:
👍

PR #564, opened 2018-03-09
MAINT: fix behavior of from_dict

jakevdp:
This is a maintenance fix for a subtle bug I came across. The issue was that when running ``from_dict``, altair would consider any subclass of ``SchemaBase`` as a potential wrapper for the generated class hierarchy.

What this meant is that if you, say, imported both vega-lite and vega wrappers, you could get cross-talk between them when using from_dict if they had common schemas.

This PR changes the auto-generated code so that each set of wrappers inherits from its own special base class, which ensures that this kind of cross-talk does not happen.


jakevdp:
Seems to work... I'm going to merge

PR #576, opened 2018-03-14
ENH: add binding interface

jakevdp:
Add a Python interface to bindings, plus an example:
ezgif-5-9d02b5a3cc

### DIFF
--- a/altair/vegalite/v2/api.py
+++ b/altair/vegalite/v2/api.py
@@ -146,6 +146,36 @@ def selection_single(**kwargs):
     return selection(type='single', **kwargs)
 
 
+@utils.use_signature(core.VgBinding)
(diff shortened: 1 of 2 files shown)


eitanlees:
I have a question. Each time we add an example could it affect the indices that are used to populate the docs homepage?

jakevdp:
Yes, at the moment that's the case. I'd like to change it so that it uses filenames rather than indices.

jakevdp:
Actually, what I'd really like is to change the ``mini-gallery`` extension so that it outputs a more dynamic gallery, similar to the top of the vega-lite page: https://vega.github.io/vega-lite/

jakevdp:
I'm going to merge this... I think it's all set.

PR #578, opened 2018-03-14
ENH: parse shorthand within alt.facet()

jakevdp:
This also makes ``facet()`` a method of ``Chart`` and ``LayerChart`` rather than a top-level function. While writing the docs for facet, it occurred to me that this is a more consistent API (it's similar to ``repeat``)


jakevdp:
I'm going to merge so I can keep iterating on docs :smile:

Issue #545, opened 2018-03-01
Lookup Transformation


eitanlees:
I was trying to make a [Choropleth example](https://vega.github.io/vega-lite/examples/geo_choropleth.html) and I am running into issues with the `LookupTransform`. 

I think my issue is that one of the attributes is 

```
    from : LookupData
        Secondary data reference.
```

but i can't use `from=alt.LookupData()` because it's a reserved word. 

Anyways, if someone can make the plot another way that would be cool.

jakevdp:
Ah, yeah that's an issue. You can get around it by doing something like this:

```
chart.transform_lookup(key=key, fields=fields, **{'from': ...})
```

But we should probably modify the ``transform_lookup`` method so that it takes a parameter called ``from_`` and builds the transform from there.

Feel free to try that modification and see if you can get the chart working!

eitanlees:
I have tried to make some modifications but I don't really know what I am doing. I think some more geo examples would be good for the gallery but most [choropleth examples](https://vega.github.io/vega-lite/examples/geo_choropleth.html) hinge on the lookup transformation.

jakevdp:
Within ``altair/vegalite/v2/api.py``, I would do something like changing
```python

    @use_signature(core.LookupTransform)
    def transform_lookup(self, *args, **kwargs):
        return self._add_transform(core.LookupTransform(*args, **kwargs))
```
to
```python

    @use_signature(core.LookupTransform)
    def transform_lookup(self, from_=Undefined, **kwargs):
        kwargs['from'] = from_
        return self._add_transform(core.LookupTransform(**kwargs))
```
it's not perfect, because tab completion will not work correctly for the function, but it would allow you to get going on the examples.

For a more long-term solution, we could add a decorator that will automatically make this kind of substitution for reserved keywords... I'd have to think about it a bit.

PR #580, opened 2018-03-14
Modify transform_lookup to take from_ argument

eitanlees:
I modified the `transform_lookup` method to take a `from_` parameter. I also included an example of a choropleth map using the method. 

This is in response to #545

### DIFF
--- a/altair/vegalite/v2/api.py
+++ b/altair/vegalite/v2/api.py
@@ -418,6 +418,8 @@ class TopLevelMixin(mixins.ConfigMethodMixin):
     @utils.use_signature(core.LookupTransform)
     def transform_lookup(self, from_=Undefined, **kwargs):
         if from_ is not Undefined:
+            if 'from' in kwargs:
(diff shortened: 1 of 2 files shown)


altair/vegalite/v2/api.py
```diff
@@ -416,8 +416,10 @@ def transform_filter(self, filter, **kwargs):
         return self._add_transform(core.FilterTransform(**kwargs))
 
     @utils.use_signature(core.LookupTransform)
-    def transform_lookup(self, *args, **kwargs):
-        return self._add_transform(core.LookupTransform(*args, **kwargs))
+    def transform_lookup(self, from_=Undefined, **kwargs):
+        if from_ is not Undefined:
+            kwargs['from'] = from_
```
jakevdp: Just to cover our bases, let's raise a ValueError if ``'from'`` is already in kwargs

```
if from_ is not Undefined:
    if 'from' in kwargs:
        raise ValueError("transform_lookup: both 'from_' and 'from' passed as arguments.")
    # ...
```

jakevdp:
Really nice, thanks!

PR #581, opened 2018-03-15
Add repeated choropleth example

eitanlees:
Another choropleth example from the vega-lite examples page. Weird dataset, but nice plot!

### DIFF
--- a/altair/vegalite/v2/examples/choropleth_repeat.py
+++ b/altair/vegalite/v2/examples/choropleth_repeat.py
@@ -1,0 +1,29 @@
+"""
+Repeated Choropleth Map
+=======================
+Three choropleths representing disjoint data from the same table.
+"""
+# category: geographic
+
+import altair as alt
+from vega_datasets import data
+
+pop_eng_hur = alt.UrlData(data.population_engineers_hurricanes.url)
+
+states = alt.UrlData(data.us_10m.url,
+                     format=alt.TopoDataFormat(type='topojson',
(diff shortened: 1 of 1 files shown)


jakevdp:
Nice! I don't really understand the dataset thought :smile:

One thing we might think about: since the ``URLData(..., TopoDataFormat(feature, ...))`` pattern is so common in these plots, it might be worth creating a shortcut for it in Altair. Since you've worked with a lot of geoshape data, do yo have any ideas for a concise function API that would cover common use-cases?

Something to think about... in the meantime I'll merge this example :smile:

eitanlees:
Something like this?

```
def UrlTopoFeature(url, feature):
    """
    Attributes
    ----------
    url : string
        An URL from which to load the data set.
     
    feature : string
        The name of the TopoJSON object set to convert to a GeoJSON feature collection. For 
        example, in a map of the world, there may be an object set named `"countries"`. 
        Using the feature property, we can extract this set and generate a GeoJSON feature 
        object for each country.
    """
    return alt.UrlData(url=url,format=alt.TopoDataFormat(type='topojson', feature=feature))
```

jakevdp:
Yes, something like that... maybe also catch other keywords and pass them along to ``TopoDataFormat``?

eitanlees:
Alright. Would I add this to the bottom of `api.py`? 
If so I can make a pull request to work out the details.

jakevdp:
That would be great!

PR #583, opened 2018-03-16
ENH: Add UrlTopoFeature to api

eitanlees:
Adding a function to address some of the boilerplate needed to extract features from a topojson url source.

### DIFF
--- a/altair/vegalite/v2/api.py
+++ b/altair/vegalite/v2/api.py
@@ -857,3 +857,21 @@ class FacetChart(TopLevelMixin, core.TopLevelFacetSpec):
         copy = self.copy()
         copy.spec = copy.spec.interactive(name=name, bind_x=bind_x, bind_y=bind_y)
         return copy
+
+
+def UrlTopoFeature(url, feature, **kwargs):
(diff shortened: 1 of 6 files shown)


altair/vegalite/v2/api.py
```diff
@@ -857,3 +857,21 @@ def interactive(self, name=None, bind_x=True, bind_y=True):
...
+    ----------
+    url : string
+        An URL from which to load the data set.
+     
+    feature : string
+        The name of the TopoJSON object set to convert to a GeoJSON feature collection. For 
+        example, in a map of the world, there may be an object set named `"countries"`. 
+        Using the feature property, we can extract this set and generate a GeoJSON feature 
+        object for each country.
+    """
```
jakevdp: here list:

```
**kwargs : 
    additional keywords passed to TopoDataFormat
```

jakevdp:
One more thing: I think since this is a function rather than a class, we should use a lowercase name to be consistent. How about ``def topo_feature``? That would also serve as a test for this.

jakevdp:
Looks good, thanks!

Issue #586, opened 2018-03-16
transform_bin requires 'as' as keyword


malramsay64:
When using the bin transformation as in the code below, the schema is expecting the keyword `as`, which is special, so setting `as='counted'`  will raise a SyntaxError.

Additionally the workaround used in the transform_compute of a trailing underscore, i.e. `as_='counted'` raises a schema validation error since it is not expecting the `as_` property.

When passing a dictionary of values as kwargs as demonstrated in the code below everything works, although I don't think the figure brings any new insights.

Doing a search of the repository it appears the `as_` trick only appears in the transform_compute so other transforms might also have this issue.

```
chart = alt.Chart(cars).mark_point().encode(
    x=alt.X('Horsepower', axis=alt.Axis(title='🐎⚡️'), bin=alt.Bin(maxbins=15)),
    y=alt.Y('counted:Q', axis=alt.Axis(title='# Cars')),
    color=alt.Color('Cylinders', legend=alt.Legend(title='# Cylinders')),
    size=alt.Size('Miles_per_Gallon', legend=alt.Legend(title='MPG'))
).transform_bin(
    **{'bin':True, 'field': 'Cylinders', 'as':'counted'}
)
```

jakevdp:
Thanks for the report!

The fix for this will be quite similar to #580. We should probably also look closely at other transforms for this issue as well...

If you'd like to take on a PR to fix this, that would be great! Otherwise I'll probably address it some time tomorrow.

malramsay64:
The other transforms also had this same issue except for the transform_calculate.

I also grepped the vega-lite v2 specification for properties that collided with python special keywords, with the only collisions not addressed in this issue or #580 being in Logical{Or,Not,And} and Selection{Or,Not,And}.

kanitw:
Btw,  IMHO we should also try to make code examples show transform before encode (because that's the right logical order) -- at least that's what we do in our example JSON :)


```
chart = alt.Chart(cars).mark_point()
  .transform_bin(
    **{'bin':True, 'field': 'Cylinders', 'as':'counted'}
  )
  .encode(
    x=alt.X('Horsepower', axis=alt.Axis(title='🐎⚡️'), bin=alt.Bin(maxbins=15)),
    y=alt.Y('counted:Q', axis=alt.Axis(title='# Cars')),
    color=alt.Color('Cylinders', legend=alt.Legend(title='# Cylinders')),
[...]

PR #589, opened 2018-03-16
Add as_ argument and handling to transform functions

malramsay64:
This addresses issue #586 

It turns out that all the transforms require the 'as' keyword argument [https://vega.github.io/vega-lite/docs/transform.html](https://vega.github.io/vega-lite/docs/transform.html).

I have added the 'as_' argument and error handling to be consistent with the changes in #580. I have also modified the transform_calculate function to make it as consistent as possible with the rest of the transforms.


jakevdp:
Looks really good – thanks for the contribution!

PR #591, opened 2018-03-16
DOC: Remove double listing of selections guide

eitanlees:


jakevdp:
Good catch!

jakevdp:
Since you seem to be looking through the docs, any thoughts overall? I went through last night with a clean conda environment and pretended I was a brand new user hitting the installation section, and tried to make it as seamless as possible. I'd love to hear any thoughts you have as you look through!

eitanlees:
I think the docs have been coming together nicely. Lots of information but laid out in an approachable way.

One thing I like to have on an installation page is a small example to verify my installation is working. Something I can quickly copy and paste. Maybe something like this 

```
import altair as alt

iris = data.iris()

alt.Chart(iris).mark_point().encode(
    x='petalLength', 
    y='petalWidth', 
    color='species'
)
```
along with the chart it should output. This way you know it's ok to move forward.

jakevdp:
Cool, I'll add that.

jakevdp:
done in b7d2388

PR #596, opened 2018-03-17
Add handing of as attribute to AggregatedFieldDef

### LINKED ISSUE #595: Collision of vega lite properties with python keywords
In #586 I was wrong that there were no additional collisions of concern, I found another in `AggregatedFieldDef`. 

There is also possibly an issue with `as` in transform_lookup [vega-lite docs](https://vega.github.io/vega-lite/docs/lookup.html#lookup-transform)

I am checking for collisions using the shell command below.

```
$ python -c "import keyword; print('\n'.join(keyword.kwlist))" | xargs -I{} grep '"{}":' -B 3 altair/vegalite/v2/schema/vega-lite-schema.json
```

malramsay64:
Fixes #595 for AggregatedFieldDef.

Additionally this adds the as_ attribute to the docstring.

### DIFF
--- a/altair/vegalite/v2/schema/core.py
+++ b/altair/vegalite/v2/schema/core.py
@@ -94,11 +94,17 @@ class AggregatedFieldDef(VegaLiteSchema):
         See the [full list of supported aggregation 
         operations](https://vega.github.io/vega-lite/docs/aggregate.html#ops) for more 
         information.
+    as_ : string
+        The output fields at which to write the aggregated values.
     """
     _schema = {'$ref': '#/definitions/AggregatedFieldDef'}
     _rootschema = Root._schema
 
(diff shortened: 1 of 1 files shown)


malramsay64:
I added `as_` as the last argument since that is less likely to break things, and that is the ordering in the specification. This does make it inconsistent with the transform_* functions which have `as_` as the first argument.

jakevdp:
Thanks!

If you look at the top of the file, you'll see a note that says that ``core/schema.py`` is automatically generated and should not be modified directly: next time we update the schema or anything else related to it, this change would be overwritten.

If this is a useful change from the user perspective, we need to re-define ``AggregatedFieldDef`` in ``api.py``.

Also, it would be nice to add an example where this is used to serve as a unit test.

malramsay64:
#598 is a more appropriate solution.

PR #607, opened 2018-03-19
DOC: the most minor of typos

phobson:
Ahead of the final release, I figured I could contribute by reviewing the docs with my somewhat naive eyes.

Here's the only change I can propose so far. 

Merge if now if you want, or I can continue going through the docs later tonight (US/Pacific TZ).

### DIFF
--- a/altair/vegalite/v2/examples/cumulative_wiki_donations.py
+++ b/altair/vegalite/v2/examples/cumulative_wiki_donations.py
@@ -2,7 +2,8 @@
 Cumulative Wikipedia Donations
 ==============================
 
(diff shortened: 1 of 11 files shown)


jakevdp:
Thanks! If you think you'll have more, I'll leave this open for a bit so they can all go on one PR.

phobson:
I'll holler back at you later tonight then. The docs really are so well done. It's not problem dig a little more.

phobson:
OK, this is as far as I'll be able to take this for a while. But I'm happy to address minor comment, or let you push to my branch.

jakevdp:
Sounds good – thanks!

PR #614, opened 2018-03-20
Allow chromedriver to work under the root user

elyase:
On systems running on root (typical for many docker/CI set ups) Chrome fails with:
```
[0320/133959.828643:ERROR:zygote_host_impl_linux.cc(90)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180
```

This option allows headless chrome to run in such cases.


jakevdp:
Thanks for the contribution!

PR #623, opened 2018-03-21
DOC: add discussion of data types and their effects on scales

jakevdp:


jakevdp:
cc/ @palewire

palewire:
I dig it. Thank you.

jakevdp:
I added a bit more in the "adjusting axis limits" section as well.

PR #627, opened 2018-03-21
DOC: Standardized import style in the basic charts examples

palewire:
Some of the basic charts are using a different import style than the other examples. It seems to me that consistency is important here, especially for basic charts that will be picked up by newbies.


jakevdp:
Looks good, thanks!

PR #637, opened 2018-03-22
Fix loading of schema on windows.

ajasja:
Import fails because schema is loaded with wrong encoding.

### DIFF
--- a/altair/vega/v2/schema/core.py
+++ b/altair/vega/v2/schema/core.py
@@ -9,7 +9,7 @@ import json
 def load_schema():
     """Load the json schema associated with this module's functions"""
     directory = os.path.dirname(__file__)
-    with open(os.path.join(directory, 'vega-schema.json')) as f:
+    with open(os.path.join(directory, 'vega-schema.json'), encoding='utf8') as f:
         return json.load(f)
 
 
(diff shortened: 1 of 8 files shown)


jakevdp:
``schema/core.py`` is automatically generated by ``tools/generate_schema.py``, so we can't modify it directly like this – the change would be overwritten next time the code is regenerated.

Please make the change in ``tools``, then run the script to generate the code. If you'd prefer me to do it, please let me know!

ajasja:
It should work now. I also committed the newly generated files.

jakevdp:
It's strange to me that the diff is so massive here... I'd have expected it to just change the one line you originally changed. Maybe windows sorts things differently than unix systems?

ajasja:
Does it matter which python version I used? The dicts are ordered in python
3 if I remember correctly. Perhaps you can try to generate the schema on
unix using the corrected  tools/generate_schema.py and see if there are any
differences?

jakevdp:
OK, I added #638 which only changes a few lines. I'm still confused as to why the diffs are so big when you run it...

Please let me know whether the changes there fix your problem!

jakevdp:
#638 is merged. I hope that fixes your issue!

ajasja:
It should, but I did not have time to test it yet (have to finally install
python 3.6 at work as well:)

PR #639, opened 2018-03-23
DOC: Update example and image shown in README.md

eitanlees:


jakevdp:
Good catch! Could you add something like this in the example code?
```python
# to use with Jupyter notebook (not JupyterLab) run the following
# alt.renderers.enable('notebook')
```

Issue #557, opened 2018-03-06
Refactor savechart()


jakevdp:
Currently, the core logic of ``savechart()`` is defined in ``altair/vegalite/v2/api`` (#556). This should be refactored to a more general utility that can be used by all versions of vega & vegalite wrappers

palewire:
I have another pedantic question. So thank you again in advance for your patience. Why not call this method something more concise like `.save()`?

jakevdp:
Seems fine. I'd be open to renaming & deprecating savechart.

PR #641, opened 2018-03-23
Refactored 'savechart' as the more concise 'save'

palewire:
As discussed in #557

### DIFF
--- a/altair/sphinxext/altairgallery.py
+++ b/altair/sphinxext/altairgallery.py
@@ -126,7 +126,7 @@ def save_example_pngs(examples, image_dir, make_thumbnails=True):
             _globals = {}
             exec(example['code'], _globals)
             chart = _globals['chart']
-            chart.savechart(image_file)
+            chart.save(image_file)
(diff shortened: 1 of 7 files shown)


jakevdp:
Looks good!

We should keep a ``savechart()`` method for the time being that raises a DeprecationWarning and then calls save().

palewire:
I *think* I can pull that off.

palewire:
I think I got that in with https://github.com/altair-viz/altair/pull/641/commits/ac121ad7c13c6161eb77fe374da613e1a6c458e8 but I haven't done one of these before so correct me where I screwed it up.

jakevdp:
Looks good – minor change, I'd say that savechart "is" deprecated, not that it will soon be...

Also, no need to test the deprecated version.

palewire:
Done in https://github.com/altair-viz/altair/pull/641/commits/492ffd08ada14cc4d85218de65450dc2e80c9da4

jakevdp:
Looks good, thanks!

Issue #649, opened 2018-03-25
save PNG/SVG in Windows


pagpires:
I encountered a problem saving PNG/SVG in Windows. I installed Selenium by 
`
conda install selenium
`
then downloaded chromedriver and put it under the system PATH variable (I think this is working). When I do chart.savechart('test.png'), there's problem message as follows:

```python-traceback
---------------------------------------------------------------------------
PermissionError                           Traceback (most recent call last)
<ipython-input-16-c16ecf6573c1> in <module>()
----> 1 chart.savechart('test.png')

G:\Software\Anaconda2\envs\py36\lib\site-packages\altair\vegalite\v2\api.py in savechart(self, fp, format, **kwargs)
    331             utils.write_file_or_filename(fp, spec_html, mode='w')
    332         elif format in ['png', 'svg']:
--> 333             utils.save_spec(self.to_dict(), fp, format=format, **kwargs)
    334         else:
    335             raise ValueError("unrecognized format: '{0}'".format(format))

G:\Software\Anaconda2\envs\py36\lib\site-packages\altair\utils\headless.py in save_spec(spec, fp, mode, format, driver_timeout)
    136                                                  spec, mode)
    137         finally:
--> 138             os.remove(htmlfile)
    139     finally:
    140         driver.close()

PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\PAGPI_~1\\AppData\\Local\\Temp\\tmpbyp9uhvp.html'
```

I suppose the temp html file is somehow being used in Windows so it's better to remove it after the chromedriver is closed? Let me know if you guys need further testing under Win environment

jakevdp:
Ah, I vaguely remember running into this problem several years ago when trying to use ``NamedTemporaryFile`` with windows in a similar context in the mpld3 project... it's something to do with Windows having some sort of lock on temporary files when they're used with a context manager. I think I ended up getting around it by writing a custom named temporary file routine from scratch.

Windows... always trying to be special.

jakevdp:
In the meantime, you can open figures in the browser and click "save to png" on the rendering.

jakevdp:
I don't have access to a windows machine to debug this, so if anyone else has time to take a look I'd appreciate it!

jakevdp:
Found it... this is a fix for essentially the same bug when it came up in another project: https://github.com/jakevdp/JSAnimation/commit/7eab0c23e237df707f02cfff3606eee8867e8665

[...]

PR #658, opened 2018-03-26
BUG: fix windows geteuid issue when saving to png/svg

jakevdp:
Fixes final issue in #649


jakevdp:
@pagpires I think once this is merged the ``save`` function should work on windows without an issue.

PR #663, opened 2018-03-27
Troubleshooting docs

jakevdp:
@mroswell I'd appreciate your thoughts on this... hopefully it would have addressed some of the issues you faced yesterday

### DIFF
--- a/altair/utils/plugin_registry.py
+++ b/altair/utils/plugin_registry.py
@@ -1,18 +1,6 @@
 from typing import Callable, Generic, List, TypeVar, Union, cast
-import textwrap
 
 import entrypoints
-
-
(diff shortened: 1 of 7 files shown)


jakevdp:
Note that I've left out the versions in the installation commands here... that's because these are the final docs that will be published once Altair version 2.0 is out. When I do another release candidate, I'll add the appropriate versions on the branch I use to build those docs.

jakevdp:
Just noticed that some plugin maintenance stuff snuck into this PR because I wasn't careful with my branches. It's main goal is to make the error message more useful when ``vega3`` is not installed.

jakevdp:
I'm going to merge this, to make it easier to continue iterating on docs.

Issue #662, opened 2018-03-27
Make examples work directly as copy-paste


jakevdp:
Currently the examples store the chart in the variable ``chart``... if a user copy-pastes the examples into a notebook, nothing will show up because the chart is never evaluated by the front-end.

I initially had a trailing ``chart`` at the end of each example, but found this to be rather ugly and annoying. I also worried that it might confuse a user who is trying to make a chart without saving it to a variable called ``chart``... we could add some sort of code comment to every file to explain this, but I couldn't think of anything that would be both succinct enough and informative enough.

I think the way I want to do this is to remove the current requirement that charts be stored in a variable called ``chart``, and instead just treat each example as if it's a normal execution cell.

The downside is that this will complicate things in unit testing, because we'll have to parse the example file and evaluate the last statement in it in order to extract the chart object, rather than just importing ``chart`` from the example.

cc/ @mroswell

jakevdp:
Closed by #665

PR #665, opened 2018-03-27
Change examples so that they work when directly pasted into notebook

jakevdp:
suggested by @mroswell 

This took a little doing:

- I moved the ``exec_then_eval`` utility from sphinxext into ``altair.utils``, renamed it to ``eval_block``, and added some simple tests
- I changed all examples so that the last statement of each evaluates to a chart object
- I updated example unit tests to use ``eval_block`` to properly extract example charts
- I updated sphinx doc generation to properly display the new examples

I generated the HTML pages locally, and everything seemed to work.

### DIFF
--- a/altair/sphinxext/altairgallery.py
+++ b/altair/sphinxext/altairgallery.py
@@ -19,6 +19,7 @@ from docutils.parsers.rst.directives import flag
 from sphinx.util.nodes import nested_parse_with_titles
 
 from .utils import get_docstring_and_rest, prev_this_next, create_thumbnail
+from altair.utils.execeval import eval_block
 from altair.vegalite.v2.examples import iter_examples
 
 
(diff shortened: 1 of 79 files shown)


jakevdp:
Fixes #662

mroswell:
I don't understand the details of what you've done, but I'm confused. Why not just add the word 
```chart``` at the bottom of each example?

jakevdp:
A few things:

- I had that in the first iteration of the docs, but I hated the aesthetics of it and so I removed it, assuming (probably wrongly) that users would read the docs and understand the package before just starting to copy and paste examples.
- Having ``chart`` hanging at the end would need some explanation. I tried and failed to write an explanation that was both complete enough to not confuse users, and succinct enough to not look weird at the end of every example.
- It would imply that the preferred way of creating a plot is to save it in a variable called ``chart`` and then to then type that variable at the end of the cell... I don't want to imply that.
- The larger change here aligns the style of the examples with the style used through the rest of the documentation.

PR #672, opened 2018-03-28
ENH: add no_limit data transformer

jakevdp:
This adds a built-in data transformer with no ``MaxRows`` limit.

I think writing to file is probably better, but I anticipate some users wanting the ability to easily enable the embedding of large datasets, and this gives them that.

Addresses #611.


jakevdp:
#758 is the alternative

PR #675, opened 2018-03-29
ENH: add selection detail example

jakevdp:


jakevdp:
Based on an example that @dhuppenkothen and I cooked up :smile:

PR #677, opened 2018-03-29
added remark based on mailing list discussion related to layered charts

ralienpp:
Previously I committed it to the wrong place.

### DIFF
--- a/doc/user_guide/compound_charts.rst
+++ b/doc/user_guide/compound_charts.rst
@@ -62,7 +62,13 @@ number of charts:
     ).interactive()
 
 The output of both of these patterns is a :class:`LayerChart` object, which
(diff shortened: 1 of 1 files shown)


jakevdp:
This looks good, but I wonder if it wouldn't be worth having a short section discussing this? For example, we could have an **Order of Layers** heading, and then show a two-panel chart demonstrating this. Something like

```
layer1 = alt.Chart(...

layer2 = alt.Chart(...

left = layer1 + layer2
right = layer2 + layer1

left | right
```

This would be much more useful to users than text alone.

Would you like to add something like that?

jakevdp:
I'm going to merge this, and work on adding the extra docs I mentioned above. Thanks for the contribution!

Issue #674, opened 2018-03-29
Mechanism to control renderer metadata


jakevdp:
It would be nice to have some mechanism to pass metadata to the renderer – i.e. a way to set the vega-embed [opts](https://github.com/vega/vega-embed#vega-embed-opt-specification-reference).

As far as API, I think this might make the most sense as part of the renderers object; for example
```
alt.renders.set_metadata(embed_opts={...})
```
though we'd have to think about how to best pass these options on to the various renderers.

jakevdp:
Related: #673

jakevdp:
Done in #776

PR #678, opened 2018-03-30
MAINT: modify renderers to always include metadata dictionary

jakevdp:
Prerequisite to addressing #674

### DIFF
--- a/altair/utils/display.py
+++ b/altair/utils/display.py
@@ -19,7 +19,7 @@ class Displayable(object):
     """A base display class for VegaLite v1/v2.
 
     This class takes a VegaLite v1/v2 spec and does the following:
-    
+
     1. Optionally validates the spec against a schema.
     2. Uses the RendererPlugin to grab a renderer and call it when the
        IPython/Jupyter display method (_repr_mimebundle_) is called.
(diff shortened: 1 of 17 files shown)


jakevdp:
My thought is to have the metadata dictionary recognize a couple keys:

- ``embed_opts`` that are passed to vega-embed
- ``tooltip_opts`` that are passed to vega-tooltip
- ``urls``, which maps library names (e.g. vega, vega-lite, vega-embed, vega-tooltip) to the URL from which the library should be loaded

And other keys available if any custom renderers want to use them.

It makes sense to me that metadata should not be set on a chart-by-chart basis, but rather on a global basis; for example

```
alt.renderer.set_options(embed_opts = {...}, tooltip_opts = {...}, ...)
```

and at that point all plots created in the runtime would use the specified metadata.

jakevdp:
Thinking about this more... I'm not certain whether this is the best model.

Basically, the question is whether metadata settings should be a property of the chart, a property of the package, a property of the renderer, or something else entirely. I'm not certain.

jakevdp:
From offline discussions, we decided that the metadata should be a property of the renderer.

Still trying to consider the best API for this... it's similar in spirit to #758

PR #680, opened 2018-03-30
MAINT: add explicit utf8 coding for generated files.

jakevdp:
These files contain unicode characters, because those characters are present in the vega
and vega-lite schemas. Python 3 defaults to UTF-8, but to make the library more broadly
compatible, it's useful to be explicit.


jakevdp:
cc/ @craigcitro that turned out to be pretty easy :smile:

PR #683, opened 2018-03-30
MAINT: use pkgutil rather than open() to access internal data files

jakevdp:
This is necessary to work properly within Google's internal Colab deployment.

### DIFF
--- a/altair/utils/display.py
+++ b/altair/utils/display.py
@@ -1,5 +1,6 @@
 import json
 import os
+import pkgutil
 from typing import Callable, Dict, Union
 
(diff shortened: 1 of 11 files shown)


jakevdp:
cc/ @craigcitro

craigcitro:
Thanks!

And 😭 on the need to decode the result for python 3.5, but happy about https://docs.python.org/3/whatsnew/3.6.html#json.

jakevdp:
Yep... just installed a py35 environment to test this locally...

jakevdp:
Looks good 😎

Issue #689, opened 2018-03-31
How to reverse the sort order of stacked bars


palewire:
Here's my draft chart.

```python
alt.Chart(latino_chart_prepped, title="Latino share of U.S. population").mark_bar().encode(
    alt.X('age:O', axis=alt.Axis(title="Age"), sort=alt.SortField(field="sort", op="sum", order="ascending")),
    alt.Y('population:Q',
        axis=alt.Axis(title='Percent of population'),
        stack='normalize'
    ),
    alt.Color('group:N', legend=None)
).properties(
    width=600
)
```

![visualization 2](https://user-images.githubusercontent.com/9993/38165591-a29aaa02-34ca-11e8-9a56-50ebfbff45b0.png)

How do I reverse the sort order so that the smaller blue bars are on the bottom of each stack in the chart?

jakevdp:
Add an ``order`` channel:
```
alt.Chart(...).encode(
    ...
    order=alt.Order('group', sort='ascending')
    ...
)
```

You may need to use ``sort='descending'`` instead... without seeing your data, it's difficult to guess.

palewire:
Thank you. That worked. Are there any examples of this in the documentation? If not, how about I add some?

jakevdp:
Nope, I'm still working on getting things like that in the docs. It's a long process.

Some examples would be useful... we also need an actual in-depth discussion of all the encodings that are available and what they are good for.

palewire:
I've submitted a quick addition of an example to #681. 

It seems to me that an explanation of the `order` option in the configuration chapter might require a fuller set of examples. Here's the ones I can think of:

* Ordering stacked bar or area chart components
* Ordering the members of a bar chart on the x or y axis (depending on if it's a horizontal or vertical bar chart)

What else?

jakevdp:
ordering of points in a line as well. I used that in an example here: https://github.com/altair-viz/altair/pull/690/commits/2098f8de3b41c794db0dae7d0076bec3151d81e0

jakevdp:
I think you mean #691 :smile:

jakevdp:
Documentation added in #693

PR #691, opened 2018-04-01
Added an alt.Order config to the horizontal stacked bar chart example.

palewire:
For #689


jakevdp:
I'm not certain this is the best place to demonstrate this... it's not clear from the example alone what the order accomplishes, and on the surface it makes it look like defining order is necessary to create a stacked bar chart.

I think a better way to address #689 would be to add a section to the encoding documentation discussing the effect of the ``order`` channel in various circumstances.

palewire:
Got it. I withdraw the pull request and will go back to the drawing board.

PR #693, opened 2018-04-01
Drafted encoding documentation for the order channel

palewire:
For #689.

### DIFF
--- a/altair/vegalite/v2/examples/connected_scatterplot.py
+++ b/altair/vegalite/v2/examples/connected_scatterplot.py
@@ -18,8 +18,7 @@ lines = alt.Chart(driving).mark_line().encode(
 
 points = alt.Chart(driving).mark_circle().encode(
     alt.X('miles', scale=alt.Scale(zero=False)),
-    alt.Y('gas', scale=alt.Scale(zero=False)),
-    order='year'
+    alt.Y('gas', scale=alt.Scale(zero=False))
 )
 
 lines + points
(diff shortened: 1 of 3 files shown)


altair/vegalite/v2/examples/connected_scatterplot.py
```diff
@@ -18,8 +18,7 @@
 
 points = alt.Chart(driving).mark_circle().encode(
     alt.X('miles', scale=alt.Scale(zero=False)),
-    alt.Y('gas', scale=alt.Scale(zero=False)),
-    order='year'
```
jakevdp: The order is really important here, because it makes the line connect points in temporal order. We should leave this in here.

jakevdp: Ah, strike that, I see what you did. Order is not necessary for points :smile:

palewire: Glad to hear I'm not crazy!

doc/user_guide/encoding.rst
```diff
@@ -349,4 +349,83 @@ Shorthand            Equivalent long-form
 ``x='sum(name)'``    ``alt.X('name', aggregate='sum')``
 ``x='sum(name):Q'``  ``alt.X('name', aggregate='sum', type='quantitative')``
 ``x='count():Q'``    ``alt.X(aggregate='count', type='quantitative')``
+``order='name'``     ``alt.Order('name')``
```
jakevdp: This table is about translation of shorthand form to full arguments, so adding order here is not really necessary.

palewire:
I've removed the order shorthand example, as you requested.

jakevdp:
This is great, thanks for the contribution!

PR #694, opened 2018-04-01
Removes trailing whitespaces from a set of files

palewire:
There are a series of other small PEP8 violations hanging around the library that could use a little housecleaning, IMHO. If you're open to it, I'd be happy to use `flake8` to tidy up the rest.

Ultimately the library could use flake8 to enforce Python style as part of the tests on TravisCI. In my experience as a package maintainer this has been a positive thing because, as pedantic as PEP8 is, it provides a common foundation that can bring some consistency to a large library and quickly settle subjective debates about style.

### DIFF
--- a/altair/vega/v2/schema/core.py
+++ b/altair/vega/v2/schema/core.py
@@ -21,7 +21,7 @@ class VegaSchema(SchemaBase):
 
 class Root(VegaSchema):
     """Root schema wrapper
-    
+
     allOf(container, Mapping(required=[]))
     """
     _schema = load_schema()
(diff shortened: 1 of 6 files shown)


jakevdp:
If you look at the top of all these files, you'll see a note that they're automatically generated by the scripts in ``tools/generate_schema_interface.py``. So any changes made directly to the files will be overwritten next time they are generated.

The corresponding changes can be made in the code generation script if you'd like them to persist.

jakevdp:
Regarding running flake8 as a part of tests... I'd be open to it, but with so much of the package programmatically generated and drawn directly from the vega and vega-lite schemas, I've found it difficult to always produce perfectly PEP8-compatible code.  But if you want to tackle that, feel free :smile:

palewire:
Thanks you for the education. Apologies for the ignorant patch!

jakevdp:
Don't worry about it... I'm just happy for your contributions!

PR #697, opened 2018-04-02
 Clean up rstrip and blank line PEP* violations in autogen code

palewire:
Also handled upstream to make the fix permanent.


jakevdp:
Looks good, thanks!

PR #701, opened 2018-04-03
MAINT: tests & code cleanup

jakevdp:


jakevdp:
Ugh... doctests have a lot of dict output whose key order is not deterministic.

I'm going to close this, and just do the cleanup commits in #702

PR #711, opened 2018-04-04
Fixed a variety of PEP8 errors in the tools folder

palewire:


jakevdp:
Same comments as in #712

PR #712, opened 2018-04-04
PEP8 fixes in the expr module

palewire:


jakevdp:
Honestly, I'm kind of against line breaks for the sake of appeasing PEP8. It makes the code harder to read in my opinion.

I also thing that peppering code with nonsense comments meant to stop PEP8 from complaining is an anti-pattern: it's confusing for users who don't know what ``# noqa`` is supposed to mean, and it's solving the wrong problem: the problem is not in the code, the problem is that the PEP8 tool is over-zealous.

That problem should not be fixed by changing the code. The problem should be fixed by changing the linting tool.

palewire:
I totally hear you. Who could agree with 100% of these rules!

From my point of view, having an authoritative source to lean on can help referee and prevent subjective debates in the future about code style. This has helped me manage projects with more than 150 outside contributors without losing my mind. It can also enforce consistency across the library in a way that makes ultimately will make it more readable for maintainers, avoiding different large sections written in different style.

Most important: Many of the rules verified by flake8 are important. It can find things a test might miss, like undeclared variables, missing imports and gnarly cruft. 

I've been taking a piecemeal approach to these pull requests to avoid some drastic rewrite led by the rulebook. It allows for gradually shaking out which ones you want to follow, and which ones it's okay to ignore. I appreciate your patience in dealing with all of them. 

From your reaction to the two latest patches, it seems to me there may be two rules that don't interest you, and could perhaps be ignored at the global level in the `setup.cfg` file. They are:

* Line length
* Unused and undeclared (i.e. `*`) imports in __init__.py files.

If that does match your feelings, I propose withdrawing this pull request and submitting a new one that would instruct flake8 to ignore those issues globally with the minimal amount of code.

That way, you don't have bend to rules you disagree with, but your library doesn't miss the benefits of having not just a code referee, but also a useful safety check to prevent errors and cruft.

jakevdp:
Maybe we could compromise and run pyflakes rather than flake8? That would find all the actual issues within the code, without the need to delve into silly rules about spaces at the end of lines, etc.

palewire:
I'm happy to give that a run if it fits your vision for this project.

I work in a profession where we are trained to be pedantic, so the rigid rulebook appeals to me. But I also got started programming with Perl, so I try to always remember its lesson that [TMTOWDTI!](https://en.wikipedia.org/wiki/There%27s_more_than_one_way_to_do_it).

jakevdp:
I think PEP8 can be useful (I generally write code that is more or less pep8 compliant) but I think spending too much time worrying about it is counter-productive. I like Raymond Hettinger's perspective on the subject: https://www.youtube.com/watch?v=wf-BqAjZb8M

palewire:
I'm withdrawing this to soon replace it with some pyflakes work.

PR #719, opened 2018-04-06
DOC: Correct scale on horizon example

eitanlees:
I think the scale needs to be set in the second area for this to work.


jakevdp:
Thanks!

PR #723, opened 2018-04-07
Fixed a number of imports that violate pyflakes rules

palewire:
This includes a configuration of `flake8` that will ignore all style rules derived PEP8 and only enforce rules brought in from pyflakes.


jakevdp:
Thanks

PR #727, opened 2018-04-08
PyFlakes fixes for Python 2/3 compatibility code

palewire:
I also trimmed out some of the compatibility module that wasn't being used, or that is already provided by the `six` library.


jakevdp:
Thanks!

PR #729, opened 2018-04-08
PyFlakes fixes to __init__.py files

palewire:
Attached are a variety of fixes to manage PyFlakes violations in ``__init__.py`` files.

My reading of the library is that ``*``'s have been used in autogenerated modules to allow for a changing number of imports. To allow that to continue, I have exempted those files from PyFlakes inspection. If you'd prefer another approach, please let me know.


jakevdp:
Thanks

PR #733, opened 2018-04-09
Fixed almost all of the remaining PyFlakes violations

palewire:
There is one separate issue yet that I will soon file a ticket on. But besides that, this is all that I see remaining in my pyflakes checks.


palewire:
The other issue I mentioned is now documented in #734.

jakevdp:
Thanks!

PR #737, opened 2018-04-10
ENH: Add Firefox option for save

### LINKED ISSUE #714: Add Firefox option for figure export
Firefox works for Altair figure export via [geckodriver](https://github.com/mozilla/geckodriver/releases). I verified it works with `savechart` after a literal "s/Chrome/Firefox" & "s/chrome/firefox" in `altair/utils/headless.py`. Would you accept a PR that adds Firefox as an option?

mihaic:
Using `options` instead of `chrome_options` follows the deprecation
advice in Selenium.

According to its docs, Firefox takes single dash options, but I verified
that it understands `--headless`.

I tested as root and Firefox does not need `--no-sandbox`, which is
convenient because it does not have such a flag.

Fixes #714.

### DIFF
--- a/altair/utils/headless.py
+++ b/altair/utils/headless.py
@@ -12,11 +12,6 @@ try:
 except ImportError:
     webdriver = None
 
-try:
-    from selenium.webdriver.chrome.options import Options as ChromeOptions
-except ImportError:
-    ChromeOptions = None
-
 
 @contextlib.contextmanager
 def temporary_filename(**kwargs):
@@ -102,7 +97,8 @@ def spec_to_image_mimebundle(spec, format, mode,
                              vega_version,
                              vegaembed_version,
                              vegalite_version=None,
-                             driver_timeout=10):
+                             driver_timeout=10,
+                             webdriver_class=None):
     """Conver a vega/vega-lite specification to a PNG/SVG image
 
     Parameters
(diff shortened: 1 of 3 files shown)


altair/utils/headless.py
```diff
@@ -122,6 +118,8 @@ def spec_to_image_mimebundle(spec, format, mode,
     driver_timeout : int (optional)
         the number of seconds to wait for page load before raising an
         error (default: 10)
+    webdriver_class : Type[Union[webdriver.Chrome, webdriver.Firefox]]
+        Webdriver to use (default: webdriver.Chrome).
```
jakevdp: I think it would be cleaner to let this argument be a string (``webdriver='chrome'`` or ``webdriver='firefox'``), and then import the associated webdriver class below.

altair/utils/headless.py
```diff
@@ -130,8 +128,8 @@ def spec_to_image_mimebundle(spec, format, mode,
 
     Note
     ----
-    This requires the pillow, selenium, and chrome headless packages to be
-    installed.
+    This requires the pillow, selenium, and chromedriver or geckodriver
+    packages to be installed.
```
jakevdp: Please make this more explicit: something like "this requires the pillow and selenium Python modules to be installed. Additionally it requires either chromedriver (if webdriver=='chrome')  or geckodriver (if webdriver=='firefox')"

PR #740, opened 2018-04-10
ENH: add theme support

jakevdp:
This addresses #718, adding support for switching between different chart themes.

For example, you can do

```python
import altiar as alt
alt.theme.enable('opaque')
```

and charts will now have a white background. Restore the default theme using

```python
alt.theme.enable('default')
```

Register custom themes using package entrypoints, or using, e.g.
```python
alt.theme.register('black-background', {'config': {'background': 'black'}})
```
and then enable it using
```python
alt.theme.enable('black-background')
```

@afonit – I'd love your input on what other themes we should make available by default. Currently I have "default" (set width and height) "opaque" (set width & height, and background color to white) and "none" (no config settings).

### DIFF
--- a/altair/vegalite/tests/test_common.py
+++ b/altair/vegalite/tests/test_common.py
@@ -58,3 +58,24 @@ def test_basic_chart_from_dict(alt, basic_spec):
 
     # remainder of spec should match the basic spec
     assert dct == basic_spec
+
+
+@pytest.mark.parametrize('alt', [v1, v2])
+def test_theme_enable(alt):
+    active_theme = alt.theme.active
+
+    try:
+        alt.theme.enable('none')
+
+        chart = alt.Chart.from_dict(basic_spec)
+        dct = chart.to_dict()
+
+        # schema should be in the top level
+        assert dct.pop('$schema').startswith('http')
+
+        # remainder of spec should match the basic spec
+        # without any theme settings
+        assert dct == basic_spec
+    finally:
+        # reset the theme to its initial value
+        alt.theme.enable(active_theme)
--- a/altair/vegalite/v1/api.py
+++ b/altair/vegalite/v1/api.py
@@ -14,6 +14,7 @@ from .schema import core, channels, Undefined, SCHEMA_URL, SCHEMA_VERSION
 from .data import data_transformers, pipe
 from ... import utils
 from .display import renderers
+from .theme import theme
 
 
 VEGALITE_VERSION = SCHEMA_VERSION.lstrip('v')
@@ -46,7 +47,6 @@ class Formula(core.Formula):
 # *************************************************************************
 
 class TopLevelMixin(object):
-    _default_spec_values = {"width": 400, "height": 300}
     _class_is_valid_at_instantiation = False
 
     def _prepare_data(self):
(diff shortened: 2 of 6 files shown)


jakevdp:
Thanks for the feedback!

I think I'm going to merge this so I can rebase #741 and fix the issues there that this will cause.

I opened #742 to track dark theme discussions.

kanitw:
These themes seem useful -- I wonder if it's worth putting these in Vega instead so all users of the Vega ecosystem can benefit from these options.  

cc: @jheer @arvind @domoritz

rpalloni:
Mmm...I get an AttributeError. I downloaded the latest github version few minutes ago.
themeerror

jakevdp:
Two things:

- it's ``alt.theme``, not ``alt.themes``
- ``'WhitTheme'`` is not a valid theme

Try
```python
alt.theme.enable('opaque')
```

rpalloni:
alt.theme worked. WhiteTheme is the one I define

jakevdp:
Now that you say this, though, I think it probably should be ``alt.themes`` rather than ``alt.theme``... all other plugin registries are plural (``alt.renderers``, alt.data_transformers``...)

rpalloni:
Discovered another issue. For some reason, DataFrame inputs produce empty charts whereas json inputs are ok.
Example: with stocks = pd.read_csv(data.stocks.url) only an empty plot with axes appears

jakevdp:
OK, #756 renames ``alt.theme`` to ``alt.themes``.

Since it's still a prerelease and the feature hasn't been documented beyond this issue, I opted not to provide any deprecation cycle on ``alt.theme``.

jakevdp:
> Discovered another issue. For some reason, DataFrame inputs produce empty charts whereas json inputs are ok.
> Example: stocks = pd.read_csv(data.stocks.url) only an empty plot with axes appears

The issue if you do this is that the date column is not automatically converted to the appropriate type. If you use ``data.stocks()``, you'll get a dataframe with the types handled appropriately. See [vega_datasets](https://github.com/altair-viz/vega_datasets#usage) for more info on this API.

domoritz:
I left a comment on vega-themes in https://github.com/altair-viz/altair/issues/742#issuecomment-381329290. I think it would be nice to collect all themes in one place.

rpalloni:
As you see, dataframe are not working.
dt = data.stocks.url is working ok

issue

jakevdp:
@rpalloni – I don't understand. Is this related to the themes issue?

rpalloni:
No, another issue -> guide reference: https://altair-viz.github.io/user_guide/troubleshooting.html#general-trouble-shooting

Anyway I found the solution: Chart() does not get a DataFrame as described in the guide but a **url** to dataframe. 
But this is not handy as, while doing analysis, you usually load data on top, go through some stats and calcualtions and than visulize them.

Basically, you cannot use the DataFrame object created with pd.read_csv into Chart().

jakevdp:
@rpalloni – please open a new issue if you have an unrelated question: this issue is about discussing theme support.

Regarding your question: ``Chart()`` can accept either a dataframe or a URL. If you've found a bug with that, please open a new issue with code that will reproduce the error you're seeing.

Thanks!

rpalloni:
Ok. #765

Casyfill:
just to be sure - is it possible to use vega-lite themes (like vox and 538?)

jakevdp:
No, not currently. Theme support is controlled by the frontend, and I don't think any of the available frontends support themes yet.

domoritz:
> I don't think any of the available frontends support themes yet

Themes are built into Vega-Embed so they should work in the notebook at JupyterLab already.

PR #752, opened 2018-04-12
Attempt to fix code block that's not rendering in the docs

palewire:
Looks like this right now:

screenshot from 2018-04-12 13-08-14


jakevdp:
Thanks! That's one solution, the shorter way to fix it is to have two colons on the preceding line. I think that was where I made the typo :smile:

PR #753, opened 2018-04-12
Typing is in requirements.txt but not setup.py

palewire:
One of my colleagues installed Altair's new release candidate in Python 2.7 today. He had to install typing himself manually afterwards. I suspect this is because it was not added to setup.py.


jakevdp:
Good catch, thanks!

Issue #754, opened 2018-04-13
How to set the axis range on histograms, shared with other subplots?


mtzl:
Hi, I am trying to make a facetted scatterplot with marginal histograms (similar to the one in https://altair-viz.github.io/user_guide/compound_charts.html?highlight=bin%20scale#horizontal-concatenation), but I can't figure out how to make the scatterplot share its x/y limits with the histograms.

This is what I've got so far:
```python
import altair as alt
from vega_datasets import data

iris = data.iris()

xrange = (3, 9)
yrange = (1, 6)

points = alt.Chart(iris).mark_circle().encode(
    alt.X('sepalLength', scale=alt.Scale(domain=xrange)),
    alt.Y('sepalWidth', scale=alt.Scale(domain=yrange)),
    color='species',
)

top_hist = alt.Chart(iris).mark_area(
    opacity=.4, interpolate='step'
).encode(
    alt.X('sepalLength:Q', 
          bin=alt.Bin(maxbins=20, extent=xrange), 
          stack=None, 
          scale=alt.Scale(domain=xrange),
         ),
    alt.Y('count(*):Q', 
          stack=None, 
         ),
    alt.Color('species:N'),
).properties(height=60)

right_hist = alt.Chart(iris).mark_area(
    opacity=.4, interpolate='step'
).encode(
    alt.Y('sepalWidth:Q', 
          bin=alt.Bin(maxbins=20, extent=yrange), 
          stack=None,
          scale=alt.Scale(domain=yrange),
         ),
    alt.X('count(*):Q', 
          stack=None, 
         ),
    alt.Color('species:N'),
).properties(width=60)

chart = top_hist & (points | right_hist)
```
which yields this (axis limits on the scatterplot are very wide to illustrate the effect better): 
![](https://gist.githubusercontent.com/mtzl/df0ef1a9017a5b577db7d2923b1c12f5/raw/179c2253d7cd6ee2e48890b1c5d5242419f487a1/scatter_marginal_hist.png)

How can I make the marginal histograms share the same x, y axis range `3-9` and `1-6`?

jakevdp:
The spec you used should behave the way you expect, but it does not due to a bug in Vega-Lite: https://github.com/vega/vega-lite/issues/3575

mtzl:
Okay thanks! 

Btw I love how easy this is in altair, compared to seaborn!

jakevdp:
I was just going to ask you about that :smile:

jakevdp:
[...]

PR #759, opened 2018-04-13
DOC add example with facetted scatterplot plus marginal histograms

mtzl:
A facetted scatterplot with marginal facetted histograms.

Beware this is affected by #754 (the binning `extent` does not set the axis scale as intended), so this example should probably not be considered before that bug is fixed in vega-lite.


altair/vegalite/v2/examples/scatter_marginal_hist.py
```diff
@@ -0,0 +1,53 @@
...
+from vega_datasets import data
+
+iris = data.iris()
+
+# We do not want to specify the exact axis manually,
+# and ``alt.Bin`` expects a range, rather than a ``alt.Scale`` instance.
+xscale = alt.Scale(padding=1, zero=False)
+yscale = alt.Scale(padding=1, zero=False)
+xlim = xscale.domain
+ylim = yscale.domain
```
jakevdp: I don't think this works the way you hope. at this point, ``scale.domain`` is ``Undefined``, which means it's the same as not providing ``xlim`` and ``ylim`` at all.

mtzl: Oh whoops, I assumed this to be a dynamic property, fixed

altair/vegalite/v2/examples/scatter_marginal_hist.py
```diff
@@ -0,0 +1,51 @@
...
+    alt.Y('sepalWidth', scale=yscale),
+    color='species',
+)
+
+top_hist = alt.Chart(iris).mark_area(**area_args).encode(
+    alt.X('sepalLength:Q',
+          # when using bins, the axis scale is set through
+          # the bin extent, so we do not specify the scale here
+          # (which would be ignored anyway)
+          bin=alt.Bin(maxbins=20, extent=xlim),
```
jakevdp: You could shorten things and do ``extent=xscale.domain`` rather than storing it in the separate ``xlim`` variable.

altair/vegalite/v2/examples/scatter_marginal_hist.py
```diff
@@ -0,0 +1,51 @@
...
+          stack=None,
+          axis=blank_axis,
+         ),
+    alt.Y('count(*):Q', stack=None, axis=blank_axis),
```
jakevdp: It's recommended now to just use ``count()`` rather than ``count(*):Q``. The ``*`` field marker is deprecated in Vega-Lite, and Altair is now smart enough to use quantitiative types for count by default.

jakevdp:
Looks good! A couple more comments inline. I think we should wait until the vega-lite bug is fixed before we merge this.

mtzl:
Adjusted, thanks!

jakevdp:
Still some issues in the frontend, but I'm going to merge this to get it into the 2.0 release. It's a great example – thanks!

Issue #768, opened 2018-04-23
Split requirements into production and dev requirements


sebastianneubauer:
currently there are typical dev requirements (e.g. pytest, docutils and ipython) in the install_requires. That means one cannot install altair without installing those development requirements which is suboptimal or even sometimes not possible in production environments.  

A best practice would be to split the requirements into the minimal set of needed requirements and additional development extra requirements.

sebastianneubauer:
I created a PR #769 which introduced a set of extra requirements for development environments. In production environments, e.g. in a backend server one can install via:
`pip install altair` 
In a development environment one can install additionally the dev requirements with:
`pip install -e .[dev]`

There is one issue remaining: To me, ipython should be a development requirement, but currently as it seems, altair is failing without ipython installed. So my suggestion would be to make ipython imports optional, e.g. something like this 
```
try:
    import ipython
except ImportError:
    NO_IPYTHON=True
```
What are your opinions?

jakevdp:
I think making the IPython import optional is fine.

The only place where that would be nontrivial is in ``altair/_magics``, because the decorators are called when the submodule is imported.

That said, I don't see any reason that ``altair._magics`` would need to be imported except when IPython is available, so we're probably fine not changing anything there.
The other places that IPython is imported, a ``try``/``except`` should be fine.

sebastianneubauer:
resolved by #769

PR #769, opened 2018-04-23
splitted requirements into production and dev requirements

sebastianneubauer:
It is important to split requirements into "hard" requirements for production and dev requirements which are only needed in development environments (e.g. pytest and ipython). 

This addresses #768.

One remaining issue is, that ipython seems to be a hard non-optional requirement. Is this really needed or could this be made optional (e.g. try import ipython, if not there ignore)?

### DIFF
--- a/altair/_magics.py
+++ b/altair/_magics.py
@@ -6,6 +6,8 @@ __all__ = ['vega', 'vegalite']
 import json
 import warnings
 
+import IPython
+from IPython.core import magic_arguments
 import pandas as pd
 import six
 from toolz import pipe
@@ -15,13 +17,6 @@ from altair.vegalite import v2 as vegalite_v2
 from altair.vega import v2 as vega_v2
 from altair.vega import v3 as vega_v3
 
-try:
-    import IPython
-    from IPython.core.magic_arguments import (magic_arguments, argument, parse_argstring)
-    IPYTHON_AVAILABLE=True
-except ImportError:
-    IPYTHON_AVAILABLE=False
-    
 try:
     import yaml
     YAML_AVAILABLE = True
@@ -80,13 +75,13 @@ def _get_variable(name):
     return ip.user_ns[name]
 
 
-@magic_arguments()
-@argument(
+@magic_arguments.magic_arguments()
+@magic_arguments.argument(
     'data',
     nargs='*',
     help='local variable name of a pandas DataFrame to be used as the dataset')
-@argument('-v', '--version', dest='version', default='3')
-@argument('-j', '--json', dest='json', action='store_true')
+@magic_arguments.argument('-v', '--version', dest='version', default='3')
+@magic_arguments.argument('-j', '--json', dest='json', action='store_true')
 def vega(line, cell):
     """Cell magic for displaying Vega visualizations in CoLab.
 
@@ -97,7 +92,7 @@ def vega(line, cell):
 
     If --json is passed, then input is parsed as json rather than yaml.
     """
-    args = parse_argstring(vega, line)
+    args = magic_arguments.parse_argstring(vega, line)
 
     version = args.version
     assert version in RENDERERS['vega']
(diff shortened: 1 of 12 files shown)


sebastianneubauer:
Ipython is now an optional dependency for interactive workloads and can be installed via:
```
pip install .[dev]
```
or 
```
pip install altair[dev]
```
Tests succeed with and without Ipython installed (tested manually on my local machine), but automatically test only run with Ipython installed so far.

altair/_magics.py
```diff
@@ -6,8 +6,28 @@
...
+            pass
+
+        def __call__(self, func):
+            pass
+
+    class argument:
+        def __init__(self, *args, **kwargs):
+            pass
+        def __call__(self, func):
+            pass
```
jakevdp: Shouldn't ``__call__`` return ``func`` for this to be a no-op?

In any case, I don't think this all of extra logic is needed: ``_magic.py`` is only ever imported by the ``load_ipython_extension()`` function in ``__init__.py``, and this will only be invoked if IPython is installed (unless a user invokes/imports this deliberately, at which point I think an ``ImportError`` is appropriate).

Unless there's something I'm overlooking, I think all the changes to this file can be reverted.

altair/vega/v2/display.py
```diff
@@ -1,6 +1,10 @@
 import os
 
-from IPython.display import display
+try:
+    from IPython.display import display
+    IPYTHON_AVAILABLE=True
+except ImportError:
+    IPYTHON_AVAILABLE=False
```
jakevdp: Here and in the other three similar places, I'd prefer to do the import within the ``vega()`` function rather than with a try/except at the top.

As currently written, if a user does not have IPython installed and tries to call ``vega()``, the result will be a ``NameError``, and they'll have read through the source code and reason that ``import IPython`` must have given an ``ImportError`` which led to ``display`` being undefined.

If we just import ``IPython`` where it's used, the result will be much clearer for the user, and it also makes the source code easier to follow.

altair/_magics.py
sebastianneubauer: That was my first approach too. Unfortunately, as the tests are in the same folder structure as the code, pytest collects `_magic.py` which results in a failure because of the decorators if IPython is not installed. 
A very pragmatic solution would be to ignore that the tests fail if IPython is missing, as normally if the test requirements are installed, IPython is installed too...Is that ok?

sebastianneubauer: So essentially, what we would loose, is the ad hoc ability to e.g. test in in travis if everything works with and without IPython. But currently we don't do this anyhow, so for me this would be acceptable.

sebastianneubauer: And still, if we want to do that, we could tweak pytest config to ignore `_magic.py` from collecting. Not so nice, but would work somehow I guess.

altair/vega/v2/display.py
sebastianneubauer: ✅ done

altair/_magics.py
jakevdp: > pytest collects _magic.py which results in a failure because of the decorators if IPython is not installed.

Running pytest depends on the dev requirements, so putting IPython there should be fine, no?

jakevdp: To be clear, what I'm saying is that ``_magic.py`` should be unchanged, and ``IPython`` should be a dev requirement, not an install requirement.

jakevdp: OK, I see what you're saying.

As written, though, won't there be a ``NameError`` if pytest tries to import ``altair._magics`` with IPython not installed? I'd prefer an ``ImportError``, because it more directly tells the user what the problem is... and that's what you'd get if you leave this file unchanged.

I don't think pytest actually grabs the contents of this file unless you run ``--with-doctests``, in which case you need all the other dev requirements as well.

doc/getting_started/installation.rst
```diff
@@ -25,7 +25,7 @@ with the classic notebook, see :ref:`installation-notebook`.
 
 To install JupyterLab and Altair with pip, run the following commands::
 
-    $ pip install jupyterlab altair
+    $ pip install jupyterlab altair[dev]
```
jakevdp: I think we should recommend that users do the simple ``altair`` installation, not the full development installation (otherwise, what's the point of splitting them out?)

sebastianneubauer: I ovelooked the jupyterlab before altair...so you are right!

jakevdp:
Looks good. Can you rebase on master to fix conflicts? I think this overlaps with some changes from #770

jakevdp:
Actually, it looks like those conflicts were addressed in the recent commits. I think we're good to go!

Thanks for the contribution here, and thanks for being patient with all the back and forth :smile:

sebastianneubauer:
No worries, this was one of the fastest merges, a pleasure! Looking forward for my next contribution! Thanks!

PR #770, opened 2018-04-23
Add support for generating vega spec in headless

### LINKED ISSUE #441: convert to VEGA
is there any way (preferably, in python) to convert Altair output to vega (not vega-lite) ?

saulshanabrook:
This closes #441 by adding support for generating Vega spec's from vega
lite specs. It does this by using a headless browser (either chrome
or firefox) and loading a hosted version of the vega lite JS API.

To test this code install seleneium (`pip install selenium`) and
chromedriver (`brew cask install chromedriver`).

I implemented this with minimal changes to the existing API by adding
the `vega` format to the `spec_to_image_mimebundle` function. So it will
return a dictionary mapping the correct vega mimetype (as defined in
https://altair-viz.github.io/user_guide/display_frontends.html#renderer-api)
to the vega spec. It doesn't really make much sense to render vega to vega,
but this is currently supported as a no-op.

### DIFF
--- a/altair/utils/headless.py
+++ b/altair/utils/headless.py
@@ -208,3 +208,4 @@ def spec_to_mimebundle(spec, format, mode,
         return {'image/svg+xml': render}
     elif format == 'vega':
         return {'application/vnd.vega.v{}+json'.format(vega_version[0]): render}
+    raise ValueError("format must be 'png', 'svg', or 'vega'")
--- a/altair/utils/headless.py
+++ b/altair/utils/headless.py
@@ -106,45 +106,7 @@ EXTRACT_CODE = {
         """}
 
 
-def spec_to_image_mimebundle(spec, format, mode,
-                             vega_version,
-                             vegaembed_version,
-                             vegalite_version=None,
-                             driver_timeout=10,
-                             webdriver='chrome'):
-    """Convert a vega/vega-lite specification to a PNG/SVG image/Vega spec
-
-    Parameters
-    ----------
-    spec : dict
-        a dictionary representing a vega-lite plot spec
-    format : string {'png' | 'svg' | 'vega'}
-        the file format to be saved.
-    mode : string {'vega' | 'vega-lite'}
-        The rendering mode.
-    vega_version : string
-        For html output, the version of vega.js to use
-    vegalite_version : string
-        For html output, the version of vegalite.js to use
-    vegaembed_version : string
-        For html output, the version of vegaembed.js to use
-    driver_timeout : int (optional)
-        the number of seconds to wait for page load before raising an
-        error (default: 10)
-    webdriver : string {'chrome' | 'firefox'}
-        Webdriver to use.
-
-    Returns
-    -------
-    output : dict
-        a mime-bundle representing the image
-
-    Note
-    ----
-    This requires the pillow and selenium Python modules to be installed.
-    Additionally it requires either chromedriver (if webdriver=='chrome') or
-    geckodriver (if webdriver=='firefox')
-    """
+def compile_spec(spec, format, mode, vega_version, vegaembed_version, vegalite_version, driver_timeout, webdriver):
     # TODO: allow package versions to be specified
     # TODO: detect & use local Jupyter caches of JS packages?
     if format not in ['png', 'svg', 'vega']:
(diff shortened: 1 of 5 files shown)


saulshanabrook:
A future version of this feature (if Python 2 support is dropped) could use [pyppeteer](https://github.com/miyakogi/pyppeteer), a wrapper around headless chrome, which removed the need to install selenium and chromedriver.

saulshanabrook:
I tried to test this on travis, but was getting this error:

> selenium.common.exceptions.WebDriverException: Message: Service chromedriver unexpectedly exited. Status code was: 1

https://travis-ci.org/altair-viz/altair/jobs/370209766#L1465

jakevdp:
This is really great... something I'd been thinking about taking on. Thanks for the contribution!

This addition makes the name of the utility make less sense... we could rename ``spec_to_image_mimebundle`` to ``compile_spec`` or something like that.

jakevdp:
Travis tests would be nice, but it's subtle to get it working... I'd be fine merging this and trying to get that working separately.

saulshanabrook:
Yeah, it is stretching the original intent of the function. It still does return a mimebundle for the vega spec, it's just not a image. So what about `spec_to_mimebundle`? After I implemented this I could it could be used by altair to return the the vega mimebundle, which I think jupyterlab has a mimerender for? But I don't know if this has any good use cases.

This commit is as far as I got on adding testing to travis, if it comes in handy later: https://github.com/saulshanabrook/altair/commit/d63a9b011e70b2cc8b65d5f3a0b716312c075e04

jakevdp:
> So what about spec_to_mimebundle?

I think we can separate this into two functions: one which is something like ``compile_spec`` that calls all the headless stuff and supports png, svg, and vega, and one that is ``spec_to_mimebundle`` that can create various mimebundles (png, svg)

jakevdp:
Exactly what I was thinking

saulshanabrook:
@jakevdp How does the change I pushed look?

altair/utils/headless.py
```diff
@@ -90,51 +90,26 @@ def temporary_filename(**kwargs):
...
-    This requires the pillow and selenium Python modules to be installed.
-    Additionally it requires either chromedriver (if webdriver=='chrome') or
-    geckodriver (if webdriver=='firefox')
-    """
+def compile_spec(spec, format, mode, vega_version, vegaembed_version, vegalite_version, driver_timeout, webdriver):
     # TODO: allow package versions to be specified
     # TODO: detect & use local Jupyter caches of JS packages?
-    if format not in ['png', 'svg']:
+    if format not in ['png', 'svg', 'vega']:
         raise NotImplementedError("format must be 'svg' and 'png'")
```
jakevdp: add vega to the error message

altair/utils/headless.py
```diff
@@ -180,12 +155,56 @@ def spec_to_image_mimebundle(spec, format, mode,
...
+    geckodriver (if webdriver=='firefox')
+    """
+
+    render = compile_spec(spec, format, mode, vega_version, vegaembed_version, vegalite_version, driver_timeout, webdriver)
     if format == 'png':
         return {'image/png': base64.decodebytes(render.split(',', 1)[1].encode())}
     elif format == 'svg':
         return {'image/svg+xml': render}
+    elif format == 'vega':
+        return {'application/vnd.vega.v{}+json'.format(vega_version[0]): render}
```
jakevdp: I'd like an else fallback here with a value error... as written that case should be caught by ``compile_spec``, but in case that code evolves in the future we should do something more than return None.

altair/utils/save.py
```diff
@@ -83,7 +83,7 @@ def save(chart, fp, vega_version, vegaembed_version,
                                              json_kwds=json_kwds)
         write_file_or_filename(fp, mimebundle['text/html'], mode='w')
     elif format in ['png', 'svg']:
-        mimebundle = spec_to_image_mimebundle(spec=spec, format=format, mode=mode,
+        mimebundle = spec_to_mimebundle(spec=spec, format=format, mode=mode,
                                               vega_version=vega_version,
                                               vegalite_version=vegalite_version,
                                               vegaembed_version=vegaembed_version,
```
jakevdp: Please make the indentation match here

altair/vegalite/v1/display.py
```diff
@@ -61,15 +61,15 @@ def json_renderer(spec):
 
 
 def png_renderer(spec):
-    return headless.spec_to_image_mimebundle(spec, format='png',
+    return headless.spec_to_mimebundle(spec, format='png',
                                              mode='vega-lite',
                                              vega_version=VEGA_VERSION,
                                              vegaembed_version=VEGAEMBED_VERSION,
                                              vegalite_version=VEGALITE_VERSION)
```
jakevdp: Please make indentation match

altair/vegalite/v1/display.py
```diff
@@ -61,15 +61,15 @@ def json_renderer(spec):
...
 
 
 def svg_renderer(spec):
-    return headless.spec_to_image_mimebundle(spec, format='svg',
+    return headless.spec_to_mimebundle(spec, format='svg',
                                              mode='vega-lite',
                                              vega_version=VEGA_VERSION,
                                              vegaembed_version=VEGAEMBED_VERSION,
```
jakevdp: indentation

altair/vegalite/v2/display.py
```diff
@@ -62,15 +62,15 @@ def json_renderer(spec):
 
 
 def png_renderer(spec):
-    return headless.spec_to_image_mimebundle(spec, format='png',
+    return headless.spec_to_mimebundle(spec, format='png',
                                              mode='vega-lite',
                                              vega_version=VEGA_VERSION,
                                              vegaembed_version=VEGAEMBED_VERSION,
                                              vegalite_version=VEGALITE_VERSION)
```
jakevdp: indentation

altair/vegalite/v2/display.py
```diff
@@ -62,15 +62,15 @@ def json_renderer(spec):
...
 
 
 def svg_renderer(spec):
-    return headless.spec_to_image_mimebundle(spec, format='svg',
+    return headless.spec_to_mimebundle(spec, format='svg',
                                              mode='vega-lite',
                                              vega_version=VEGA_VERSION,
                                              vegaembed_version=VEGAEMBED_VERSION,
```
jakevdp: indentation

saulshanabrook:
Sounds good, addressed those.

jakevdp:
Looks good, thanks!

PR #774, opened 2018-04-24
Replaced the existing, complicated stacked area example with a simpler example.

palewire:
I've substituted the existing example with one that doesn't require a transformation of the sample data.

screenshot from 2018-04-24 14-18-45


jakevdp:
Looks good, thanks!

Issue #742, opened 2018-04-10
ENH: add a dark theme


jakevdp:
#740 added theme support along with a few basic themes (default, none, opaque)

It would be useful to add a dark theme: i.e. one with white axes and labels, that would play well with black backgrounds.

kanitw:
These themes seem useful (esp. if someone makes a dark theme) -- so I wonder if it's worth putting these in Vega instead so all users of the Vega ecosystem can benefit from these options.  

cc: @jheer @arvind @domoritz

domoritz:
I like the idea of having a few predefined themes in Vega embed that people could choose from.

afonit:
@jakevdp 
Note - this looks ugly - I know that, just playing with colors and identifying parts of the chart.

I have been trying to put this through its paces to see if I can reach all the attributes.
When I do a faceted chart I can't seem to reach the:
* titles of the facets
* the right most grid line in the facets.

I also can't seem to reach the title's color, I tried putting it both at the top level config, then tried to do it following the hierarchy noted here: https://vega.github.io/vega-lite/docs/title.html#config

See here:
![image](https://user-images.githubusercontent.com/1798497/38615941-5c05ad0a-3d60-11e8-857d-158fa85926f7.png)

```python
import altair as alt
import pandas as pd
alt.theme.register('dark_enhanced', 
                   {'config':
                    {'title': {'color': 'orange'}, #NOT WORKING
                    'axis': {'titleColor': 'orange', #axis titles
                             'domainColor': 'green', #domain color colors the bar on axis labels
                             'gridColor': 'red', #grid color is most of the grid except the right hand side
                             'labelColor': 'orange', #text labels of the axes
                             'tickColor': 'yellow'} #tick marks
                   }}
                  )
alt.theme.enable('dark_enhanced')

from vega_datasets import data

source = data.movies.url

chart = alt.Chart(source).mark_point().encode(
    x='Worldwide_Gross:Q',
    y='US_DVD_Sales:Q',
    column='MPAA_Rating:N'
)
chart
```

jakevdp:
Thanks for looking into this... Maybe one of the Vega-Lite folks can weigh-in.

jakevdp:
Note that ``alt.theme`` has been renamed to ``alt.themes``.

domoritz:
[...]

Issue #673, opened 2018-03-29
Hide Export as PNG/ View Source/ Open in Vega Editor


bgruenefeld:
Hi, 
i recently discovered Altair and I'm playing around with it. 
It looks quite impressive and easy to use!

Under every visualization i create an Export as PNG/ View Source/ Open in Vega Editor footer is shown.
I already searched the documentation but couldn't find where to disable this "footer row".
Could anybody give me a hint how i can turn off this Footer?

Kind regards
Bernd

jakevdp:
There is currently no way to do this. We've been talking about a model to give users control over chart metadata, but it's not implemented yet.

yashagrawal3:
@jakevdp I wish to work on this issue. Can you tell me how do you want the user to control it?

yashagrawal3:
I notice you have already been working on it on #678 , anything I can do to help?

jakevdp:
Hi @yashagrawal3 - this issue is a bit complicated because it requires defining a Python-side API for setting metadata, defining a way for that metadata to be passed to the frontends, and updating all the frontends to handle metadata in that format. None of those three things has been figured out yet, though #678 is one proposal for how to handle the first of the three related issues.

jakevdp:
I'm hoping to tackle related things this week, so hopefully there will be some progress here :smile:

yashagrawal3:
OK @jakevdp , please let me know if I can contribute something there.

repmax:
@bgruenefeld You can hide the chart footer by inserting `<style>.vega-actions {display: none}</style>` in the html.

jakevdp:
Note: as soon as #776 is merged, there will be a better story for controlling plot metadata without manually adjusting CSS

ggada:
Didn't want to open a new issue about this so commenting here.

This works in the notebook, however, `chart.save('viz.html')` still has the footer. Is there a way to turn it off in the HTML, or should we add CSS as mentioned above?

In my specific use case, I am using a vega html export in a reveal.js slide, so I would love the HTML export to not have the footer.

Best,
Gaurav

jakevdp:
If you want to customize the HTML export, you can export the JSON and embed it in any HTML template you wish to use. The default one can be seen here: https://github.com/altair-viz/altair/blob/358f06e1ac13b343e0afa1b1799346c3931e8edb/altair/utils/html.py

jakevdp:
[...]

PR #776, opened 2018-04-25
[MRG] Generalized Plugin Interface

jakevdp:
The goal of this work is to generalize the plugin interface to allow specification of additional options and/or metadata. See below for examples.

TODO list:

- [x] Require all plugins to return a callable function or object (mainly affects theme registry)
- [x] Make ``plugins.enable()`` able to be used as a context manager
- [x] Change the theme plugins to a function returning a spec
- [x] Add ability to specify extra options when calling ``plugins.enable()``
- [x] Specialize the data_transformers plugins
- [x] Allow ``max_rows`` to be set for data transformers
- [x] Specialize the renderers plugins
- [x] Allow metadata to be set for renderers (e.g. vega-embed args and vega-tooltip args)
- [x] Test setting of metadata
- [x] Modify html output to make use of that metadata

**External TODOS:**

- update ipyvega & jupyterlab to use ``embed_options`` from mimebundle metadata

---

This PR addresses (or partially addresses) #758, #742, #674, #673, and #240.

Most of this is internal reorganization, but there are a few primary user-side API enhancements that this PR introduces out-of-the-box.

## Temporary plugin enabling via context manager
With this PR, ``plugins.enable`` can optionally be used as a context manager to temporarily enable a specified plugin. For example:
```python
with alt.data_transformers.enable('json'):
    do_something()  # executed with the json data transformer
do_something_else()  # executed with the previous data transformer
```

## Data transformer settings
After this PR, you can directly set the ``max_rows`` parameter of the default data transformer:
```python
alt.data_transformers.enable('default', max_rows=10000)
```
Alternatively, you can use
```python
alt.data_transformers.enable('default', max_rows=None)
```
and no check of the number of rows will be performed.

## Renderer metadata
This provides a mechanism to pass arbitrary metadata to the renderer. For the Colab renderer, which is built-in to the package, vegaEmbed [options](https://github.com/vega/vega-embed#options) can be passed using the ``embed_options`` argument; e.g.
```python
alt.renderers.enable('colab', embed_options={'actions': False})
```
this will hide the action links (i.e. save as png, open in vega editor, etc.).

As I've designed it here, these additional arguments will be passed in the metadata dictionary of other mimebundle representations, so that renderers like jupyterlab-vega and ipyvega can make use of them as well. For example:

```python
>>> import altair as alt
>>> alt.renderers.enable('default', embed_options={'actions': False})
>>> alt.Chart('data.csv').mark_point()._repr_mimebundle_(None, None)
({'text/plain': '<VegaLite 2 object>\n\nIf you see this message, it means the renderer has not been properly enabled\nfor the frontend that you are using. For more information, see\nhttps://altair-viz.github.io/user_guide/troubleshooting.html\n',
  'application/vnd.vegalite.v2+json': {'config': {'view': {'width': 400,
     'height': 300}},
   'data': {'url': 'data.csv'},
   'mark': 'point',
   '$schema': 'https://vega.github.io/schema/vega-lite/v2.3.0.json'}},
 {'embed_options': {'actions': False}})
```
Notice the second element in the tuple contains the embed options.

I would propose we adopt this ``embed_options`` metadata key as a standard to be accepted in the metadata and handled appropriately by jupyterlab-vega, ipyvega, and any other renderers.

### DIFF
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -45,7 +45,10 @@ class MaxRowsError(Exception):
 
 @curry
 def limit_rows(data, max_rows=5000):
-    """Raise MaxRowsError if the data model has more than max_rows."""
+    """Raise MaxRowsError if the data model has more than max_rows.
+
+    If max_rows is None, then do not perform any check.
+    """
     check_data_type(data)
     if isinstance(data, pd.DataFrame):
         values = data
@@ -54,8 +57,11 @@ def limit_rows(data, max_rows=5000):
             values = data['values']
         else:
             return data
-    if len(values) > max_rows:
-        raise MaxRowsError('The number of rows in your dataset is greater than the max of {}'.format(max_rows))
+    if max_rows is not None and len(values) > max_rows:
+        raise MaxRowsError('The number of rows in your dataset is greater '
+                           'than the maximum allowed ({0}). '
+                           'For information on how to plot larger datasets '
+                           'in Altair, see the documentation'.format(max_rows))
     return data
 
 
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -4,8 +4,19 @@ import uuid
 
 import pandas as pd
 from toolz.curried import curry, pipe  # noqa
+from typing import Callable
 
 from .core import sanitize_dataframe
+from .plugin_registry import PluginRegistry
+
+
+# ==============================================================================
+# Data transformer registry
+# ==============================================================================
+DataTransformerType = Callable
+
+class DataTransformerRegistry(PluginRegistry[DataTransformerType]):
+    pass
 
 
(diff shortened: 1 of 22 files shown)


jakevdp:
ping @ellisonbg – I'd love to get this reviewed and merged so we can push forward with the 2.0 release!

jakevdp:
I'm planning to merge this PR this afternoon, because I'd like to get a 2.0 release out by maybe tomorrow. Let me know if you see any issues here.

jakevdp:
I'm going to merge this so we can start working on renderer updates & documentation in preparation for the 2.0 release.

jakevdp:
If folks have any comments or other feedback, hopefully we can address them from master!

PR #788, opened 2018-04-28
DOC: Reorganized the example gallery

palewire:
For #775. You'll find an explicitly defined table of contents order for the gallery list page, with the simple charts first. 

Here's the draft order I settled on. With newbies in mind, I aimed for an increasing order of complexity as you move through the list.

* Simple Charts
* Bar Charts
* Line Charts
* Area Charts
* Scatter Plots
* Histograms
* Maps
* Interactive Charts
* Other Charts

There are some emerging categories of charts, like heatmaps or "table bins," that could potentially grow into their own categories as more examples are added.

Other than adjusting their categories, I did not get into adding or editing charts as part of this patch. I have plenty of thoughts about that, but they can wait for later.

Issue #775, opened 2018-04-24
Are you open to reorganizing the Example Gallery?


palewire:
In my humble opinion, [the example gallery](https://altair-viz.github.io/gallery/index.html) -- which I love and see as the most important section of the documentation -- has grown beyond its current structure. 

My proposal: Break it up into sections that are based on chart type.

The "Simple" section at the top, for newbies, could stay in place. It would feature a single, simple example of each chart type, with very little customization and option usage in the altair code.

Following that, the "General" and "Interactive" sections would be removed and their elements reorganized into sections like:

* Bar charts
* Line charts
* Area charts
* Scatterplots
* Histograms
* Maps

My current feeling is that this would provide a more flexible framework for further expanding the examples, and make the page easier to navigate for users. 

If you're interested in this idea, I'd be happy to take the lead in attempting to put together a draft.

kanitw:
I think finer grain taxonomy is useful, but the taxonomy should not omit layered plots and multi-view plots, which is the key benefit of visualization grammars like Altair and Vega-Lite over "chart template" libraries.  

I also just notice that the Vega-Lite gallery is organized a bit differently from Altair.  https://vega.github.io/vega-lite/examples/

There is definitely a room for discussion & improvement and perhaps we can improve the organization of both VL and Altair galleries.

kanitw:
That's true.  One of the thing that I like about the categories in Vega-Lite is that they are quite mutually exclusive.  That said, I wanna split basic into true basic one and another one that's less basic (but can't think of a better name yet.)

palewire:
An excercise that can help think through this kind of challenge is to imagine the "use cases" of the readers we expect to arrive at this page. There are three that occur to me off the top of my head. In no particular order:

1) "Curious chart makers" checking out Altair for the first time, but have never installed it. 
2) "Newbies" who have installed Altair for the first time and are looking to quickly begin experimenting
3) "Regular users" who are looking for reminder or tutorial to answer their question: "How does Altair make chart X?"

To me, each of these cases would be well served by a compact "Basic charts" section at the top of the page, which can serve as quick overview of what's possible. 

Regular users, who I think should be thought of as the most important use case, would be best served, IMHO, by a taxonomy that allows them quickly find the type of chart they're interested in making. Then they can narrow down and scan pick through all the variations for that chart to find the one they're interested in mimicking. These users are not served as well by the current hodge podge of charts.

[...]

PR #789, opened 2018-04-28
DOC: Example gallery reorganization

palewire:
Replaces #788 (where you can find a fuller discussion). For #775.

### DIFF
--- a/altair/sphinxext/altairgallery.py
+++ b/altair/sphinxext/altairgallery.py
@@ -2,6 +2,7 @@ import hashlib
 import os
 import json
 import random
+import collections
 from operator import itemgetter
 
 import jinja2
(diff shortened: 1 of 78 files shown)


kanitw:
FYI, we're now also [restructuring the Vega-Lite gallery](https://github.com/vega/vega-lite/pull/3679#pullrequestreview-116151298) -- obviously motivated by this change here.  :)

jakevdp:
I finally had a chance to download and build this... looks really good!

I wonder if it wouldn't be worth adding a half dozen more simple charts for the first category. At the very least we should demo various marks that are available.

jakevdp:
I'm going to go ahead and merge this, and then separately add a few more examples to the "Simple" category.

palewire:
I agree that adding more simple charts would be good.

One idea: Each subsection has a corresponding simple chart.

PR #811, opened 2018-05-03
Added an example of data with error bars

dhuppenkothen:
This is a really simple example, but almost all of my data is of this type and it took me a bit to figure out `transform_calculate`, so I thought it might be useful for someone else to see. 

This is kind of a simpler version of the [errorbar example](https://altair-viz.github.io/gallery/error_bars_with_ci.html), where data points and uncertainties don't have to be calculated from a distribution, but are given as (x, y, yerr).

### DIFF
--- a/altair/vegalite/v2/examples/simple_scatter_with_errorbars.py
+++ b/altair/vegalite/v2/examples/simple_scatter_with_errorbars.py
@@ -11,7 +11,7 @@ import altair as alt
 from altair import datum
 
 import numpy as np
-
+import pandas as pd
 
 # generate some data points with uncertainties
 np.random.seed(0)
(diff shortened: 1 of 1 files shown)


jakevdp:
Thanks!

I'd put that in the "Scatter Plots" category... the "Simple Charts" category is mainly for simple examples without layering, transforms, etc.

dhuppenkothen:
Wherever you think it fits best. Is there something I should change in the files or the description to make that work?

altair/vegalite/v2/examples/simple_scatter_with_errorbars.py
```diff
@@ -0,0 +1,46 @@
+"""
+Simple Scatter Plot with Errorbars
+----------------------------------
+
+A simple scatter plot of a data set with errorbars.
+
+"""
+# category: simple charts
+
```
jakevdp: Change this to ``# category: scatter plots``

jakevdp:
Looks good, thanks!

PR #814, opened 2018-05-04
added example that links an interval selection in a histogram with a scatter plot

dhuppenkothen:
Here's something else that took me a while to figure out yesterday, while I was trying to reproduce the [cross-highlights example](https://altair-viz.github.io/gallery/interactive_cross_highlight.html) on some of my own data: the behaviour is different when selecting a bar chart made from nominal categories to when there's a histogram binning step of quantitative information in the middle. 

I found that the only way I could make it work is by applying the `transform_bin` method to both charts. I don't know if that's the correct way to do it, or if there's a better way, but in either case, I thought it might be nice to have an example that shows this behaviour.

### DIFF
--- a/altair/vegalite/v2/examples/scatter_with_histogram.py
+++ b/altair/vegalite/v2/examples/scatter_with_histogram.py
@@ -6,11 +6,10 @@ This example shows how to link a scatter plot and a histogram
 together such that an interval selection in the histogram will 
 plot the selected values in the scatter plot. 
 
-Note that the binning  has to be applied to *both* 
-the scatter plot and the bar chart via the `transform_bin` mehtod 
-in order for this to work! 
-
-
+Note that both subplots need to know about the `mbin` field created 
+by the `transform_bin` method. In order to achieve this, the data is
+not passed to the `Chart()` instances creating the subplots, but 
+directly in the `hconcat()` function, which joins the two plots together.
 """
 # category: interactive charts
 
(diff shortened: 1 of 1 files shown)


jakevdp:
A cleaner way to make the transformed data available to both panels is to apply the transform once at the top level:
```python
"""
Scatter Plot and Histogram with Interval Selection
==================================================
This example shows how to link a scatter plot and a histogram 
together such that an interval selection in the histogram will 
plot the selected values in the scatter plot. 
Note that the binning  has to be applied to *both* 
the scatter plot and the bar chart via the `transform_bin` mehtod 
in order for this to work! 
"""
# category: interactive charts

import altair as alt
import pandas as pd
import numpy as np

x = np.random.normal(size=100)
y = np.random.normal(size=100)
m = np.random.normal(15, 1, size=100)

df = pd.DataFrame({"x": x, "y":y, "m":m})

pts = alt.selection(type="interval", encodings=["x"])

# left panel: scatter plot
points = alt.Chart().mark_point(filled=True, color="black").encode(
    x='x',
    y='y'
).transform_filter(
    pts.ref()
).properties(
    width=300,
    height=300
)

# right panel: histogram
mag = alt.Chart().mark_bar().encode(
    x='mbin:N',
    y="count()",
    color=alt.condition(pts, alt.value("black"), alt.value("lightgray"))
).properties(
    selection=pts,
    width=300, 
    height=300
)

# build the chart:
alt.hconcat(points, mag,
    data=df
).transform_bin(
    "mbin",
    field="m",
    bin=alt.Bin(maxbins=20)
)
```

jakevdp:
Or maybe this is clearer?

```python
chart = alt.HConcatChart(data=df).transform_bin(
    "mbin",
    field="m",
    bin=alt.Bin(maxbins=20)
)

# left panel: scatter plot
chart |= alt.Chart().mark_point(filled=True, color="black").encode(
    x='x',
    y='y'
).transform_filter(
    pts.ref()
).properties(
    width=300,
    height=300
)

# right panel: histogram
chart |= alt.Chart().mark_bar().encode(
    x='mbin:N',
    y="count()",
    color=alt.condition(pts, alt.value("black"), alt.value("lightgray"))
).properties(
    selection=pts,
    width=300, 
    height=300
)

chart
```

dhuppenkothen:
Aaah, that was the thing I failed to figure out: how to apply transforms to both charts! Thanks! I personally find your first example more straightforward to follow, but that might just be me? 

Do you want me to update the PR with that?

jakevdp:
Sure, sounds good! And maybe update the description as well: one of the confusing points here is that you build each chart **without** a dataset, and they only get connected to a dataset when added to the compound chart.  I think the second formulation makes that slightly more clear, but I could go either way.

jakevdp:
Hey @dhuppenkothen – this is really close. Will you have a chance to make those changes? Otherwise I'm happy to take it from here. Thanks!

dhuppenkothen:
Oops, I lost track of this. I can make the changes this weekend. :)

dhuppenkothen:
Updated this PR with the first of your two suggested solutions (which might be less elegant, but which I find much easier to understand). Let me know if there are any other changes that would be useful!

jakevdp:
Looks great, thanks!

PR #817, opened 2018-05-04
DOC: Normalized Stacked Area Chart example

palewire:
The same as the simple area chart example, but normalized. 

```python
import altair as alt

source = data.unemployment_across_industries.url

alt.Chart(source).mark_area().encode(
    x="date:T",
    y=alt.Y("count:Q", stack="normalize"),
    color="series:N"
)
```

visualization 17

### DIFF
--- a/altair/vegalite/v2/examples/layered_area_chart.py
+++ b/altair/vegalite/v2/examples/layered_area_chart.py
@@ -1,0 +1,16 @@
+"""
+Layered Area Chart
+------------------
+This example shows a layered area chart.
+"""
+# category: area charts
+import altair as alt
+from vega_datasets import data
+
+iowa = data.iowa_electricity()
+
+alt.Chart(iowa).mark_area(opacity=0.3).encode(
+    x="year:T",
+    y=alt.Y("net_generation:Q", stack=None),
+    color="source:N"
+)
--- a/altair/vegalite/v2/examples/normalized_stacked_area_chart.py
+++ b/altair/vegalite/v2/examples/normalized_stacked_area_chart.py
@@ -3,7 +3,7 @@ Normalized Stacked Area Chart
 -----------------------------
 This example shows how to make a normalized stacked area chart.
 """
-# category: simple charts
+# category: area charts
 import altair as alt
 from vega_datasets import data
 
--- a/altair/vegalite/v2/examples/normalized_stacked_area_chart.py
+++ b/altair/vegalite/v2/examples/normalized_stacked_area_chart.py
@@ -1,0 +1,16 @@
+"""
+Normalized Stacked Area Chart
+-----------------------------
+This example shows how to make a normalized stacked area chart.
+"""
+# category: simple charts
+import altair as alt
+from vega_datasets import data
+
+source = data.unemployment_across_industries.url
+
+alt.Chart(source).mark_area().encode(
+    x="date:T",
+    y=alt.Y("count:Q", stack="normalize"),
+    color="series:N"
+)
--- a/altair/vegalite/v2/examples/normalized_stacked_area_chart.py
+++ b/altair/vegalite/v2/examples/normalized_stacked_area_chart.py
@@ -7,10 +7,10 @@ This example shows how to make a normalized stacked area chart.
 import altair as alt
 from vega_datasets import data
 
-source = data.unemployment_across_industries.url
+iowa = data.iowa_electricity()
 
-alt.Chart(source).mark_area().encode(
-    x="date:T",
-    y=alt.Y("count:Q", stack="normalize"),
-    color="series:N"
+alt.Chart(iowa).mark_area().encode(
+    x="year:T",
+    y=alt.Y("net_generation:Q", stack="normalize"),
+    color="source:N"
 )
--- a/altair/vegalite/v2/examples/simple_stacked_area_chart.py
+++ b/altair/vegalite/v2/examples/simple_stacked_area_chart.py
@@ -7,10 +7,10 @@ This example shows how to make a simple stacked area chart.
 import altair as alt
 from vega_datasets import data
 
-source = data.unemployment_across_industries.url
+iowa = data.iowa_electricity()
 
-alt.Chart(source).mark_area().encode(
-    x="date:T",
-    y="count:Q",
-    color="series:N"
+alt.Chart(iowa).mark_area().encode(
+    x="year:T",
+    y="net_generation:Q",
+    color="source:N"
 )
--- a/altair/vegalite/v2/examples/trellis_area.py
+++ b/altair/vegalite/v2/examples/trellis_area.py
@@ -1,23 +1,17 @@
 """
 Trellis Area
 -----------------
-This example shows stock prices of four large companies as a small multiples of area charts.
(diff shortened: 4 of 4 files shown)


eitanlees:
I like the example, but wonder if there is a better dataset to use for the area chart examples. I don't like how the colors are repeated. I feel the same way about the "Simple Stacked Area Chart".

jakevdp:
The vega-lite example uses a different color scheme, which makes it a bit better: https://vega.github.io/vega-lite/examples/stacked_area_normalize.html

eitanlees:
hmmm, and there are concerns with the `d3.schemeCategory20*` categorical color schemes as well. From the [D3 5.0.0 release notes](https://github.com/d3/d3/releases/tag/v5.0.0)

> The twenty-color schemes were removed because their grouped design often falsely implied non-existent relationships in the data

It's tough to have a good color scheme for a large number of categories. 

My only issue is that for simple examples I think a dataset with fewer categories would be better a better choice.

palewire:
I agree with you but I was unable to find a better dataset among the Vega
options.

To me the best possible set would: 1) Require no data transformations 2) Be
a time series 3) Have a small number of categories 4) Tell an obvious
visual "story."

This dataset only checks the first two boxes.

Is there another dataset from Vega that you think hits all four?

If not, this might be a spot for me to dig into our past news work to find
a better example set, which I can then submit upstream to Vega's
maintainers.

eitanlees:
Currently `stocks.csv` ticks a lot of boxes but it's a little jarring. A new dataset with a better "story" would be a breath of fresh air. I guess I don't know if it's better to use the same dataset for many examples or use different datasets. From a beginners perspective the homogeneity might be nice but also might be boring. The benefit of diverse datasets is the appeal to a wider audience. As we iterate on the examples/documentation these are things I have been thinking about. 

Anyways, here is a mock up with `stocks.csv`

```python
import altair as alt

source = data.stocks()

alt.Chart(source).mark_area().encode(
    x="date",
    y=alt.Y("price", stack="normalize"),
    color="symbol"
)
```
norm_area

palewire:
I agree it better meets some of the requirements I spelled out, but there's the problem that those values aren't percentages of a whole. I suppose that should be additional requirement. To me, the ideal dataset would easily lead to an ideal example of this type of chart.

palewire:
I'll spent a little time thinking about what small, simple news datasets might be introduced as a better example.

palewire:
Here's an experiment with a dataset I acquired earlier today from the Energy Information Administration: [U.S. energy production by source since 1949](https://gist.github.com/palewire/95f9b4b86060aea82e8649a9ecab8f7b). 

I can imagine a more dramatic chart than this -- and maybe that's what we should be aiming for -- but I think this is a little more in the ballpark. Let me know.

```python
import pandas as pd
import altair as alt

data = pd.read_csv("https://gist.githubusercontent.com/palewire/95f9b4b86060aea82e8649a9ecab8f7b/raw/cbe3e3d2e4235a4d21d9fa477166abe861ce7a61/production-by-sector.csv")

alt.Chart(data).mark_area().encode(
    x="year:O",
    y="production:Q",
    color="type:N"
)
```

visualization 5

```
alt.Chart(data).mark_area().encode(
    x="year:O",
    y=alt.Y("production:Q", stack="normalize"),
    color="type:N"
)
```

visualization 6

The chart is also wider than we might prefer. We could include a width configuration in the example, but I'd prefer not to. We could also simple truncate the dataset.

palewire:
Here's another stab that I think is closer to the mark.

I explored the EIA data until I found a state that has seen a significant shift in its electricity generation mix in recent years. The goal is to yield a more dramatic chart with a shorter time span.

I give you, Iowa.

```python
data = pd.read_csv("https://gist.githubusercontent.com/palewire/6a8c512c518b98f85358242684584a81/raw/22ee97dba4a59a4cac8c02c70c68ecfbce50aba9/iowa.csv")
alt.Chart(data).mark_area().encode(
    x="year:O",
    y="generation:Q",
    color="type:N"
)
```

visualization 7


```python
alt.Chart(data).mark_area().encode(
    x="year:O",
    y=alt.Y("generation:Q", stack="normalize"),
    color="type:N"
)
```

visualization 8

palewire:
My pull request with vega-datasets upstream is in [here](https://github.com/vega/vega-datasets/pull/35).

domoritz:
This is an interesting discussion. Note that even though category 20 should be avoided, it is better than using the same color again and unfortunately the jobs dataset has more than 10 categories. To avoid this confusion, I'm strongly in favor of a simpler dataset and added some comments to the PR in vega-datasets. Thanks everyone for the discussion here!

palewire:
I've just filed a pull request adding the new `iowa-electricity` dataset to the Python release of `vega_datasets`.

palewire:
Once the finalized dataset is merged into the Python version of `vega_datasets` and released, here are a newly revised set of area charts I'd like to submit via this pull request.

```python
import pandas as pd
import altair as alt
# This data import would obviously be replaced with a vega_datasets call
data = pd.read_csv("https://raw.githubusercontent.com/vega/vega-datasets/gh-pages/data/iowa-electricity.csv")
```

First, a simple stacked area chart for the top section.

```python
alt.Chart(data).mark_area().encode(
    x="year:T",
    y="net_generation:Q",
    color="source:N"
)
```
visualization 9

Then, in the stacked area chart section below, a normalized version.

```python
alt.Chart(data).mark_area().encode(
    x="year:T",
    y=alt.Y("net_generation:Q", stack="normalize"),
    color="source:N"
)
```
visualization 10

A layered version

```python
alt.Chart(data).mark_area(opacity=0.3).encode(
    x="year:T",
    y=alt.Y("net_generation:Q", stack=None),
    color="source:N"
)
```
visualization 11

And a trellis one.

```python
alt.Chart(data).mark_area().encode(
    x="year:T",
    y="net_generation:Q",
    row="source:N"
)
```

visualization 12

Finally, a tricked out version with enough configuration to show what a publishable version might look like.

```python
alt.Chart(data, title="Iowa's renewable energy boom").mark_area().encode(
    x=alt.X(
        "year:T",
        timeUnit="year",
        axis=alt.Axis(title="Year")
    ),
    y=alt.Y(
        "net_generation:Q",
        stack="normalize",
        axis=alt.Axis(title="Share of net generation", format=".0%"),
    ),
    color=alt.Color(
        "source:N",
        legend=alt.Legend(title="Electricity source"),
    )
)
```

visualization 13

jakevdp:
This is great!

I think it would be worth adding a few of these to the gallery, once the dataset is in the vega_datasets release.

I think this might also be useful in a narrative form under the "Case Studies" heading in the docs, exploring the process of going from the simple to more complicated charts. Let me know if that's something you'd like to contribute... otherwise I can use your code and put it together!

palewire:
Happy to do it, but I think it will require a new version of `vega_datasets` being released on PyPI.

palewire:
I pushed again with the four examples from the thread above now included. The more complex case study idea can wait for a separate pull request, I think.

jakevdp:
Looks like you're hitting test failures related to #846. If you rebase on master they should go away

palewire:
Done.

PR #830, opened 2018-05-13
ENH: support multiple fields for single encoding.

### LINKED ISSUE #825: ENH: encodings with multiple fields (e.g. tooltip)
We need to update the ``encode()`` method so that it will handle multiple encoding fields in cases where it is supported.

For example, in the most recent vega-lite release, it is possible to pass multiple fields to the tooltip encoding:
```python
cars = data.cars()

chart = alt.Chart(cars).mark_point().encode(
    x='Horsepower',
    y='Miles_per_Gallon',
    color='Origin'
)

chart.encoding.tooltip = [{'field': 'Name', 'type': 'nominal'},
                          {'field': 'Origin', 'type': 'nominal'}]

chart
```
screen shot 2018-05-08 at 10 37 35 am

Altair should make this available via a more simplified API, such as
```python
cars = data.cars()

chart = alt.Chart(cars).mark_point().encode(
    x='Horsepower',
    y='Miles_per_Gallon',
    color='Origin',
    tooltip=['Name', 'Origin']
)
```

jakevdp:
This is valid for the 'detail' and 'tooltip' encodings.

Fixes #825

### DIFF
--- a/altair/vegalite/v1/schema/channels.py
+++ b/altair/vegalite/v1/schema/channels.py
@@ -15,6 +15,12 @@ class FieldChannelMixin(object):
         context = context or {}
         if self.shorthand is Undefined:
             kwds = {}
+        elif isinstance(self.shorthand, list):
+            # If given a list of shorthands, then transform it to a list of classes
+            kwds = self._kwds.copy()
+            kwds.pop('shorthand')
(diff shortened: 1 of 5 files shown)


jakevdp:
Thanks!

PR #836, opened 2018-05-14
MAINT: update vega-lite to version 2.4.3

jakevdp:


jakevdp:
Note that this has a small backward incompatibility: ``alt.MarkProperties`` is removed, and the existing ``alt.MarkConfig`` is used in its place. This is an internal class, not used anywhere in the documentation or examples, and so I think that will be OK.

Issue #876, opened 2018-05-21
DOC: Line chart with point markers


eitanlees:
I thought a nice adition to the documentation would be a simple line chart with point markers.

As an example let's use the [simple line chart](https://altair-viz.github.io/gallery/simple_line_chart.html) from the example gallery. 

```python
import altair as alt
import numpy as np
import pandas as pd

x = np.arange(100)
data = pd.DataFrame({'x': x,
                     'sin(x)': np.sin(x / 5)})
```

I found out there are two ways to do this. 

- Create the line and points seperately, and layer

```python
base = alt.Chart(data).encode(
    x='x',
    y='sin(x)'
)

base.mark_line() + base.mark_circle()
```
![layer](https://user-images.githubusercontent.com/2041969/40310112-c5562d82-5cd9-11e8-8cf4-d89f8a3b5041.png)

- Create the line with `point=True` as a parameter. 

```python
alt.Chart(data).mark_line(point=True).encode(
    x='x',
    y='sin(x)'
)
```
![pointstrue](https://user-images.githubusercontent.com/2041969/40310150-d807a280-5cd9-11e8-9b32-41246dd185d4.png)

I am not sure which is prefered. 

There are a few other examples with points and lines where changes could be made such as

- [Connected Scatterplot (Lines with Custom Paths)](https://altair-viz.github.io/gallery/connected_scatterplot.html)

- [Multiple Marks](https://altair-viz.github.io/gallery/multiple_marks.html)

jakevdp:
This would be a nice example!

jakevdp:
I think either approach is fine.

eitanlees:
Closed by #877

PR #877, opened 2018-05-21
DOC: Add line chart with markers example

eitanlees:
Added a simple example of a line chart with markers discussed in #876 

I can also update 

- [Connected Scatterplot (Lines with Custom Paths)](https://altair-viz.github.io/gallery/connected_scatterplot.html)

- [Multiple Marks](https://altair-viz.github.io/gallery/multiple_marks.html)

if you think using `mark_line(point=True)` is preferred to the layering method.

### DIFF
--- a/altair/vegalite/v2/examples/connected_scatterplot.py
+++ b/altair/vegalite/v2/examples/connected_scatterplot.py
@@ -2,7 +2,8 @@
 Connected Scatterplot (Lines with Custom Paths)
 -----------------------------------------------
 
-This example show how the order encoding can be used to draw a custom path. The dataset tracks miles driven per capita along with gas prices annually from 1956 to 2010. It is based on Hannah Fairfield's article 'Driving Shifts Into Reverse'. See https://archive.nytimes.com/www.nytimes.com/imagepages/2010/05/02/business/02metrics.html for the original. 
+This example show how the order encoding can be used to draw a custom path. The dataset tracks miles driven per capita along with gas prices annually from 1956 to 2010. 
+It is based on Hannah Fairfield's article 'Driving Shifts Into Reverse'. See https://archive.nytimes.com/www.nytimes.com/imagepages/2010/05/02/business/02metrics.html for the original. 
 """
 # category: scatter plots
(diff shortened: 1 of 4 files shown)


kanitw:
> mark_line(point=True)

is better because the point markers will have opacity = 1 be default.  (The layering method will make point's opacity a bit semi-transparent like in scatterplots.)

jakevdp:
Yes, it would be great to update those other examples.

jakevdp:
There's also a couple instances in the sphinx docs that would be worth updating as well

eitanlees:
I fixed the two examples listed earlier. Now for the sphinx docs. 

I see one example in the [Layered Charts](https://altair-viz.github.io/user_guide/compound_charts.html#layered-charts) section. 

Any others?

altair/vegalite/v2/examples/connected_scatterplot.py
```diff
@@ -2,23 +2,16 @@
 Connected Scatterplot (Lines with Custom Paths)
 -----------------------------------------------
 
-This example shows how layering can be used to build a plot. This dataset tracks miles driven per capita along with gas prices annually from 1956 to 2010. It is based on the May 2, 2010 New York Times article 'Driving Shifts Into Reverse'. See http://mbostock.github.io/protovis/ex/driving.html .
+This example show how the order encoding can be used to draw a custom path. The dataset tracks miles driven per capita along with gas prices annually from 1956 to 2010. It is based on Hannah Fairfield's article 'Driving Shifts Into Reverse'. See https://archive.nytimes.com/www.nytimes.com/imagepages/2010/05/02/business/02metrics.html for the original.
```
jakevdp: Can you break this line?

jakevdp:
Looks good aside from the one long line. Thanks!

altair/vegalite/v2/examples/connected_scatterplot.py
eitanlees: Sure thing

jakevdp:
Looks great, thanks!

PR #881, opened 2018-05-22
DOC: Added simple highlighted bar chart example

palewire:
I got a question today from a journalism colleague. His question: How can I quickly pick a column in a bar chart for highlighting?

Here's a method I developed that I propose for the example gallery. I don't know if this is the pattern you'd prefer to popularize, but it worked well for me today.

visualization 32

```python
import altair as alt

population = data.population()

alt.Chart(population).mark_bar().transform_calculate(
    # Using a simple boolean test to calculate a new column on the fly
    # allows the chart creator to decide which bar they'd like to highlight.
    "highlight",
    alt.datum.year == 2000
).encode(
    x="year:O",
    y="sum(people):Q",
    # The highlight is then applied by using the calculated column
    # to set the color.
    color="highlight:N"
)
```


kanitw:
I think it is better to use [`condition`](https://vega.github.io/vega-lite/docs/condition.html ) as it will use less memory.  

I can't find the documentation for the Atlair syntax. But here is a similar VL example that makes rect colors depend on a test condition: https://vega.github.io/editor/#/examples/vega-lite/layer_text_heatmap

palewire:
Got it. If someone can help with the Altair syntax I will amend my pull request.

jakevdp:
I think you can just put the expression within the condition; e.g.

```python
color=alt.condition(alt.datum.year == 2000,
                    alt.value('orange'),
                    alt.value('steelblue'))
```

palewire:
I've pushed a revised example.

visualization 14

```python
import altair as alt

population = data.population()

alt.Chart(population).mark_bar().encode(
    x="year:O",
    y="sum(people):Q",
    # The highlight will be set on the result of a conditional statement
    color=alt.condition(
        alt.datum.year == 1970,  # If the year is 1970 this test returns True,
        alt.value('orange'),     # which sets the bar orange.
        alt.value('steelblue')   # And if it's not true it sets the bar steelblue.
    )
)
```

jakevdp:
Looks great! Seems that some pipfile stuff snuck into the PR. Remove that and we'll be good to go.

palewire:
Done.

Issue #883, opened 2018-05-22
DOC: Some of the examples should be considered for retitling or reorganization


palewire:
IMHO, some of the examples in the Example Gallery are framed too tightly around complex projects to make sense as generic examples in the categories we have now.

The best examples, IMHO, are the ones that have a clear title describing the Altair feature or chart trick being documented, rather than the data content of the chart.  

I believe the titles below should either be regrouped in a "Case Studies" section that features more intricate projects, or retitled with more focus on the techniques they teach.

* Cumulative Wikipedia Donations
* Anscombe’s Quartet
* Becker’s Barley Trellis Plot
* Locations of US Airports
* London Tube Lines
* One Dot Per Zipcode
* U.S. state capitals overlayed on a map of the U.S
* Seattle Weather Interactive
* US Population Over Time
* Binned Movie Rating Heatmap

eitanlees:
- [Falkensee](https://altair-viz.github.io/gallery/falkensee.html)
- [Gapminder Bubble Plot](https://altair-viz.github.io/gallery/gapminder_bubble_plot.html)
- [Natural Disasters](https://altair-viz.github.io/gallery/natural_disasters.html)

might also fall into this category

jakevdp:
I think this was addressed by @palewire's work last year

PR #903, opened 2018-05-28
Reorganized complex examples into a new case studies section

palewire:
As proposed in #883

### DIFF
--- a/altair/sphinxext/altairgallery.py
+++ b/altair/sphinxext/altairgallery.py
@@ -244,6 +244,7 @@ def main(app):
         'Histograms': [],
         'Maps': [],
         'Interactive Charts': [],
+        'Case Studies': [],
         'Other Charts': []
     })
     for d in examples:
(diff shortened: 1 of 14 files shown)


jakevdp:
I think users of the site might be confused by what "case studies" means in this context.

For example, we already have a "case studies" tab in the documentation that links to a narrative walkthrough of exploring a dataset using progressively more sophisticated charts. When you mentioned turning complicated charts into case studies, I thought you were talking about giving them that kind of treatment.

jakevdp:
I do think it makes sense to somehow delineate between highly-tweaked charts and simple charts, though.

jakevdp:
I think "Customized Charts" might make sense. Is that the main thing that separates these?

kanitw:
FWIW, we plan to add "Customized Chart" / "Custom Visual Design" to Vega-Lite gallery too once we have a few more examples.

palewire:
My thought is that this set is primarily about creating the best possible
chart to illustrate a particular dataset.

That is a contrast to the bulk of examples, which are the reverse: The best
possible dataset to illustrate a particular chart technique.

jakevdp:
Sounds good – I think "case studies" will work then... it might be worth trying to add some mechanism to add summary text at the top of each example section, but that can come later.

Thanks!

Issue #695, opened 2018-04-02
Better access to logical operand predicates


jakevdp:
Vega-Lite has a number of logical operands; e.g.

- ``alt.LogicalAndPredicate``
- ``alt.LogicalOrPredicate``
- ``alt.LogicalNotPredicate``

These come up, for example, in constructing filter transforms. It would be useful to make these available to Altair users via Python's logical operators, ``&``, ``|``, and ``~``, respectively.

There are two challenges:

1. How to implement them effectively... maybe create some sort of ``PredicateMixin`` for these classes? If we were constructing the wrapper classes by hand, it would make sense for ``alt.Predicate`` to be the base class of each of these logical predicates... perhaps we should change the code generation framework so that this kind of class hierarchy is (optionally) inferred in generated classes? Sounds hard to do well in a general way, though.

2. The interaction with selection operands (e.g. ``alt.SelectionOr``, ``alt.SelectionAnd``, ``alt.SelectionNot``, and ``alt.NamedSelection``) makes this a bit tricky, because ``NamedSelection`` objects serve two purposes in Altair (they act both as a ``alt.SelectionDef`` and a ``alt.SelectionPredicate`` depending on the context).

kanitw:
I think using `&`, `|`, `~` is definitely way more natural. 

We have to do `{"and/or": [...]}` in Vega-Lite because we're limited to be just a JSON format.

pagpires:
A question regarding the **combination of multiple operands**: can we do that in altair now?
I met an error when I wanted to combine 'or' and 'not':
```python
pop = vega_datasets.data.population()
alt.Chart(pop).mark_line().encode(
    x='age:O',
    y='sum(people):Q',
    color='year:O'
).properties(
    width=600, height=200
).transform_filter(
    {'or': [alt.FieldRangePredicate(field='year', range=[1850, 1880]), 
            alt.FieldRangePredicate(field='year', range=[1960, 2000])],
     # if I remove the following line and only keep 'or', it will work
    'not': alt.FieldRangePredicate(field='year', range=[1870, 1880])
    }    
)
```

jakevdp:
Yes, you can do them in Altair. But no, you can't have ``or`` and ``not`` within the same object... I'm not sure how to interpret your intent here.

Is this what you were intending to do?
```
{'and': [
        {'or': [alt.FieldRangePredicate(field='year', range=[1850, 1880]), 
                alt.FieldRangePredicate(field='year', range=[1960, 2000])]},
[...]

PR #923, opened 2018-06-07
WIP: update Vega-Lite to version 2.5

jakevdp:
A few tasks before merging

- [ ] ~wrap new field predicates~ (actually, field predicates are not wrapped at the moment; see #695)
- [x] handle ``.dsv`` data format
- [ ] deal with ``SortField`` backward incompatibility somehow?
- [x] update docs on above features


jakevdp:
I can't think of any good way to deal with the ``SortField`` change. If ``SortField`` were now not used in the schema, we could alias it to ``EncodingSortField`` so that old code will continue working. But since ``SortField`` still exists but refers to something different, that will serve to create confusion in the case that someone wants to use the new version of ``SortField``.

I think we just need to stomach the API breakage and move on, though it will cause issues for some people when upgrading from 2.1 to 2.2.

I know @palewire has used ``SortField`` before, and others probably have too...

kanitw:
We could think of another name and rename SortField in VL.

Ill be out of town for a week though so might nit be able to change that until i cole back

palewire:
At this stage, I'm okay with eating API changes. I just wouldn't want to lose the functionality. What is it changing to?

jakevdp:
OK, I think we'll just swallow the ``SortField`` -> ``EncodingSortField`` change at this point, then.

Issue #929, opened 2018-06-10
scale_factor causes chromedriver to timeout intermittently


jploudre:
I'm generating about 1200 graphs with a loop. When I try to use:

chart.save with scale_factor = 2,

it repeatedly crashes after about 300 graphs -- not at the same spot. Replicated on 2 different computers. Timeouts resolve when removing scale_factor.

(That said, **Love** scale_factor.)

jakevdp:
Thanks for the report

jakevdp:
Fixed in #939

PR #939, opened 2018-06-11
Longer webdriver timeout

jploudre:
Allows altair to not time out on less responsive development computers (like my slow laptop). Addresses Issue #929


jakevdp:
Thanks – looks like this was based on an older commit. Can you resolve the conflicts?

Issue #940, opened 2018-06-12
Sorting of alt.Columns should be similar to alt.Colors?


brainstorm:
While trying to apply a custom ordering to the `columns` chart spec here:

```python
logplot = alt.Chart(dataset_curated).mark_bar().encode(
        alt.X('technology:N', scale=alt.Scale(rangeStep=15), axis=alt.Axis(title='')),
        y = 'log:Q',
        color = alt.Color('sample:N'),
        column = 'chromosome:N'
        #column = alt.Column('chromosome:N', sort=alt.SortField(order=chr_array))
)
```

The column strings seem to be ordered lexicographically, not ideal for i.e, chromosomes:

<img width="918" alt="skarmavbild 2018-06-12 kl 10 49 10" src="https://user-images.githubusercontent.com/175587/41264112-4afadcac-6e2e-11e8-8d33-2ea5eccd9c28.png">

So my pandas Categorical ordering, which works for raw dataframe ordering (chromosomes ordered by array's positions):

```python
chr_array = ['chr'+str(x) for x in range(1,23)]
chr_array.append('chrX')
chr_array.append('chrY')
chr_array.append('chrM')

dataset_curated["chromosome"] = pd.Categorical(dataset_curated["chromosome"], chr_array)
dataset_curated = dataset_curated.sort_values("chromosome")
```

Is ignored by Altair's column ordering :_/ Also, according to the API reference for [alt.Column](https://altair-viz.github.io/user_guide/API.html?highlight=altair%20column#altair.Column), that only supports `ascending` and `descending` `sort=` arguments, unlike i.e [alt.Colors](https://altair-viz.github.io/user_guide/API.html?highlight=altair%20color#altair.Color) which supports a manually ordered `List(String)`.

I'm currently trying to figure out some workaround while reading through https://github.com/altair-viz/altair/issues/245, https://github.com/altair-viz/altair/pull/899, https://github.com/altair-viz/altair/issues/898 and https://github.com/altair-viz/altair/issues/397, I just wanted this usecase to be noted ;)

jakevdp:
Altair data is encoded as JSON, which does not have a categorical type, so all pandas categories will be converted to simple strings.

If you'd like to specify the explicit order in which a nominal category will appear, you can use the ``sort`` argument. For example:
```python
import altair as alt
import pandas as pd

data = pd.DataFrame({'a': list('ABC'),
                     'b': [28, 55, 43]})

alt.Chart(data).mark_bar().encode(
[...]

PR #941, opened 2018-06-12
Do not convert categoricals to strings

brainstorm:
Related with #940 as I explore a way to solve the columnar sorting...

https://github.com/pandas-dev/pandas/issues/10778 is fixed upstream now

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -87,7 +87,6 @@ def sanitize_dataframe(df):
 
     * Make a copy
     * Raise ValueError if it has a hierarchical index.
-    * Convert categoricals to strings.
     * Convert np.bool_ dtypes to Python bool objects
     * Convert np.int dtypes to Python int objects
     * Convert floats to objects and replace NaNs/infs with None.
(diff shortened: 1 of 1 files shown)


jakevdp:
What's the benefit of removing this conversion?

brainstorm:
Perhaps down the line Altair will be able to sort (columns) based on categoricals? If we cast to `str` that metadata (ordering) is lost?... but granted, I don't have a global understanding of Altair yet :-S

jakevdp:
After sanitization, the data is converted to JSON, in which case categoricals are converted to strings anyway. JSON does not have a categorical type, so there's no way to maintain sorting of categories when converting the data to the format needed for Altair.

The sanitization code serves to ensure that people with older versions of Pandas don't get an error – it doesn't change anything about how data is represented in the end.

brainstorm:
Oh, I see, backwards compat issues, gotcha :-S

In any case, that bug has been solved 2 years ago, so worst case it'll prompt people to upgrade their pandas installations? If the conversion to str happens downstream anyway when converting to JSON, this PR could be labeled as a maintenance enhancement since it does remove `XXX`'s.

jakevdp:
I don't really see any benefit to the change, and there is the distinct downside that it will re-introduce a bug that was already fixed for previous pandas versions. There are many situations in which people might be using old library versions and not be able to easily upgrade.

I'm going to close this – thanks for the contribution!

PR #983, opened 2018-06-28
docs: fix chart not displaying and delete mention of field * in Seattle case study

lincolnfrias:
- changing 'month(date:T)' to 'month(date):T'
- deleting mention of special field *


jakevdp:
Thanks!

PR #992, opened 2018-07-03
MAINT: update vega-lite to version 2.6.0

jakevdp:
Update vega-lite and auto-generated wrappers.


jakevdp:
This doesn't contain any backward-incompatible changes as far as I can tell; tests pass so I think it's safe to merge into master.

Issue #976, opened 2018-06-27
Is there a way to hide the top and right border "axis" lines?


palewire:
If I export a transparent backed SVG from Altair it still carries a thin stroke on the top and right axes. Is there a way to remove them?

iliatimofeev:
I guess recipe from #969 would help.

lincolnfrias:
Use `.configure_view(strokeOpacity=0)`. See example in #975.

palewire:
That worked,@lincolnfrias. Thank you. 

This should be documented at the very least. Where could that best go? I'd be happy to add it.

Also, am I crazy, or should there be a more directly API for this. If I can say `grid=False` should there be a similar option for this? Or perhaps the default should not have stroke?

jakevdp:
This example could go in the configuration docs: https://altair-viz.github.io/user_guide/configuration.html

Source here: https://github.com/altair-viz/altair/blob/master/doc/user_guide/configuration.rst

Currently it mostly contains a list of available keywords, but adding some examples would be very useful.

nemo-omen:
I know this is pretty old at this point, but I want to leave an answer here for anyone who's looking.

You can remove the axis borders by setting `domainOpacity=0`, in `configure_axis`, or one of its varients.

dzonimn:
Hi, sorry to resurrect an old thread, but I found this to be the most relevant thread out of all the other similar ones that I've found. 

I'm trying to achieve something like this, but only on one specific chart type within a concatenated layered chart. If I specify `configure_view(strokeOpacity=0)` on the desired chart, it's not possible as concat charts would rather you specify top-level configurations at the end. If I specify it at the end, every single chart in the concatenation will be affected.

Perhaps I am approaching this the wrong way, so I think it would be more helpful to specify what I'm trying to do. I want to be able to generate a concatenated vertical chart like so:

## Title of VConcat Chart
### 1st Image and Subtitle (encoded as a mark_image + mark_text layered chart)
1st Chart (with view strokeOpacity = 1)
### 2nd Image and Subtitle
2nd Chart

and so on..

I have tried giving each of the chart objects a title, but that gives me no ability (to my knowledge) to append an image/icon to the front of the title. It also leads to strange alignment issues as the system I'm working with (Kibana) works on an older Vega spec that doesn't have text anchors, so I decided to try it this way. At worst, I will just set everything to strokeOpacity=0. I am curious if there is some other way to approach this.

PR #1009, opened 2018-07-10
DOC: Example of hiding the border using configure_view.

palewire:
For #976.

This is beyond the scope of this patch, but I noticed that the automatically generated documentation for `configure_view` is a little confusing.

It describes the default of `strokeOpacity` as "The stroke opacity (value between [0,1]).  Default value: (none)." But an unconfigured chart does has a stroke. So how can the default be None?

### DIFF
--- a/doc/user_guide/configuration.rst
+++ b/doc/user_guide/configuration.rst
@@ -171,7 +171,9 @@ View Configuration
 Removing the border
 ~~~~~~~~~~~~~~~~~~~
 
-By default, charts have an outside border. Hide it by setting the `strokeOpacity` option to `0`.
+By default, charts have both a grid and an outside border. To create a chart with no border, you will need to remove the both.
+
+As an example, let's start with a simple scatter plot.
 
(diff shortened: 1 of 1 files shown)


jakevdp:
Thanks!

I think this example should probably be modified a bit: because of the grid, I don't see any difference between strokeOpacity of zero and strokeOpacity of 1 with this chart.

Regarding the documentation: that comes directly from the ``description`` field in the Vega-Lite schema (see https://github.com/vega/schema/blob/master/vega-lite/v2.6.0.json#L1701) Any modification to that would have to be made in Vega-Lite itself.

palewire:
Thanks for the quick feedback. I'll take another pass.

jakevdp:
Maybe a progression would be helpful – here's how you remove gridlines, but there's still a border. Here's how you remove the border

palewire:
Changes are in.

jakevdp:
This looks great. Thanks so much!

PR #1019, opened 2018-07-14
DOC: Use relative links for images

pratapvardhan:
Earlier image links were talking `/`  root path. This pull request changes it to use relative URLs, making the docs work locally, or as under subdirectory of a website as well.

### DIFF
--- a/altair/sphinxext/altairgallery.py
+++ b/altair/sphinxext/altairgallery.py
@@ -204,7 +204,7 @@ class AltairMiniGalleryDirective(Directive):
             if size:
                 examples = examples[:size]
 
-        include = MINIGALLERY_TEMPLATE.render(image_dir='/_static',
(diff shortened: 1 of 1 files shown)


jakevdp:
Sorry for such a delay in reviewing this.

Looking more closely, I'm a bit confused about the intent of this PR. Within sphinx, I believe that the ``_static`` directory generally appears only in the root of the site. So by switching from absolute paths to relative paths, it actually makes the location of the gallery *less* flexible. After this change, for example, if I try to create a minigallery anywhere but the root directory, it will no longer work. Before this change, the minigallery will work anywhere within a sphinx website.

What problem were you having that was fixed by this change?

jakevdp:
Since there's been no response here, I'm going to close this PR. Feel free to re-open if you'd like to continue the discussion

PR #1020, opened 2018-07-14
REF: stem and leaf example with better pandas

pratapvardhan:
Larges two changes to stem and leaf plot example

- Compact random data construction, stem and leaf using string methods
- Use pandas cumcount instead of apply


jakevdp:
Looks great – thanks!

Issue #1023, opened 2018-07-16
Save charts without internet connection


SyntaxRules:
My current use case with altair is that I must save charts without an internet connection. This means the `chart.save(...)` function does not work for me. I'd like to change this, and I'm happy to invest some time to do so. Before working on a PR, I want to make sure that my work is going in the right direction. 

As for implementation here is what I'm thinking at the moment. The `generate_schema_wrapper.py` file will also download the necessary *.js files from `https://cdn.jsdelivr.net/npm/{library}@{version}` and store them with the other version specific generated files. These files will then be referenced in the `chart.save()` method.

Does this implementation sound like the right way to go or is there a better approach? Is there any opposition to including these `*.js` files in the release package?

jakevdp:
I think I would prefer to not bundle javascript sources with Altair, which is what would be required for the solution you have in mind.

Note that currently all javascript sources are in the frontend renderers (jupyter notebook, jupyterlab, vegascope, nteract, etc.) and not in Altair.

Maybe the way to go would be to, for example, somehow detect if the jupyterlab extension is available, and use the JS sources from there if possible?

SyntaxRules:
I'm terribly unfamiliar with how jupyterlab extensions are used and imported, so any pointers you have to start me in that direction are appreciated. 

As I stated in the linked PR, the solution I introduced was to download the `*.js` files along side of the `*-schema.json` files when using the `generate_schema_wrapper.py` script. This works for my short term needs, and should be sustainable in a longer term *if altiar is ok with packaging and sending `*.js` files along with the `*-schema.json` files* and the rest of the code.

In any case https://altair-viz.github.io/user_guide/saving_charts.html needs to be updated to cover whatever solution we come up with here. (Even if that is "Internet connection required")

jakevdp:
This is now possible via http://github.com/altair-viz/altair_saver

My hope is to incorporate this into the next release of Altair.

PR #1025, opened 2018-07-16
feat(save): Change save to work offline

SyntaxRules:
Does this by saving the appropriate *.js to a js folder when the schema
is generated. Packages these with pip to enable there use locally.

It might be better to get these sources from an already installed dependency as @jakevdp suggested but I haven't figured that out yet.

Addresses #1023


jakevdp:
I think I'd like to avoid bundling javascript sources with Altair. For the time being, I'd suggest creating a separate small package that does offline saving of vega/vega-lite charts. It could be hosted in this org if you would like. We may even think about moving all of the selenium/headless logic into that separate package.

jakevdp:
You can currently use Altair and save figures without Chromedriver – the dependency is entirely optional.

SyntaxRules:
As it's been decided that the offline save feature would be better suited in a separate package. I'm closing this PR. I'll open another one when I get the separate package working.

PR #1026, opened 2018-07-16
DOC: Standardized stacked bar examples

palewire:
I've standardized three of the stacked bar chart examples so they use the same the dataset and similar, stripped down configurations. I believe this will make it easier for beginners to grok.


jakevdp:
Great idea – I think this will be much clearer for people learning the library.

Issue #981, opened 2018-06-27
ENH: Move all datasets to top level by default?


jakevdp:
#951 added this as an option; I think it should be the default.

Advantage is that it leads to more compact specifications for compound charts.

Disadvantage is that it might break things, though it seems to work just fine in all the cases I've tried.

PR #1046, opened 2018-07-26
Enable dataset consolidation by default

jakevdp:
Addresses #981


jakevdp:
I've re-run all charts in the example gallery using this change, and confirmed that they render without error... and at least by eye seem to look right.

PR #1063, opened 2018-07-31
DOC: use autosummary to clean up API docs

jakevdp:
The API.html file was becoming quite large (3.8MB in master). This PR uses sphinx autosummary to split the API docs between multiple files, with a nice table summary in the index.


jakevdp:
Last couple issues:

- ~700 sphinx warnings of the form ``WARNING: Duplicate explicit target name: single``. Fixing this requires a patch to m2r (see https://github.com/miyakogi/m2r/pull/26)
- ``altair.require.rst`` and ``altair.Require.rst`` conflict on non-linux systems. Maybe should use two ``generated`` directories?

PR #1076, opened 2018-08-09
Docs regarding width of bar marks

mattijn:
Includes explanation and some examples as discussed in here https://github.com/altair-viz/altair/issues/1050

### DIFF
--- a/doc/user_guide/customization.rst
+++ b/doc/user_guide/customization.rst
@@ -290,3 +290,88 @@ You can remove the legend entirely by submitting a null value.
       y='petalLength',
       color=alt.Color('species', legend=None),
   )
+
+Adjusting the width of Bar Marks
+--------------------------------
+The width of the bars in a bar plot are controlled through the `size` property in the `mark_bar()`:
+
+.. altair-plot::
+
+  import altair as alt
+  import pandas as pd
+
+  data = pd.DataFrame({'name': ['a', 'b'], 'value': [4, 10]})
+
+  alt.Chart(data).mark_bar(size=10).encode(
+      x='name:O',
+      y='value:Q'
+  )
+
+But since `mark_bar(size=10)` only controls the width of the bars, it might become possible that the width of the chart is not adjusted accordingly:
+
+.. altair-plot::
+
+  import altair as alt
+  import pandas as pd
+
+  data = pd.DataFrame({'name': ['a', 'b'], 'value': [4, 10]})
+
+  alt.Chart(data).mark_bar(size=30).encode(
+      x='name:O',
+      y='value:Q'
+  )
+
+The width of the chart containing the bar plot can be controlled through two mechanisms:
+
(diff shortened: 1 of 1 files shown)


doc/user_guide/customization.rst
```diff
@@ -290,3 +290,88 @@ You can remove the legend entirely by submitting a null value.
...
+  )
+
+But since `mark_bar(size=10)` only controls the width of the bars, it might become possible that the width of the chart is not adjusted accordingly:
+
+.. altair-plot::
+
+  import altair as alt
+  import pandas as pd
+
+  data = pd.DataFrame({'name': ['a', 'b'], 'value': [4, 10]})
```
jakevdp: This data should still be available to this plot directive since it is defined in the previous one... no need to re-define it here.

Same comment applies in several places below.

doc/user_guide/customization.rst
```diff
@@ -290,3 +290,88 @@ You can remove the legend entirely by submitting a null value.
...
+  alt.Chart(data).mark_bar(size=30).encode(
+      x='name:O',
+      y='value:Q'
+  )
+
+The width of the chart containing the bar plot can be controlled through two mechanisms:
+
+1. Setting the `width` of chart, so the width of the bars are adjusted to fit the width of the chart.
+
+2. Setting the `rangeStep` of the bars, so the width of the chart becomes the number of bars multiply the `rangeStep`.
```
jakevdp: I would clarify explicitly here that `rangeStep` means the width (in pixels) that the chart allocates for each bar.

doc/user_guide/customization.rst
```diff
@@ -290,3 +290,88 @@ You can remove the legend entirely by submitting a null value.
       y='petalLength',
       color=alt.Color('species', legend=None),
   )
+
+Adjusting the width of Bar Marks
+--------------------------------
+The width of the bars in a bar plot are controlled through the `size` property in the `mark_bar()`:
```
jakevdp: Please change to double back-ticks for literals like ``size``.

For referencing methods, please use
```
:meth:`~Chart.mark_bar`
```
which will automatically be turned into a clickable a cross-reference.

doc/user_guide/customization.rst
```diff
@@ -290,3 +290,88 @@ You can remove the legend entirely by submitting a null value.
...
+
+  data = pd.DataFrame({'name': ['a', 'b'], 'value': [4, 10]})
+
+  alt.Chart(data).mark_bar(size=30).encode(
+      alt.X('name:N', scale=alt.Scale(rangeStep=100)),
+      y='value:Q'
+  )
+
+The width of the bars are set using `mark_bar(size=30)` and the width of the chart is set using `alt.Scale(rangeStep=100)`
+
```
jakevdp: I think everything below here could be left out, and replaced with a note block:

```rst
.. note::

   If both ``width`` and ``rangeStep`` are specified, then ``rangeStep`` will be ignored.
```

PR #1078, opened 2018-08-10
ADD: Trellis Area Sort Chart Example

pratapvardhan:
Example of Trellis Area Chart with sort order.

```python
import altair as alt
from altair.expr import datum


alt.Chart(data.stocks.url).transform_filter(
    datum.symbol != 'GOOG'
).mark_area().encode(
    x='date:T',
    y='price:Q',
    color='symbol:N',
    row=alt.Row('symbol:N', sort=['MSFT', 'AAPL', 'IBM', 'AMZN']
    )
).properties(height=50, width=400)
```
image


jakevdp:
Very nice, thanks!

PR #1095, opened 2018-08-15
WIP: Update to Vega-Lite 3.0

jakevdp:
~~Current blocker is the fact that VL 3.0 somewhere uses a list of schemas within an ``items`` argument for a ``type: 'array'``. This is not supported by schemapi, so the code cannot be generated.~~

~~I'm looking into the best way to fix this... it's a bit delicate because it breaks the fundamental (incorrect) assumption in schemapi that every schema argument is a dict.~~

Update: fixed it!


jakevdp:
Everything seems to work... note that the ``vegalite.v3`` mimetype is not yet supported by the notebook or jupyterlab extensions, so I've been testing things by running ``chart.serve()`` or ``chart.save('chart.html')``

jakevdp:
We should wait until the VL 3.0 final release before merging this. A few things also to take care of:

- Make sure VL 3.0 JupyterLab & Jupyter Notebook extensions are available
- add ``transform_`` methods for new transforms
- add ``mark_`` methods for new composite marks
- other API changes?

jakevdp:
Closing because there are a lot of moved files that have been changed in master; it will be better to start fresh.

PR #1107, opened 2018-08-21
Appropriately handle non-string column names

jakevdp:
Replaces #899


jakevdp:
@ellisonbg let me know if this looks OK

PR #1112, opened 2018-08-23
Add link to altair slack

domoritz:
I'm not sure what you think is the right role of the slack channel but I thought it would be good to have the link in the readme.


jakevdp:
Thanks!

PR #1120, opened 2018-08-27
Upgrade Python syntax with pyupgrade

hugovk:
https://github.com/asottile/pyupgrade is a tool to automatically upgrade syntax for newer versions of Python.

### DIFF
--- a/altair/_magics.py
+++ b/altair/_magics.py
@@ -59,7 +59,7 @@ def _prepare_data(data, data_transformers):
     elif isinstance(data, six.string_types):
         return {'url': data}
     else:
-        warnings.warn("data of type {0} not recognized".format(type(data)))
+        warnings.warn("data of type {} not recognized".format(type(data)))
         return data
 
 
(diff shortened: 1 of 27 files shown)


jakevdp:
Interesting... aside from one dict comprehension, it seems like the main effect is to change format strings of the form ``"... {0} ... {1} ..."`` to ``"... {} ... {} ..."``

I personally prefer the former over the latter (explicit is better than implicit, etc.) Do you know if there's any compelling reason to make this change?

hugovk:
I find it more readable: 

If I see `"{} ... {}"` I instantly know they are in order.

If I see numbers, I know they might not be in order, and have to pause to read the numbers to see if they're in order -- `"{0} ... {1}"` -- or in reverse order -- `"{1} ... {0}"`.

Or if there's more than one:

* `"{0} ... {0} ... {0}"`
* `"{0} ... {0} ... {1}"`
* `"{0} ... {1} ... {1}"`
* `"{1} ... {0} ... {1}"`
* etc.

It's that little bit quicker to read: 

* No numbers mean in order. 
* Numbers mean something special is happening.

jakevdp:
Thanks – I didn't know that!

I'll go ahead and merge then

PR #1122, opened 2018-08-29
README to show how to edit the documentation

kindofluke:
Unsolicited lessons learned from updating the altair documentation.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -84,7 +84,7 @@ With this convention, your example will build automatically and be placed on the
 
 When updating the user guide the following conventions are helpful: 
 
-* Add the directive ``.. currentmodule:: altair`` somewhere near the top of your documentation page so that you inform the renderer how to link to docstrings in the classes and functions. pip
+* Add the directive ``.. currentmodule:: altair`` somewhere near the top of your documentation page so that you inform the renderer how to link to docstrings in the classes and functions.
 * When referring to an Altair class or function preface it with ``:class:`` or ``:func:`` and the rendering engine will automatically link it to the appropriate section in the API reference. 
 * Link to examples using the following syntax 
     ```
--- a/doc/README.md
+++ b/doc/README.md
@@ -84,6 +84,7 @@ With this convention, your example will build automatically and be placed on the
 
 When updating the user guide the following conventions are helpful: 
 
+* Add the directive ``.. currentmodule:: altair`` somewhere near the top of your documentation page so that you inform the renderer how to link to docstrings in the classes and functions. pip
 * When referring to an Altair class or function preface it with ``:class:`` or ``:func:`` and the rendering engine will automatically link it to the appropriate section in the API reference. 
 * Link to examples using the following syntax 
     ```
--- a/doc/README.md
+++ b/doc/README.md
@@ -1,0 +1,93 @@
+# Contributing to Altair Documentation
+
+Altair documentation is written in [reStructuredText](http://docutils.sourceforge.net/rst.html) and compiled into html pages using [Sphinx](http://www.sphinx-doc.org/en/master/). Contributing to the documentation requires some extra dependencies and we have some conventions and plugins that are used to help navigate the docs and generate great Altair visualizations. 
+
+## Building the Documentation locally
+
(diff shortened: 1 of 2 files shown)


doc/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+# more plot code
+```
+With this convention, your example will build automatically and be placed on the gallery page. 
+
+**Updating the User Guide**
+
+When updating the user guide the following conventions are helpful: 
+
+* When referring to an Altair class or function preface it with ``:class:`` or ``:func`` and the rendering engine will automatically link it to the appropriate section in the API reference.
```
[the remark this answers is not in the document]
jakevdp: Maybe also note that the ``.. currentmodule: altair`` directive needs to be used somewhere in the file if you don't want to type the full function name.

doc/README.md
```diff
@@ -0,0 +1,93 @@
...
+pip install -r requirements.txt
+pip install -e . 
+pip install -r doc/requirements.txt
+```
+In addition, the sphinx documentation builder needs access to Selenium to support generating images of sample visualizations. You may have selenium already installed but if you do not it can also be installed via pip but is not in the requirements file because of its size and dependencies. 
+
+```
+pip install selenium
+```
+
```
eitanlees: I tried doing a fresh install and I had to run 
```
pip install recommonmark
```
as well to make things work. Should `recommonmark` be added to `doc/requirements.txt`? @jakevdp

kindofluke:
@eitanlees thanks for your review. 

I just pushed some changes based on your feedback. Specifically in the first paragraph, I referred people to the main `README` located in the parent directory where there is a section on setting up for development. 

My thoughts are a discussion of setting up a virtual env and branch checkouts is ideally handled there. The most important thing to know when changing the docs is that you need to set-up all the dependencies for Altair, locally install Altair, AND set-up dependencies for the docs and I think this version currently gets that across. 

Another reason I am glossing over it slightly is that I don't use Anaconda nor Windows a discussion of the virtual environments really needs some notes about both those cases. 

@jakevdp added server command and model import 

also added recommonMark to requirements though I am not sure where its actually used so maybe want to keep it out.

eitanlees:
I think pointing to the main `README` is a good idea. It cuts down on repetition.

As far as `recommonMark` it might be something weird on my end. Somehow I always manage to mess up virtual envs and fresh installs.

doc/README.md
jakevdp: I think ``recommonmark`` is part of ``requirements_dev.txt``. We should probably either add ``pip install requirements_dev.txt`` to the list of steps to follow, or add the sphinx-related requirements to ``doc/requirements.txt``

jakevdp:
Thanks @kindofluke and @eitanlees - sorry for the delayed review, I was traveling last week.

As I mentioned above, the ``recommonmark`` dependency is listed in ``requirements-dev.txt``. Thinking about it, it probably does not make sense to require people to install ``requirements-dev`` for simply building the documentation... so maybe ``recommonmark`` could be listed in ``doc/requirements.txt`` as well.

doc/README.md
```diff
@@ -0,0 +1,99 @@
...
+  linkcheck  to check all external links for integrity
+  doctest    to run all doctests embedded in the documentation (if enabled)
+  coverage   to run coverage check of the documentation (if enabled)
+  dummy      to check syntax errors of document sources
+
+```
+For most updates, run `make html` and the documentation will generate in a sub folder `_build`. You can open `_build/html/index.html` to view the documenation as local static files which should give you a good preview of how any updates will look. Even better, use the command below to serve the static files and you will be able to navigate the whole documentation website. 
+
+```
+cd doc/_buld/_html
```
timocb: `_buld` should be `_build`

mattijn:
@binste, I saw your comment regarding improving the READE (https://github.com/altair-viz/altair/discussions/2599#discussioncomment-4503613). Is this draft PR of interested for you?

binste:
Thanks @mattijn, it is indeed! You can close this PR and I will add the content from @kindofluke (with them as author) to the `CONTRIBUTING.md` file which is referenced from the README.

PR #1123, opened 2018-08-31
Refactor HTML output methods

jakevdp:
- adds ``alt.Chart.to_html()`` method
- adds Kaggle renderer, and refactors Colab renderer
- adds ability to specify whether html output is full or partial, as well as whether or not it uses require.js
- adds ``jinja2`` to the core requirements for better HTML templating

### DIFF
--- a/altair/utils/__init__.py
+++ b/altair/utils/__init__.py
@@ -5,11 +5,11 @@ from .core import (
     use_signature,
     update_subtraits,
     update_nested,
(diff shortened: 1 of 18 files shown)


jakevdp:
I checked this branch on colab and confirmed it works properly... still need to check on kaggle.

jakevdp:
I also checked this branch on kaggle and confirmed that it works correctly.

PR #1127, opened 2018-09-05
ENH: add nbviewer renderer

jakevdp:


jakevdp:
I'm pausing on this because I wonder if we shouldn't try to provide something like a "universal HTML renderer" that works whether requirejs is present or not. An example is in the mpld3 package.

jakevdp:
Superseded by #1793

PR #1131, opened 2018-09-11
Example of a waterfall chart

cuchoi:
Inspired in the Gantt chart example I created a waterfall chart: 
image
Let me know if you think it is a relevant example.

### DIFF
--- a/altair/vegalite/v2/examples/waterfall_chart.py
+++ b/altair/vegalite/v2/examples/waterfall_chart.py
@@ -13,7 +13,6 @@ data = pd.DataFrame([
     {"costs": "Fixed costs", "start": 3, "end": 8},
     {"costs": "Other costs", "start": 8, "end": 10},
     {"costs": "Total costs", "start": 0, "end": 10}
-
 ])
 
 waterfall_bars = alt.Chart(data).mark_bar(size=100).encode(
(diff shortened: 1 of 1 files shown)


cuchoi:
If this example is not insightful enough to justify a PR I can implement this waterfall: https://vega.github.io/vega-lite/examples/window_sum_waterfall_chart.html

The pro is showing Altair's flexibility, the disadvantage is the additional complexity of the example.

jakevdp:
Thanks for this example. One thing that gives me pause about this example is that the "start" and "end" are manually encoded, which makes it not all that applicable to many situations.

What do you think about reworking it so that you start with a single column of values, and then compute the start, end, and total either using pandas operations to preprocess the data, or using Altair transformers within the spec?

cuchoi:
Happy to do that. I am planning to do it with Pandas because I am not as familiar with Altair transformers. Will get back to you.

jakevdp:
Sounds good!

cuchoi:
Does this original dataset looks better? Copied below for convenience.

`data = pd.DataFrame([
    {"Fixed costs": 5,
     "Advertising costs": 3,
     "Other costs": 2},
])`

A know issue is that the x-axis labels are sorted alphabetically, so this would only look right if the "Total cost" variable happens to be the last one in alphabetical order. Should I fix that? Didn't want to make the example more complex than it should be.

cuchoi:
Happy to use a different starting dataset if this doesn't cut it.

mattijn:
Thanks for this PR @cuchoi! Since this PR:https://github.com/altair-viz/altair/pull/2756 regarding a waterfall chart was merged recently I will close this one. Thanks again.

PR #1134, opened 2018-09-13
 add python2 support for saving a chart as png 

arlenk:
simple fix to all saving png files in python 2.7 per https://github.com/altair-viz/altair/issues/994


jakevdp:
Looks good, thanks!

PR #1149, opened 2018-09-29
DOC: add sorted bar chart example

eitanlees:
I thought an example of a sorted bar chart would be a good addition to the examples.


jakevdp:
Looks good, thanks!

stsievert:
Thanks for this @eitanlees!

PR #1156, opened 2018-10-08
allow other transformers to work when saving to other formats

afonit:
The changes the logic so that if you have registered a custom data_transformer - you will still be able to save the plot to the allowed formats (png, html etc...).

I ran into this when using gpdvega, I would get this error when saving to other formats:
```
TypeError: Object of type 'Polygon' is not JSON serializable
```

With this logic change - I can now save using another registered data_transformer.

### DIFF
--- a/altair/vegalite/v2/api.py
+++ b/altair/vegalite/v2/api.py
@@ -403,11 +403,11 @@ class TopLevelMixin(mixins.ConfigMethodMixin):
         # By default we override the data transformer. This makes it so
         # that save() will succeed even for large datasets that would
         # normally trigger a MaxBinsError
-        if override_data_transformer:
+        if data_transformers.active != 'default':
+            result = save(**kwds)
+        else:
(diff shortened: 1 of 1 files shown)


jakevdp:
The reason this switch was put here is because saving does not work if the JSON transformer is enabled.

I think with this change, that common use-case would be broken again.

Can you not fix the error you're getting by setting ``override_data_transformer=False``, rather than changing the source code?

afonit:
@jakevdp you are quite right - I was really over thinking this.

Thank you and I apologize for the noise. I will close the pull request.

jakevdp:
One way we could do this is to special-case the json and csv transformers, and only do the temporary switch to default for those. What do you think?

afonit:
I actually really like how you have it now - your comments gave me the perspective I needed and cleared up the misunderstandings I had.  I really just got too tied up on an idea and was not seeing the bigger picture.

PR #1159, opened 2018-10-13
DOC: Create Filled Step Chart example

palewire:
One chart variation I find myself making occasionally is a step chart with its area filled in. I've added one as an example in this patch by making a few modifications to the existing step chart example. 

visualization

### DIFF
--- a/altair/vegalite/v2/examples/filled_step_chart.py
+++ b/altair/vegalite/v2/examples/filled_step_chart.py
@@ -1,0 +1,20 @@
+"""
+Filled Step Chart
+-----------------
+This example shows Google's stock price over time as a step chart with its area filled in. 
+"""
+# category: line charts
+import altair as alt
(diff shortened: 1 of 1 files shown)


jakevdp:
One thing that was added in the most recent vega-lite release was a ``line=True`` option for ``mark_area``. It can make this chart a bit more concise:

```python
import altair as alt
stocks = data.stocks()

alt.Chart(stocks).mark_area(
    color="lightblue", line=True
).encode(
    x='date',
    y='price'
).transform_filter(
    alt.datum.symbol == 'GOOG'
).configure_mark(interpolate='step-after')
```

We lack examples of that in the docs currently, so this would be a good candidate!

palewire:
I just learned something. Thanks for the tip. I'll make that change.

jakevdp:
Very nice, thanks!

PR #1160, opened 2018-10-13
Removed datum imports from example gallery

palewire:
A handful of examples in the gallery used imports like this:

```python
from altair.expr import datum
```

I changed them to use `alt.datum` instead.

I propose this patch because I think it will reduce the number of imports newbies need to learn and enforce consistency in our examples, which is a good thing in itself.


jakevdp:
I like that idea – thanks!

PR #1161, opened 2018-10-13
DOC: Standardize data import style and variable names

palewire:
Our examples have evolved to have a number of different styles for importing and naming data from vega_datasets. To avoid confusion and make life easier for the n00bs, I've attempted to standardize them here.  The style I'm aiming for is:

```python
source = data.whatever_dataset()
```

An obvious exception here are examples with multiple datasets, where I think more descriptive variable names are called for.


palewire:
I had to push through a couple bug fixes after my initial pull request. Apologies for that. I think the path is now ready for review.

jakevdp:
I think regularizing data imports is a great idea.

One complicating factor: in general in the examples I have been using URLs for data that does not come bundled with vega_datasets, to prevent flaky tests when they are run on travis or in places with no internet connectivity. A few of the changes in this PR broke this convention.

palewire:
Ah. I did not know that was a rule. I had crafted with the only using URL for datasets with greater than the 5,000 record limit. Should we add them to vega_datasets? If they're in the examples here, it seems like they merit inclusion there.

jakevdp:
There are a few issues with adding these to the bundled vega datasets:

- adding large files makes the vega_datasets installation larger. I don't want that package size to balloon too large.
- licensing is an issue. the vega_datasets package is MIT licensed, and in order to bundle the data with the code we need to make sure the data is public domain or otherwise compatibly licensed. Many of the datasets in vega/vega-datasets are of murky provenance, so that complicates these things.

palewire:
Ah. What do you think of redesigning the examples to only use data that is packaged in vega_datasets and meets our licensing requirement?

jakevdp:
> What do you think of redesigning the examples to only use data that is packaged in vega_datasets and meets our licensing requirement?

My feeling is that it's probably not worth the trouble. I think having examples of datasets loaded by URL is probably useful, as it shows that pandas dataframes are not a requirement.

palewire:
Okay, per our discussion, I have revised this pull request to use "url" imports for all vega_datasets that are not packaged with the Python package.

PR #1162, opened 2018-10-13
DOC: Trimmed slope_graph.py example

palewire:
The method for handling the x-axis layout offered by the Slope Graph example can be improved, I think, but substituting our handy dandy shorthand.


jakevdp:
very nice – thanks!

PR #1163, opened 2018-10-13
DOC: Removed pandas and numpy imports from examples

palewire:
A number of examples are cluttered by pandas and numpy imports at the top. I have removed them. In their place you will find references to the libraries via the `alt` variable. For instance, this:

```python
import pandas as pd

source = pd.DataFrame([whatever])
```

Has been replaced by:

```python
source = alt.pd([whatever])
```

I propose this change because it seems to me that pandas and numpy usage in the examples is primarily to generate test data. For that reason, including pandas code only distracts the user from the core purpose of the example, demonstrating Altair configuration options. Therefore, the less we see of it the better.

### DIFF
--- a/altair/vegalite/v2/examples/bar_chart_sorted.py
+++ b/altair/vegalite/v2/examples/bar_chart_sorted.py
@@ -5,9 +5,8 @@ This example shows a basic sorted bar chart.
 """
 # category: bar charts
 import altair as alt
-import pandas as pd
 
-data = pd.DataFrame({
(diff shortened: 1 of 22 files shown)


jakevdp:
I'm -1 on this change. The fact that ``pd`` is in the ``alt`` namespace is not intentional, and I'd rather use more idiomatic import styles for core libraries.

palewire:
Okie dokie. Withdrawn.

PR #1164, opened 2018-10-14
DOC: Standardize the creation of compound charts in examples

palewire:
The example gallery includes a variety of different ways for layering marks on top of each other. It would be nice if they all did the same way so that n00bs don't get confused and to popularize what we think are the best practices in chart assembly.

I took a pass at the ones I could spot to try to get them all on the same page. Take a look.


altair/vegalite/v2/examples/histogram_with_a_global_mean_overlay.py
```diff
@@ -7,15 +7,17 @@
 import altair as alt
 from vega_datasets import data
 
-source = data.movies.url
+source = data.movies()
```
jakevdp: We should use the URL, since this data is not cached locally in ``vega_datasets`` (and it will cause tests to fail as a result of poor web connectivity)

altair/vegalite/v2/examples/layered_chart_bar_mark.py
```diff
@@ -1,34 +1,33 @@
...
-LayerChart with Bar and Tick
-----------------------------
-This example shows how to layer two charts on top of one another.
+Bar and Tick Chart
+------------------
+How to layer a tick chart on top of a bar chart.
 """
 # category: bar charts
 import altair as alt
-import pandas as pd
```
jakevdp: I would prefer to import pandas explicitly.

altair/vegalite/v2/examples/simple_scatter_with_errorbars.py
```diff
@@ -6,36 +6,38 @@
 
 """
 # category: scatter plots
-
 import altair as alt
 
-import numpy as np
-import pandas as pd
```
jakevdp: Keep explicit imports of numpy and pandas

palewire:
The edits you have requested have been made.

jakevdp:
Looks good! I think your changes in #1161 conflicted with changes here. Can you fix the conflicts?

palewire:
I believe the conflicts have been resolved.

PR #1182, opened 2018-10-27
update troubleshooting guide for empty charts caused by adblocker

corriebar:
This would have helped me in this situation.

### DIFF
--- a/doc/user_guide/troubleshooting.rst
+++ b/doc/user_guide/troubleshooting.rst
@@ -240,6 +240,8 @@ by which a field can be specified within the full schema, and it is too difficul
 to account for all corner cases. Improving the user experience in this is a
 priority; see https://github.com/vega/vega-lite/issues/3576.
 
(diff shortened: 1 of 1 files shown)


jakevdp:
I think this is a useful addition, but it would be better if there were a few sentences about what causes the issue and how it can be fixed.

corriebar:
Unfortunately, I'm not quite sure what causes it. I assume it is because of some Javascript code that is blocked by common adblocker. If this is enough as an explanation, I can add it.

jakevdp:
Even if there's no explanation of why it causes an issue, a sentence about how to fix the issue would be useful. I don't know how to fix it, because I've never seen this issue myself.

corriebar:
I added a short sentence on how to fix the empty plot (deactivate adblocker).

jakevdp:
Thanks!

Kaligule:
Congratulations to @corriebar to your first successfull merge request. You can now safly delete your fork of altair. :tada:

PR #1185, opened 2018-11-01
"typing" package apparently creates issues if installed on Python3.7+

stonebig:
see  https://ci.appveyor.com/project/ericpre/hyperspy-bundle/builds/19660402/job/gs9l31okjhplckur

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,4 @@ numpy
 pandas
 six
 toolz
-typing
+typing;python_version<"3.5"


ericpre:
To be more precise, the issue only occurs with python 3.7 AFIAK and typing is a standard library since python 3.5 (https://docs.python.org/3/library/typing.html) so the fix of this PR seems appropriate.

jakevdp:
Thanks for the fix!

PR #1196, opened 2018-11-14
Fix the test notebooks

domoritz:
The notebooks were using an undefined `vegalite` property. The Vega-Lite 1 notebook used the v2 api.


domoritz:
Unfortunately, I could not render some of the older charts. I'll try again with https://github.com/jupyterlab/jupyter-renderers installed.

domoritz:
I'm stuck at https://github.com/jupyterlab/jupyter-renderers/issues/167. Can you rerun the notebooks for me?

domoritz:
I fixed them but there is a JS error in VL 1. Not sure what's the problem.

jakevdp:
I'd honestly be fine removing these notebooks if they're causing a problem. They're not part of any automated tests, and are of little value as documentation.

domoritz:
I deleted them. I think it's still good to have a separate commit with the fixed notebooks as the last commit before they get deleted.

jakevdp:
OK, thanks!

Issue #986, opened 2018-06-29
Examples not rendering in Altair's example gallery


louis-vines:
Hi, having recently switched to using Python for work and coming from an R background I've been really keen to start using Altair as it seems to be the closest python library to ggplot.

However quite a few of the examples aren't rendering in the examples gallery in my browser. For example, the simple highlighted bar chart:

https://altair-viz.github.io/gallery/simple_highlighted_bar_chart.html

here is screen grab of the page:

![image](https://user-images.githubusercontent.com/10738810/42117427-44e94366-7bf4-11e8-801b-0dd86ff7ebf2.png)

There are plenty of other examples that are exhibiting the same behaviour.

Having read a bit of the docs, this behaviour seems to be consistent with a chart that's incorrectly specified which would imply that the examples are out of date with the current version of altair/vega-lite. Or alternatively something is wrong with my setup, since these pages are already rendered this implies it must be a compatability issue with my browser - I'm running an up to date version of Google Chrome (Version 67.0.3396.87).

Any insight into this would be much appreciated as I'm quite keen to start adopting Altair into my workflows!

louis-vines:
Actually having looked into this a bit, I've found the problem - the current version of Altair doesn't seem to be able to accept URLs pointing to json. All the examples in the example gallery that use the `data.<dataset>.url` approach to pointing to the data seem to be failing.

If I change line 4 in the example above to:

```
population = data.population()
```
it works totally fine.

jakevdp:
All the examples work fine for me, and I'm also on Chrome 67.0.3396.

Do you have adblock turned on? Sometimes that prevents javascript from accessing external URLs, which is important in the case of Altair.

lincolnfrias:
I had a similar problem. The solution was to update the vega_datasets package.

jakevdp:
> I had a similar problem. The solution was to update the vega_datasets package.

I don't think that would have any effect on the rendering of charts on the website.

jakevdp:
@louis-vines were you able to check whether adblock affects this, and whether there were helpful errors in the javascript console? Thanks

gwdekker:
Confirming @jakevdp hypothesis for my case: I had the same problem, which got fixed by disabling my adblocker. I propose to add this to the FAQ section of the documentation.

FlorianWetschoreck:
The issue was fixed in the course of #1199 and can be closed

louis-vines:
[...]

PR #1198, opened 2018-11-15
fix data imports in the documentation in order to (partially) fix #1199 and #986

### LINKED ISSUE #1199: Documentation plots are not working with activated AdBlocker
I am using the AdBlock extension in Google Chrome.
This is blocking the loading of the external dataset. For example in
https://altair-viz.github.io/user_guide/compound_charts.html

The data and the plot is shown, if the data is loaded directly like in #1198

FlorianWetschoreck:
Currently, the documentation samples are broken:
- I added a fix via calling the data directly and not via the external JSON. However, this has inferior performance if I understand it correctly.
- other solutions might be discussed in #1199

example link for the broken documentation:
https://altair-viz.github.io/user_guide/compound_charts.html

### DIFF
--- a/doc/user_guide/compound_charts.rst
+++ b/doc/user_guide/compound_charts.rst
@@ -37,7 +37,7 @@ same data; for example:
     from altair.expr import datum
 
     from vega_datasets import data
-    stocks = data.stocks.url
+    stocks = data.stocks()
 
     base = alt.Chart(stocks).encode(
         x='date:T',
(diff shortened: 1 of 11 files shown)


FlorianWetschoreck:
The error was due to an interaction with the AdBlocker in the client browser, see #1199

jakevdp:
One reason I chose to use URLs here rather than explicitly embed the data is because when the data is used multiple times in once page, embedding it means that the page itself becomes very large. If we reference the data by URL multiple times in the page, good browsers will use cacheing and only do the download once.

So, to be clear, the tradeoff with this change is that we're serving adblocker users at the expense of mobile/low-bandwidth users. It's not obvious to me that that's the right choice.

PR #1207, opened 2018-11-21
DOC: Add example of bar chart with negative values

palewire:
visualization 1

```python
import altair as alt

source = "https://raw.githubusercontent.com/datadesk/vega-datasets/gh-pages/data/us-employment.csv"

    x="month:T",
    y="nonfarm_change:Q",
    color=alt.condition(
        alt.datum.nonfarm_change > 0,
        alt.value("steelblue"),  # The positive color
        alt.value("orange")  # The negative color
    ),
).properties(width=600)
```

This should not be merged until the next version of vega_datasets is released and the source line can be updated as follows:

```python

source = data.us_unemployment()
```

### DIFF
--- a/altair/examples/negative_bar_chart.py
+++ b/altair/examples/negative_bar_chart.py
@@ -1,0 +1,19 @@
+"""
+Bar Chart with Negative Values
+==============================
+This example shows a bar chart with both positive and negative values.
+"""
+# category: bar charts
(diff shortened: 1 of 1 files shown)


palewire:
This patch has been updated to use the new version of vega_datasets. If it meets your standards, it is ready to be merged.

PR #1212, opened 2018-11-21
DOC: Tidy up line chart with points example, move it down the gallery

palewire:
IMHO, the simple charts section at the top should only have one very basic example per chart type. Everything else should go below in the sections dedicated to each chart type.


jakevdp:
Makes sense, thanks!

PR #1213, opened 2018-11-21
DOC: Standardize highlighted bar chart examples

palewire:
Our highlighted bar chart examples are drastically different. Just as we improved our stacked bar examples by switching to a standard data source, I've done the same here. I've also cleaned up the titles and docs to bring them more into line.


jakevdp:
Thanks!

PR #1214, opened 2018-11-21
DOC: Regroup error bars chart in the example gallery

palewire:
IMHO, this is a not a traditional bar chart. Grouping it where we have it now has the potential to confuse readers.


jakevdp:
Looks good, thanks

PR #1224, opened 2018-11-24
DOC: Standardize the horizontal bar chart example

palewire:
I changed the horizontal bar chart example to bring it inline with recent edits to other bar charts, and clarify its documentation and mission.


jakevdp:
Looks good!

PR #1230, opened 2018-11-25
DOC: Change the default category to Other Charts

eitanlees:
I tried a fresh doc build to see all of the changes that have happened recently. It threw a `keyerror: 'General'` when running `altairgallery.py`

I poked around to try to figure out what was going on. It seems that when populating the examples the default category was 'general' but that category has since been abandoned. 

I changed the default to be 'other charts' which fixed the error when building the docs. 

Not sure if that is the best solution but it's a start.


jakevdp:
Looks good, thanks!

PR #1231, opened 2018-11-26
DOC: Added more fully developed us_employment case study

palewire:
I'm still not sure how to get the January bar on the left to start exactly where I'd like it, but here's a run at further developing this new dataset into a case study example.

visualization 4

```python
import altair as alt
import pandas as pd

source = data.us_employment()
presidents = pd.DataFrame([
    {
        "start": "2006-01-01",
        "end": "2009-01-19",
        "president": "Bush"
    },
    {
        "start": "2009-01-20",
        "end": "2015-12-31",
        "president": "Obama"
    }
])

bars = alt.Chart(source, title="The U.S. employment crash during the Great Recession").mark_bar().encode(
    x=alt.X("month:T", axis=alt.Axis(title="")),
    y=alt.Y("nonfarm_change:Q", axis=alt.Axis(title="Change in non-farm employment (in thousands)")),
    color=alt.condition(
        alt.datum.nonfarm_change > 0,
        alt.value("steelblue"),
        alt.value("orange")
    )
)

rule = alt.Chart(presidents).mark_rule(
  color="black",
  strokeWidth=2
).encode(
  x='end:T'
).transform_filter(alt.datum.president == "Bush")

text = alt.Chart(presidents).mark_text(
    align='left',
    baseline='middle',
    dx=7,
    dy=-135,
    size=11
).encode(
    x='start:T',
    x2='end:T',
    text='president',
    color=alt.value('#000000')
)

(bars + rule + text).properties(width=600)
```

### DIFF
--- a/altair/examples/us_employment.py
+++ b/altair/examples/us_employment.py
@@ -8,7 +8,7 @@ import altair as alt
 import pandas as pd
 from vega_datasets import data
 
-source = "https://vega.github.io/vega-datasets/data/us-employment.csv"
+source = data.us_employment()
 presidents = alt.pd.DataFrame([
     {
         "start": "2006-01-01",
--- a/altair/examples/us_employment.py
+++ b/altair/examples/us_employment.py
@@ -1,0 +1,56 @@
+"""
+The U.S. employment crash during the Great Recession
+----------------------------------------------------
+This example is a fully developed bar chart with negative values using the sample dataset of U.S. employment changes during the Great Recession.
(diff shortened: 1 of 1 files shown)


jakevdp:
I think the test failure is a result of the pandas import not being used.

instead of ``alt.pd.DataFrame`` please use ``pd.DataFrame``.

jakevdp:
Ah, so it is.

Looks good, thanks!

PR #1237, opened 2018-11-28
Add example of custom data transformer to docs

JarnoRFB:
The example shows how the json data transformer can be used to store the
data files in a separate directory.


jakevdp:
Very nice – thanks!

PR #1238, opened 2018-11-29
DOC: Change grouped bar chart example

eitanlees:
I changed the grouped bar chart example as requested in https://github.com/altair-viz/altair/issues/1236 also I moved the previous grouped bar chart example to the case studies section.


jakevdp:
Looks good, thanks!

PR #1239, opened 2018-11-29
DOC: Replace link of pdvega repo

JarnoRFB:
pdvega is no under altair-viz. The old link was not working anymore.


jakevdp:
Thanks!

PR #1240, opened 2018-11-30
DOC: Standardized grouped bar chart example

palewire:
* Substituted dataset common to bar examples
* Reduced code complexity
* Fixed "too wide" bug
* Added more docs
* Added second horizontal version of the example


palewire:
Looks like @eitanlees and I were on the same page today. I noticed #1238 after I filed the PR. I think mine still should be considered as an addendum, though he clearly beat me to the punch.

jakevdp:
Thanks for looking at this – I think I'd prefer to have a minimal grouped bar chart example that doesn't require any data transforms: it's a common enough question that I'd like to have a bare-bones example like the one added in #1238.

palewire:
Fair. I still think it's worth adding a companion version that runs on the other axis as well as the inline comments I've put in. Let me loop back and make another run at this patch.

palewire:
My revisions are in. Let me know what you think.

jakevdp:
Looks good, thanks!

PR #1241, opened 2018-11-30
DOC: Relabel line chart variable

eitanlees:
I was a little confused by the labels used on the simple line chart example. The label was `'sin(x)'` but then the equation was `sin(x / 5)`. I think it would make more sense just to have `'x'` and `'f(x)'` as labels. I modified the other example where this occurs.


jakevdp:
Makes sense

PR #1243, opened 2018-12-01
DOC: Cleaned up trellis scatter example

palewire:
The current example is too wide and breaks out of the template. I'd also like to try to gradually standardize this section of examples to use a common local dataset. Cars seems like a good one we're already using in other examples, and it fits pretty well here.


jakevdp:
Looks good

PR #1251, opened 2018-12-04
ENH: set defaultStyle to True by default in HTML output

jakevdp:
This means that the actions menu appears in the upper-right corner rather than as text links below the chart. Note that this also affects the Colab and Kaggle renderers.

cc/ @domoritz

### DIFF
--- a/altair/utils/html.py
+++ b/altair/utils/html.py
@@ -90,14 +90,13 @@ def spec_to_html(spec, mode,
                  vega_version, vegaembed_version, vegalite_version=None,
                  base_url="https://cdn.jsdelivr.net/npm/",
                  output_div='vis', embed_options=None, json_kwds=None,
-                 fullhtml=True, requirejs=False):
+                 fullhtml=True, requirejs=False, default_style=True):
     """Embed a Vega/Vega-Lite spec into an HTML page
 
     Parameters
     ----------
     spec : dict
         a dictionary representing a vega-lite plot spec.
-    fullhtml : boo
     mode : string {'vega' | 'vega-lite'}
         The rendering mode. This value is overridden by embed_options['mode'],
         if it is present.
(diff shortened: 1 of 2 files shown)


domoritz:
Great! I was thinking of making this the default in the next version of Embed, btw.

jakevdp:
> I was thinking of making this the default in the next version of Embed

Oh, in that case I might not merge this... when will you know for sure?

domoritz:
We just finished (5 minutes ago) the meeting where we decided to change the default. I'm going to make the change today and release.

jakevdp:
Sounds good – I'm going to merge the testing improvements here, but rely on vega-embed defaults for this default style.

domoritz:
https://github.com/vega/vega-embed/pull/112

jakevdp:
Leaving this open as a reminder to upgrade the vega-embed dependency

domoritz:
Released as Vega-Embed 3.25

jakevdp:
Thanks! Looking at it, we currently pin https://cdn.jsdelivr.net/npm/vega-embed@3, so I think this should automatically grab the newest version.

domoritz:
Excellent. https://github.com/altair-viz/altair/pull/1251/files#diff-c850a6dcd6935988601601ad65f58d62L100 should still be applied, no?

jakevdp:
Since the argument defaults to True in vega-embed, I think there's no reason to explicitly set that default in the Altair code.

domoritz:
What is `fullhtml`, though?

jakevdp:
It's documented further down.
```
    fullhtml : boolean (optional)
        If True (default) then return a full html page. If False, then return
        an HTML snippet that can be embedded into an HTML page.
```

jakevdp:
I removed the stray text in #1252

PR #1259, opened 2018-12-08
DOC: change axis=Axis(title=title) to title=title

jakevdp:
Instead of setting encoding titles like this:
```python
x=alt.X('x', axis=alt.Axis(title='title'))
```
we should set them like this:
```python
x=alt.X('x', title='title')
```


palewire:
I didn't know you could do that! Excellent!

palewire:
I've resolved the conflicts and issues here.

PR #1261, opened 2018-12-10
[Paper] Add missing author to paper

domoritz:
The [corresponding archive](https://zenodo.org/record/2030098) has already been updated.


jakevdp:
Thanks!

PR #1284, opened 2019-01-07
DOC: Added new line chart examples with points over the series

palewire:
I've added three new line chart examples lifted from the ones upstream in the Vega Lite docs. In the interest of uniformity, I've also modified the multi-series example and ensured that the three new ones have a matching dataset.

If you like these, I'll make some more line charts change soon so we can give it the same attention we recently gave the bar charts.

### DIFF
--- a/altair/examples/line_chart_with_interpolation.py
+++ b/altair/examples/line_chart_with_interpolation.py
@@ -1,6 +1,6 @@
 """
-Line Chart with Points
-----------------------
+Line Chart with Interpolation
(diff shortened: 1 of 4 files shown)


palewire:
If you want to see the examples in action, I have them in a [Collab scratchpad](https://colab.research.google.com/drive/1QdV23HqCY1dK5_TSLlCaqA7u-g9eLkV5).

palewire:
This has been superceded by #2644

PR #1297, opened 2019-01-14
DOC: Really move Seattle weather heatmap example into case studies

eitanlees:
I think a mistake was made in #1286 where the weather heat map was moved to the line chart section instead of the case studies.


jakevdp:
Looks good, thanks

PR #1324, opened 2019-02-07
DOC: Add window transform example

iaindillingham:
This example shows that aggregation functions aren't only for aggregate transforms. By using two aggregation functions within a window transform we are able to compute z-scores within a calculate transform.

See also: https://vega-js.slack.com/archives/C0NG6SXFX/p1548451285140300

### DIFF
--- a/doc/user_guide/transform.rst
+++ b/doc/user_guide/transform.rst
@@ -699,6 +699,38 @@ last_value    None       Assigns a value from the last data object in the curren
 nth_value     Number     Assigns a value from the nth data object in the current sliding window frame. If no such object exists, assigns ``null``. Requires a non-negative integer parameter that indicates the offset from the start of the window frame. This operation must have a corresponding entry in the `fields` parameter array.
 ============  =========  =========================================================================================================================================================================================================================================================================================================================
 
+Notice that when we use an aggregation function within an aggregate transform, the data objects are collapsed.
+However, when we use an aggregation function within a window transform, the data objects are augmented.
+It's by augmenting, rather than collapsing, that we can compute statistics, such as `z-scores`_.
+For example, consider the following time series of stock prices:
+
+.. altair-plot::
+
+    import altair as alt
+    from vega_datasets import data
+
+    alt.Chart(data.stocks.url).mark_line().encode(x='date:T', y='price:Q', color='symbol:N')
+
+It's hard to see the overall pattern in the above example, because Google's stock price is much higher than the other stock prices.
+If we plot the z-scores of the stock prices, rather than the stock prices themselves, then the overall pattern becomes clearer:
+
+.. altair-plot::
+
+    import altair as alt
+    from altair import expr
+    from vega_datasets import data
+
+    alt.Chart(data.stocks.url).transform_window(
+        mean_price='mean(price)',
+        stdev_price='stdev(price)',
+        frame=[None, None], groupby=['symbol'],
+    ).transform_calculate(
+        z_score=(expr.datum.price - expr.datum.mean_price) / expr.datum.stdev_price,
+    ).mark_line().encode(x='date:T', y='z_score:Q', color='symbol:N')
+
+By using two aggregation functions (``mean`` and ``stdev``) within the window transform, we are able to compute the z-scores within the calculate transform.
+
 For more information about the arguments to the window transform, see :class:`WindowTransform` and `the Vega-Lite documentation <https://vega.github.io/vega-lite/docs/window.html>`_.
 
 .. _Vega expression: https://vega.github.io/vega/docs/expressions/
+.. _z-scores: https://en.wikipedia.org/w/index.php?title=Z-score
--- a/doc/user_guide/transform.rst
+++ b/doc/user_guide/transform.rst
@@ -723,7 +723,6 @@ If we plot the `z-scores`_ of the stock prices, rather than the stock prices the
 .. altair-plot::
 
     import altair as alt
-    from altair import expr
     from vega_datasets import data
 
     alt.Chart(data.stocks.url).transform_window(
--- a/doc/user_guide/transform.rst
+++ b/doc/user_guide/transform.rst
@@ -671,7 +671,10 @@ For example, consider the following cumulative frequency distribution:
         sort=[{'field': 'IMDB_Rating'}],
         frame=[None, 0],
         cumulative_count='count(*)',
-    ).mark_area().encode(x='IMDB_Rating:Q', y='cumulative_count:Q')
+    ).mark_area().encode(
+        x='IMDB_Rating:Q',
+        y='cumulative_count:Q',
+    )
 
 First, we pass a sort field definition, which indicates how data objects should be sorted within the window.
 Here, movies should be sorted by their IMDB rating.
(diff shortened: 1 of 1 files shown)


doc/user_guide/transform.rst
```diff
@@ -699,6 +699,38 @@ last_value    None       Assigns a value from the last data object in the curren
 nth_value     Number     Assigns a value from the nth data object in the current sliding window frame. If no such object exists, assigns ``null``. Requires a non-negative integer parameter that indicates the offset from the start of the window frame. This operation must have a corresponding entry in the `fields` parameter array.
 ============  =========  =========================================================================================================================================================================================================================================================================================================================
 
+Notice that when we use an aggregation function within an aggregate transform, the data objects are collapsed.
+However, when we use an aggregation function within a window transform, the data objects are augmented.
+It's by augmenting, rather than collapsing, that we can compute statistics, such as `z-scores`_.
+For example, consider the following time series of stock prices:
```
jakevdp: I'm not entirely clear what is meant by "collapsed" and "augmented" in this context.

doc/user_guide/transform.rst
```diff
@@ -699,6 +699,38 @@ last_value    None       Assigns a value from the last data object in the curren
...
+
+.. altair-plot::
+
+    import altair as alt
+    from vega_datasets import data
+
+    alt.Chart(data.stocks.url).mark_line().encode(x='date:T', y='price:Q', color='symbol:N')
```
jakevdp: Please put each encoding argument indented on its own line.

doc/user_guide/transform.rst
```diff
@@ -699,6 +699,38 @@ last_value    None       Assigns a value from the last data object in the curren
...
+    import altair as alt
+    from altair import expr
+    from vega_datasets import data
+
+    alt.Chart(data.stocks.url).transform_window(
+        mean_price='mean(price)',
+        stdev_price='stdev(price)',
+        frame=[None, None], groupby=['symbol'],
+    ).transform_calculate(
+        z_score=(expr.datum.price - expr.datum.mean_price) / expr.datum.stdev_price,
```
jakevdp: Use ``alt.datum`` rather than ``expr.datum``

doc/user_guide/transform.rst
```diff
@@ -699,6 +699,38 @@ last_value    None       Assigns a value from the last data object in the curren
...
+    ).mark_line().encode(x='date:T', y='z_score:Q', color='symbol:N')
```
jakevdp: encodings on their own line, as above

doc/user_guide/transform.rst
iaindillingham: Yes, good question. What I want to communicate is that an aggregate transform with an aggregation function is like an SQL `GROUP BY` but a window transform with an aggregation function is like broadcasting in Pandas. In the latter case, each data object has a new property (is augmented with), however, there are the same number of data objects. In the former case, there are (probably) a different number of objects.

iaindillingham:
I've addressed your comment about encoding arguments, and extended it slightly for consistency. I've attempted to clarify the aggregate/window difference with more descriptive concepts of `GROUP BY` and broadcasting, although couldn't find a definition of the latter in the Pandas documentation. Hopefully the combination of text and example is sufficient: it's the `GROUP BY`/broadcasting behaviour that initially confused me about aggregate/window transforms.

doc/user_guide/transform.rst
jakevdp: Ah, maybe something like "While aggregate transforms compute a single value summarizing the dataset, window transforms effectively add a new column to the dataset, constructed from local aggregates at each row."

iaindillingham: I'm not sure what you mean by "local aggregates around each value in the column". Local to the data object, which is synonymous with adding a new column? I updated this section in a previous commit, but Github's UI unfortunately isn't displaying it here. Would you mind taking a look? I'll comment it now.

doc/user_guide/transform.rst
```diff
@@ -702,9 +702,10 @@ last_value    None       Assigns a value from the last data object in the curren
...
 ============  =========  =========================================================================================================================================================================================================================================================================================================================
 
-Notice that when we use an aggregation function within an aggregate transform, the data objects are collapsed.
-However, when we use an aggregation function within a window transform, the data objects are augmented.
-It's by augmenting, rather than collapsing, that we can compute statistics, such as `z-scores`_.
+Notice that using an aggregate transform is like using an SQL ``GROUP BY``:
+the input data objects are replaced by the output data objects, which contain the results of applying the aggregation function.
+However, using a window transform is like broadcasting the results to the input data objects.
+By using a window transform, and preserving the input data objects, we can compute statistics, such as `z-scores`_.
 For example, consider the following time series of stock prices:
```
iaindillingham: Does this address your original comment @jakevdp?

doc/user_guide/transform.rst
jakevdp: By "local aggregates" I was trying to say "windowed aggregates" without being so tautological :smile:

jakevdp: It's local in the sense that the values nearby the point are aggregated and put in the new column.

iaindillingham: Ah, I see! I've replaced the section with the following:

> While an aggregate transform computes a single value that summarises all data objects, a window transform adds a new property to each data object. This new property is computed from the neighbouring data objects: that is, from the data objects delimited by the window field definition.

I felt that "neighboring" was more descriptive than "local", but I'm happy to change it. Hopefully the reference to the window field definition isn't too tautological.

jakevdp: Looks good, thanks!

doc/user_guide/transform.rst
jakevdp: I don't really understand the connection here between window functions and broadcasting. To me they seem like distinct concepts: for example, you cannot compute a convolution through broadcasting except via a very contrived process, but you can easily express a convolution via a window transform. I like your description elsewhere of window transforms as aggregates computed in the neighborhood of each row of the data.

Regarding group by & aggregation, I think this is also a confusing comparison. An altair aggregate can express the same thing as a groupby followed by an aggregation (if you specify the ``groupby`` parameter in the transform) but in general aggregation is not equivalent to a group-by.

iaindillingham: I think we're out of step with our comments and commits. I've removed the references to `GROUP BY` and broadcasting, as well as the original references to collapsed and augmented. My most recent commit incorporated your suggested amendments.

iaindillingham: I've managed to get into a tangle with PRs and MRs on GitHub before. Sorry for the confusion!

jakevdp: Ah, I see. Sorry about that!

doc/user_guide/transform.rst
```diff
@@ -699,6 +702,46 @@ last_value    None       Assigns a value from the last data object in the curren
...
+        color='symbol:N',
+    )
+
+It's hard to see the overall pattern in the above example, because Google's stock price is much higher than the other stock prices.
+If we plot the `z-scores`_ of the stock prices, rather than the stock prices themselves, then the overall pattern becomes clearer:
+
+.. altair-plot::
+
+    import altair as alt
+    from altair import expr
```
jakevdp: Import is no longer used

iaindillingham: Good spot!

iaindillingham: Removed in latest commit.

iaindillingham:
Okay, I think we're there! Thanks for your feedback, and, as every for your hard work on Altair.

jakevdp:
Thanks!

PR #1343, opened 2019-02-21
add borderless example

### LINKED ISSUE #1342: Completely remove all bottom axis and top axis (e.g. in Charts)?
I can turn off grids, labels, ticks, ... but the line is still there. My axes are in arbitrary units - is there a way to get rid of the lines completely?

jona-sassenhagen:
Closes #1342


jakevdp:
Looks good, thanks!

PR #1364, opened 2019-03-08
example using selection for mark_area

reckoner:
Moving this from issues to an example of how to use interval section with `mark_area`.


jakevdp:
This is great - thanks so much for the contribution!

PR #1373, opened 2019-03-18
Create grouped_bar_chart_with_error_bars.py

mc-robinson:
As mentioned in https://github.com/altair-viz/altair/issues/1362 . I have included a change of title, in order to suppress the CIs showing up in the title. However, this may not fit with the style of the examples, so that might need a change.

### DIFF
--- a/altair/examples/grouped_bar_chart_with_error_bars.py
+++ b/altair/examples/grouped_bar_chart_with_error_bars.py
@@ -1,6 +1,6 @@
 """
 Grouped Bar Chart with Error Bars
----------------------------
+---------------------------------
 This example shows a grouped bar chart with error bars.
 """
 # category: bar charts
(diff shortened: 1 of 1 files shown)


altair/examples/grouped_bar_chart_with_error_bars.py
```diff
@@ -0,0 +1,27 @@
+"""
+Grouped Bar Chart with Error Bars
+---------------------------
```
jakevdp: This underline needs to be the same length as the line above or the RST->HTML conversion will complain.

altair/examples/grouped_bar_chart_with_error_bars.py
```diff
@@ -0,0 +1,27 @@
...
+
+source = data.barley()
+
+bars = alt.Chart().mark_bar().encode(
+    x='year:O',
+    y=alt.Y('mean(yield):Q', title='Mean Yield'),
+    color='year:N',
+)
+
+error_bars = bars.mark_rule().encode(
```
jakevdp: Since you're overwriting all encodings and marks in the bars object, there's no reason to build this of the ``bars`` object. Instead use ``alt.Chart().mark_rule()...``

jakevdp:
Looks great, thanks!

PR #1377, opened 2019-03-20
Adding example of sorted error bars plot.

mc-robinson:
An adaptation of the https://github.com/altair-viz/altair/blob/master/altair/examples/error_bars_with_ci.py file. It took me a while to figure out how to do this, so figured I would submit it for others' benefit. 

I tried to stick to the style of the original file, but let me know if that is too many strangely placed parentheses.


jakevdp:
Looks great, thanks!

PR #1383, opened 2019-03-23
Add basic Vega-Lite v3 wrappers

jakevdp:
This doesn't change any of the public API, but adds Python wrappers for Vega-Lite version 3.0.0rc13. They can be used as follows:
```python
import altair.vegalite.v3 as alt
```
and then use ``alt`` as normal. Currently all examples in the gallery work correctly with v3.


jakevdp:
This PR also removes the vegalite v1 wrappers.

PR #1388, opened 2019-03-26
Make Altair default to using vega-lite 3

jakevdp:
Also bump VL version to 3.0.2


jakevdp:
Bulk of the diffs are due to the 3.0.0->3.0.2 version change.

PR #1396, opened 2019-03-28
ENH: Add Error Bars with Standard Deviation example

ImadDabbura:
Using the new error bar mark to show the standard deviation.


jakevdp:
Looks great, thanks!

PR #1399, opened 2019-03-28
ENH: Add William Playfair Wheat and Wages example

pratapvardhan:
image

### DIFF
--- a/altair/examples/bar_and_line_with_dual_axis.py
+++ b/altair/examples/bar_and_line_with_dual_axis.py
@@ -2,6 +2,8 @@
 Bar Chart with Line on Dual Axis
 --------------------------------
 This example shows how to combine two plots and keep their axes.
+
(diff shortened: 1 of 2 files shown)


jakevdp:
Looks great! We have a simpler version of this chart here: https://altair-viz.github.io/gallery/bar_and_line_with_dual_axis.html

I still think this more polished version is a good addition to the examples. Would you mind linking to that in the description? If you use something like the following, the sphinx build will generate the correct link:

```
This is a more polished version of the simpler chart in :ref:`_gallery_bar_and_line_with_dual_axis`.
```

jakevdp:
In the other example header, you could also add something like this:
```
For a more polished version of this chart, see :ref:`_gallery_wheat_wages`.
```

pratapvardhan:
Updated with crosslinks.

jakevdp:
Looks great, thanks!

Issue #1366, opened 2019-03-10
Encode point on Map


noklam:
I cannot figure out how to encode points/circle with calculation on a map. The exampe https://altair-viz.github.io/gallery/airports.html only works for fix value, what I am hoping is the size can be calculate with mean of a field (Group by long,lat)

(Updated)
Notebook to reproduce (data included):
https://github.com/noklam/Gallery/blob/master/temp.ipynb

![image](https://user-images.githubusercontent.com/18221871/54083831-de2fab00-4363-11e9-9a20-f54e04ac0111.png)

![image](https://user-images.githubusercontent.com/18221871/54083836-ea1b6d00-4363-11e9-9eda-9f05be8ad2e3.png)

![image](https://user-images.githubusercontent.com/18221871/54083844-f3a4d500-4363-11e9-91cd-b2a77a439132.png)

mattijn:
Can you share your data and make some code snippets so I can copy it in a notebook?

noklam:
> Can you share your data and make some code snippets so I can copy it in a notebook?

@mattijn Thank you for offering help, I have included the data and notebook here. It surprise me I can only encode size with no aggregation.
https://github.com/noklam/Gallery/blob/master/temp.ipynb

mattijn:
Sorry, this is a bug at the Vega-lite repository: https://github.com/vega/vega-lite/issues/4368

noklam:
@mattjin I saw it is fixed in vega-lite, any idea when altair will be able to catch up these changes?

mattijn:
Once Vega-Lite releases v3.0, Altair will follow shortly. When this happens? No idea, hopefully before end of Q2!

noklam:
@mattijn Thanks for the quick response, I guess I can only wait then.

noklam:
The issue is fixed now, could it be an example for the gallery? The only problem is I cannot find a good vega dataset with both geolocation and quantitative columns.
https://github.com/noklam/Gallery/blob/master/Circles%20Overlay%20on%20Map.ipynb
![image](https://user-images.githubusercontent.com/18221871/55251638-05efa000-528c-11e9-9ad4-3ca8f2055d03.png)

jakevdp:
This would be a great example to add, if you'd like to work on it!

PR #1407, opened 2019-03-30
add example for points overlay on maps

noklam:
I cannot find a complete fit dataset from vega-dataset, so I have to do a little by of transformation to demonstrate this example

image

#1366

### DIFF
--- a/altair/examples/airports_count.py
+++ b/altair/examples/airports_count.py
@@ -1,0 +1,44 @@
+"""
+Locations of US Airports
+========================
+This is a layered geographic visualization that shows the positions of US
+airports on a background of US states.
+"""
+# category: maps
+import altair as alt
+from vega_datasets import data
+
+airports = data.airports()
+airports_counts = airports[['state','latitude', 'longitude']]
+
+# Take the mean as the position of state
+agg = {'latitude': 'mean',
+         'longitude': 'mean'
+                  }
+airports_counts = airports.groupby(['state']).agg(agg) 
+agg = {'latitude': 'count',
+      
+                  }
+
+airports_counts['Number of Airports'] =  airports.groupby(['state']).agg(agg)
+
+states = alt.topo_feature(data.us_10m.url, feature='states')
+
+# US states background
+background = alt.Chart(states).mark_geoshape(
+    fill='lightgray',
+    stroke='white'
+).properties(
+    width=500,
+    height=300
+).project('albersUsa')
+
+# airport positions on background
+points = alt.Chart(airports_counts).mark_circle().encode(
+    longitude='longitude:Q',
+    latitude='latitude:Q',
+    size='Number of Airports:Q',
+    color=alt.value('steelblue')
+)
+
+background + points
--- a/altair/examples/airports_count.py
+++ b/altair/examples/airports_count.py
@@ -36,13 +36,4 @@ points = alt.Chart(airports).transform_aggregate(
     title='Number of airports in US'
 )
 
-chart = background + points 
-chart = chart.configure_title(
-    fontSize=24,
-    color='black'
-).configure_legend(
-    titleFontSize=14,
-    labelFontSize=12,
-)
-
-chart
+background + points
--- a/altair/examples/airports_count.py
+++ b/altair/examples/airports_count.py
@@ -39,6 +39,9 @@ points = alt.Chart(airports_counts).mark_circle().encode(
     latitude='latitude:Q',
     size='Number of Airports:Q',
     color=alt.value('steelblue')
+).properties(
+    title='Number of airports in US'
 )
+
 
 background + points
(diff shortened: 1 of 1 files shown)


altair/examples/airports_count.py
```diff
@@ -0,0 +1,47 @@
...
+========================
+This is a layered geographic visualization that shows the positions of US
+airports on a background of US states.
+"""
+# category: maps
+import altair as alt
+from vega_datasets import data
+
+airports = data.airports()
+airports_counts = airports[['state','latitude', 'longitude']]
```
jakevdp: This line can be deleted, because the variable is redefined below.

altair/examples/airports_count.py
```diff
@@ -0,0 +1,47 @@
...
+# Take the mean as the position of state
+agg = {'latitude': 'mean',
+         'longitude': 'mean'
+                  }
+airports_counts = airports.groupby(['state']).agg(agg) 
+agg = {'latitude': 'count',
+      
+                  }
+
+airports_counts['Number of Airports'] =  airports.groupby(['state']).agg(agg)
```
jakevdp: This would be easier to follow if you did all three aggregates in a single operation.

jakevdp:
I think this example would be better if the data transformations were done as a transform within the chart, rather than as a preprocessing step. Something like this:
```python
import altair as alt

airports = data.airports.url
states = alt.topo_feature(data.us_10m.url, feature='states')

# US states background
background = alt.Chart(states).mark_geoshape(
    fill='lightgray',
    stroke='white'
).properties(
    width=500,
    height=300
).project('albersUsa')

# airport positions on background
points = alt.Chart(airports).transform_aggregate(
    latitude='mean(latitude)',
    longitude='mean(longitude)',
    count='count()',
    groupby=['state']
).mark_circle().encode(
    longitude='longitude:Q',
    latitude='latitude:Q',
    size=alt.Size('count:Q', title='Number of Airports'),
    color=alt.value('steelblue')
).properties(
    title='Number of airports in US'
)

background + points 
```

altair/examples/airports_count.py
noklam: yes, thank you, I have overlooked it.

altair/examples/airports_count.py
noklam: I did it because the earlier aggregation is using the same column, while seperating this aggregration allows me to assign it to a new column in one line. 

I would love to know is there a cleaner pandas way to do so? I agree using transform_aggregate is cleaner in this case.

```
agg = {'latitude': 'mean',
         'longitude': 'mean'
                  }
```

noklam:
tooltips now show the state instead of lat, long. Adjust the font size to make title and legend a bit bigger, I tried to adjust symbolSize too but that seems not working...

altair/examples/airports_count.py
jakevdp: Your third aggregate is count, so it can be computed on any column. For example, this gives the same result:
```python
airports_counts = airports.groupby('state').agg({
    'latitude': 'mean',
    'longitude': 'mean',
    'name': 'count'
}).rename({'name': 'Number of Airports'}, axis=1)
```
But overall I like the transform approach below better.

altair/examples/airports_count.py
```diff
@@ -0,0 +1,48 @@
...
+chart = background + points 
+chart = chart.configure_title(
+    fontSize=24,
+    color='black'
+).configure_legend(
+    titleFontSize=14,
+    labelFontSize=12,
+)
+
+chart
```
jakevdp: For the sake of this example, I think I'd avoid minor stylistic tweaks like these. I see the primary purpose of this example as demonstrating overlaying points on a map, and the additional configuration might be a bit distracting.

altair/examples/airports_count.py
noklam: Didn't know we can pass count directly without an actual column. I try to use  'latitude': ['mean','count'] before but that result in multi-index columns and is less easy to rename.

jakevdp: You do have to pass a column; in this case I passed the column called ``"name"``, which has the same count as latitude (or as any of the other columns)

altair/examples/airports_count.py
noklam: Okay, I removed the styling part for this example. But I think it would be a great idea to add an example just to demonstrate some common styling operation. It is much easier than going through the doc. (Most of the time we just want a slightly larger Title, label, etc.)

jakevdp:
Thanks!

dazzag24:
Apologies for slightly off topic question, but it is related.  Is it possible to overlay polylines on top of the map?  For example if I wanted to draw aircraft trajectories between the airports as a series of points with lines connecting them.  Can altair do this?

jakevdp:
Yes, here is an example of drawing lines over geoshapes: https://altair-viz.github.io/gallery/london_tube.html

jakevdp:
You might also get inspiration from vega-lite examples, e.g. https://vega.github.io/vega-lite/examples/geo_rule.html

PR #1409, opened 2019-03-30
add example for normalized parallel coordinates

pagpires:
Add example for normalized parallel coordinates to demo `.transform_joinaggregate()` usage:
image

### DIFF
--- a/altair/examples/normed_parallel_coordinates.py
+++ b/altair/examples/normed_parallel_coordinates.py
@@ -13,6 +13,8 @@ where the y-axis shows the value after min-max rather than the raw value. It's a
 simplified Altair version of `the VegaLite version <https://vega.github.io/vega-lite/examples/parallel_coordinate.html>`_
 """
 # category: other charts
+import altair as alt
+from vega_datasets import data
 from altair import datum
 
 source = data.iris()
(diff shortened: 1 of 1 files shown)


altair/examples/normed_parallel_coordinates.py
```diff
@@ -0,0 +1,36 @@
+"""
+Normalized Parallel Coordinates Example
+----------------------------
```
jakevdp: Line needs to be the same length as the text above, or the sphinx RST->html tool will complain.

jheer:
Cool! Not particularly critical, but it might be nice to synchronize the design with the existing example here (scroll down to the last chart): https://observablehq.com/@vega/vega-lite-api

altair/examples/normed_parallel_coordinates.py
pagpires: yep, please see the revised code

pagpires:
> Cool! Not particularly critical, but it might be nice to synchronize the design with the existing example here (scroll down to the last chart): https://observablehq.com/@vega/vega-lite-api

Was actually trying to replicate that one but found involving other techniques like multi-layers and custom axis and thought it might be too complicate for a single usage demo.  Will spend more time looking into it and make a exact replicate later.

jakevdp:
Thanks!

PR #1410, opened 2019-03-31
example of wrapped facet

pagpires:
Replicate of https://vega.github.io/vega-lite/docs/facet.html#facet-full to demo `columns` usage.  It's very similar from the original barley trellis plot, but the `columns` documentation is not ready now, so this example might be helpful.  Let me know if any modification is preferred to make it more useful.
image


jakevdp:
This is great - thanks!

PR #1412, opened 2019-03-31
Update saving_charts.rst

jploudre:
It's not obvious how to save with Firefox (kept getting a missing chromedriver error.) Pulled this tidbit from the code.


jploudre:
Maybe a little more documentation about using Firefox to save would help others? It was confusing to me!

jakevdp:
Thanks!

I don't use firefox, and the option was a one-off addition from another Altair user. If you're a firefox user who is interested in improving the documentation for other firefox users, that would be a welcome contribution.

PR #1424, opened 2019-04-10
Specify vega==1.4.0 in notebook installation instructions

sid-kap:
I found that altair looks for vega in the entrypoint 'altair.vegalite.v2.renderer' (implying it doesn't support vegalite v3), but the PyPI package vega now puts its entrypoint in 'altair.vegalite.v3.renderer' in all version starting at 2.0.0 (see [ipyvega commit](https://github.com/vega/ipyvega/commit/d01f00217f0166ef9b304d195bea17c4bd844d41#diff-2eeaed663bd0d25b7e608891384b7298R37)).
So to use altair in a Jupyter notebook, we must specify an older version of vega. Otherwise `alt.renderers.enable('notebook')` fails with a message saying that vega cannot be found.

(We should probably update the error message as well.)

### DIFF
--- a/doc/getting_started/installation.rst
+++ b/doc/getting_started/installation.rst
@@ -87,7 +87,10 @@ To install the notebook and Altair with conda, run the following command::
 
 To install the notebook and Altair with pip, run the following command::
 
-    $ pip install -U altair vega_datasets notebook vega
+    $ pip install -U altair vega_datasets notebook vega==1.4.0
+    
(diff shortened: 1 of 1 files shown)


jakevdp:
Sorry about that, we're in the process of updating to Vega-Lite 3.0, and all of the versions have strange interactions. I don't plan to update the Altair v2 docs at this point, but they will soon be replaced by Altair version 3 and this will not be an issue.

PR #1433, opened 2019-04-13
Clean up alt.selection (fixes #1431)

### LINKED ISSUE #1431: Altair selections resolve incorrectly as scale domain values.
I am observing this issue in Altair 2.x. Apologies if this issue has already been addressed for v3!

The following code for an overview + detail plot fails:

```python
brush = alt.selection_interval(encodings=['x']);

base = alt.Chart().mark_area().encode(
    x=alt.X('date:T', title=None),
    y='price:Q'
).properties(
    width=700
)
  
alt.vconcat(
    base.encode(alt.X('date:T', title=None, scale=alt.Scale(domain=brush))),
    base.add_selection(brush).properties(height=60),
    data='https://vega.github.io/vega-datasets/data/sp500.csv'
)
```

The errors indicate that Altair is expanding the brush to the full selection definition (what you would use for `add_selection`) rather than generating a selection reference. If I add `name='brush'` to the selection definition and then replace `domain=brush` with `domain={'selection': 'brush'}`, the example works.

jakevdp:
This cleans up the syntax of ``alt.selection_*()``. The main effects are

- you no longer need to use ``selection.ref()`` when referencing the selection (e.g. as a scale domain). Concretely, if ``selection=alt.selection_interval()`` then ``alt.Scale(domain=selection.ref())`` can now be written ``alt.Scale(domain=selection)``.
- it's no longer possible to pass a selection directly as a property; it must be passed via the ``add_selection`` function
- it's no longer possible to use the addition operator to combine two selections; so, rather than ``chart.properties(selection=sel1+sel2)``, we instead do ``chart.add_selection(sel1, sel2)``.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,8 @@
 # Altair Change Log
 
-## Version 3.0.0rc1 (prerelease)
+## Version 3.0.0 (unreleased)
 
-Update to Vega-Lite 3.0 and Vega 5.0 & support all new features. See
+Update to Vega-Lite 3.2 and Vega 5.0 & support all new features. See
 https://github.com/vega/vega-lite/releases/tag/v3.0.0 for Vega-Lite
 feature lists.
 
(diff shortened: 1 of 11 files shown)


jakevdp:
cc/ @jheer

PR #1443, opened 2019-04-21
MAINT: Fix some minor nits

genneth:
Mostly a test that I've got the dev env set up correctly, but whilst I'm there...
* Some pep8 violations
* Some deprecation warnings that show up in py.test

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -271,7 +271,7 @@ def parse_shorthand(shorthand, data=None, parse_aggregates=True,
     if parse_types:
         patterns = list(itertools.chain(*((p + ':{type}', p) for p in patterns)))
 
-    regexps = (re.compile('\A' + p.format(**units) + '\Z', re.DOTALL)
+    regexps = (re.compile(r'\A' + p.format(**units) + r'\Z', re.DOTALL)
                for p in patterns)
 
     # find matches depending on valid fields passed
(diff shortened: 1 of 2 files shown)


altair/utils/schemapi.py
```diff
@@ -99,14 +99,15 @@ def __str__(self):
             return six.text_type(self).encode("utf-8")
 
 
-
```
jakevdp: The ground truth for this file is https://github.com/altair-viz/altair/blob/master/tools/schemapi/schemapi.py

Changes to the file should be made there, and then propagated here by running ``tools/generate_schema_wrapper.py``.

genneth: Have done. However, running `generate_schema_wrapper.py` also generates some diffs to various core.py files, I suspect because upstream has subtly changed the wording in some docs? I've not committed those changes here. Should I?

jakevdp:
FYI, we generally use flake8 to check code style, and the ignored attributes are listed here: https://github.com/altair-viz/altair/blob/f1e59fac1e986bb86f3f6b23c7e014e5b0ec4ae4/setup.cfg#L1-L5

altair/utils/schemapi.py
jakevdp: There shouldn't be any diffs, because it's targeted to specific version tags. It might be some small difference in the version of m2r or some other package.

genneth:
Yeah, I see now how the workflow is meant to go. No worries. I'll probably rework this branch to try to clean up the py.test warnings, and if you don't mind, sort out any pep8 issues in only the files I touch?

jakevdp:
Looks good - is this ready to merge?

genneth:
Yep, good as-is.

PR #1463, opened 2019-04-30
Create brushing scatter plot to show data on a table example

sterlinm:
Create an Altair implementation of the Vega-Lite example [Brushing Scatter Plot to show data on a table](https://vega.github.io/vega-lite/examples/brush_table.html).

### DIFF
--- a/altair/examples/scatter_linked_table.py
+++ b/altair/examples/scatter_linked_table.py
@@ -24,14 +24,12 @@ points = alt.Chart().mark_point().encode(
 # Base chart for data tables
 ranked_text = alt.Chart().mark_text().encode(
     y=alt.Y('row_number:O',axis=None)
-).transform_window(
-    row_number='row_number()'
 ).transform_filter(
     brush
 ).transform_window(
-    rank='rank(row_number)'
+    row_number='row_number()'
 ).transform_filter(
-    alt.datum.rank<20
+    alt.datum.row_number<20
 )
 
 # Data Tables
--- a/altair/examples/scatter_linked_table.py
+++ b/altair/examples/scatter_linked_table.py
@@ -1,0 +1,57 @@
+"""
+Brushing Scatter Plot to show data on a table
+---------------------------------
+A scatter plot of the cars dataset, with data tables for horsepower, MPG, and origin. 
+The tables update toreflect the selection on the scatter plot.
+"""
+# category: scatter plots
+
+import altair as alt
+from vega_datasets import data
+
+source = data.cars()
+
+# Brush for selection
+brush = alt.selection(type='interval')
+
+# Scatter Plot
+points = alt.Chart().mark_point().encode(
+    x='Horsepower:Q',
+    y='Miles_per_Gallon:Q',
+    color=alt.condition(brush, 'Cylinders:O', alt.value('grey'))
+).add_selection(brush)
+
+# Base chart for data tables
+ranked_text = alt.Chart().mark_text().encode(
+    y=alt.Y('row_number:O',axis=None)
+).transform_window(
+    row_number='row_number()'
+).transform_filter(
+    brush
+).transform_window(
+    rank='rank(row_number)'
+).transform_filter(
+    alt.datum.rank<20
+)
+
+# Data Tables
+horsepower = ranked_text.mark_text().encode(text='Horsepower:N')
+mpg = ranked_text.mark_text().encode(text='Miles_per_Gallon:N')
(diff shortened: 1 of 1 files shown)


altair/examples/scatter_linked_table.py
```diff
@@ -0,0 +1,57 @@
...
+    y=alt.Y('row_number:O',axis=None)
+).transform_window(
+    row_number='row_number()'
+).transform_filter(
+    brush
+).transform_window(
+    rank='rank(row_number)'
+).transform_filter(
+    alt.datum.rank<20
+)
```
jakevdp: I think this can be accomplished with one less transform: 
```python
).transform_filter(
    brush
).transform_window(
    row_number='row_number()'
).transform_filter(
    alt.datum.row_number<20
)
```

altair/examples/scatter_linked_table.py
```diff
@@ -0,0 +1,57 @@
...
+
+# Data Tables
+horsepower = ranked_text.mark_text().encode(text='Horsepower:N')
+mpg = ranked_text.mark_text().encode(text='Miles_per_Gallon:N')
+origin = ranked_text.mark_text().encode(text='Origin:N')
```
jakevdp: The ``mark_text()`` in each of these can be removed, because the ``ranked_text`` chart already specifies that.

Also, you can use the ``properties`` to set the title in this line; putting it all together it might look like

```python
text = alt.hconcat(
  ranked_text.encode(text='Horsepower:N').properties(title='Horsepower'),
  ranked_text.encode(text='Miles_per_Gallon:N').properties(title='MPG'),
  ranked_text.encode(text='Origin:N').properties(title='Origin'),
)
```

sterlinm: Thanks! I thought there must be a better way to add the title, and I didn't realize I could layer encode in that way which will help with making my code less awkward all around. I'll try to make those changes (I'm an inexperienced contributor but will try to figure it out). Thanks again!

altair/examples/scatter_linked_table.py
sterlinm: I think you're right. I was following the [Vega-Lite example](https://vega.github.io/vega-lite/examples/brush_table.html) somewhat literally and was nervous that I had missed some edge case where you couldn't just use the row_number, but I think it's fine so long row_number is set after the brush filter is applied.

sterlinm:
There are two other things I noticed that could be better but aren't part of the example that inspired this.

First, we could filter out records where Horsepower or Miles_per_Gallon are null from the table, since they aren't visible on the scatter plot.

Second, I think the behavior of the table when you select an interval and then click away from it is a little confusing. If you select some points and then click to clear the selection, the tables don't revert to the initial default state. I think what happens is that the records that were in the filter keep the same value for row_number and the records that were outside of it get new row_numbers. If you filter to say 5 records and then clear, the 5 records stay at the top and the remaining 15 are populated from the rest of the data I think.

I'd be happy to try to fix those but don't want to make the example too confusing.

jakevdp:
>  I think what happens is that the records that were in the filter keep the same value for row_number and the records that were outside of it get new row_numbers.

Ah, you're right. I think that's an argument in favor of your original approach to the transformations: in that case the row number is constant regardless of the selection, and I think it works correctly. I'd be in favor of restoring that behavior.

sterlinm:
> Ah, you're right. I think that's an argument in favor of your original approach to the transformations

I played around with it a bit and using my original order for the transformations changes things slightly but it's still not exactly what I'd expect in terms of the behavior. It helps slightly in that it can use the row_number value to order the records in the table, but it never goes back to the original view before you select anything.

When the chart first loads it lists the records with row number ranging from 1 to 19. If I then select two points with the brush (say row_number 162 and 163) it shows just those two rows. Then when I clear the selection it shows 1 through 17, and then at the bottom 162 and 163. It looks like rank is not recalculated for points that were selected when you clear it, as the rank of records 162 and 162 are 1 and 2 respectively.

jakevdp:
Thanks - that looks like an issue with propagation of selection signals in vega. I've opened an issue there.

sterlinm:
Okay thanks! Let me know if there are any other improvements you think I should make, but otherwise I think it's okay from my end.

jakevdp:
Would you like to change it back to your original set of transforms?

sterlinm:
Okay done. My fork is behind master now and I couldn't seem to get it to update (git fetch upstream led to a permission denied error) but hopefully it's okay. Thanks.

jakevdp:
Looks good, thanks!

PR #1480, opened 2019-05-07
MAINT: fix spelling of several new encoding channels

jakevdp:


jakevdp:
Still TODO: add deprecated versions of the old names for backward compatibility.

jakevdp:
We can use something like this to deprecate the functions:
```python
import warnings
import functools


def _deprecated(obj, message=None, name=None):
  name = name or obj.__name__
  message = message or "{} is deprecated: use {} instead".format(name, obj.__name__)
  if isinstance(obj, type):
    return type(name, (obj,),
                {'__doc__': obj.__doc__,
                 '__init__': _deprecated(obj.__init__, message)})
  elif callable(obj):
    @functools.wraps(obj)
    def new_obj(*args, **kwargs):
        warnings.warn(message, DeprecationWarning)
        return obj(*args, **kwargs)
    return new_obj
  else:
    raise ValueError("Cannot deprecate object of type {}".format(type(obj)))

Xerror = _deprecated(channels.XError, name='Xerror')
```

PR #1492, opened 2019-05-09
DOC: mention jupyterlab pre-release in troubleshooting docs

jakevdp:


jamesmyatt:
Is there a way of doing this with conda?

jakevdp:
I don't think there's any way to install the pre-release with conda. But you can use pip within your conda environment.

PR #1504, opened 2019-05-13
Create simple_scatter_large_square.py

yoonghm:
The example could be used in Mark example for mark_square() in https://github.com/altair-viz/altair/blob/master/doc/user_guide/marks.rst#marks


jakevdp:
Looks good – I think this should go in ``category: scatter plots`` rather than ``category: simple charts``. The simple charts category is for charts without additional style tweaks.

yoonghm:
Ok, i will rename the file and combine two pull requests into one.

PR #1517, opened 2019-05-17
It stripplot example

iliatimofeev:
visualization - 2019-05-17T222608 137
Analog of https://seaborn.pydata.org/generated/seaborn.stripplot.html


altair/examples/stripplot.py
```diff
@@ -0,0 +1,39 @@
+"""
+Stripplot
+-----------------
+This example shows how to make a Stripplot.
+"""
+# category: scatter plots
+import altair as alt
+from vega_datasets import data
+
+source = data.movies()
```
jakevdp: Since the movies data is not local, please use ``data.movies.url`` instead (this is to keep the test suite from requiring http access on the Python side)

altair/examples/stripplot.py
```diff
@@ -0,0 +1,39 @@
...
+
+stripplot = (
+    alt.Chart(source, width=40)
+    .mark_circle(size=8)
+    .encode(
+        alt.X(
+            shorthand="jitter:Q",
```
jakevdp: Please pass the shorthand as a positional arg rather than by keyword.

altair/examples/stripplot.py
iliatimofeev: fixed

altair/examples/stripplot.py
iliatimofeev: fixed

altair/examples/stripplot.py
```diff
@@ -0,0 +1,39 @@
...
+        header=alt.Header(
+            labelAngle=-90,
+            titleOrient='top',
+            labelOrient='bottom',
+            labelAlign='right',
+            labelPadding=3,
+        ),
+    ),
+).transform_calculate(
+    jitter='pow(random()-0.5,3)'
```
jakevdp: This looks a bit better IMO with gaussian noise, which you can generate with a Box-Muller transform:
```python
transform_calculate(
   # Generate Gaussian jitter with a Box-Muller transform
   jitter='sqrt(-2*log(random()))*cos(2*PI*random())'
)
```

altair/examples/stripplot.py
```diff
@@ -0,0 +1,39 @@
+"""
+Stripplot
+-----------------
```
jakevdp: This line needs to be the same length as the title above, or sphinx will complain when generating the HTML.

jakevdp:
Looks great, thanks!

Issue #1500, opened 2019-05-10
Better UX around layers & facets


jakevdp:
One of the more common questions comes from people who are trying to layer two faceted charts together. For example:
```python
base = alt.Chart('data.csv').encode(
    x='x:Q',
    y='y:Q',
    column='label:N'
)

base.mark_bar() + base.mark_text()
```
```
---------------------------------------------------------------------------
SchemaValidationError                     Traceback (most recent call last)
/usr/local/lib/python3.6/dist-packages/altair/vegalite/v3/api.py in to_dict(self, *args, **kwargs)
    376         if dct is None:
    377             kwargs['validate'] = 'deep'
--> 378             dct = super(TopLevelMixin, copy).to_dict(*args, **kwargs)
    379 
    380         # TODO: following entries are added after validation. Should they be validated?

/usr/local/lib/python3.6/dist-packages/altair/utils/schemapi.py in to_dict(self, validate, ignore, context)
    268                 self.validate(result)
    269             except jsonschema.ValidationError as err:
--> 270                 raise SchemaValidationError(self, err)
    271         return result
    272 

SchemaValidationError: Invalid specification

        altair.vegalite.v3.api.LayerChart->layer->items, validating 'anyOf'

        {'data': {'url': 'data.csv'}, 'mark': 'bar', 'encoding': {'column': {'type': 'nominal', 'field': 'label'}, 'x': {'type': 'quantitative', 'field': 'x'}, 'y': {'type': 'quantitative', 'field': 'y'}}} is not valid under any of the given schemas
        
LayerChart({
  layer: [Chart({
    data: 'data.csv',
    encoding: FacetedEncoding({
      column: Column({
        shorthand: 'label:N'
      }),
      x: X({
        shorthand: 'x:Q'
      }),
      y: Y({
        shorthand: 'y:Q'
      })
    }),
    mark: 'bar'
  }), Chart({
    data: 'data.csv',
    encoding: FacetedEncoding({
      column: Column({
        shorthand: 'label:N'
      }),
      x: X({
        shorthand: 'x:Q'
      }),
      y: Y({
        shorthand: 'y:Q'
      })
    }),
    mark: 'text'
  })]
})
```

At the very least, we should ensure that this raises a more understandable error (for example, checking in ``alt.layer`` whether the subcharts have facet encodings and raising an informative ValueError).

More than this, we might think about whether we could automatically do the right thing here: for example, detect that the layers have identical encodings, and construct a faceted layer chart from this code.

[...]

PR #1521, opened 2019-05-20
ENH: move layer & concat chart data to the top level when possible.

jakevdp:
Addresses one part of #1500.

The enhancement here is that now you can do something like this:
```python
base = alt.Chart(df).encode(
    x='x', 
    y='y'
)
alt.layer(
    base.mark_point(),
    base.mark_line()
).facet(
    'category'
)
```
Previously, you had to do
```python
base = alt.Chart().encode(
    x='x', 
    y='y'
)
alt.layer(
    base.mark_point(),
    base.mark_line()
).facet(
    'category',
    data=df
)
```
or else the ``facet()`` would lead to an obscure error about missing data. With this change, both these options work correctly.


jakevdp:
Note one consequence is that there is now a slight difference between ``a + b + c`` and ``alt.layer(a, b, c)``. In particular. If ``a`` and ``b`` both specify the same data and ``c`` specifies no data, then the former will move data to the top level, and the latter will not. I anticipate this being very confusing for any user that hits it, but I also think it will be a very rare situation. Unfortunately, I can't think of any way to change the behavior that doesn't introduce corner cases that are even worse.

Issue #1524, opened 2019-05-21
Weird behavior with alt.Fill


romainmartinez:
Hi,

I'm replicating the example of PR #1519 , but I came across a strange behavior of `alt.Fill`.

Consider the following function (which replicates exactly #1519):

```python
def ridge_plot(d, value, groupby, step=20, overlap=1):
    return (
        alt.Chart(d)
        .transform_joinaggregate(mean_value=f"mean({value})", groupby=[groupby])
        .transform_bin(["bin_max", "bin_min"], value)
        .transform_aggregate(
            value="count()", groupby=[groupby, "mean_value", "bin_min", "bin_max"]
        )
        .transform_impute(
            impute="value", groupby=[groupby, "mean_value"], key="bin_min", value=0
        )
        .mark_line(
            interpolate="monotone", fillOpacity=0.8, stroke="lightgray", strokeWidth=0.5
        )
        .encode(
            alt.X("bin_min:Q", bin="binned"),
            alt.Y("value:Q", scale=alt.Scale(range=[step, -step * overlap]), axis=None),
            alt.Fill("mean_value:Q", legend=None),
            alt.Row(
                f"{groupby}:N",
                title=None,
                header=alt.Header(labelAngle=0, labelAlign="right", format="%B"),
            ),
        )
        .properties(bounds="flush", height=step)
        .configure_facet(spacing=0)
        .configure_view(stroke=None)
    )
```
With randomly numbers sampled from a normal distribution, the graph looks good:

```python
def gen_data(size, random_function):
    return pd.DataFrame(
        {
            "value": random_function(size=size),
            "category": np.array(
                [np.repeat("a", size / 2), np.repeat("b", size / 2)]
            ).ravel(),
        }
    )

ridge_plot(
    gen_data(size=2000, random_function=np.random.normal),
    value="value",
    groupby="category",
)
```
![norm](https://user-images.githubusercontent.com/17710617/58134335-867bae80-7bf4-11e9-8b58-6934a3c183de.png)

But with exponential distributions, the fill is wrong:
```python
ridge_plot(
    gen_data(size=2000, random_function=np.random.exponential),
    value="value",
    groupby="category",
)
```
![exp](https://user-images.githubusercontent.com/17710617/58134344-8bd8f900-7bf4-11e9-96ae-548f0f6d3047.png)

It seems that when the ends of the distribution do not tend towards zero, `alt.Fill` is reversed.

jakevdp:
[...]

PR #1527, opened 2019-05-23
Use mark_area in ridgeline plot example

romainmartinez:
As discussed in #1524, using `mark_area` instead of `mark_line` ensures that the fill is between the line and the x-axis.


jakevdp:
Thanks!

PR #1542, opened 2019-05-31
Clarify that vega needs to be installed for server as well

tkf:
I propose to add the following note in "Quick Start: Altair + Notebook" section:

> If you have multiple Python environments, you need to install the
> vega package in the Python environment that runs ipykernel *and*
> the Python environment that runs the Jupyter Notebook server.  You
> only need to install Altair in the Python environment that runs
> ipykernel.


jakevdp:
Looks good, thanks!

PR #1548, opened 2019-06-03
Added an example of the default modified boxplot

yardsale8:
I just found out that the new boxplot mark creates a modified boxplot by default.  I have added a separate example for this default behavior.  Let me know if you would like me to make any changes.


jakevdp:
Thanks - I think I'd like to avoid having two examples in the gallery that are so similar. The information you added here would work better within a narrative documentation section about the boxplot mark, where these chart modifications could be viewed in the same context. Is that something you'd be willing to work on?

yardsale8:
Sure, where should I insert this narrative?  

Also, would you be willing to change the box plot example to the default behavior (i.e. plot outliers as points)?  I believe that most users are more likely to be looking for this behavior and I personally look to the gallery first when trying to make a specific plot.

jakevdp:
> Sure, where should I insert this narrative?

It could go in this section: https://github.com/altair-viz/altair/blob/master/doc/user_guide/marks.rst

yardsale8:
What are your thoughts on the number of examples.  [This is what I am thinking](https://github.com/WSU-DataScience/SDSS19-dataviz-workshop/blob/master/notebooks/boxplot_documentation.ipynb)  

Too many examples for one mark?

If that is acceptable, I can submit that as a separate PR.

jakevdp:
That looks great - feel free to submit a PR (or modify this one) and let me know if you run into any issues translating that content to .rst

Feel free to modify the existing example chart as well if you think it could be made more useful.

Thanks!

yardsale8:
The changes discussed above have now been added.

1. Documentation on `mark_boxplot` was added to `doc/user_guide/marks.rst`
2. Renamed `boxplot_max_min.py` to `boxplot.py` and changed the `extent` property to the default behavior.

Note the name change on the file. I couldn't find any references to this file with grep, let me know if this name change will cause any problems and I will fix them or change the name back.

doc/user_guide/marks.rst
```diff
@@ -60,7 +60,73 @@ arguments to represent points as red semi-transparent filled circles:
        y='Miles_per_Gallon:Q'
    )
 
-The additional arguments to ``mark_*()`` methods are passed along to an
+Similarly, the compound mark :meth:`~Chart.mark_boxplot` can be used to create a boxplot without having to specify each part of the plot (box, whiskers, outliers) separately.
```
jakevdp: I would put all of this in its own section:
```
Compound Marks
--------------

BoxPlot
~~~~~~~
```

yardsale8: Good idea, I made the change and also added a section on mark properties (to help the transition).

jakevdp:
Looks great, thanks! I'll merge as soon as the tests are green.

PR #1551, opened 2019-06-05
DOC: Change detail encoding example

eitanlees:
The [Selection Detail Example](https://altair-viz.github.io/gallery/select_detail.html#gallery-select-detail) does not actually use the detail encoding so I changed the example to [Ranged Dot Plot](https://altair-viz.github.io/gallery/ranged_dot_plot.html) which does.


jakevdp:
Good catch, thanks!

PR #1572, opened 2019-06-20
Truncate top-level chart repr

jakevdp:
Any time there is a rendering error Jupyter defaults to printing the chart repr. This can sometimes be _very_ long, and obscure the actual error. This change suppresses the full chart repr.


mcnuttandrew:
10 thumbs up

PR #1583, opened 2019-06-27
DOC: Add example for Connections Among U.S. Airports Interactive

manzt:
From the Vega-Lite examples: https://vega.github.io/vega-lite/examples/airport_connections.html

NOTE: I can't seem to get the tooltip to work on the points layer like in the Vega-Lite Example.

### DIFF
--- a/altair/examples/airport_connections.py
+++ b/altair/examples/airport_connections.py
@@ -53,7 +53,8 @@ points = alt.Chart(flights_airport).mark_circle().encode(
     latitude="latitude:Q",
     longitude="longitude:Q",
     size=alt.Size("routes:Q", scale=alt.Scale(range=[0, 1000]), legend=None),
-    order=alt.Order("routes:Q", sort="descending")
+    order=alt.Order("routes:Q", sort="descending"),
+    tooltip=['latitude:Q', 'longitude:Q', 'routes:Q']
 ).transform_aggregate(
     routes="count()", 
     groupby=["origin"]
--- a/altair/examples/airport_connections.py
+++ b/altair/examples/airport_connections.py
@@ -1,0 +1,69 @@
+"""
+Connections Among U.S. Airports Interactive
+-------------------------------------------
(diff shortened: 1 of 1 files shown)


altair/examples/airport_connections.py
```diff
@@ -0,0 +1,69 @@
...
+    as_=["state", "lat2", "lon2"]
+).transform_filter(
+    select_city
+)
+
+points = alt.Chart(flights_airport).mark_circle().encode(
+    latitude="latitude:Q",
+    longitude="longitude:Q",
+    size=alt.Size("routes:Q", scale=alt.Scale(range=[0, 1000]), legend=None),
+    order=alt.Order("routes:Q", sort="descending")
```
jakevdp: Add `tooltip=['latitude:Q', 'longitude:Q', 'routes:Q']` here to get the tooltips. We chose in Altair to disable tooltips unless they are added explicitly; Vega-Lite 3 adds tooltips by default.

altair/examples/airport_connections.py
```diff
@@ -0,0 +1,70 @@
...
+).transform_filter(
+    select_city
+)
+
+points = alt.Chart(flights_airport).mark_circle().encode(
+    latitude="latitude:Q",
+    longitude="longitude:Q",
+    size=alt.Size("routes:Q", scale=alt.Scale(range=[0, 1000]), legend=None),
+    order=alt.Order("routes:Q", sort="descending"),
+    tooltip=['latitude:Q', 'longitude:Q', 'routes:Q']
```
jakevdp: One more thought... might be nice to put ``'origin:N'`` in this list, so the tooltip contains the three-letter code for the airport you are hovering over.

It might also be worth removing lat/lon from the list, as I think its more distracting than useful in terms of what the visualization is communicating.

What do you think?

manzt: yes. that's a good idea!

jakevdp:
Thanks again :smile:

PR #1584, opened 2019-06-27
DOC: Add example for World Cup rankings with window transformation.

manzt:
From Vega-Lite examples: https://vega.github.io/vega-lite/examples/window_rank.html


manzt:
I'm thinking these examples with various transformations can be useful in the docs.

jakevdp:
Looks good, thanks!

Issue #1013, opened 2018-07-11
Possible to use the slider values in transform_calculate?


KODeKarnage:
Forum would be a better place for this question, sorry, but I couldn't find one yet :)

From what I can see, the slider can only control selection, is that correct?

I want to be able to give my users the option to scale the chart within a certain range, but would prefer not to have to pre-calculate the results for each point in that range.

If I could get the slider position, I could use that in a transform_calculate to do exactly what I need and reduce the size of the html by 50 times!

jakevdp:
No, I don't think that is possible. Unless I'm mistaken, Vega-Lite can only bind slider values to selections, which cannot be used to set arbitrary chart properties.

I believe you can do something along those lines with Vega, though. It might be worth asking on the vega slack channel.

breadbaron:
Along the same lines - is it possible to use a slider's values to select subsets of data? 

E.g. I'd like to make a plot like this, but instead of only selecting the year selected in the slider, select all years >= that value:

https://altair-viz.github.io/gallery/us_population_over_time.html?highlight=slider

adrielvieira:
Did you managed to do that @breadbaron?

breadbaron:
@adrielvieira no I don't believe so

adrielvieira:
I managed to change the color based on the value of the slider, following your mention of value comparison:

```
slider = altair.binding_range(min=0, max=100, step=1)
selector = altair.selection_single(name="SelectorName", fields=['field_name'],
                                   bind=slider, init={'score': 100})

altair.Chart(dataframe).mark_bar().encode(
    y='other_field_name',
    x='field_name',
    color=altair.condition('datum.field_name< SelectorName_ field_name', altair.value('lightblue'), alt.value('lightgray')),
).add_selection(selector)
```

You can use the same logic to filter data instead of change color:

```

slider = altair.binding_range(min=0, max=100, step=1)
selector = altair.selection_single(name="SelectorName", fields=['field_name'],
                                   bind=slider, init={'score': 100})

altair.Chart(dataframe).mark_bar().encode(
    y='other_field_name', 
    x='field_name').add_selection(
    selector
).transform_filter(
    'datum.field_name < SelectorName_ field_name'
)

```

jakevdp:
Interesting... didn't know this was possible!

It might be useful to think about adding a high-level syntax for doing this; e.g.
```
chart.transform_filter(
   alt.datum.field_name < slider.value
)
```
or something along those lines.

[...]

PR #1599, opened 2019-07-05
ENH: add selector value access for expressions

jakevdp:
Addresses #1013 

This PR adds the ability to access selector values within filter & condition expressions; for example:
```python
import altair as alt
import pandas as pd
import numpy as np

df = pd.DataFrame({
    'xval': range(100),
    'yval': np.random.randn(100).cumsum()
})

slider = alt.binding_range(min=0, max=100, step=1, name='cutoff')
selector = alt.selection_single(fields=['cutoff'], bind=slider, init={'cutoff': 50})

alt.Chart(df).mark_point().encode(
    x='xval',
    y='yval',
    color=alt.condition(alt.datum.xval < selector.cutoff, alt.value('red'), alt.value('blue'))
).add_selection(
    selector
)
```
Screen Shot 2019-07-05 at 8 18 50 AM

### DIFF
--- a/altair/expr/core.py
+++ b/altair/expr/core.py
@@ -7,10 +7,10 @@ class DatumType(object):
         raise ValueError("must use datum.attribute")
 
     def __getattr__(self, attr):
-        return ValueExpression(attr)
+        return GetAttrExpression("datum", attr)
 
     def __getitem__(self, attr):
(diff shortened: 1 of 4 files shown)


jakevdp:
It would be useful to add an example and some documentation as well.

PR #1610, opened 2019-07-14
Set minimum pandas version.

MarkDaoust:
If you have an old pandas installed, plots just fail with `infer_dtype() takes no keyword arguments`.

The `skipna` arg was added in pandas `0.21.0`. 

ref:  https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.api.types.infer_dtype.html

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ entrypoints
 jinja2
 jsonschema
 numpy
-pandas
+pandas>=0.21.0
 six
 toolz
 typing>=3.6;python_version<"3.5"


jakevdp:
Thanks - alternatively we could adjust this logic to handle version 0.20.0 correctly: https://github.com/altair-viz/altair/blob/0ddab1e9ad54990978a8ea14a714325bac4402f7/altair/utils/core.py#L17-L24

jakevdp:
So it turns out it's only Pandas 0.20.X that is broken currently; I think it would be better to modify the code to support 0.20.X rather than exclude it as a dependency. Here's a version of the utility that works on all pandas versions back to at least 0.16 (https://colab.research.google.com/drive/1SaigNiN-W8DY54zzKoou0cPc5XvjxMSt):
```python
try:
    from pandas.api.types import infer_dtype as _infer_dtype
except ImportError:
    # Import for pandas < 0.20.0
    from pandas.lib import infer_dtype as _infer_dtype
    
    
def infer_dtype(value, skipna=False):
    if not hasattr(infer_dtype, '_supports_skipna'):
        try:
            _ = _infer_dtype([1], skipna=False)
        except TypeError:
            # pandas < 0.21.0 don't support skipna keyword
            infer_dtype._supports_skipna = False
        else:
            infer_dtype._suppots_skipna = True
    if infer_dtype._supports_skipna:
        return _infer_dtype(value, skipna)
    else:
        return _infer_dtype(value)
```

jakevdp:
Replaced by #1611. Thanks for pointing out the issue!

MarkDaoust:
Thanks Jake!

PR #1620, opened 2019-07-23
Support Altair in Apache Zeppelin

zjffdu:
This PR is to support Altair in Apache Zeppelin. I create a new render for Zeppelin. The changes I made in HTMLTemplate is especially for Zeppelin. Without this change. The rendering only works for the time, afterwards, I will get js error of `el is already defined`.

Here's the screenshot in Zeppelin.

image

### DIFF
--- a/altair/utils/html.py
+++ b/altair/utils/html.py
@@ -58,7 +58,7 @@ requirejs.config({
     });
     {%- endif %}
     require(['vega-embed'], function(vegaEmbed){
-    {%- endif %}
+    {%- endif %} (function(root) {
       var spec = {{ spec }};
       var embedOpt = {{ embed_options }};
 
@@ -73,6 +73,7 @@ requirejs.config({
       const el = document.getElementById('{{ output_div }}');
       vegaEmbed("#{{ output_div }}", spec, embedOpt)
         .catch(error => showError(el, error));
+    })(window);
     {%- if requirejs %}
     });
     {%- endif %}
--- a/altair/utils/html.py
+++ b/altair/utils/html.py
@@ -58,7 +58,7 @@ requirejs.config({
     });
     {%- endif %}
     require(['vega-embed'], function(vegaEmbed){
-    {%- endif %} (function(root) {
+    {%- endif %} (function(vegaEmbed) {
       var spec = {{ spec }};
       var embedOpt = {{ embed_options }};
 
@@ -73,7 +73,7 @@ requirejs.config({
       const el = document.getElementById('{{ output_div }}');
       vegaEmbed("#{{ output_div }}", spec, embedOpt)
         .catch(error => showError(el, error));
-    })(window);
+    })(vegaEmbed);
     {%- if requirejs %}
     });
     {%- endif %}
--- a/altair/vegalite/v3/display.py
+++ b/altair/vegalite/v3/display.py
@@ -74,10 +74,10 @@ colab_renderer = HTMLRenderer(mode='vega-lite',
                               vegalite_version=VEGALITE_VERSION)
 
 zeppelin_renderer = HTMLRenderer(mode='vega-lite',
-                              fullhtml=True, requirejs=False,
-                              vega_version=VEGA_VERSION,
-                              vegaembed_version=VEGAEMBED_VERSION,
-                              vegalite_version=VEGALITE_VERSION)
+                                fullhtml=True, requirejs=False,
+                                vega_version=VEGA_VERSION,
+                                vegaembed_version=VEGAEMBED_VERSION,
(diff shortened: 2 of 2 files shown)


zjffdu:
I only do it for v3, if this approach is correct, I will do it for v2, v4, v5 as well.

jakevdp:
The reason it only works once is because you set ``full_html=True``, which causes the script tags to be re-embedded each time a chart is rendererd. Better would be to use ``full_html=False`` and load the Javascript resources separately. Even better would be to make Zeppelin recognize ``_repr_mimebundle_()`` and use that mechanism to display charts.

jakevdp:
Note that for Kaggle, requirejs is loaded by the top page, and so vega-related resources will only be loaded in the first require statement. For Colab, every output is in its own iframe, and so full_html is appropriate as each output is effectively a standalone webpage. JupyterLab and Jupyter Notebook don't use HTML-based rendering, but rather have frontend extensions that recognize the vega-lite mimetype.

zjffdu:
@jakevdp Thanks for the review. I have tried `full_html=False`, but it would fail without my change in `html.py`. Here's error message
image

And I also tried to use `requirejs`, but would hit this error. https://requirejs.org/docs/errors.html#mismatch

jakevdp:
>  I have tried full_html=False

Yes, with full_html=False you need to load the libraries elsewhere at the top level.

For full_html=True, you cannot create multiple charts in the same document, unless you also have requirejs=True.

You'll need to figure out how to address this in order to base a zeppelin renderer on the HTML options built into Altair: I'd suggest either adding a frontend extension that recognizes vega-lite mimetypes, or deciding on a mechanism to load the vega-lite libraries once per browser session.

zjffdu:
>  For full_html=True, you cannot create multiple charts in the same document.

Actually, I can create multiple charts in the same document if don't set `output_div` (see the screenshot in my PR description), so that every time it will create unique `output_div`. 

Adding frontend extension to recognize the mimetype is an ideal solution, but I am afraid it would take a large effort to do that. Because Zeppelin doesn't have that extension capability for now. How to use html rendering first? And Zeppelin community will try to introduce frontend extension later for the long term solution.

jakevdp:
OK, I see, thanks

altair/utils/html.py
```diff
@@ -58,7 +58,7 @@
     });
     {%- endif %}
     require(['vega-embed'], function(vegaEmbed){
-    {%- endif %}
+    {%- endif %} (function(root) {
```
jakevdp: Looks like the root variable is never used. What would you think about passing `vegaEmbed` so the same function definition can be used both with and without requirejs?

Also note that none of this javascript rendering code is tested as part of the automated tests, so it will be important to validate the outputs by exporting a chart with ``fullhtml`` and opening in a browser.

zjffdu: @jakevdp  I have updated the PR.
* Change `root` to `vegaEmbed`
* Verify the exported html
image

jakevdp: You'll also need to change the function argument to pass ``vegaEmbed`` rather than ``window``. And my suggestion was to combine it with the anonymous function definition just above to avoid code duplication.

zjffdu: Thanks @jakevdp I have updated it

altair/vegalite/v3/display.py
```diff
@@ -73,6 +73,12 @@ def svg_renderer(spec, **metadata):
                               vegaembed_version=VEGAEMBED_VERSION,
                               vegalite_version=VEGALITE_VERSION)
 
+zeppelin_renderer = HTMLRenderer(mode='vega-lite',
+                              fullhtml=True, requirejs=False,
+                              vega_version=VEGA_VERSION,
+                              vegaembed_version=VEGAEMBED_VERSION,
+                              vegalite_version=VEGALITE_VERSION)
```
jakevdp: Fix indentation to line up with ``mode`=`

altair/utils/html.py
jakevdp: Looks good.

altair/vegalite/v3/display.py
zjffdu: Addressed

jakevdp: Thanks

zjffdu:
Thanks @jakevdp for helping merge this PR.

zjffdu:
BTW, do you mean need to it in v2,v5, v6 ? And documentation is also needed as well I think. @jakevdp

jakevdp:
It could be added in ``vegalite.v2`` as well, though that's rarely used and will go away with the next release.

zjffdu:
What about vega v4, v5 ?

PR #1636, opened 2019-08-02
Fix facet() and resolve() methods

jakevdp:
- remove ``resolve_*`` methods from non-compound chart
- add ``facet()`` method to all top-level charts

This more closely matches what is allowed by the schema.


jakevdp:
Turns out this was a bit over-zealous: ``resolve_scale()`` should be included in ``alt.Chart`` for the sake of setting resolution on facet charts.

PR #1643, opened 2019-08-07
BUG: Charts serialized as dict/json should be equivalent to original …

bergtholdt:
…instance

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -250,8 +250,7 @@ class SchemaBase(object):
 
     def __eq__(self, other):
         return (type(self) is type(other)
-                and self._args == other._args
-                and self._kwds == other._kwds)
+                and self.to_dict() == other.to_dict())
 
     def to_dict(self, validate=True, ignore=None, context=None):
         """Return a dictionary representation of the object


jakevdp:
What is the application in which you need round-tripped charts to evaluate as equal to unserialized ones? One issue I see with this change is that, as written, it does not allow equality checks on schema objects unless they are valid. Equality comparison of invalid objects would result in a ``SchemaValidationError``.

jakevdp:
Here's an example of what I'm talking about:
```python
>>> import altair as alt
>>> base = alt.Chart().mark_point().encode(x='x')
>>> base == base
True
>>> base.to_dict() == base.to_dict()
<ipython-input-3-4073fbee472b> in <module>()
----> 1 base1.to_dict() == base2.to_dict()

9 frames
/usr/local/lib/python3.6/dist-packages/altair/vegalite/v3/schema/channels.py in to_dict(self, validate, ignore, context)
     46                                      "the type cannot be automatically inferred because "
     47                                      "the data is not specified as a pandas.DataFrame."
---> 48                                      "".format(self.shorthand))
     49         else:
     50             # Shorthand is not a string; we pass the definition to field,

ValueError: x encoding field is specified without a type; the type cannot be automatically inferred because the data is not specified as a pandas.DataFrame.
```

bergtholdt:
Hi Jake,

I see that. My use-case,  as stated in the issue #1681, is to save and load charts from the file-system. I think with the fix to #1681 using the pull request in #1706 that can be considered as resolved.

PR #1645, opened 2019-08-07
Update crossfilter to gray/blue scheme

domoritz:
Same as in https://vega.github.io/editor/#/examples/vega-lite/interactive_layered_crossfilter


jakevdp:
Thanks!

PR #1686, opened 2019-09-06
Add API wrapper for sphere generator

eitanlees:
I noticed the sphere generator was missing from the data gen wrappers.

### DIFF
--- a/altair/vegalite/v3/api.py
+++ b/altair/vegalite/v3/api.py
@@ -1988,9 +1988,6 @@ def graticule(**kwds):
   return core.GraticuleGenerator(graticule=graticule)
 
 
-def sphere(**kwds):
+def sphere():
   """Sphere generator."""
-  if not kwds:
-      # sphere: True indicates default parameters
-      sphere = True
-  return core.SphereGenerator(sphere=sphere, **kwds)
+  return core.SphereGenerator(sphere=True)
--- a/altair/vegalite/v3/api.py
+++ b/altair/vegalite/v3/api.py
@@ -94,7 +94,8 @@ def _prepare_data(data, context):
     # if data is still not a recognized type, then return
     if not isinstance(data, (dict, core.Data, core.UrlData,
                              core.InlineData, core.NamedData,
-                             core.GraticuleGenerator, core.SequenceGenerator)):
+                             core.GraticuleGenerator, core.SequenceGenerator,
+                             core.SphereGenerator)):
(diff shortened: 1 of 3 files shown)


altair/vegalite/v3/api.py
```diff
@@ -1984,4 +1985,12 @@ def graticule(**kwds):
...
\ No newline at end of file
+  return core.GraticuleGenerator(graticule=graticule)
+
+
+def sphere(**kwds):
+  """Sphere generator."""
+  if not kwds:
+      # sphere: True indicates default parameters
+      sphere = True
+  return core.SphereGenerator(sphere=sphere, **kwds)
```
jakevdp: Possible for ``sphere`` to be undefined here.

altair/vegalite/v3/tests/test_api.py
```diff
@@ -858,4 +858,9 @@ def test_graticule():
...
 
     data = alt.graticule(step=[15, 15])
-    assert data.to_dict() == {'graticule': {'step': [15, 15]}}
\ No newline at end of file
+    assert data.to_dict() == {'graticule': {'step': [15, 15]}}
+
+
+def test_sphere():
+    data = alt.sphere()
+    assert data.to_dict() == {'sphere': True}
```
jakevdp: Test passing of kwargs as well - anything is allowed according to the schema, and ``alt.sphere(key='val').to_dict()`` should result in ``{'sphere': {'key': 'val'}}``

eitanlees:
Thanks for the review! I have a question. After looking a little closer at the documentation for the [Sphere Generator](https://vega.github.io/vega-lite/docs/data.html#sphere) it seems that the generator doesn't take any additional parameters.

Should `alt.sphere` take no parameters? Should it take in `**kwds` and then raise an error if any are passed? 

Thanks again for any advice.

jakevdp:
Ah, you're right: the ``sphere`` parameter of a ``SphereGenerator`` definition can be either the boolean True or an empty dict – I'd missed the ``additionalProperties: false`` when I looked before.

https://github.com/altair-viz/altair/blob/2d74e4a189b61e051da2d453357d38b5e8016106/altair/vegalite/v3/schema/vega-lite-schema.json#L10781-L10795

So yeah, the ``sphere()`` function should probably just take no keywords.

eitanlees:
Alright, done and done

jakevdp:
Looks good, thanks!

PR #1707, opened 2019-09-26
BUG: composing multiple selections

mattijn:
This PR has still some issues, but it adds some documentation + inline example to compose multiple selections as discussed in https://github.com/altair-viz/altair/issues/1704.

I tried to emulate the Vega-Lite example from [here](https://vega.github.io/vega-lite/docs/selection.html#compose), but preferred to hold the ctrl-key or shift-key to separately initiate the selections.

I don't know how to make the following Vega-Lite example with bitwise operators work in Altair:
```
"selection": {"not": {"and": ["alex", "morgan"]}}
```
I tried some (eg. `~alex & ~morgan` and `~[alex & morgan]`), but got none working.

### DIFF
--- a/altair/vegalite/v3/api.py
+++ b/altair/vegalite/v3/api.py
@@ -177,17 +177,17 @@ class Selection(object):
         return {'selection': self.name}
 
     def __invert__(self):
-        return core.SelectionNot(**{'not': self.name})
+        return Selection(core.SelectionNot(**{'not': self.name}), self.selection)
 
     def __and__(self, other):
         if isinstance(other, Selection):
             other = other.name
-        return core.SelectionAnd(**{'and': [self.name, other]})
+        return Selection(core.SelectionAnd(**{'and': [self.name, other]}), self.selection)      
 
     def __or__(self, other):
         if isinstance(other, Selection):
             other = other.name
-        return core.SelectionOr(**{'or': [self.name, other]})
+        return Selection(core.SelectionOr(**{'or': [self.name, other]}), self.selection)      
    
     def __getattr__(self, field_name):
         return expr.core.GetAttrExpression(self.name, field_name)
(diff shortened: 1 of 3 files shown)


jakevdp:
I think ``~(alex & morgan)`` should work (or equivalently, ``~alex | ~morgan`` following De Morgan's law)

jakevdp:
Actually, it looks like those compound operands are not fully implemented currently.

jakevdp:
Relevant issue is https://github.com/altair-viz/altair/issues/695

mattijn:
I think I got the compound operators working:



```python
class Foo(object):
    def __init__(self, name):
        self.name = name      
    
    def __invert__(self):
        return Foo({'not': self.name})
        
    def __repr__(self):
        return str(self.name)
    
    def __or__(self, other):
        return Foo({'or': [self.name, other]})
    
    def __ror__(self, other):     
        return Foo({'or': [other, self.name]})
    
    def __and__(self, other):
        return Foo({'and': [self.name, other]})
    
    def __rand__(self, other):     
        return Foo({'and': [other, self.name]})    


x = Foo('s1')
y = Foo('s2')
z = Foo('s3')
```


```python
x | ~y
```




    {'or': ['s1', {'not': 's2'}]}




```python
~x | y
```




    {'or': [{'not': 's1'}, s2]}




```python
~x | ~y
```




    {'or': [{'not': 's1'}, {'not': 's2'}]}




```python
~(x & y)
```




    {'not': {'and': ['s1', s2]}}




```python
~z | x & ~y
```




    {'or': [{'not': 's3'}, {'and': ['s1', {'not': 's2'}]}]}

jakevdp:
Yeah, that's how to implement them. The challenge is that the relevant classes are auto-generated, so we'll need to create wrappers in order to define the operators correctly. And there are a *lot* of relevant classes.

mattijn:
I've changed the operators (`__invert__`, `__and__`, and `__or__`) to return instances of `Selection`. With this I'm able to chain the operations to invoke further operations on them.

I add some tests to verify, but (!) also had to change some tests to make all tests pass. Please have a look.

PR #1717, opened 2019-10-02
ENH: update properties in project shorthand

mattijn:
This PR aligns the possible property options in the api function `project()` with the `Projection` class from the Vega-lite schema.


jakevdp:
Thanks!

PR #1756, opened 2019-10-28
Update install instructions to include Vega

firasm:
A few of my students had trouble installing altair and getting plots to show up because `vega` wasn't included as a required package. This was the case in jupyterlab. Folks on jupyter seemed to be fine. 

Request adding this to the documentation

### DIFF
--- a/doc/getting_started/installation.rst
+++ b/doc/getting_started/installation.rst
@@ -31,11 +31,11 @@ Altair version 3 works best with JupyterLab version 1.0 or later.
 
 To install JupyterLab and Altair with conda, run the following command::
 
-    $ conda install -c conda-forge altair vega_datasets jupyterlab
+    $ conda install -c conda-forge altair vega vega_datasets jupyterlab
 
 To install JupyterLab and Altair with pip, run the following command::
 
-    $ pip install -U altair vega_datasets jupyterlab
+    $ pip install -U altair vega vega_datasets jupyterlab
 
 Once this is finished, run::
 


jakevdp:
vega is not required for use with Jupyterlab, but it is required for use with Jupyter notebook. Vega is included in the install command for "Altair and Jupyter Notebook", further down in the page.

firasm:
Few notes about this:

- As advised [here](https://altair-viz.github.io/user_guide/faq.html), I asked students to add this line to the top of their Jupyterlab notebooks: `alt.data_transformers.enable('json')` and then added the `altair-data*.json` to their .gitignore_global file 

- The issue that affected about 40-50 students out 125 was that a "blank plot" showed up when they tried to run a cell in Jupyter lab. I had a look at [this page](https://altair-viz.github.io/user_guide/troubleshooting.html#display-troubleshooting) and the issue we were seeing wasn't the same as any of those.

- In general: for the students having this issue, the temporary (not practical longterm) workaround was to comment out `alt.data_transformers.enable('json')` and restart kernel. This produced a plot and they could continue with their lab. Could this be a hint on another issue that could be causing this?

- Though some students were using Jupyter notebook (rather than lab), I can confirm that in at least one case that installing `vega` through `conda` resolved the issue. This student was using Jupyterlab > 1.0 

- I am of course not able to reproduce this issue on my machine so it makes it hard to diagnose. If you have other ideas, I am open to trying things out.

(perhaps this discussion doesn't belong in a PR but here we are...)

Continuing to troubleshoot this week.

firasm:
Working with another student's computer (using the UW worksheets to test)

Here is the notebook output: 
notebook

Here is the lab output: 

lab

jakevdp:
> As advised here, I asked students to add this line to the top of their Jupyterlab notebooks: alt.data_transformers.enable('json') and then added the altair-data*.json to their .gitignore_global file

I don't see where in that page it advises setting the JSON transformer unconditionally. It's definitely a good thing to use in some cases.

> The issue that affected about 40-50 students out 125 was that a "blank plot" showed up when they tried to run a cell in Jupyter lab. I had a look at this page and the issue we were seeing wasn't the same as any of those.

This sounds like https://github.com/jupyterlab/jupyterlab/issues/7017, which was fixed in a recent Jupyterlab release. I'd suggest those students update their jupyterlab installation.

> In general: for the students having this issue, the temporary (not practical longterm) workaround was to comment out alt.data_transformers.enable('json') and restart kernel. This produced a plot and they could continue with their lab. Could this be a hint on another issue that could be causing this?

Yes, I think it's a much better idea to only enable the JSON data transformer when you actually need it, rather than doing it unconditionally.

> Though some students were using Jupyter notebook (rather than lab), I can confirm that in at least one case that installing vega through conda resolved the issue. This student was using Jupyterlab > 1.0

Yes, as detailed in the documentation, the pip-installed vega package is required when working with Altair in Jupyter notebooks (not in JupyterLab). The presence or absence of the vega package has no effect on Jupyterlab, because jupyterlab has an entirely distinct frontend extension model from jupyter notebook.

Regarding the blank chart you mentioned in the subsequent comment: that looks consistent with the issue reported in https://github.com/jupyterlab/jupyterlab/issues/7017. Update jupyterlab and it should be fixed.

---

A meta point: I too am quite frustrated by the state of frontend extensions in the juptyer notebook and jupyterlab; imagine the troubles you're having with 50 students, and multiply that by the number of Altair users in the world, and you'll get a sense of the support burden involved :grin:

Honestly, my recommendation would be to teach using a cloud-based notebook system in which the compute environment is a known factor. There are several possibilities, but I'm biased toward http://colab.research.google.com (full disclosure: I'm employed by Google to work on Colab) because it supports Altair and other common data science tools out of the box, and is free to use for anyone with a Google-backed email account (true of all students and employees at UW).

firasm:
Thanks for the detailed answer Jake, I'll close this PR now. It's entirely possible updating jupyterlab update will fix this and I'll try this solution tomorrow.

We will be working with fairly large datasets so I think in our context it's good to get the JSON transformer working because otherwise they'll be pulling and pushing huge notebooks (their labs) through GitHub LMS.

firasm:
just got another report of the blank chart, student is fully updated on jupyterlab:

```
(base) Annys-MBP:~ annychih$ jupyter --version
jupyter core     : 4.6.0
jupyter-notebook : 6.0.1
qtconsole        : 4.5.5
ipython          : 7.8.0
ipykernel        : 5.1.2
jupyter client   : 5.3.4
jupyter lab      : 1.1.4
nbconvert        : 5.6.0
ipywidgets       : 7.5.1
nbformat         : 4.4.0
traitlets        : 4.3.3
```

student also said she got errors in her terminal related to an untrusted kernel. (just a note for myself)

jakevdp:
It would be helpful to know what errors appear in the browser's JS console in these cases.

PR #1761, opened 2019-10-30
Add example - Cumulative count

baogianghoangvu:
Hi.

I was looking to plot a cumulative count axis and only found a Vega example, so I thought I might as well PR a simple adaptation.

Hope this helps! Thank you.

### DIFF
--- a/altair/examples/cumulative_count_chart.py
+++ b/altair/examples/cumulative_count_chart.py
@@ -13,7 +13,7 @@ from vega_datasets import data
 source = data.movies.url
 
 alt.Chart(source).transform_window(
-    cumulative_count="count(count)",
+    cumulative_count="count()",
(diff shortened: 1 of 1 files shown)


altair/examples/cumulative_count_chart.py
```diff
@@ -0,0 +1,21 @@
...
+"""
+# category: area charts
+
+import altair as alt
+from vega_datasets import data
+
+source = data.movies.url
+
+alt.Chart(source).transform_window(
+    cumulative_count="count(count)",
```
jakevdp: It would be clearer for this to be ``cumulative_count="count()"`` – no need to use any field name as a standin

jakevdp:
Looks good, thanks

PR #1762, opened 2019-10-30
Missing character in util-core doc

baogianghoangvu:
Hi,

I'm not sure if this really counts as a typo. While reading through the source code I noticed this tiny missing ">" in a function docstring. Turns out it's the only occurrence of its kind in the whole codebase.

Hope this helps!


jakevdp:
Thanks!

PR #1777, opened 2019-11-08
Add minimum version to pandas to account for dependency on RangeIndex

yajaru:
This commit pins the minimum supported version of pandas to be 23.4.
Currently there is a hard dependency on pandas.RangeIndex in /altair/utils/core.py:179, and version 23.4 is the earliest pandas version that I was able to find online documented with that feature.
Note that I did not do an exhaustive search over pandas versions so this limit could be moved lower on a more thorough search. Alternatively, it might need to be moved higher as I did no testing beyond the features we use in our product.
Testing should probably be done for other unpinned libraries as well as it is doubtful that early versions of those libraries would be able to run Altair.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ entrypoints
 jinja2
 jsonschema
 numpy
-pandas
+pandas>=0.23.4
 six
 toolz
 typing>=3.6;python_version<"3.5"


jakevdp:
I'm not certain this is correct... RangeIndex seems to be available going back to at least version 0.18, confirmed by the following:
```
for version in [25, 24, 23, 22, 21, 20, 19, 18]:
  !python3 -m pip install -qq pandas~=0.{version}.0
  !python3 -c "import pandas as pd; print(pd.__version__, pd.RangeIndex)"
```
```
0.25.3 <class 'pandas.core.indexes.range.RangeIndex'>
0.24.2 <class 'pandas.core.indexes.range.RangeIndex'>
0.23.4 <class 'pandas.core.indexes.range.RangeIndex'>
0.22.0 <class 'pandas.core.indexes.range.RangeIndex'>
0.21.1 <class 'pandas.core.indexes.range.RangeIndex'>
0.20.3 <class 'pandas.core.indexes.range.RangeIndex'>
0.19.2 <class 'pandas.indexes.range.RangeIndex'>
0.18.1 <class 'pandas.indexes.range.RangeIndex'>
```

jakevdp:
Could you give more information about the error you're seeing that prompted this PR?

yajaru:
https://github.com/altair-viz/altair/blob/e2eb5224996bcb4646387dea955ceeed80495551/altair/utils/core.py#L181

That line references pandas RangeIndex, which doesn't exist prior to pandas 0.18.1. I came across this b/c we are using pandas 0.17 and we updated altair to a newer version that added that check and it  broke a report. I've got a work around though, a temporary class that I monkey patch into pandas, to make altair happy.

jakevdp:
OK – I'd support pinning pandas to ``>=0.18.0`` if that would help you.

PR #1793, opened 2019-11-18
ENH: add universal HTML renderer

jakevdp:
This adds the ``html`` renderer to Altair. If you use
```python
alt.renderers.enable('html')
```
The chart will render correctly, with the correct versions of Vega/Vega-Lite/Vega-Embed, in a variety of frontends.

I have tested that this renderer displays charts correctly in Jupyter Notebook, JupyterLab, Colab, Kaggle, nbviewer, and nbconvert HTML, without the need for any separately installed frontend extensions.

The downside is that unlike the vega labextension and nbextension the HTML renderer loads JS resources from the web, and so it cannot be used offline.

### DIFF
--- a/altair/utils/html.py
+++ b/altair/utils/html.py
@@ -90,10 +90,12 @@ HTML_TEMPLATE_UNIVERSAL = jinja2.Template("""
     const outputDiv = "{{ output_div }}";
     const baseUrl = "{{ base_url }}"
     const element = document.getElementById(outputDiv);
+    const vegaVersion = "{{ vega_version }}";
+    const vegaLiteVersion = "{{ vegalite_version }}";
+    const vegaEmbedVersion = "{{ vegaembed_version }}";
 
     function loadScript(url) {
-      var fullUrl = `${baseUrl}/${url}`
-      console.log(`loadScript("${fullUrl}")`);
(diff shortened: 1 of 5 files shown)


altair/utils/html.py
```diff
@@ -10,12 +10,6 @@
 <head>
 {%- endif %}
   <style>
-    .vega-actions a {
-        margin-right: 12px;
-        color: #757575;
-        font-weight: normal;
-        font-size: 13px;
-    }
```
jakevdp: Removed this because it's unnecessary with current versions of vega-embed.

domoritz:
This is a great fallback for when you need a specific version.

@jasongrout even if the Vega developers (probably me) miss backwards compatibility accidentally, there will be a viable fallback option for users of newer jupyterlab versions that don't ship with old versions of the Vega renderer.

domoritz:
This is a great fallback for when you need a specific version.

@jasongrout even if the Vega developers (probably me) miss backwards compatibility accidentally, there will be a viable fallback option for users.

jakevdp:
After some discussions offline & on slack among maintainers, we will likely make this the default renderer in Altair 4.0. But either way we want this in as an option.

PR #1813, opened 2019-11-27
Remove Python 2 support

jakevdp:


jakevdp:
There is still a lot of cleanup that can happen (particularly in moving comment-based type annotations to py3 type annotations) but I'll call this good for now.

PR #1824, opened 2019-12-01
DOC: update docs for V4 changes

jakevdp:
Still need to add sections on new transforms, etc. Will do that as another PR.


jakevdp:
There are some strange definition names introduced in this schema; I've opened an issue here: https://github.com/vega/vega-lite/issues/5603

For now, we'll go with them, and hope they are fixed in the next schema version.

PR #1830, opened 2019-12-02
DOC: add image mark example

eitanlees:
I added an example and also edited the table on the marks page to link to it


jakevdp:
Looks good, thanks!

PR #1831, opened 2019-12-02
DOC: add qq example

eitanlees:


jakevdp:
Nice! I was thinking of adding something like this.

PR #1832, opened 2019-12-03
DOC: add poly fit example with regression

eitanlees:
In light of the new transformations I have revisited some of the examples. Here I have added a new example based on the Polynomial Fit Plot. 

The only difference in the output is that there is no legend.


jakevdp:
I think I would rather replace the old polynomial fit example with this one. What do you think?

Also, it may be cleaner to use a fold transform + color encoding rather than explicit layers with manual colors.

eitanlees:
Good idea. 

I've got it working now. 

visualization (24)

jakevdp:
Looks good! I was initially hoping we could apply the regression transform repeatedly to the same chart, but it looks like that's not possible due to how the regression dataflow works. This is a nice solution.

PR #1833, opened 2019-12-03
Deprecation Warning Cleanup

eitanlees:
Cleaned up a few deprecation warnings I found when running `pytest`.

Specifically:
1. Added a `Loader` parameter when loading `yaml`
2. Changed `getargspec` to `getfullargspec` and the `keywords` attribute to `varkw`
3. Changed `import collections` to `from collections.abc import Mapping` and modified the corresponding `Mapping` calls.


jakevdp:
LGTM, thanks!

PR #1863, opened 2019-12-10
DOC: add section on customizing chart size

jakevdp:


jakevdp:
This includes setting chart divs to 100% so that dynamic width charts can scale appropriately.

PR #1864, opened 2019-12-10
DOC: refactor some examples which use EncodingSortField

eitanlees:
I read in `changes.rst` that now you can sort by an encoding channel directly. I thought a few of the examples could be refactored to reflect this. 

Note there were two places where I couldn't replace `EncodingSortField`

- in `beckers_barley_trellis_plot.py` it didn't seem to work with the row encoding. 
- in `sorted_error_bars_with_ci.py` I had trouble because the x encoding was being used twice. 

After going through this process I think the sorting documentation could use an update. Maybe I will make an issue.

### DIFF
--- a/altair/examples/bar_chart_sorted.py
+++ b/altair/examples/bar_chart_sorted.py
@@ -11,12 +11,5 @@ source = data.barley()
 
 alt.Chart(source).mark_bar().encode(
     x='sum(yield):Q',
-    y=alt.Y(
-        'site:N',
-        sort=alt.EncodingSortField(
-            field="yield",  # The field to use for the sort
-            op="sum",  # The operation to run on the field prior to sorting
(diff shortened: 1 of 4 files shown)


jakevdp:
I think it doesn't work in those examples because they require an aggregate.

Sorting doc update would be welcome! If you do it quickly we can get it into the 4.0 release (I'm hoping to get it out tonight)

jakevdp:
(totally unrelated, but FYI I removed your image mark example from a few days ago because it broke the doc build due to https://github.com/vega/vega/issues/2238. We can add that back when the bug is fixed)

eitanlees:
Cool, I will update the sorting docs now. I figured a new release was iminate! 

No worries about the broken example. c'est la vie!

jakevdp:
Thanks!

PR #1869, opened 2019-12-12
DOC: add area chart with gradient fill example

eitanlees:
I noticed we didn't have a gradient fill example.


jakevdp:
Looks good, thanks!

PR #1874, opened 2019-12-14
DOC: update candlestick example to use ohlc dataset

eitanlees:
I think this example is much better now.


jakevdp:
Looks great, thanks for this! (and fore all the multi-package prep work that enabled it)

eitanlees:
No problem! It’s great when a plan comes together!

Issue #1899, opened 2020-01-06
How to hide Axis Labels?


odoublewen:
I've searched through the docs and online forums, but can't accomplish what I thought should be a simple chart customization:  I would like to remove all **labels** from the axes -- but I still want the axis title and ticks to be drawn (so this is different from #1342).

I'm very sure it's possible and I'm pretty sure it's in your docs (which are very good).  But could this be a worked example on the [customization examples](https://altair-viz.github.io/user_guide/customization.html#adjusting-axis-labels) page? 

Many thanks!

jakevdp:
You can hide the axis labels by setting ``labels=False`` within the axis. For example:
```python
import altair as alt
from vega_datasets import data
cars = data.cars()

alt.Chart(cars).mark_point().encode(
    x=alt.X('Horsepower', axis=alt.Axis(labels=False)),
    y=alt.Y('Miles_per_Gallon', axis=alt.Axis(labels=False)),
    color='Origin',
).interactive()
```
![visualization (37)](https://user-images.githubusercontent.com/781659/71837446-4b3acc80-306b-11ea-9284-abf462d47e6e.png)

jakevdp:
It would be worth adding an example like this to the docs.

odoublewen:
:Doh:    Right in front of my eyes, sorry to have bothered you.

I'd be happy to make a PR to put this example to your docs, but maybe you've already done it.

Thank you.

jakevdp:
> I'd be happy to make a PR to put this example to your docs

That would be great! Please let me know if you need any help getting started.

IndianBOYinDetroit:
@jakevdp how can I hide alternate label  ?For example if my list has following values(0,5,10,15,20,25,30), on x0axis I want to show 0,  blank, 10, blank, 20, blank 30....

jakevdp:
Set `alt.Axis(tickMinStep=10)`

jayanti-disys:
> You can hide the axis labels by setting `labels=False` within the axis. For example:
> 
> ```python
> import altair as alt
> from vega_datasets import data
> cars = data.cars()
> 
> alt.Chart(cars).mark_point().encode(
>     x=alt.X('Horsepower', axis=alt.Axis(labels=False)),
>     y=alt.Y('Miles_per_Gallon', axis=alt.Axis(labels=False)),
>     color='Origin',
> ).interactive()
> ```
> 
[...]

PR #1901, opened 2020-01-06
customization example - how to remove axis labels

odoublewen:
When plots have categorical data on one or both axis, and when there are a lot of categories, the axis labels may not be useful.  So this is an easy way to remove them.

For this example, I'm just using the same plot from the above example.  If this is too contrived an example, I could come up with something more reasonable.

#1899


jakevdp:
Looks great, thanks!

PR #1907, opened 2020-01-08
DOC: Fix interval selection bug

### LINKED ISSUE #1906: Interval Selection Example Bug
I am having trouble with the the [Interval Selection Example](https://altair-viz.github.io/gallery/interval_selection.html). 
interval-selection

```python
import altair as alt

source = data.sp500.url

brush = alt.selection(type='interval', encodings=['x'])

upper = alt.Chart(source).mark_area().encode(
    alt.X('date:T', scale=alt.Scale(domain=brush)),
    y='price:Q'
).properties(
    width=600,
    height=200
)

lower = upper.properties(
    height=60
).add_selection(brush)

upper & lower
```

The example looks correct so I am unsure what is causing this behavior.

eitanlees:
closes #1906

### DIFF
--- a/altair/examples/interval_selection.py
+++ b/altair/examples/interval_selection.py
@@ -13,15 +13,19 @@ source = data.sp500.url
 
 brush = alt.selection(type='interval', encodings=['x'])
 
-upper = alt.Chart(source).mark_area().encode(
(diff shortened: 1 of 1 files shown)


jakevdp:
Nice catch - thanks!

PR #1912, opened 2020-01-13
Fixed test_save by closing file descriptor

General-Sax:
All test conditions of vegalite/v3/test_api/test_save were previously failing when run. Determined that an open file descriptor automatically returned by tempfile.mkstemp() was blocking read/write access to temp file in question, preventing test from running nominally.

Added line to immediately close the file descriptor to resolve this; afterward all tests in vegalite/v3/test_api.py run and passed.


jakevdp:
Thanks!

PR #1913, opened 2020-01-13
Fix broken test: test_api.test_save()

General-Sax:
All test conditions of vegalite/v4/test_api/test_save were previously failing when run. Determined that an open file descriptor automatically returned by tempfile.mkstemp() was blocking read/write access to temp file in question, preventing test from running nominally.

Added line to immediately close the file descriptor to resolve this; afterward all tests in vegalite/v4/test_api.py run and passed.


General-Sax:
Sorry for submitting these as two separate PR's; at least they're essentially identical and only one line each!

jakevdp:
Thanks!

Issue #1922, opened 2020-01-20
pandas nullable integer type in sanitize_dataframe


vadella:
When given an array of [nullable integers](https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html), `sanitize_dataframe` raises a `TypeError: data type not understood`.

Minimal example:

```
>>> import altair as alt
>>> import pandas as pd
>>> df = pd.DataFrame({"x": [1,2,3, None, 5]}, dtype="Int64")
>>> df
     x
0    1
1    2
2    3
3  NaN
4    5
>>> alt.utils.sanitize_dataframe(df)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Miniconda3\lib\site-packages\altair\utils\core.py", line 223, in sanitize_dataframe
    elif np.issubdtype(dtype, np.integer):
  File "C:\Miniconda3\lib\site-packages\numpy\core\numerictypes.py", line 393, in issubdtype
    arg1 = dtype(arg1).type
TypeError: data type not understood
```

I've also tested this with pandas version 0.25 and 1.0.0rc0 and altair 3.2.0 and 4.1.0dev0

`alt.utils.sanitize_dataframe(df.astype(float))` works as expected

jakevdp:
Nullable integers didn't exist when this code was written.

Are you interested in preparing a pull request?

PR #1924, opened 2020-01-21
Nullable integers and string dtype

vadella:
Add support for the new dtypes in `utils.sanitize_dataframe` to solve #1922

- [nullable integer datatype](https://pandas.pydata.org/pandas-docs/version/0.25/whatsnew/v0.24.0.html#optional-integer-na-support) 
- [dedicated string dtype](https://pandas.pydata.org/pandas-docs/version/1.0.0/whatsnew/v1.0.0.html#dedicated-string-data-type)

### DIFF
--- a/altair/utils/tests/test_utils.py
+++ b/altair/utils/tests/test_utils.py
@@ -167,3 +167,29 @@ def test_sanitize_string_dtype():
         "string_object_null": ["a", "b", None, "d"],
         "string_string_null": ["a", "b", None, "d"],
     }
+
+
+@pytest.mark.skipif(
+    not hasattr(pd, "BooleanDtype"),
+    reason="Nullable boolean dtype not supported in pandas v{}".format(pd.__version__)
+    )
+def test_sanitize_boolean_dtype():
+    df = pd.DataFrame(
+        {
+            "bool_none": pd.array([True, False, None], dtype="boolean"),
+            "none": pd.array([None, None, None], dtype="boolean"),
+            "bool": pd.array([True, False, True], dtype="boolean"),
+        }
+    )
+
+    df_clean = sanitize_dataframe(df)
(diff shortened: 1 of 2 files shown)


vadella:
To make the tests pass for pandas versions which do not support the new dtypes, I surrounded it with 
```
try:
    ...
except TypeError:  # dtype not supported
    return
```

An alternative here would be 
```
try:
    ...
except TypeError:  # dtype not supported
    assert pd.__version__ <= <min_pandas_version_for_feature>
    return
```

Would that be better?

altair/utils/tests/test_utils.py
```diff
@@ -106,3 +106,61 @@ def test_sanitize_dataframe_infs():
     df_clean = sanitize_dataframe(df)
     assert list(df_clean.dtypes) == [object]
     assert list(df_clean['x']) == [0, 1, 2, None, None, None]
+
+
+def test_sanitize_nullable_integers():
```
jakevdp: Rather than silently passing if there is a type error, it would be better to explicitly skip this test for older versions of pandas. The cleanest way would be to add a decorator like:
```python
@pytest.mark.skipif(not hasattr(pd, "Int64Dtype"), "Nullable integers not supported in pandas v{}".format(pd.__version__))
def test_sanitize_nullable_integers():
    ...
```
and then remove the ``try...except`` statement.

Similar for the string test below.

vadella:
I've changed it to use `pytest.mark.skipif`

I've also added the behaviour to the docstring.

Initially I forgot the nullable boolean dtype. 
For clarity, I added this as a separate path in the `elif` tree, instead of adding it to  one of the other branches even though what happens is the same as for example the `category` dtype.

jakevdp:
Looks great, thanks!

PR #1955, opened 2020-02-07
DOC: add scatter with loess example

eitanlees:
visualization (50)

I thought an example of adding multiple grouped LOESS lines to a scatter plot could be a useful addition to the docs.

### DIFF
--- a/altair/examples/scatter_with_loess.py
+++ b/altair/examples/scatter_with_loess.py
@@ -12,14 +12,17 @@ import numpy as np
 
 np.random.seed(42)
 
-source = pd.DataFrame(np.cumsum(np.random.randn(100, 3), 0).round(2),
-                    columns=['A', 'B', 'C'], index=pd.RangeIndex(100, name='x'))
-source = source.reset_index().melt('x', var_name='category', value_name='y')
+source = pd.DataFrame(np.cumsum(np.random.randn(100, 3), 0),
+                      columns=['A', 'B', 'C'], 
+                      index=pd.RangeIndex(100, name='x')).reset_index()
 
-base = alt.Chart(source).mark_circle(opacity=0.5).encode(
-    alt.X('x'), 
-    alt.Y('y'), 
-    alt.Color('category')
+base = alt.Chart(source).mark_circle(opacity=0.5).transform_fold(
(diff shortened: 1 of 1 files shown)


jakevdp:
Nice example! What would you think about replacing the ``melt()`` operation with a fold transform?

eitanlees:
Sure thing. I based this example on [Multi-Line Tooltip](https://altair-viz.github.io/gallery/multiline_tooltip.html) which used `melt()`.

Now that the transforms have improved some of the examples are outdated in subtle ways. In general do we want to use built in transforms?

jakevdp:
>  In general do we want to use built in transforms?

Good question... thinking about it, I'd say we should use built-in transforms whenever they make the examples clearer. I think the fold transform here is a bit clearer than ``pd.melt``; I also think that ``reset_index()`` here is probably clearer than using ``transform_window(x='count()')``.

eitanlees:
Personally I like the example given from the [Fold Transform](https://altair-viz.github.io/user_guide/transform/fold.html#user-guide-fold-transform) page where the data is intialized as such. 

```python
import numpy as np
import pandas as pd
import altair as alt

rand = np.random.RandomState(0)
data = pd.DataFrame({
    'date': pd.date_range('2019-01-01', freq='D', periods=30),
    'A': rand.randn(30).cumsum(),
    'B': rand.randn(30).cumsum(),
    'C': rand.randn(30).cumsum(),
})

alt.Chart(data).transform_fold(
    ['A', 'B', 'C'],
).mark_line().encode(
    x='date:T',
    y='value:Q',
    color='key:N'
)
```

I feel this is a nice way to build a toy dataframe that doesn't need any index resetting. If you agree I can change this example to have that form.

eitanlees:
Maybe I should revisit the idea of an Altair examples style guide.

jakevdp:
Yeah, that looks great: everything is very explicit and understandable, even to a reader who's not very familiar with pandas.

Would you be willing to update this PR with that style?

eitanlees:
Absolutely! I think this example is good to go! 

visualization (51)

jakevdp:
Thanks!

PR #1974, opened 2020-02-17
FIX: change datum repr to fix doc build

eitanlees:
discussed in https://github.com/altair-viz/altair/issues/1972


jakevdp:
Looks good, thanks!

PR #1983, opened 2020-02-23
DOC: fix vconcat example on compound charts guide

eitanlees:
I found another place where the interval selection bug (https://github.com/altair-viz/altair/pull/1907) occurred.


jakevdp:
Nice catch - thanks

PR #1985, opened 2020-02-24
Add a convenience method for tooltips

hannahburkhardt:
Hello, I thought it would be nice to have a simple method that adds a tooltip with all variables, as I find myself doing this fairly frequently. The new method I am suggesting is based on Chart.interactive(). I have only used pandas DataFrames, so that is the only one I added for now. Let me know what you think.

### DIFF
--- a/altair/vegalite/v4/api.py
+++ b/altair/vegalite/v4/api.py
@@ -1776,7 +1776,7 @@ class Chart(TopLevelMixin, _EncodingMixin, mixins.MarkMethodMixin,
         Returns
         -------
         chart :
-            self, with interactive axes added
+            self, with tooltips added for every column in data
 
         """
         if not isinstance(self.data, pd.DataFrame):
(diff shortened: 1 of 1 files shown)


jakevdp:
There's already a way to do this succinctly, using the ``tooltip`` argument to ``mark_*()`` methods. For example:
```python
import altair as alt
cars = data.cars()

alt.Chart(cars).mark_point(tooltip=True).encode(
  x='Horsepower',
  y='Miles_per_Gallon',
  color='Origin'
).interactive()
```
You can also use
```
tooltip=alt.TooltipContent('encoding')
```
to just get tooltips for the encodings, or
```
tooltip=alt.TooltipContent('data')
```
to get tooltips for the entire dataset.

The benefit of this is that it works regardless of whether the data is specified as a dataframe or by URL.

hannahburkhardt:
This is perfect, thank you!

PR #1990, opened 2020-02-26
WIP: Update to Vega-Lite 4.8

jakevdp:
Tests will not pass until VL 4.8 is part of altair_viewer.


jakevdp:
Merging for now; we can iterate and then work on the release.

PR #1991, opened 2020-02-26
Add Black formatting for most files

jakevdp:
This adds a reformat step to ``make test``, as well as a format github action.

This initial commit leaves out ``altair/examples`` (because we want to keep our examples formatted as-is), and also leaves out ``altair/vegalite``, and ``altair/vega`` for now. As we think more deeply about how to handle auto-generated code, we may add these last two to the mix.


jakevdp:
@ellisonbg @eitanlees any objections?

This would mean that new contributions could not be merged unless they're run through the black formatter, examples & auto-generated code aside.

eitanlees:
I am ok with introducing a strict fromater as long as it's documented in the [CONTRIBUTING.md](https://github.com/altair-viz/altair/blob/master/CONTRIBUTING.md) so that people know what to expect.

jakevdp:
OK - updated contributing (and made it a bit more general).

Thanks for the suggestion!

PR #2038, opened 2020-03-23
Add an example showing how to create a minimap

jonathangjertsen:
Implements what was discussed here: https://github.com/altair-viz/altair/issues/2037

I think this is a useful recipe for large time-series data -- if you only have one view, you kind of need to choose between having enough detail and having enough context.


jakevdp:
This is a nice example - thanks!

PR #2063, opened 2020-04-04
BUG: support pdf format in chart.save

### LINKED ISSUE #2062: chart.save does not support pdf as shown in docs
[The docs](https://altair-viz.github.io/user_guide/saving_charts.html#png-svg-and-pdf-format) for saving charts show `chart.save('chart.pdf')` as one of the options, but running that results in `ValueError: unrecognized format: 'pdf'`, even though `altair_saver` is installed and `from altair_saver import save; save(chart, "chart.pdf")` works.

By contrast, `chart.save("chart.svg")` works fine.

Looking at the [docstring of `Chart.save`](https://altair-viz.github.io/user_guide/generated/toplevel/altair.Chart.html#altair.Chart.save), it looks like pdf isn't even supposed to work: "Supported formats are json, html, png, svg". But maybe it's just a matter of wiring it up?

If the issue is more complicated than that (maybe because `Chart.save` is currently hardwired to use the Selenium backend, which doesn't offer pdf export?), then I would suggest just removing `chart.save('chart.pdf')` from the code example and refer users who are interested in pdf export functionality to `altair_saver`'s docs :)

jakevdp:
Fixes #2062

### DIFF
--- a/altair/utils/save.py
+++ b/altair/utils/save.py
@@ -73,7 +73,7 @@ def save(
             format = fp.split(".")[-1]
         else:
             raise ValueError(
-                "must specify file format: " "['png', 'svg', 'html', 'json']"
+                "must specify file format: " "['png', 'svg', 'pdf', 'html', 'json']"
             )
 
     spec = chart.to_dict()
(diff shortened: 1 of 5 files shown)


ADR777:
Hi Jake

I am still not able to save pdf. Do you know if this issue has been resolved?

jakevdp:
This is merged to master but not yet part of a release.

PR #2069, opened 2020-04-04
DOC: tiny update

yogabonito:


jakevdp:
Thanks!

yogabonito:
Thank you, Jake, for this awesome package!
There might be something else on this doc-page needing an update: the plots in [this section](https://altair-viz.github.io/user_guide/encoding.html#effect-of-data-type-on-axis-scales) are intended to demonstrate that `scale=alt.Scale(zero=False)` makes a difference, but the left part of the first plot looks exactly the same as the plot below.

jakevdp:
Thanks for pointing that out! That was a behavior change in one of the recent vega-lite updates. Are you interested in putting together a PR either fixing or removing that?

yogabonito:
Done :) (#2076)

PR #2074, opened 2020-04-08
DOC: tiny updates

yogabonito:


jakevdp:
Thanks!

I've also been meaning to adjust that section and point to the native interactive legends that were introduced in version 4.0

PR #2087, opened 2020-04-13
Added example with line chart on a logarithmic scale

palewire:
A new line chart example that shows how to set [a log scale](https://en.wikipedia.org/wiki/Logarithmic_scale).

```python
import altair as alt

source = data.population()

alt.Chart(source).mark_line().encode(
    x='year:O',
    y=alt.Y(
        'sum(people)',
        scale=alt.Scale(type="log")  # Here the scale is applied
    )
)
```

visualization (4)


jakevdp:
Thanks!

PR #2088, opened 2020-04-13
Added example of bar chart with rolling mean

palewire:
```python
import altair as alt

source = data.wheat()

    x='year:O',
    y='wheat:Q'
)

line = alt.Chart(source).mark_line(color='red').transform_window(
    # The field to average
    rolling_mean='mean(wheat)',
    # The number of values before and after the current value to include.
    frame=[-9, 0]
).encode(
    x='year:O',
    y='rolling_mean:Q'
)

(bar + line).properties(width=600)
```

visualization (5)


jakevdp:
Looks good, thanks!

PR #2091, opened 2020-04-14
Added example of a cumulative line chart

palewire:
```python
import altair as alt

source = data.wheat()

alt.Chart(source).mark_line().transform_window(
    # Sort the data chronologically in this case
    sort=[{'field': 'year'}],
    # Include all previous records before the current record and none after
    frame=[None, 0],
    # What we want to add up
    cumulative_wheat='sum(wheat)',
).encode(
    x='year:O',
    # Plot the calculated field created by the transformation
    y='cumulative_wheat:Q'
).properties(width=600)
```

visualization (6)

### DIFF
--- a/altair/examples/line_chart_with_cumsum.py
+++ b/altair/examples/line_chart_with_cumsum.py
@@ -1,0 +1,23 @@
+"""
+Line Chart with Cumulative Sum
+------------------------------
+This chart creates a simple line chart from the cumulative sum of a fields.
+"""
+# category: line charts
+import altair as alt
+from vega_datasets import data
+
+source = data.wheat()
+
+alt.Chart(source).mark_line().transform_window(
+    # Sort the data chronologically in this case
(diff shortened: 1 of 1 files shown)


altair/examples/line_chart_with_cumsum.py
```diff
@@ -0,0 +1,23 @@
...
+import altair as alt
+from vega_datasets import data
+
+source = data.wheat()
+
+alt.Chart(source).mark_line().transform_window(
+    # Sort the data chronologically
+    sort=[{'field': 'year'}],
+    # Include all previous records before the current record and none after
+    frame=[None, 0],
```
jakevdp, line 16: Maybe note that this is the default value, so specifying ``[None, 0]`` here is optional

palewire: Done

jakevdp:
Thanks!

PR #2093, opened 2020-04-16
DOC: tiny update

yogabonito:
The mentioned example is no longer a good example (due to a Vega-Lite update). Thus, the link was removed.


jakevdp:
Thanks!

PR #2106, opened 2020-04-21
DOC: Update "Line Chart with Points" example.

### LINKED ISSUE #2100: Options of `point` argument
Documentation of `point` argument in marks isn't very easily discoverable.

Recently I tried to create a chart with colored line and points. My first try was with `mark_line(color="red", point=True)`, similarly to [this example](https://altair-viz.github.io/gallery/line_chart_with_points.html), but it produced red line with blue points. This turned out to be current default behavior of Vega-lite (found in [this issue](https://github.com/vega/vega-lite/issues/6111)).

After some digging, I found that `point` can be "an object defining the properties of the overlayed points." (which seems to imply `OverlayMarkDef`, but simple dictionary showed to be enough). I think it would be really helpful to have this slightly easier to find. A "Line chart with points" example seems like a good place.

So, what do you think about adding some sort of variation of the following code snippet in that example?
```python
import altair as alt
import numpy as np
import pandas as pd

x = np.arange(100)
source = pd.DataFrame({"x": x, "f(x)": np.sin(x / 5)})

# Only this part will be added
alt.Chart(source).mark_line(
    color="red", point={"color": "red"}
).encode(
    x="x",
    y="f(x)"
)
```
issue_altair_point-argument_example

If this is OK, I would like to help with PR. Thanks.

echasnovski:
This is a PR to fix #2100.
I couldn't dare to add second chart as it seems that only one per example is desired. So I decided to update text description. I hope that is OK.

### DIFF
--- a/altair/examples/line_chart_with_points.py
+++ b/altair/examples/line_chart_with_points.py
@@ -1,7 +1,9 @@
 """
 Line Chart with Points
 ----------------------
-This chart shows a simple line chart with points marking each value.
+This chart shows a simple line chart with points marking each value. Use
+``point=True`` for points with default appearance or customize it with
+``OverlayMarkDef()``.
 """
 # category: line charts
 import altair as alt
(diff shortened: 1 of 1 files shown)


jakevdp:
Looks good, thanks!

PR #2169, opened 2020-05-25
CLN: Use f-strings over format

dsaxton:
I'm not sure if this project prefers f-strings, but if so here's a PR to use those instead of format in a few places. Also fixes a typo in the contributing guidelines.


jakevdp:
Looks good - just to be safe, I'd like to get the chrome version bug fixed before merging this, in order to make sure this doesn't break any tests.

PR #2170, opened 2020-05-25
Don't cast pandas.Categorical to object

dsaxton:
I think the pandas bug mentioned in this comment is fixed so it may be possible to skip the conversion to object entirely

https://github.com/altair-viz/altair/issues/245#issuecomment-633165346

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -282,10 +282,7 @@ def sanitize_dataframe(df):  # noqa: C901
 
     for col_name, dtype in df.dtypes.iteritems():
         if str(dtype) == "category":
-            # XXXX: work around bug in to_json for categorical types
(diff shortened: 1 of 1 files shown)


RoyalTS:
How come this PR was never merged? The pandas bug that's referenced in comment really does seem to have been fixed and recasting categories to object screws up e.g. ordering of axes.

jakevdp:
This was not merged because I was in the middle of an interstate move during a global pandemic and I ignored github for a while :grin:

Can you confirm that this fixes any issue? JSON doesn't support categorical, so I don't think converting to string a different way will change how outputs are ordered. Also, I'm not certain which pandas release has this fix, and it's useful to support older versions of dependencies unless breaking that support is absolutely necessary.

PR #2212, opened 2020-06-18
compound chart changes due to movies dataset column name change

johnyrufus:
movies dataset column names have changed to remove the _ replaced by spaces. So the user guide compound chart is not rendering correctly.


jakevdp:
Thanks!

PR #2221, opened 2020-06-25
Broader data support in Chart

Rik-de-Kort:
I use Altair pretty much exclusively as a visualization library. One action that comes up moderately often is having to cast things to a DataFrame before Altair will accept them. For me, a nice feature would be to have broader support for different data types in `Chart`.

* I've implemented this support for `pd.Series`, allowing rug plots, histograms, etc. without the call to `to_frame()`.
* I plan on experimenting with pure Python data types. I know we can pass a `Data` object, but this is still mystical to me. Maybe all that comes out of this is some updates to the documentation.
* I tend to make good use of pandas multiindex functionality. I'm a bit mystified why having a multiindex in the index raises an error: it gets dropped anyway in the call `to_records`. Is there any specific reason for that?
* On the subject of indices, maybe we could make sure the index data gets preserved. For instance, currently it is quite a lot of lines to make a density plot of categorical variable. It would be quite nice to just type `alt.Chart(df.column.value_counts(normalize=True)).mark_bar().encode(x="index", y="column")` instead of having to reset the index and rename the column before passing it into `Chart`.

Let me know what you think!


jakevdp:
Regarding pure Python data types: the simplest data specification (and what all others get turned into in the end) is a list of dicts:
```
data = {"values": [{"val": 1, "label": "A"}, {"val": 2, "label": B"}]}
```

altair/utils/core.py
```diff
@@ -240,21 +240,113 @@ def sanitize_geo_interface(geo):
     return geo
 
 
+def sanitize_series(s):
```
jakevdp, line 243: Could the goal of this change be accomplished by leaving ``sanitize_dataframe`` untouched, and adding this?
```
def sanitize_series(s):
  return sanitize_dataframe(s.to_frame())
```

jakevdp: Or perhaps ``sanitize_dataframe(s.reset_index())`` since visualization of series data rarely makes sense without its index?

Rik-de-Kort: Yeah, that would be doable.

I refactored `sanitize_dataframe` because for someone coming into the codebase fresh it would make more sense have `sanitize_dataframe` depend on `sanitize_series` than the other way around. Let me know if you still prefer for me to change it.

In the context of `reset_index`, I think that if it is done for series, it should also be done for dataframes to make sure things stay consistent.

jakevdp: I disagree on defaulting to reset_index. Size of serialized data is a concern in Alatir; not serializing the index by default allows people to choose whether they want the index included (by using `reset_index` or not). If we reset the index by default, there is no way to opt-out of including an extra column in every data source derived from a dataframe.

Rik-de-Kort: Looking at it from my daily work, I am pretty much *always* typing `reset_index()` because my indices have meaningful information about the data such as the iteration number, the year, or a personal identifier that I want to include in my visualizations *and* use as indices.  
I understand the need to be efficient with respect to the size of serialized data. By adding `reset_index`, worst case scenario is formed with an inline dataset with 1 column and many rows that gets repeated a lot. But (at least for me) it is quite rare to have very few columns and the relative damage is a lot less with more columns. If the user is that close to the limits of inline data, they can easily start using the json transformer.
I should note I always use the json transformer so maybe I'm not feeling the pain of serialization issues.

With regards to the comment about series support, I think that's a point well-taken. I took Series because it seemed to be the closest one to DataFrame and thus the easiest to implement by mimicking what is done for DataFrame.

jakevdp: I don't see any downside to Altair supporting series input by calling `to_frame()` to convert it to a dataframe. I think that would be a convenient feature.

jakevdp: Regarding the JSON data transformer: it doesn't work consistently across different display frontends, so I've stopped recommending it as a general solution to larger datasets.

Rik-de-Kort: I've rolled back most of the sanitize series support, so it's now a simple call to `reset_index()`. Of course I'll also add the docs for it. The weird thing about it still is that you need to explicitly state the encoding type even though pandas can infer the dtypes of a series easily.

I also had a look at dictionaries. Do you think it's worth to include the cast to a Dataframe there as well (simply calling `pd.DataFrame` on the underlying dict)? Or to take a list of records and adding the `{"values": <stuff>}` bit automatically, or casting that list to a DataFrame as well?  

Additionally I discovered that the error for a dict not including a "values" key (in the `to_values` transformer) will never get raised: first schema validation fails, and if validation is turned off, the `_prepare_data` function in `api.py` doesn't route dicts through the transformers.

In regards to the JSON data transformer, any idea as to potential universal solutions? Altair server?

jakevdp: Altair server is a good option, except when it isn't :grin: (for example, when using a jupyterhub with no port forwarding).

Rik-de-Kort: Any comments on the dict and list ideas? If not, I'll take them out and complete the PR with just the series functionality.

altair/utils/data.py
```diff
@@ -148,19 +150,28 @@ def to_values(data):
...
+        data = sanitize_dataframe(data.reset_index())
+        return {"values": data.to_dict(orient="records")}
+    elif isinstance(data, dict) and "values" in data:
         return data
+    elif isinstance(data, dict):
+        data = pd.DataFrame(data)
+        data = sanitize_dataframe(data)
+        return {"values": data.to_dict(orient="records")}
+    elif isinstance(data, list):
+        data = pd.DataFrame(data, columns=[f"x{i}" for i in range(len(data))])
```
jakevdp, line 163: I think I'd prefer not to support generic lists like this. In particular, given how central column names are to Altair chart encodings, it seems unhelpful to generate generic ones in a way that is opaque to the user.

Issue #2213, opened 2020-06-19
Update docs for vega_datasets v2


eitanlees:
The [vega/vega-datasets](https://github.com/vega/vega-datasets) repository recently released a major update. We are working on updating [altair-viz/vega_datasets](https://github.com/altair-viz/vega_datasets) to reflect v2. As a result many examples throughout the docs will need to be updated. 

After https://github.com/altair-viz/vega_datasets/issues/39 is closed we can begin the process of cleaning up the docs. 

For now I just wanted to make note of the transition.

jakevdp:
Related: #2210, #2212

jakevdp:
Note that docs are currently broken due to the changes in vega/vega-datasets, due to using URL data that has changed or been removed.

eitanlees:
I have compiled a big list of everything that needs to be changed once https://github.com/altair-viz/vega_datasets/pull/40 is merged. 

## From the Example Gallery
### Broken
- [ ] [Binned Scatterplot](https://altair-viz.github.io/gallery/binned_scatterplot.html#binned-scatterplot) due to `movies`
- [ ] [Interactive Chart with Cross-Highlight](https://altair-viz.github.io/gallery/interactive_cross_highlight.html#interactive-chart-with-cross-highlight) due to `movies`
- [ ] [Seattle Weather Heatmap](https://altair-viz.github.io/gallery/weather_heatmap.html#seattle-weather-heatmap) due to `seattle_temps`
- [ ] [Top K Items](https://altair-viz.github.io/gallery/top_k_items.html#top-k-items) due to `movies`
- [ ] [Top-K plot with Others](https://altair-viz.github.io/gallery/top_k_with_others.html#top-k-plot-with-others) due to `movies`
- [ ] [Binned Heatmap](https://altair-viz.github.io/gallery/binned_heatmap.html#binned-heatmap) due to `movies`
### Needs Update
- [ ] [Faceted Density Estimates](https://altair-viz.github.io/gallery/density_facet.html#faceted-density-estimates) due to `iris`
- [ ] [Stacked Density Estimates](https://altair-viz.github.io/gallery/density_stack.html#stacked-density-estimates) due to `iris`
- [ ] [Multifeature Scatter Plot](https://altair-viz.github.io/gallery/multifeature_scatter_plot.html#multifeature-scatter-plot) due to `iris`

## From the docs
### Broken
[...]

PR #2231, opened 2020-07-09
docs: Iris to penguins

baldwint:
Maybe someone is already working on this, but to help with #2213 I found it fun to rewrite three of the (currently broken) compound chart documentation sections using the new penguins dataset instead of iris.

I used the mapping
```python
{
    "sepalLength": "Beak Length (mm)",
    "sepalWidth": "Beak Depth (mm)",
    "petalLength": "Flipper Length (mm)",
    "petalWidth": "Body Mass (g)",
    "species": "Species",
}
```
since this seemed more or less faithful to the patterns in iris. I also chose to un-anchor the scales from zero since the penguin means are not very close to zero.

visualization

While working on this, I found it useful to make a [modified version](https://github.com/eitanlees/vega_datasets/compare/update-to-vega-datasets-v2...baldwint:altair-doc-shim) of `vega_datasets` that retains some of the older datasets so that the docs can build without failing on examples that haven't been updated yet.


jakevdp:
Thanks! This will have to wait for a vega-datasets release that includes the penguins data.

mattijn:
Pretty awesome to merge a 5 year old PR. Thanks for all the hard work here!

PR #2242, opened 2020-07-17
Add new implementation of horizon graph

jesserobertson:
I was interested in the horizon graph paper but the implementation on the altair website didn't include the negative offset parts as in the paper. So I implemented that myself!

Only change is to the code in the example. It generates a set of random sine wave data visualizations stacked together to show how you can scan across a number of horizon graphs to increase data density.

Apologies if the style isn't quite correct - I'm learning altair myself at the moment (and enjoying it so far!).


jesserobertson:
Oops forgot to include some output
visualization

PR #2262, opened 2020-08-15
fix: remove duplicate /

domoritz:
Right now, the generated html contains double `/`.

Screen Shot 2020-08-15 at 13 39 13


mattijn:
Thanks @domoritz. I'd noticed this as well before. I rebased so Github Actions could do a re-run. All passing. Let me merge this.

PR #2293, opened 2020-09-17
maint: black formatting

eitanlees:
I noticed #2292 failed due to formatting errors. 

I simply ran `black .` on the root directory, and it reformatted a few files.


jakevdp:
Probably changes in black version. Thanks!

PR #2310, opened 2020-10-07
Update URLs and field names to match new vega datasets

joelostblom:
This fixes the broken weather sections in the docs mentioned in https://github.com/altair-viz/altair/issues/2213. I upated based on my understanding of https://github.com/altair-viz/vega_datasets/pull/40.

### DIFF
--- a/altair/examples/weather_heatmap.py
+++ b/altair/examples/weather_heatmap.py
@@ -8,7 +8,7 @@ import altair as alt
 from vega_datasets import data
 
 # Since the data is more than 5,000 rows we'll import it from a URL
-source = data.seattle_temps.url
+source = data.seattle_weather_hourly_normals.url
 
 alt.Chart(
     source,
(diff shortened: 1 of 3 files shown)


joelostblom:
All test failures are related to me using the new name for the seattle data set from https://github.com/altair-viz/vega_datasets/pull/40, and should be fixed once that is merged.

jakevdp:
Thanks for this.

What needs to happen to fix this issue:

- a new minor release of vega-datasets 1.X that pins the old URLs, so that it can work with older Altair versions
- a new micro release of Altair with accompanying doc build with the older vega-datasets path
- a new minor release of Altair that makes all the required updates to work with vega-datasets 2.X

I've not done that yet, because it's a ~half day of work to get right, and I've not had an open half day in half a year.

jakevdp:
I should say: thanks for the contribution. I really appreciate it. But we can't merge changes like this until 1 & 2 are done.

joelostblom:
Thank you for responding so quickly! I understand that you are super busy and I didn't want to stress, more just checking in if I could do something helpful for moving this forward. But it sounds like most things you mentioned are release related, which sounds a bit tricky to help with.

mattijn:
fixed by https://github.com/vega/altair/pull/3859

PR #2313, opened 2020-10-09
Update renderer descriptions for 4.0 and fix mimebundle typo

joelostblom:
This is based of my understanding of [your comment here](https://github.com/altair-viz/altair/issues/2311#issuecomment-705877316) and issue https://github.com/altair-viz/altair/issues/2234. If `'colab'` is the same as `'html'` I can change to reflect that.

close https://github.com/altair-viz/altair/issues/2234


jakevdp:
Thanks, this looks much better!

PR #2320, opened 2020-10-18
Remove redundant column name

joelostblom:
I don't think this is needed is it? It seems to work for me without the column name.


jakevdp:
Looks good, thanks!

PR #2334, opened 2020-11-09
Use `line_scale` to set customized colors

morberg:
`line_scale` was defined but never used in the chart. The PR looks large since the file is reformatted using black. The only actual change is to add `scale=line_scale` to `alt.Color`.

### DIFF
--- a/altair/examples/london_tube.py
+++ b/altair/examples/london_tube.py
@@ -35,7 +35,7 @@ labels = alt.Chart(centroids).mark_text().encode(
 
 line_scale = alt.Scale(domain=["Bakerloo", "Central", "Circle", "District", "DLR",
                                "Hammersmith & City", "Jubilee", "Metropolitan", "Northern",
-                               "Piccadilly", "Victoria", "Waterloo & City" ],
+                               "Piccadilly", "Victoria", "Waterloo & City"],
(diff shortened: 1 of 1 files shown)


altair/examples/london_tube.py
```diff
@@ -9,48 +9,72 @@
...
-            orient='bottom-right',
-            offset=0
+lines = (
+    alt.Chart(tubelines)
+    .mark_geoshape(filled=False, strokeWidth=2)
+    .encode(
+        alt.Color(
+            "id:N",
+            legend=alt.Legend(title=None, orient="bottom-right", offset=0),
+            scale=line_scale,
```
morberg, line 77: This is the only actual change, the rest of the diff is due to black.

morberg:
Hi @jakevdp , I haven’t seen any feedback on this. Would you prefer a smaller PR without running black on the existing example?

jakevdp:
Yeah, sorry - can you make the change without reformatting the whole file? The altair examples generally are not in black format.

morberg:
I'm not sure how to squash these two commits once they are on github. Perhaps it would be cleaner to submit a new PR?

morberg:
I tried to do a local squash merge and then push that, but that [ended up even messier](https://xkcd.com/1597/)... I'll submit a new PR.

PR #2338, opened 2020-11-13
Corrected field names to allow example to work

quaesitor-scientiam:
Field names in the code had character '_' while the field names in the data uses a space character causing a mismatch and the example did not show correct data.


mattijn:
Not sure what happened at that time, but this is showing for me fine currently, https://altair-viz.github.io/gallery/interactive_cross_highlight.html. Closing this one. Thanks again @quaesitor-scientiam

PR #2355, opened 2020-11-30
allow Paths in save()

### LINKED ISSUE #2354: Allow Paths in save()
Instead of allowing only string paths or file-likes, allow pathlib.Paths to be passed to `save()`.

Are these two the only places that would have to be changed?

https://github.com/altair-viz/altair/blob/54e03d403c1cec9ce2f2e8b14dc3d936c6686128/altair/utils/save.py#L8
https://github.com/altair-viz/altair/blob/54e03d403c1cec9ce2f2e8b14dc3d936c6686128/altair/utils/save.py#L72

RoyalTS:
Fixes #2354

### DIFF
--- a/altair/utils/save.py
+++ b/altair/utils/save.py
@@ -1,11 +1,13 @@
 import json
+import pathlib
 
 from .mimebundle import spec_to_mimebundle
 
 
 def write_file_or_filename(fp, content, mode="w"):
-    """Write content to fp, whether fp is a string or a file-like object"""
-    if isinstance(fp, str):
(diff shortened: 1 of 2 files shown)


jakevdp:
This is great, thanks! It would be useful to add a test as well; the current test of saving to a string filepath is here: https://github.com/altair-viz/altair/blob/4f6dd28b3d8d540533d59ad1b638f4da8480b87e/altair/vegalite/v4/tests/test_api.py#L284-L292

You could add another few lines of code below this to save via a path object as well.

jakevdp:
Looks great, thanks!

PR #2375, opened 2020-12-23
Docs: Flip extent='max-min' to extent='min-max'

akanz1:
Fixed a small mistake in user_guide/marks

- extent='max-min' is not a valid argument -> Changed to extent='min-max'


jakevdp:
Looks good, thanks!

PR #2386, opened 2021-01-12
Add white background to cars.png to support dark mode Github

jonathangjertsen:
cars.png is not legible when using Github in dark mode because the background is transparent. Here is a screenshot of the diff when pointing README.md to https://raw.githubusercontent.com/jonathangjertsen/altair/master/images/cars.png instead of https://raw.githubusercontent.com/altair-viz/altair/master/images/cars.png:

Capture

There may be other images with the same issue in the docs etc, I haven't checked.


jakevdp:
Looks good, thanks!

Issue #2395, opened 2021-01-24
[Question] Draw scatterplot of wedges with encoded size and angle


dkapitan:
I want to implement [Zan Armstrong's comet chart](https://www.zanarmstrong.com/infovisresearch) in Altair, for which I need to draw a scatterplot of wedges with encoded start-point and end-points.  I used glyphs in [an earlier implementation in Bokeh](https://github.com/dkapitan/comet-chart).

I have read about `wedge` as one of the available marks, and managed to draw wedges, all with the same configuration e.g. just changing the angle.

```python
import altair as alt
import numpy as np
import pandas as pd

df = pd.DataFrame(
    np.hstack((np.random.randn(100, 2), np.random.randint(0, 360, (100, 1)))),
    columns=["x", "y", "angle"],
)

(alt.Chart(df).mark_point(shape='wedge', angle=135).encode(x="x:Q", y="y:Q"))
```

I was wondering whether it is possible to encode the wedges individually, where the length and direction are determined by the start and endpoint in the dataset. As a first test, I tried to encode `angle='angle:Q` in the example above, but that property isn't defined.

Vega-lite shape mark looks like another possible solution (see [this SO post](https://stackoverflow.com/a/62610941/3197404)) and I see it is accessible in Altair as shown [here](https://altair-viz.github.io/gallery/isotype.html)) but I don't know how to connect the dots.

If someone could point me in the right direction, I am happy to try and contribute a comet chart viz to Altair.

mattijn:
Altair compiles to Vega-lite version 4.8, where the `angle` channel encoding was added in Vega-lite 4.9. See release [docs](https://github.com/vega/vega-lite/releases/tag/v4.9.0) and [wind vector map](https://vega.github.io/vega-lite/examples/point_angle_windvector.html) for an example.

dkapitan:
@mattijn 
I think the solution using SVG paths seems do-able. This provides the basic mechanism I am looking for (thanks to @jakevdp solution on SO I found earlier):

```python
df2 = pd.DataFrame(
    {
        "a": [10,20,30,40],
        "b": [10,20,30,40],
        "c": ["M -1 0 L0 10 L1 0", "M -1 0 L0 -5 L1 0", "M -1 0 L0 -20 L1 0", "M -1 0 L0 -1 L1 0"],
    })
alt.Chart(df2).mark_point().encode(x="a", y="b", shape=alt.Shape('c', scale=None))
```

[...]

PR #2397, opened 2021-01-31
add comet chart example

dkapitan:
Here's the example resulting from https://github.com/altair-viz/altair/issues/2395.


morberg:
Comet chart is a quite new concept to many people. It seems wrong to use that label for this type of graph which is not (as you correctly pointed out in #2395 ) actually a comet graph. This example may be useful for other reasons, but it should be called something else.

jakevdp:
This looks great, thanks!

PR #2400, opened 2021-02-05
Force double_precision=15 in utils.data:_data_to_json_string

sjmielke:
Leaving it implied at pandas' [default of 10](https://github.com/pandas-dev/pandas/blob/5d65e0a3092f3c30169d0c0a3d0227e985104a56/pandas/io/json/_json.py#L55) makes the altair_data_server that uses this method drop numbers between 9.999e-16 and 5e-11, leading to unexpected and imo bad behavior like in https://github.com/altair-viz/altair_data_server/issues/45.


jakevdp:
Looks good, thanks!

PR #2401, opened 2021-02-09
Correct boxplot example to reflect title

joelostblom:
If you prefer I can change the title instead, but currently it does not match with the example code. I think it is more useful to see the non-default behavior and expanded a bit on the note regarding the default, but up to you.


jakevdp:
Looks good, thanks!

PR #2404, opened 2021-02-10
Use colors specified in line_scale

morberg:
This change will use the colors specified in `line_scale`. Previously `line_scale` was never used.


jakevdp:
Looks good, thanks!

PR #2417, opened 2021-02-23
Add issue templates

joelostblom:
Here is a suggestion for what the issue templates could look like.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -1,0 +1,16 @@
+---
+name: 'Bug report'
+about: 'Create a bug report to help us improve'
+labels: 'bug'
+---
+
+Type here...
+
+---
+
+Please follow these steps to make it more efficient to solve your issue:
+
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,16 @@
...
+about: 'Create a bug report to help us improve'
+labels: 'bug'
+---
+
+Type here...
+
+---
+
+Please follow these steps to make it more efficient to solve your issue:
+
```
jakevdp, line 12: We should add something here about including complete code to reproduce the problem

jakevdp (approved):
This is great, thanks!

.github/ISSUE_TEMPLATE/bug-report.md
joelostblom: I do have it in the last check box below. Do you want me to put it on top instead?

jakevdp: Oh, I see. That should be fine!

Issue #2411, opened 2021-02-20
Layered Histogram fails to represent null values


harabat:
## Issue

The method for making layered histograms [suggested on altair-viz.github.io](https://altair-viz.github.io/gallery/layered_histogram.html) seems to fail to take into account null values within the range of the data: bins that should be empty are represented as having 1 observation. 

---

## Examples

Altair:
```python
import pandas as pd
import altair as alt
import numpy as np

np.random.seed(42)

# Generating Data
source = pd.DataFrame({
    'Trial A': np.random.normal(0, 0.8, 1000),
    'Trial B': np.random.normal(-2, 1, 1000),
    'Trial C': np.random.normal(3, 2, 1000)
})

alt.Chart(source).transform_fold(
    ['Trial A', 'Trial B', 'Trial C'],
    as_=['Experiment', 'Measurement']
).mark_area(
    opacity=0.3,
    interpolate='step'
).encode(
    alt.X('Measurement:Q', bin=alt.Bin(maxbins=100)),
    alt.Y('count()', stack=None),
    alt.Color('Experiment:N')
)
```
![layered_histogram_altair](https://user-images.githubusercontent.com/45575009/108603170-45eb2f00-739e-11eb-88fe-0b1468d33a1c.png)
---

Seaborn:
```python
import seaborn as sns

source_melt = source.melt(var_name='Experiment', value_name='Measurement')
sns.displot(
    source_melt,
    x='Measurement',
    binwidth=.2,
    hue='Experiment',
    aspect=2,
    element='step'
)
```
![layered_histogram_seaborn](https://user-images.githubusercontent.com/45575009/108603174-4b487980-739e-11eb-9db2-c52d958f6297.png)

---

## Solutions

The easiest solution would be to suggest another method. The one below solves the null values in bins, but is much more verbose, and requires manually setting the x-axis and the legend. 

```python
trial_a = alt.Chart(source).mark_bar(
    opacity=0.5, color='#c9d6e5'
).encode(
    alt.X('Trial A:Q', bin=alt.Bin(maxbins=50), axis=None),
    y='count()'
)

trial_b = alt.Chart(source).mark_bar(
    opacity=0.5, color='#fcdab9'
).encode(
    alt.X('Trial B:Q', bin=alt.Bin(maxbins=50), axis=None),
    y='count()'
)

trial_c = alt.Chart(source).mark_bar(
    opacity=0.5, color='#f7cccc'
).encode(
[...]

PR #2419, opened 2021-02-24
fix histograms

harabat:
Implements solution discussed in issue #2411 (based on https://github.com/vega/vega-lite/issues/5403) for the examples [Layered Histogram](https://altair-viz.github.io/gallery/layered_histogram.html) and [Facetted Scatterplot with marginal histograms](https://altair-viz.github.io/gallery/scatter_marginal_hist.html).

### DIFF
--- a/altair/examples/scatter_marginal_hist.py
+++ b/altair/examples/scatter_marginal_hist.py
@@ -16,7 +16,7 @@ base = alt.Chart(source)
 xscale = alt.Scale(domain=(4.0, 8.0))
 yscale = alt.Scale(domain=(1.9, 4.55))
 
-area_args = {'opacity': .3, 'interpolate': 'step'}
+area_args = {'opacity': .3, 'binSpacing': 0}
 
 points = base.mark_circle().encode(
     alt.X('sepalLength', scale=xscale),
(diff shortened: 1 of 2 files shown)


joelostblom:
I built locally and it looks good to me, thanks! You can see that the binned histograms extends a bit longer than the area one, which I believe is the most correct approach so that the last bin is not half width as previously:

image

image

joelostblom (approved):

harabat:
Hadn't noticed that half-bin, good catch.

altair/examples/scatter_marginal_hist.py
```diff
@@ -16,15 +16,15 @@
 xscale = alt.Scale(domain=(4.0, 8.0))
 yscale = alt.Scale(domain=(1.9, 4.55))
 
-area_args = {'opacity': .3, 'interpolate': 'step'}
+area_args = {'opacity': .3, 'binSpacing': 0}
```
jakevdp, line 19: nitpick: change this to `bar_args`?

harabat: Good catch, I'll change this.

harabat: @jakevdp done

PR #2445, opened 2021-04-04
Add Hexbin Example to docs

FranzFlueckiger:
I like to use hexbin charts, and thought it would be nice to include a example to altair's documentation. Please feel free to review and eventually correct or extend my contribution, if you think it is worth merging with your master branch.

As always, thank you for your great work!

### DIFF
--- a/altair/examples/hexbins.py
+++ b/altair/examples/hexbins.py
@@ -1,0 +1,43 @@
+"""
+Hexbin Chart
+-----------------
+This example shows a hexbin chart.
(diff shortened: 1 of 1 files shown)


jakevdp:
There's no way to create hexagonally-binned charts in Altair, because it's not supported by Vega-Lite. If you're interested, you could request this kind of chart in the Vega project.

jakevdp:
This is a cool example - thanks!

PR #2466, opened 2021-05-16
fix: remove webdriver default argument to save

Midnighter:
This PR addresses https://github.com/altair-viz/altair_saver/issues/72


mattijn:
Thanks for this PR @Midnighter! Since this PR there was merged another PR: https://github.com/altair-viz/altair/pull/2701 that made vl-convert the new default for saving Altair Charts to png/svg as it does not require external dependencies. Github docs here: https://github.com/altair-viz/altair/blob/master/doc/user_guide/saving_charts.rst#png-svg-and-pdf-format. Do you still think it is useful to have this PR included within Altair?

Midnighter:
I don't know to be honest. It's been a while since I looked at this and I haven't really followed altair development since then.

If the webdriver is still used at all, then I think it might be useful to apply this PR such that the webdriver can pick the available rendering engine rather than being forced to use chrome. Ultimately, it's up to you/maintainers to decide whether it's still useful. I won't be cross if you just close it.

mattijn:
I updated this branch to be in-sync with the main repo. Since the GitHub Actions all pass, tested against a chrome webdriver, I am OK in merging this. Thanks @Midnighter!

PR #2480, opened 2021-07-22
DOC: add `mark_trail` to user guide's list of marks

ehudkr:
I hope I'm not missing anything obvious, but it seems trail mark is missing from the [list of marks in the user guide](https://altair-viz.github.io/user_guide/marks.html).


jakevdp:
Thanks!

PR #2488, opened 2021-08-18
Bump chart example

ChristopherDavisUCI:
I made a bump chart for the example gallery.  It's a mix of this Vega example [Bump Chart](https://vega.github.io/vega-lite/examples/line_bump.html), which I don't think has a direct analogue in the Altair gallery, and this example from the Altair gallery [Window Rank Line Chart](https://altair-viz.github.io/gallery/window_rank.html).

This is my first contribution (to any repository) so I'm very happy for any feedback.

bump

### DIFF
--- a/altair/examples/bump_chart.py
+++ b/altair/examples/bump_chart.py
@@ -1,0 +1,29 @@
+"""
+Bump Chart
+----------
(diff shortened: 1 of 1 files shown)


jakevdp:
This is a very cool example – thanks for contributing it!

Issue #2425, opened 2021-03-02
Upgrade to Vega-Lite 5


mattijn:
How can we prepare Altair and its community to upgrade towards the new [release](https://github.com/vega/vega-lite/releases/tag/v5.0.0) of Vega-Lite 5?

It's probably wishful thinking that @jakevdp has the time available to do this completely by himself, but it's probably also wishful thinking that someone can do this completely without the help of Jake. 

The skills and knowledge required to change the Vega-Lite schema into workable Python code is quite high. While Altair's codebase is flexible by nature to accommodate little changes within Vega-Lite, bigger changes such as `datum` and the recently introduced `params`, require more changes in the codebase.

Maybe someone is just about to finish all the work required for VL5 or maybe we have to become creative`¹`.

I'm sure more people have thoughts on this!

Useful references:
Interesting comments regarding development in the API of Altair can be found in these issues/PRs: https://github.com/altair-viz/altair/issues/2148, https://github.com/altair-viz/altair/issues/1567, https://github.com/altair-viz/altair/pull/1629

And here is the maintenance guide for Altair internals for people that already like to give it a go: https://github.com/altair-viz/altair/blob/master/NOTES_FOR_MAINTAINERS.md

`¹` such as:
- asking Jake for a code-explanation tour in a virtual setting so people with an interest in the project have a guideline to succeed in core-work PRs
- set up an Altair sponsor page to establish a fund for paid upgrading

joelostblom:
Thanks for starting this discussion! I would love to see Altair being updated to work with the new exciting additions to Vega-Lite! I am not sure if I can contribute meaningfully in terms of the needed core changes, but I am more than happy to help out with other tasks that would free up the time for those working on this.

In terms of sponsorship, maybe Altair and/or Vega-Lite would [qualify to become a Numfocus project]( https://numfocus.org/projects-overview)? They support Python and JS projects among others and both Matplotlib and Bokeh receive their sponsorship.

jakevdp:
Thanks! As far as Altair internals go... there's unfortunately not much of a process beyond what's documented in https://github.com/altair-viz/altair/blob/master/NOTES_FOR_MAINTAINERS.md

When it the update script leads to errors, the process is essentially:

- play around until you fix the code generation errors
- run the unit tests to see what backward-incompatible changes are implied by the updated Vega-Lite spec
- do your very best to accommodate those without breaking the existing Python API

[...]

PR #2513, opened 2021-11-01
Update to Vega-Lite 4.17.0

ChristopherDavisUCI:
Hi again,

@mattijn and I have tried to update the [Pull Request from last week](https://github.com/altair-viz/altair/pull/2509), taking into account your suggestions and trying to get all the tests (that I know about) to work.

Here are some of the main changes from the current Altair release:

* Changed the Vega-Lite schema version to 4.17.0.
* Added definitions of `DatumSchemaGenerator` and `DatumChannelMixin` in `generate_schema_wrapper.py`.
* Updated the loop in `generate_vegalite_channel_wrappers` [here](https://github.com/ChristopherDavisUCI/altair/blob/771e0362b56da2de40e88fa868b49f4635431b2b/tools/generate_schema_wrapper.py#L471) to allow for the possibility of a datum.
* Updated `infer_encoding_types` from `altair/utils/core.py` [here](https://github.com/ChristopherDavisUCI/altair/blob/771e0362b56da2de40e88fa868b49f4635431b2b/altair/utils/core.py#L668) to recognize Datum class names.
* Changed `get_valid_identifier` in `tools/schemapi/utils.py` to deal with some symbols like `[]` appearing in certain names in the Vega-Lite schema.  (Removing those symbols led to some duplicated class names.)
* Updated `TopLevelMixin` from `altair/vegalite/v4/api.py` to allow for `layer` to be repeated, in addition to `row` and `column`.
* Updated the *Encoding Channel Options* part of the docs [here](https://github.com/ChristopherDavisUCI/altair/commit/0a15c5d454efebf386ef4f3038dc83bb8edd95ee).  (I wasn't sure if there was a way to auto-generate these groupings of for example `Row` with `Column` with `Facet`.)
* Added some examples to the example gallery [here](https://github.com/ChristopherDavisUCI/altair/commit/cec423ecd3dffbf1e122c7aac12c5f8bcf0df023) and some additional documentation about `mark_arc` and layering in repeat [here](https://github.com/ChristopherDavisUCI/altair/commit/9a365d41b7d7073679decdb0674ae248572615df).
* Applied a temporary fix and linked Vega-Lite issue to get `test_vegalite_to_vega_mimebundle` to work [here](https://github.com/ChristopherDavisUCI/altair/commit/6fafc330a77f4337e0db737b7b58e7fbea80cd43).

Main outstanding **issue** that we know of:
* Starting with `v4.9`, the format of `TopLevelRepeatSpec` changed in the Vega-Lite schema.  We have some [code to deal with this](https://github.com/ChristopherDavisUCI/altair/blob/771e0362b56da2de40e88fa868b49f4635431b2b/tools/generate_schema_wrapper.py#L609) by editing the definition of `TopLevelRepeatSpec` in the downloaded JSON file.  We've tried asking on the Vega-Lite Slack channel and as a Vega-Lite GitHub [issue](https://github.com/vega/vega-lite/issues/7775), but it seems like the issue is on the Altair side, not the Vega-Lite side.  If the rest of the changes seem mostly in good shape, I can focus on finding a more adequate solution.

This is not part of the current Pull Request, but to get some of the  tests to work, we made the following changes to altair_viewer:
* Save https://unpkg.com/vega-lite@4.17.0/build/vega-lite.min.js as `vega-lite-4.17.0.js` in `altair_viewer\altair_viewer\scripts`
* Add "4.17.0" in `listing.json` in `altair_viewer\scripts`

Thank you for any feedback!


jakevdp:
Regarding the testing: the way I've done it in the past is to add a branch to altair viewer with the necessary changes, and temporarily adjust the CI in this branch to point to that. Then when we're happy with everything, we cut a new altair-viewer release and then change this CI back to normal.

mattijn:
From here: https://github.com/altair-viz/altair/pull/1990/commits/3fc9a8610667d0cded4712fd9c748f78278ebd2d I see you update master of altair_viewer to include support for VL4.8.1. Once https://github.com/altair-viz/altair_viewer/pull/43 is merged we can follow same principle.

ChristopherDavisUCI:
Thank you for the comments!

Does the new definition of [RepeatChart](https://github.com/ChristopherDavisUCI/altair/blob/b3e30dae8d2960fdb1ff0042a82c4a13a0465a98/altair/vegalite/v4/api.py#L2101) look like what you had in mind?

My change has broken one test.  Currently the [test_chart_from_dict](https://github.com/ChristopherDavisUCI/altair/blob/b3e30dae8d2960fdb1ff0042a82c4a13a0465a98/altair/vegalite/v4/tests/test_api.py#L685) test from altair/vegalite/v4/tests/test_api.py is failing on `base.repeat(["c", "d"])`.  I am having a hard time understanding how `from_dict` works.

If this looks like the approach you had in mind, I will try to get this `test_chart_from_dict` working.

jakevdp:
So I think the new definition of RepeatChart, inheriting from `VegaLiteSchema` and setting `_schema = {'$ref': '#/definitions/TopLevelRepeatSpec'}`, is essentially equivalent to the old definition, inheriting from `TopLevelRepeatSpec`. All that inheritence does is to set the `_schema` to the associated definition.

If the behavior changed, it's probably due to removing the decorator, which adds some input validation to `__init__`. Maybe just removing the decorator is easier?

ChristopherDavisUCI:
With the previous method I never could get `test_chart_from_dict` to pass.  The problem seemed to be that _FromDict.from_dict was not able to recognize when a dictionary satisfied the RepeatChart schema.  My theory is that because TopLevelRepeatSpec is a union of two schemas, but those two schemas are never named, that they are never included in the list of possible options by the _FromDict hash function.

The [most recent commit](https://github.com/altair-viz/altair/pull/2513/commits/df162dea409a5d11ada4acbbab02f104cc1360e5) gets around this by only providing one merged schema for RepeatChart.

It felt strange to paste that long dictionary directly into the code.  Would it be better to load and edit from the json file?  It felt a little inefficient to load the entire json file since only a small piece of it was needed.

mattijn:
Yes I realized the same, I thought maybe setting one of the two schemas here for `repeat`: https://github.com/altair-viz/altair/blob/33e0c90df4c58f8e169183e888442b62ae962529/altair/vegalite/v4/api.py#L562, but to no avail.

jakevdp:
> `_FromDict.from_dict` was not able to recognize when a dictionary satisfied the RepeatChart schema.

Ah, yeah that makes sense. A better approach might be to do pass `cls` as an argument here: https://github.com/altair-viz/altair/blob/159831869f84298af19b6d231c35204cf1c7b40d/tools/schemapi/schemapi.py#L554-L558

Then the top level class will be used as a wrapper even in the case of an `anyOf` schema.

mattijn:
Hm, it does enter the `anyOf` schema already. I notice that the `spec` attribute is interpreted as `FacetUnitSpec`, but should this not be `RepeatSpec`?
Around here: 
https://github.com/altair-viz/altair/blob/159831869f84298af19b6d231c35204cf1c7b40d/tools/schemapi/schemapi.py#L550

jakevdp:
Maybe I wasn't clear: I mean that you should remove the schema patch, in which case the issue I think is that the `_ToDict` method finds the `RepeatChart` schema and class, but then since it has an `anyOf`, it recursively calls `self.to_dict()`, and the `cls` is not passed along. If we pass that along there, the schema patch should not be necessary.

mattijn:
I removed the patch, but the failing spec
```
base = alt.Chart("data.csv").mark_point().encode(x="x:Q", y="y:Q")
base.repeat(["c", "d"])
```
resolves into this:
```python
{'config': Config({
  view: ViewConfig({
    continuousHeight: 300,
    continuousWidth: 400
  })
}), 'repeat': ['c', 'd'], 'spec': FacetedUnitSpec({
  data: UrlData({
    url: 'data.csv'
  }),
  encoding: FacetedEncoding({
    x: PositionFieldDef({
      field: FieldName('x'),
      type: StandardType('quantitative')
    }),
    y: PositionFieldDef({
      field: FieldName('y'),
      type: StandardType('quantitative')
    })
  }),
  mark: Mark('point')
}), '$schema': 'https://vega.github.io/schema/vega-lite/v4.17.0.json'}
```
And I think I should expect `RepeatSpec`  instead of `FacetedUnitSpec`

jakevdp:
Nice – I think making that resolve correctly will require some changes to how `_ToDict` works.

jakevdp:
At a high level: basically you want `cls` to be passed through on `anyOf` in some cases, but not in others. Which cases? If the specifications within `anyOf` have their own classes associated, you want to use those more specialized classes. If not, you want to use the higher-level class.

So maybe what we need is some kind of `default_cls` argument to `_ToDict.to_dict()` that controls what `cls` defaults to if there are no matches for the particular schema.

mattijn:
But what if a `dct` validates with a `possible_schema` but it is not the correct `possible_schema`?
From my limited understanding this is happening.

jakevdp:
Yeah, I see now. I'm surprised we haven't run into that before.

jakevdp:
Wait, but if it is a `RepeatSpec`, it is required to have a `repeat` property, so I don't think it's correct that `FacetedUnitSpec` should be `RepeatSpec` in that output.

mattijn:
There is one, `'repeat': ['c', 'd']`, but higher level and not within `spec`, which gets validated

jakevdp:
The issue in that output spec is that at the top level, it is a raw dictionary, meaning `cls` was set to `_passthrough`. I believe my recommendations from above are what's needed to fix the issue.

mattijn:
Great, let me see if I or Chris can get a step further with the recommendations. Its complicated matter.

ChristopherDavisUCI:
When I run `black .` it seems to behave differently from how the GitHub `black` behaves.  For example, it changes 
```
.encode(
            x="x:Q",
            y="y:Q",
        )
```
to
```
.encode(x="x:Q", y="y:Q",)
```
I'm happy to hear what I should be doing differently...

ChristopherDavisUCI:
Oops it looks like I was editing _FromDict in the wrong file (in an auto-generated file).  I will correct that.

ChristopherDavisUCI:
I tried to implement your suggestion @jakevdp 
https://github.com/ChristopherDavisUCI/altair/blob/a23a8ff3cf0575bfead109403e5076e7333bade3/tools/schemapi/schemapi.py#L539-L552
That is where `default_class` gets set, and it only ever gets passed to `from_dict` from within an `anyOf` or `oneOf` block.  Does that look like what you had in mind?  It worked in the cases I tested, but I don't have a full understanding of all the different possible scenarios.

jakevdp:
Yeah, that seems like the right logic - very nice!

mattijn:
From [VL4.14](https://github.com/vega/vega-lite/releases/tag/v4.14.0) onwards Vega-Lite can infer types within a Data object, although I think its good practice, what will Altair do [here](https://github.com/ChristopherDavisUCI/altair/blob/sample417/altair/vegalite/v4/schema/channels.py#L31:L47)?

jakevdp:
That's a good point... I think the logic vega-lite uses for type inference is slightly different than the logic that Altair uses, so for now I'm leaning toward leaving Altair's logic in place, to avoid breaking charts in cases where the logic produces different types. What do you think?

altair/examples/line_chart_with_datum.py
```diff
@@ -0,0 +1,31 @@
...
+    y = alt.Y("price"),
+    color = "symbol"
+)
+
+xrule = alt.Chart(no_source).mark_rule(color='cyan', strokeWidth=2).encode(
+    x=alt.XDatum(alt.DateTime(year=2006, month = "November"))
+)
+
+yrule = alt.Chart(no_source).mark_rule(strokeDash=[12,6], size=2).encode(
+    y = alt.YDatum(350)
```
jakevdp, line 27: Altair currently has an `alt.value(x)` that can be used in place of the `alt.YValue()` classes. I wonder if it would make sense to have `alt.datum(x)` for use in place of `alt.YDatum(x)`?

It's maybe a bit strange, because `alt.datum` already exists for a different purpose, but we could do this by adding a `__call__` method to [`DatumType`](https://github.com/altair-viz/altair/blob/159831869f84298af19b6d231c35204cf1c7b40d/altair/expr/core.py#L4) with an implementation similar to this: https://github.com/altair-viz/altair/blob/8a8642b2e7eeee3b914850a8f7aacd53335302d9/altair/vegalite/v4/api.py#L209-L211

What do you think?

tools/schemapi/schemapi.py
```diff
@@ -539,7 +541,15 @@ def _passthrough(*args, **kwds):
             # Our class dict is constructed breadth-first from top to bottom,
             # so the first class that matches is the most general match.
             matches = self.class_dict[self.hash_schema(schema)]
-            cls = matches[0] if matches else _passthrough
+            if matches:
+                cls = matches[0]
+            elif default_class:
```
jakevdp, line 546: maybe
```
else:
  cls = default_class or _passthrough
```

tools/schemapi/schemapi.py
```diff
@@ -555,6 +565,7 @@ def _passthrough(*args, **kwds):
                         dct,
                         schema=possible_schema,
                         rootschema=rootschema,
+                        default_class=default_class,
```
jakevdp, line 568: To localize the logic a bit, I'd remove lines 550-552 and write `default_class=cls or default_class` here.

jakevdp (commented):
This is looking really great! Just a few minor comments...

The last overall thing would be to populate the changelog to outline the changes in API. It looks like things are for the most part backward compatible (given the lack of modification in examples), but highlighting the arc marck and datum encoding would be useful.

ChristopherDavisUCI:
Thanks a lot for looking through it so quickly!

I'm very inexperienced in this process.  Are these above changes things that @mattijn or I should make in the code, or are you able to make them directly?  I have plenty of time later this afternoon to make edits.

altair/examples/line_chart_with_datum.py
ChristopherDavisUCI: I find myself using `alt.value` very often, like for example if I just want to set an explicit color (not sure if that's good style or not), so having an `alt.datum` seems nice.  If the change breaks a lot of existing code then I guess it's not worth it here.  (I admit I don't really see the big picture of why both Datum and Value exist on the Vega-Lite side.  I think I haven't digested the essential differences between the concepts.)

tools/schemapi/schemapi.py
ChristopherDavisUCI: My thinking was that we might want to check if `default_class` had been set or not, and I wasn't sure if something like `default_class == _passthrough` was safe.  But maybe we never need to check that anyway.

But if we can remove some of the cases then that sounds good to me.

jakevdp:
I'm fine either way - if you'd like to make the changes, please do so! Otherwise I'm happy to give it a try maybe this evening or towmorrow.

mattijn:
No rush. I'm also not against pausing a bit and see what we actually have. Lots of progress in one week!

ChristopherDavisUCI:
I'll try to make some changes later this afternoon or evening and let you know how far I get.

altair/examples/line_chart_with_datum.py
ChristopherDavisUCI: Now that I understand this option (and that it doesn't break the existing `alt.datum`) I really like it.  Do you think any/all of the three datum examples should be changed to use `alt.datum` instead of the longer syntax (like `alt.Theta2Datum`)?

tools/schemapi/schemapi.py
ChristopherDavisUCI: I implemented about half of these changes, make `default_class` default to `_passthrough`.  (To get that to work I had to move the definition of `_passthrough` to the outside.)  From my understanding of the logic, we aren't able to use `default_class=cls or default_class`, because `cls` might get changed earlier, and I think we only want to use `cls` here if it arrived as an input argument.

ChristopherDavisUCI:
I tried making most of those changes discussed above.  Still to do maybe:
* Should I change any/all of the new examples to use the shorter `alt.datum` syntax as opposed to the `alt.ColorDatum` syntax?
* See if the `_FromDict.from_dict` code can be further simplified (I don't see how to simplify it further). https://github.com/altair-viz/altair/blob/18d0c1238556f87983ef82ae4dc62e3b5e0d67e0/tools/schemapi/schemapi.py#L526-L528

mattijn:
I added some more docs+examples, including a special section for circular plots in the example gallery.

Since ```default_class=_passthrough``` is defined as function parameter
https://github.com/altair-viz/altair/blob/18d0c1238556f87983ef82ae4dc62e3b5e0d67e0/tools/schemapi/schemapi.py#L526-L528
Then line `549` can be removed or not?
https://github.com/altair-viz/altair/blob/18d0c1238556f87983ef82ae4dc62e3b5e0d67e0/tools/schemapi/schemapi.py#L545-L551

ChristopherDavisUCI:
> Then line `549` can be removed or not?
> https://github.com/altair-viz/altair/blob/18d0c1238556f87983ef82ae4dc62e3b5e0d67e0/tools/schemapi/schemapi.py#L545-L551

My understanding is that line `549` still needs to be there, because if `default_class` shows up as some value other than `_passthrough`, we don't want that value to persist to the next level of the recursion.

jakevdp:
I think line 549 should be removed: with it there, a class with a schema containing a nested `anyOf` may not be correctly instantiated.

jakevdp:
Regarding the default data... I think we need to be careful about this. For example, you can currently write something like this:
```python
import altair as alt

alt.layer(
  alt.Chart().mark_point(),
  alt.Chart().mark_line(),
).encode(
    x='Horsepower:Q',
    y='Miles_per_Gallon:Q'
).properties(
    data=data.cars()
)
```
If `alt.Chart()` filled-in an empty dataset by default, this would no longer work.

Fundamentally this is a vega-lite issue: the schema for `TopLevelUnitSpec` should not list `data` as required, because it's not truly required.

But perhaps we can figure out a way to inject data conditionally to make this `datum` example more convenient?

jakevdp:
Also, after the logic to define `cls`, shouldn't `default_class = cls` unconditionally?

jakevdp:
Actually, I believe lines 549-551 can be removed: https://github.com/altair-viz/altair/blob/18d0c1238556f87983ef82ae4dc62e3b5e0d67e0/tools/schemapi/schemapi.py#L549-L551

And then line 567 can be changed to `default_class=cls` https://github.com/altair-viz/altair/blob/18d0c1238556f87983ef82ae4dc62e3b5e0d67e0/tools/schemapi/schemapi.py#L563-L568

Why? The purpose of the `default_class` argument is to pass down the `cls` from higher levels, so that if a new `cls` isn't identified, the previously determined one is used instead of `_passthrough`.

ChristopherDavisUCI:
Thanks for these comments!

I think I should leave this `default_class` part to one of you.  I think I understand the logic needed to fix `TopLevelRepeatSpec`, but I don't think I understand the overall logic of how this should behave.

mattijn:
On injecting data conditionally for `datum`; I tried a bit, but honestly I've no real idea where to start poking. Furthermore its more or less fine by me like this.

tools/schemapi/schemapi.py
```diff
@@ -564,7 +561,7 @@ def from_dict(
                         dct,
                         schema=possible_schema,
                         rootschema=rootschema,
-                        default_class=default_class,
+                        default_class=cls,
                     )
 
         if isinstance(dct, dict):
```
jakevdp, line 567: This change needs to be applied to the copy in the package as well in order for the tests to cover it.

jakevdp:
> On injecting data conditionally for datum; I tried a bit, but honestly I've no real idea where to start poking. Furthermore its more or less fine by me like this.

I think the way to do it would be to use the `context` argument of `to_dict` to pass on information about the parent chart, and if there's no parent then insert dummy data as appropriate. That said, it's orthogonal to the work in this PR so I think we can leave that for later.

tools/schemapi/schemapi.py
ChristopherDavisUCI: I copied it over by running the `generate_schema_wrapper.py` script.

ChristopherDavisUCI:
Hmm, this isn't an error I've seen before; is it obvious what it means?  Trying it on my computer I got a similar error, and it seemed to point to line 278 here: https://github.com/ChristopherDavisUCI/altair/blob/25baa86dbd921a30761bcc5300b0c06397eae99d/altair/sphinxext/altairplot.py#L275-L282

jakevdp:
Ah, yeah I think the issue there is that our default class wrapper thing is too aggressive now. `alt.InlineDataSet` has a type signature that looks something like `anyOf(List(float), List(string), List(boolean), string)`, which means that any object that in `from_dict()`, any object that matches these types will be wrapped in an `InlineDataSet` class.

We probably don't want that.

So we'll need a way to distinguish between `anyOf` schemas that we do want to wrap with the parent type vs. ones that we don't.

jakevdp:
Another piece of context here: it's very tempting to introduce specific patches or checks to fix this kind of bug as it comes up... the problem with that is that as Vega-Lite evolves, those kinds of one-off changes become very difficult to manage and maintain. It's a mistake I made early on in writing Altair. That's why, for each of these issues, my inclination is to try to solve the most general version of the problem that's surfacing, because then maybe future upgrades will be just a bit easier.

ChristopherDavisUCI:
I experimented a little with some of the examples and most of them still seem to work correctly.  Here is the smallest example I could find that does not work:

```
d = {'data': {'values': [{}]},
       'mark': {'type': 'point'}}

alt.Chart.from_dict(d).to_json()
```
Here is the result of `alt.Chart.from_dict(d).to_dict()`:
```
{'config': {'view': {'continuousWidth': 400, 'continuousHeight': 300}},
 'data': {'name': 'data-a21cafb4c405e6997671a02e578b9b1e'}, 
 'mark': {'type': 'point'}, 
 '$schema': 'https://vega.github.io/schema/vega-lite/v4.17.0.json', 
 'datasets': {'data-a21cafb4c405e6997671a02e578b9b1e': InlineDataset([{}])}}
```

jakevdp:
Ah, that’s helpful. My guess is there’s a to_dict call missing somewhere in the dataset handler.

mattijn:
I'm not sure if my commit resolves the effect or the cause. But before, this:
```
'data': {'values': [{}]}
```
resolved into `alt.InlineData({[]})`.
Where it now resolves into `alt.InlineData(alt.InlineDataset({[]})`.

So I apply `alt.InlineDataset({[]}).to_dict()` within the data consolidation function.

altair/vegalite/v4/api.py
```diff
@@ -47,7 +47,10 @@ def _consolidate_data(data, context):
 
     if isinstance(data, core.InlineData):
         if data.name is Undefined and data.values is not Undefined:
-            values = data.values
+            if isinstance(data.values, core.InlineDataset):
+                values = data.values.to_dict()
+            else:
+                values = data.values
```
jakevdp, line 53: Nice find! This seems like the right fix. Maybe make it a bit more robust and do
```
values = data.to_dict()['values']
```

mattijn: Done

jakevdp:
I think this is looking pretty good! What do you think? Anything missing at this point?

Before merging, I'll create a new altair_viewer release and we can change the CI requirements back to normal.

ChristopherDavisUCI:
> I think this is looking pretty good! What do you think? Anything missing at this point?

Nothing missing that I'm aware of!

mattijn:
Merge-ready! +1

joelostblom:
This is awesome! Thank you all so much for working on this PR, the notifications from this thread have been my favorite morning reading over the last few days :heart:

jakevdp:
In https://github.com/altair-viz/altair_viewer/pull/45 I added the most recent vega & vega-embed versions to altair viewer. I triggered a re-run of the CI here to make sure there are no issues.

jakevdp:
Alright, altair-viewer 0.4.0 is released! We can remove the temporary reference to its github branch from this PR

mattijn:
I checked if the tests use `altair_viewer` version 0.4.0, and this seems to be the the case. All good.

jakevdp:
Thanks all - I think we should merge this!

Given the number of commits, I think I'll opt to squash and merge – I believe that will maintain correct authorship (@ChristopherDavisUCI will be the author, @mattijn will be listed in "Co-authored By" and I would be listed as the committer). Is everyone happy with that?

ChristopherDavisUCI:
Great!  That definitely sounds good to me.  (It will be especially nice to get rid of some of those commits going back and forth between different fixes for TopLevelRepeatSpec.)

mattijn:
Perfect

jakevdp:
Alright, it's done! Thanks all, this is really great work 🎉

jakevdp:
Regarding a release... I'd like to iron-out the "default data" issue discussed above. I also did a bit of maintenance stuff this morning (updating CI jobs, etc.) but yes, it would be great if we could get a release out sometime this week.

Regarding VL5 - that would be great if you want to open a draft PR! The biggest issue I anticipate there is that VL5 replaced "selections" with a more general "params" specification, so I suspect a number of things will have to change to accommodate that.

ChristopherDavisUCI:
I've experimented a little with Vega-Lite v5 on [this branch](https://github.com/ChristopherDavisUCI/altair/tree/sample51), although I might want to redo some things now that I better understand what's going on.

```
no_data = pd.DataFrame()

    width=100
)

c+c
```

mattijn:
Hi Chris, if you move your last comment to here #2425, or start a new pull request titled 'WIP: update to Vega-Lite 5' and add your observations there, than I'm more than happy to join the process (and maybe others as well!). It would be nice if we can get further with fewer help from Jake.

Its indeed super awesome to align a few brains and go forward, but be careful, it's addicting and if its your spare-time than other things also probably like attention;)

But if we can give ourselves a Christmas gift with a new Altair release, either based on VL4.17 or 5, then I think we are doing great.

ChristopherDavisUCI:
Thanks @mattijn and yes, "addicting" is a good word for it!  Thanks for the suggestion; I made a new draft PR #2516

PR #2515, opened 2021-11-09
Inject empty data at top level for easier datum encodings

jakevdp:
.cc @mattijn 

I think this is roughly what you had in mind in the previous discussion (https://github.com/altair-viz/altair/pull/2513#issuecomment-960643083). We may want a more sophisticated version of this rather than *always* adding empty data at the top level.

### DIFF
--- a/altair/examples/line_chart_with_datum.py
+++ b/altair/examples/line_chart_with_datum.py
@@ -11,22 +11,21 @@ import altair as alt
 from vega_datasets import data
 
 source = data.stocks()
-no_source = alt.Data(values=[{}])
 
 lines = (
     alt.Chart(source)
     .mark_line()
-    .encode(x=alt.X("date"), y=alt.Y("price"), color="symbol")
+    .encode(x="date", y="price", color="symbol")
 )
 
 xrule = (
-    alt.Chart(no_source)
+    alt.Chart()
     .mark_rule(color="cyan", strokeWidth=2)
     .encode(x=alt.datum(alt.DateTime(year=2006, month="November")))
 )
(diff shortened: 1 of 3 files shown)


ChristopherDavisUCI:
I saw in the Vega-Lite schema that "type": "null" is an allowable option for the `data` property.  Is there a standard way to use that in Altair?  For me writing something like `alt.Chart(None)` is just as nice (or maybe even nicer since it's more explicit) as writing `alt.Chart()` if we want to specify empty data.

jakevdp:
Interesting - where do you see that `{"type": null}` is a valid option for data? When I look at the schema, I only see URLData (which requires `"url"`), NamedData (which requires `"name"`), and InlineData (which requires `"values"`)

ChristopherDavisUCI:
Maybe you're one level deeper in the schema than I am?  I see `"type": "null"` under:
`TopLevelUnitSpec->properties->data`.

jakevdp:
Ah, yeah I see. I think the meaning of `data=None` is different than what we want here, though:
```
"description": "An object describing the data source. Set to `null` to ignore the parent's data source. If no data is set, it is derived from the parent."
```

ChristopherDavisUCI:
I see!  I often find those schema descriptions difficult to parse, so I've often gotten in the habit of not even looking at them.

If there's anything you'd like me to try related to this empty encoding, just let me know (I couldn't tell if there was more to be done or not).  But for now I'll go back to thinking about version 5.

mattijn:
I played a bit with this; I think it works good.

PR #2516, opened 2021-11-09
WIP: update to Vega-Lite 5

ChristopherDavisUCI:
(Draft version only!)

```
no_data = pd.DataFrame()

    width=100
)

c+c
```

Other things:
* This is silly but I don't understand why it lists so many commits.  Most of those old commits were already merged.  Is there an easy way to fix that?  If I were left to my own, I would just create a new branch and move over the changed files (there are only a few I made changes in).
* I usually try some tests by running things in Jupyter notebook, but since making the change to Vega-Lite 5 that hasn't worked for me.  Instead I get the following message in red the first time I try to display a chart, and then subsequent times I just get a blank response: `Error loading script: Script error for "vega-util", needed by: vega-lite http://requirejs.org/docs/errors.html#scripterror` It does work in Jupyter Lab.
* Some of the code changes have been experiments trying to learn how the old `selection` fits with the new `parameter`.  It might be best to redo this code later now that I see more of the big picture.


mattijn:
> This is silly but I don't understand why it lists so many commits. Most of those old commits were already merged. Is there an easy way to fix that? If I were left to my own, I would just create a new branch and move over the changed files (there are only a few I made changes in).
I think the proper method is to fetch master first, rebase your branch and then do a pull request, but I think I would do what you wrote.

ChristopherDavisUCI:
Thank you @mattijn I won't tell you which of the methods I ended up using :)
I made a new pull request #2517 that now only shows two commits, so that's better.  I will close this one.

PR #2517, opened 2021-11-09
WIP: update to Vega-Lite 5

ChristopherDavisUCI:
(Draft version only!)

```
no_data = pd.DataFrame()

    width=100
)

c+c
```

Other things:
* I usually try some tests by running things in Jupyter notebook, but since making the change to Vega-Lite 5 that hasn't worked for me.  Instead I get the following message in red the first time I try to display a chart, and then subsequent times I just get a blank response: `Error loading script: Script error for "vega-util", needed by: vega-lite http://requirejs.org/docs/errors.html#scripterror` It does work in Jupyter Lab.
* Some of the code changes have been experiments trying to learn how the old `selection` fits with the new `parameter`.  It might be best to redo this code later now that I see more of the big picture.


mattijn:
`width` within a layered mark seems the culprit, observe:
```python
# python 3
from urllib.request import urlopen
import json
import jsonschema  # 3.2.0

def validate(vl_spec, vl_schema="v5.1.0"):
    schema = json.load(urlopen(f'https://vega.github.io/schema/vega-lite/{vl_schema}.json'))
    spec = json.loads(vl_spec)
    jsonschema.validate(spec, schema)
    
vl_spec = """
{
  "layer": [{"mark": "circle", "width": 100}, {"mark": "circle"}],
  "data": {"values": []}
}
"""
validate(vl_spec)
```

> ```
> ValidationError: Additional properties are not allowed ('mark' was unexpected)
> On instance:
>     {'mark': 'circle', 'width': 100}
> ```

The above validates still OK in v4.17.0 (`validate(vl_spec, vl_schema='v4.17.0')`. 

When placing the `width` outside the layered mark it validates without error in v5.1.0, like this:
```python
vl_spec = """
{
  "width": 100,
  "layer": [{"mark": "circle"}, {"mark": "circle"}],
  "data": {"values": []}
}
"""
validate(vl_spec)
```

jakevdp:
If we wanted, we could add logic to `alt.layer` that will move any width/height specification in child charts to the top level. I believe there is already some existing logic along these lines for other properties

ChristopherDavisUCI:
Thank you!  Do you have an example of a property where we do something similar to this?  (It doesn't have to be in LayerChart.)  If I have a template to use, I should be able to do the same thing for width and height.

So far at least with `selection` it's seemed straightforward to keep everything backwards compatible.

jakevdp:
What I had in mind was the `_combine_subchart_data` step in `alt.LayerChart` - basically this looks for common datasets in the layers, and moves them up to the top level. This so you can do things like this:
```python
base = alt.Chart(dataframe).encode(...)
chart = base.mark_point() + base.mark_line()
```
and only embed a single copy of the serialized dataframe in the output. We could do something similar for `width` and `height`: extract them from the underlying layers & apply them to the parent chart, raising an error if there are any disagreements.

joelostblom:
> I've read the list of "breaking changes" for the Vega-Lite 5.0.0 release and don't see anything that seems related to this, so it does make me wonder if maybe I misunderstand the cause of the problem.

I looked around a bit and I think this was [initially introduced in VL4.1.1](https://github.com/vega/vega-lite/pull/5797), but then [reverted and included with a deprecation warning because Jake reported that it broke Altair](https://github.com/vega/vega-lite/issues/5854). That [deprecated code was removed in 5.0 in this PR](https://github.com/vega/vega-lite/pull/7062), which is probably why it is breaking now. It seems like in addition to `width` and `height`, the `view` property is also affected so maybe we need to include that in this PR as well.

ChristopherDavisUCI:
Thank you @joelostblom for tracking that down.  Would you be able to give me an example where we use this `view` property in Altair?  That would help with my testing.  I tried searching through the documentation examples and didn't immediately find one.

jakevdp:
One relevant piece of info here is that the Vega-lite *renderer* maintains backward compatibility for some things (such as passing `width` and `height` to within a layer) even though the Vega-Lite schema now forbids them. Because we validate based on the schema, Altair is actually a bit stricter than the Vega-Lite renderer.

We have a couple options for dealing with that:

1. Accept that Altair is stricter than vega-lite, and pass on those errors to users (the "do nothing" approach)
2. maintain a list of allowable departures from the schema.
3. do some preprocessing at initialization for specific issues
4. stop validating based on the schema

I would lean toward (3) here, as (1) would lead to poorer usability, and my sense is that (2) would add an undue maintenance burden, and (4) would be a large departure from Altair's current implementation

joelostblom:
@ChristopherDavisUCI I am not sure how much it is actually used, but `alt.Chart` does accept a `view` parameter, so I am guessing we should handle that the same as `height` and `width` (since it was deprecated in the same VL PR) and move it up to the top level of layered charts:

```py
alt.Chart(alt.Data(values=[{}]), height=200, view=alt.ViewConfig()).mark_bar().to_dict()
```

```
{'config': {'view': {'continuousWidth': 400, 'continuousHeight': 300}},
 'data': {'values': [{}]},
 'mark': 'bar',
 'height': 200,
 'view': {},
 '$schema': 'https://vega.github.io/schema/vega-lite/v4.8.1.json'}
```

and using `view` instead of `width` in the validation example by @mattijn above gives the same `ValidationError`:

```
vl_spec = """
{
  "layer": [{"mark": "circle", "view": {}}, {"mark": "circle"}],
  "data": {"values": []}
}
"""
```

ChristopherDavisUCI:
@joelostblom Thank you!  

@jakevdp Thanks for the pointer to `_combine_subchart_data`.  I think I was able to adapt that to get the layer examples working.

Aside from rendering examples (literally zero of which are working), there's just one example that doesn't work that I know of, `Scatter Plot with Minimap`, because it explicitly specifies `"selection": zoom.name`.  If we change it to `"param": zoom.name` then it works.  Do you see a robust way to deal with that?  I looked through the schema and I think `"selection"` as a property name only exists in `"Config"`.

```
"""
Scatter Plot with Minimap
-------------------------
This example shows how to create a miniature version of a plot
such that creating a selection in the miniature version
adjusts the axis limits in another, more detailed view.
"""
# category: scatter plots

import altair as alt

source = data.seattle_weather()

zoom = alt.selection_interval(encodings=["x", "y"])

minimap = (
    alt.Chart(source)
    .mark_point()
    .add_selection(zoom)
    .encode(
        x="date:T",
        y="temp_max:Q",
        color=alt.condition(zoom, "weather", alt.value("lightgray")),
    )
    .properties(
        width=200,
        height=200,
        title="Minimap -- click and drag to zoom in the detail view",
    )
)

detail = (
    alt.Chart(source)
    .mark_point()
    .encode(
        x=alt.X(
            "date:T", scale=alt.Scale(domain={"selection": zoom.name, "encoding": "x"})
        ),
        y=alt.Y(
            "temp_max:Q",
            scale=alt.Scale(domain={"selection": zoom.name, "encoding": "y"}),
        ),
        color="weather",
    )
    .properties(width=600, height=400, title="Seattle weather -- detail view")
)

detail | minimap
```

ChristopherDavisUCI:
Here's a first example of using `VariableParameter`.  Is that basically how we want it to work?   Should we try to add functionality so `opacity=alt.ExprRef("sample/1000")` can be replaced by maybe something like `opacity=alt.expr("sample/1000")` or `opacity=alt.expr(f"{size_var}/1000")`?
```
import altair as alt

cars = data.cars.url

year_slider = alt.binding_range(min=0, max=1000, step=5)
size_var = alt.variable(bind=year_slider, value=200, name="sample")

c = alt.Chart(cars).mark_circle(size=size_var, opacity=alt.ExprRef("sample/1000")).encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q',
).add_variable(size_var)

c
```
sample

jakevdp:
That's nice! I think it would be cool if we could use the variables directly in the expression, i.e. something like `opacity = size_var / 1000` which would basically output that same `ExprRef`.

We do a similar thing with `alt.datum.value / 1000`

mattijn:
I played a bit in https://github.com/ChristopherDavisUCI/altair/commit/0b34ec62391b80ce6c4be6e0f657fae5da7a6078 and now this works:
```python
cars = data.cars.url

year_slider = alt.binding_range(min=0, max=1, step=0.05)
size_var = alt.variable(bind=year_slider, value=0.2, name="sample")

c = alt.Chart(cars).mark_circle(
    size=size_var*1000,
    opacity=size_var + 0.2
).encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q',
).add_variable(size_var)

c
```
but this won't, `opacity=size_var + 0.2 + 0.1`:
```TypeError: unsupported operand type(s) for +: 'ExprRef' and 'float'```

Because the `__add__` function includes the `ExprRef`:
```python
    def __add__(self, other):
        return ExprRef(expr.core.BinaryExpressionReference("+", self.name, other))
```
It get stuck in the _second round_ of addition, since `size_var + 0.2` has become an `ExprRef` and cannot add the `0.1` . Another solution is needed, but maybe you its a step forward for someone else.

mattijn:
I can completely imagine this should be done very different, but now this also works:
```python
import altair as alt
year_slider = alt.binding_range(min=0, max=1000, step=5)
size_var = alt.variable(bind=year_slider, value=200, name="sample")
expr = size_var ** 2 + 0.1
expr
```

> ```
> ExprRef({
>   expr: pow(sample,2) + 0.1
> })
> ```

Where `size_var ** 2 + 0.1` can be used eg. for the `opacity` channel in above example.

Changed code in this commit https://github.com/ChristopherDavisUCI/altair/commit/530a11a7fc48df891977a8ecabcdabe72ceb7951

ChristopherDavisUCI:
Thanks @mattijn!  I glanced and didn't immediately follow, but I will look more slowly later.

Is it obvious to you why the charts don't display for me in Jupyter notebook?  Do you think it's in the same category as the changes you made related to `vegalite_mimebundle` from last round?

jakevdp:
If you're using the jupyterlab/mimebundle renderer, that will not work, because the vega-lite extension does not yet support vl 4.17. Switch to the default renderer.

Even with the default renderer, you can still run into issues if charts from previous altair versions are displayed in the notebook. Try clearing all outputs, saving the notebook, then reloading the page.

If you're using jupyterlab, the same is true of *any notebook tab you have open in the session*, because notebooks are not sandboxed from each other. So close all notebooks, clear outputs from ones you plan to open, reload the browser, ensure the default renderer is enabled, and then you should be good to go.

mattijn:
Basic idea is to add a number and comparison protocol to the `Variable` and `ExprRef` class. This includes methods for mathematical operations such as addition and instance comparison like equality checking. Addition (`+`) is represented as `__add__(self, other)` which carry out `self + other`.
So an expression `x + y` is interpreted as `x.__add__(y)`.
In this case, within this `__add__` method we use the defined `name` of the `Variable` (eg. `sample`) append the operand `+` and the definition of `other` and return this as an `ExprRef`. Given the following `Variable`:
```python
import altair as alt
year_slider = alt.binding_range(min=0, max=1000, step=5)
size_var = alt.variable(bind=year_slider, value=200, name="sample")
size_var
```

> ```
> Variable('sample', VariableParameter({
>   bind: BindRange({
>     input: 'range',
>     max: 1000,
>     min: 0,
>     step: 5
>   }),
>   name: 'sample',
>   value: 200
> }))
> ```
And:
```python
size_var.name
```

> ```
> 'sample'
> ```
Then `size_var + 1` is interpreted as `size_var.__add__(1)`. Where the `__add__` method does `size_var.name + 1` which becomes `sample + 1`. This result is returned as `ExprRef("sample + 1")`.
Therefor
```python
er = size_var + 1
er
```
> ```
> ExprRef({
>   expr: sample + 1
> })
> ```
And then the same works if these mathematical operations are also defined as methods for `ExprRef`, where the defined `expr` is used instead of `name` (`ExprRef.expr`).
So:
```python
er + 2
```
becomes er.__add__(2), which does `er.expr` (in this case this is `sample + 1`) append `+ 2`, so it becomes `sample + 1 + 2` and is again returned as `ExprRef` as such:
```
ExprRef({
  expr: sample + 1 + 2
})
```

ChristopherDavisUCI:
In the following example (which doesn't work at the moment), `size_var` is of type `Variable`.  What's the right way to define the `Variable` class so that the following code will work.  (Notice that `size=size_var` is being specified inside of `encode`, not inside of `mark_circle`.  The example does work if you move `size=size_var` into the `mark_circle` method.)
```
import altair as alt

cars = data.cars.url

year_slider = alt.binding_range(min=0, max=1, step=0.05)
size_var = alt.variable(bind=year_slider, value=0.2, name="sample")

c = alt.Chart(cars).mark_circle().encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q',
    size=size_var,
).add_variable(size_var)

c
```
I'd like to get something like this working and then add on the arithmetic functionality that @mattijn described above.

mattijn:
Hmm, per https://vega.github.io/vega-lite/docs/parameter.html, I don't think an expression reference is supported as input for an encoding channel. Within an encoding channel it can can only be included [as predicate](https://vega.github.io/vega-lite/docs/parameter.html#as-predicates) and for [data extents](https://vega.github.io/vega-lite/docs/parameter.html#data-extents) (but then as `param` and not as `expr`).

ChristopherDavisUCI:
@mattijn Ah, thank you, very interesting!

Playing around on the [Vega editor](https://vega.github.io/editor/), I think something like this is allowed, where the `"expr"` has been wrapped as a `"value"`:
```
{
  "params": [
    { "name": "barHeight", "value": 5}
  ],
  "data": {
    "values": [
      {"a": "A"}, {"a": "B"}, {"a": "C"}
    ]
  },
  "mark": {
    "type": "bar"
  },
  "encoding": {
    "x": {"field": "a", "type": "nominal"},
    "y": {"value": {"expr": "barHeight"}}
  }
}
```
I guess on the Altair side, we should allow something like `alt.value(bar_variable)`.  Do you think we should allow further abbreviations?  Maybe not, since we don't allow `y = 5`, for example (only `y = alt.value(5)`)?

mattijn:
Ah yes I see, yes that make sense. So then you can eventually do the same for `datum` and `field`. Something like this:
```python
import altair as alt

cars = data.cars.url

mpg_slider = alt.binding_range(min=0, max=50, step=1)
rule_mpg = alt.variable(bind=mpg_slider, value=0.2, name="mpg_rule")

dots= alt.Chart(cars).mark_circle().encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q'
)

yrule = alt.Chart().mark_rule(strokeDash=[12, 6], size=2).encode(
    y=alt.datum(rule_mpg)  # <-- how to infer type? eg. alt.Y(alt.datum(rule_mpg), type='quantitative') ?
).add_variable(rule_mpg)

dots + yrule
```

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAFgAY2AacaYsiygAlMiRoRQBmDgF9p3AA4BDAE6KAtpRQBtUFHWYUINfIQB9ZaQA2B7gCMSdPCBLzS45CFVQENo4oAeKACsXEYkKKGUhJjyKACMciBoipakBsgATEHSALrclooAnpjK2qB0ioSKTqTKloZihPKUyAD0rWB0UAB0AFaUdJiWcGjK3VCYhK1Q8mqtaJgIigC0FVWUk5QAAmhx3RkAnN1srWuKHSqUfZTQIIlqKgDWhmBwymDWINyYUJB0JEhUCBAkD4ENHB4ALJwayUUzyEqmADiKUst24hEKCMMAEdSIoiHAqsQFnduIUnGDLBCQIIIMoNvIINgSl8QJjsR48QTiMSRgZZHJypVqkC9Gp0iBMMZMWS-MpnkCOZKLJ9uJQ4AAvdIZdWEZQQR6YAAiikozGQWjiuoAbDlEj8-gCnCDhYRSGonJh-PJSh5jGZVQKMVjJdzCXzSYK8iAACSUMCMaWikCNZpteaLRTdBBExikWzdOAQVoJpMPTNLZbDaLzFjdOIAdmO13RIDOG0ImiBZ2WBxtmAAHEFG4obXEwDbbDaMhJbIoYJhZyx2HQgoOMooQmwyiBKIU1LYIPUoQBlABiABU2WtJRkOGxlmw4k+4peOMgPxw2b64GB0hIRyDgknCgPuh7HoYkIXte3C3oY94cE+GSvu+bCfuh34KMof4ATa3QSNkoF7geR4niA0FXjelR3g+T4SKhX6YTu2G4awEj7BkQokRB5GUbB7Y0QhdFsCwjHMRhP44f+KAZIOBGNtx4FkVBMHUdEwlIWwQTiRhkmsTJmRBN0LBEWBpGQWeVFwUJHiIY+bA2rpTFSWxyASBk3RBCwSkWXxak2RpdkiY2znMa5hlySZvm8ap1mCUFID2U+g5hfpIC-pF8k+cRymWRRAUJbRWkHGlWEZdJOr1kEEgxSpVkCfBwVIXEDlvkx6WZTq8lsFxuV+XFjW2UldFxC+z5oXp5VdbJHESDl5mxQ16nFY+1plSxFVuU23Q2mZPH1QV8VNSNz6vhtEVVd0g4LQd+X8StmnjShE0ddNlWybdeX+cdw2IeNDGvRJ71uRknlZHV92FSd-1PmJQNTZtM2ZM2QQ2pDP1DYlsPaRdBndQ2X0DctgXFeNTkIy5+OyUcin9UtR1Y2TT6hZT4XU5k+EHLV9OHQ9pNPSleNbZFHEZCBi189Df0cONpVs51H2ZGw119ZLUO-djsvLK1wvI7Ou37d9g2Pc1L5jXrSsZPh1oYybAtmzrL3tcDSNW1zRvE4zpunc7lug8ZBxq3dmM+4hL3O5NVMizqHGNjz6uhw7vv0f7kXVQnIf20VmkvfDLuI5dsl7NbdskznzUvTpCsg5FKuNhLWflzDHAvRTBfR-r4Oewz-MVynbCsx37Mx-ERxBMHxvNzLvVCzXbvbUcBx04n2ct7PbDy8Pivbc2jYHGX3vJ+HOttVHI-63sjY91LmvFShFvz0XmQcSw6O8xrTO507ac6nsbCZynkffuiFAbjW3rXdIcQjgSEbkAvu69AaRzegvQy0CCJE17tLLWADU5Pw5ugxs79V7TxwYDfO58d512uivJuwDEFPmrhA1BOoVab0PggmegN26UMgbJFW3MOHYOKoDIevCWHFy8sQuhnCyFz2Yc-cWhshF300oDLe4jFGeRtAfD+ScQEcAYrrfBo9Mh7FMior+zUjHgM0RzJRcCvayJET-ExXcTLSPgcIzS+df6yW0bokh9CZ7w2Qa7RRV9aFeNUc1eGYC7GmPvB4yxYcODwwoSgiJJkb6f1SaJRhfizFeUAU47xsSnw8MyfY1+mDb5WNOvDMRVTElzUHCk4+aT5EJPcSwKJpSYkNKfBo5p7iiHtIMfk4xCjqnHEnv0+piExKP2mYk+snj5l5KWWEwu9jqpzKwQMxCTDCkZHrGNcZ69q7bM7lbDicQcn6MuXglZ7jTkXJntXDJ4SZlyXeTg6uTDum3NVn84q1dKnfMSZ5QRei14fJZicjiNpam5I6bjNgqU3FW2MoRUFmlq7DMhfrDig4SkHIWRwHSUygWgxJXi5qVLbEjOxd0Bu9LTpUuuRfK29ZDjssQu3E5NsUWPJnhTLlVCdTGVEvyjgFN4nMoDgRdZ5K8kUy+TsqFXlZWOQKVi0GexoE6ophCzVl9rrGoRfqyKnkgiONVWiimmKXk8obGSupaqhknOMjaNpsLSHFSctSxVkUuYqo9Y6nWTKiVW2bDacNqKJlBolXwzI8k4jusTevIeJyjiDx1azFNEjOa7UzaKnBrMFUxtBvhXF-rgkVrhic+SZa4WNvRS60GRxzn1ucZpVmpqbmg2bAcB5bbiqsyadWyKqMC1dJDTqZs8052b2bccEV47+2n0KYRY4raA1buWTSwyHljiBJkWU06oV1rWoAhug9zVnU7sNfuhtxVMVFsUfhfNvbL2IUxVWs1sb9j2ojRMzFGqh0zowTqzFgKF2yWbBDX9hyOCYsHdymtJk+kOvA1azt6dWVjofadTFzrj06lrQm8t76vW3tksZRsfqgl9sfdu+jRSRywajYU9B8buM3oI1A+SBxQNZpnlvXj+EbSvtY6deWn6CHGWAjq+WgGoNQJHee6JFLV35I4+gm6qm9VCf4V5bTGy0Xyww5K2aJl9lgfXvLKdQHaUNmYxe1DenyMIcyPWFgsm-0cHloS1zkVjIsAs7hpz7HTMoxMh5nTeTSpHt84cfYOHHMSdcXFqQMGUMUqmWu7SsrjGKcSfJG0mXxNa2MepzDWULX1oAIKQgAFoADk8ltUKciqRh9Wude68hXr8lBzVfIoNrraKHL1ds8gKrA32vTYmQ5SDDX0hBGMkalry3hsdoo6wFtYnJt7ZmxUndtaTuGCm-tlzGnJAq1trtob52MWFJYHsYjIBbtvdCw99y2V70-bO6t2Lh33LSYc-lX7YPUvTqujopbr24fleRvc-rL2Vvr3Abxpd12PCw5xyNgzT2ouHSJzPAGvGVazmR9jqnTaDOMcS17SnODxrwYR-EaT5OYeg+J7qgzZzvvs+ZoPXjnlIv07yeNHz3PkAieB2LwWem4uBwmzdgXjPg0K65qzhmKvHbw7C1AvYHkZdovNmjpWrUQVY7yX7YXDY+enZR+vCOvGzmW4mS9Obqa4j1lfUbge635txF5z7j3JmIdxHknTh3aK25Sf2KL7XOCXr3Y2-Efzaf3czxevL03OfugHGV+n++dG4sY5hUEkPJ9dfF+QOgiQ1Gtf54zzxjjBs6114r9-QTsexuu-bwzuRZ1u97GHFHrhJO4tg2OATkHHeXG4O79+4P-frFM-n5VkfhOt+nUBlzpvu7Bz7+X2P1fNnU2t4ItDt3V+1H4Yh593a5eV-P-exxt+9+Z-j-+2z2QBYGyk30-230bwB28hMjAKfwgOjSbxqlL01wP3AKPxywhwnl2gf1H02XOg4xCGikTwmVCQ+w4mfH-2KjiTIP2ANwp0P0WR31fyRUoJ8Rj18xAO1WIPXnhhv2LWgI-zgMGQlw4wkBExwNQKEMYO-1y3TUELwLV1fxVkHDb0kLwMgKAPmjdVYPKRNwB0Al2joP5zQMYMHw4N5Qv3r0pXwLiwC04h0I5Tnwh13TjwcKOWeWcN5RQMvzyU+R3U8lbzcOsIO1813TtSCPRT4OfgkHIMsIYOCKz3m1MgbDzykOCKLwBxYE8m8KsPRUAKSJxQiKpTPlCKAlUJ8LRUZQ+2ynKNyM5Q+2bHuQiMFR-3rA3GaKcNCObBXA6LX1y3whql6LD1v2MnFl6JP30LWSX1yJNR3UqyMMf09REPn3wnP16IyM0OUNgKWPyNv08mez7xMLlXB1CJ6jiKON1T0KAPYHy0OLSMuJtzcgMLfgiJzVELjRyPiJEMeJPW7XkLRUrR3VpnOPuNZmGOLXHAIm2IBPYL1z+NeIuw419RMmmK+MnUKSYy8n+ImVZg2Pm33lL1RIuNZl2OLW5hSIRI0Pm3PwpO4JnmvQQIBwOBVjRkpJ+PSCDnfwiKfQ40bD3m5M6L1380+IuIAwxPNwFP0zi0xJk0lImKAOAmwMlKiI5njihMlMSNTSIVuLoVyLI0KUHGUKJPuMxVJOfkbGyNqK+NSipNTRHH2GxPXhtMZOuOyIFLMIVztVoIiMkwILGwkIqImQU163rH3h9I8N8zVJ6LpJwXlnBOfkNK8gDNyPlnlPxK03DKFzi1JV2itIuOcwNMYzzPuPljxNTUVJ2zuOSyrwh3JMCJjOKhSxKIVzGjG2NOrKuPD13VE3DI9KbwzT2SCKKwMyyBRKHIcnZPiDjyIKrI6Tq0l39MPgAEkAAhSEfbF08PDgJM5ctc-bSc5ATkntIJVc9ct7f3YtVqG2JfU8-beMjmA4Jeb7W8t7NM1NfsAiG8vct7FU0xA4FtV9F8sHZY2PbcxsYHIC9eByMsy86cmcXcs84Cs0ghZ8F3BC4bW04tQceSZ878uHTc8ss5AMyCmeNaA87U3pdCq3Gw0C+ucokijnQUpvGk-8qiiZanDjFQ0tNiwXe80xVqOOHixnEIls7c9o+tBi8XX89HXqAiXCxCwXTUskto7wyS1XGC5+Zk1PISxixQ3zDce3E8vC4nTCzS1GBYwwNS43Aiy80SLgoyhSnXA81qGBei4ygvGi3zZeLEnSyvDebMxciS9yzvPo2skXXy7+KUiHa+QyuhKygeN84tC05AiKyuREuLPaT81KgeJS5+GTa6NyxykKjSjmZEiebKk+PSz08gr8oqyvUy1Uq+Qqx3LvaU7tZqpPDAyMlkiyjweK0BTyhXBuewoKuqr-cil9CqwxCMoar7VS4K1fPi5GPk7DKatfRKhMlWEAta7hDEiLYihar-XK1U3lXqkAfq6amQ6K7bA6sa7fKq5igRWqvJIxZs5ivYfsHa1qiHHMjFL6vsgHYcY4eShQmyzS-+DqkgpiwG+Seau64Qi8hM-zZ6tFdJA0-+CCw68pES968zNa3g9GtC0ahQ460xRMysuKrG4Qkq0xTEpsfGmsyM7tQCqm0wt6wGzyIzYm1G76-SxokGnmgGoA0TBsSGp5CfbM2tC-C69FA8oGpjNa6uRGjmG6XMxWpg7q1lOG3w2EpvRsJ7M6mW8FUbUvFm+G9wkCyM1GM2nWq6yM6VbWyoxmoa0M9Whq2m-CZDBynWzs8s6FQ21m6woW-E-CeONalo6U5TaWwOoXci6VaO82440Kq2hsR2pNDWoa5sErbm9OnGgHGKmVHO9eWY3k-YgW3O0m5at0ousVedIa2tcu4u52vWpDMW2u925GcbY4NunBINMGlWj2cOrqhXEWhOvJN4uLcksMmu9tOW9NAOxOkQ5Wsmo4aM72mEqKry-XNa1mDalCr7TGxegdGnWnRu+kl-XzVqCLHeuu-ssaHymeidZugHMabRG+ju23Y8ymo+3m0SgRd+5ygBVlNank6vWS6+EB6GoAlyrkx+zSMUgzO3cbSBzelsiPfYG2tFODSXbbQ+vJdDXjQPXaFByu23A2cClBmm9HaciPFB5C-ijNYpSBj+tyLSt5OBtjX2pKo4PaZhuWsuta307M1enuxsqB+bLSi3Dh+TGapvUdM9QRjO-s1CvlaRxCVMmnNos+2M9K0CmqxRy2tB2tPBqzW+l++mtR4LZ+6B6crbRRlhtBdBbyex-u-i0c6e9eoM4e-svLKRzxluYc6vJRafSxsrSXRo0R4Seckc4yOx+tJEAAeQSaRAUKodt1kokG8MSeSYUPofR3S3KOyZScFvZuga0uRUPiKfUNcfR0VNYviaSeKahuDoD05I8boSqadoltj3kfgoaZya6ecq7sfMqcadtuXtqZZKX06YmT8O73vFGYGdmd1oBwyBWrWcWaafFukqtjL1pKCRmfFtIdBi7rGX6a2fhTtoVzkkKbGa6byaVlgXwhePOZ9tKfm0yf2OB0OfhS4efnYHrB7NeadpaeLU+zaNuaWabu6Y4NPX42BdzoPNb20QDJ+d7tkf0OZM2aWKWqVk+xHW+budzr3tMUyan1RaJehZ2bck+xgVfTRcDQvoVzEOk2xcjSudP0xMGIRehYeZpdQovwZc0iDXedTV6SlrZfTr+Y5hALaOmcpfboPLfi2rOqFeagnswNQu+zVavXEbFci2yUlezQxeuPXFFqNfPtQab16XN1VYVfbRJeRkixHXlahctepcMgnmroOftafsMcQNagtfbTSbcjtVGO1d9a3QeqgKqwyyDafocfSH1qycjfVd-qbx0XpdTd1dBbypFu2p5ZnlAcwL6yzbdZwQ-QaMNTtfLdo2ToVxXC2pTdrfgaUagPAv2EFezf-RWaAPXGU3jdbazMwOtns0HbY39cyLVIgcLYrbMeuPytUZ9ZbYnb5cMnYBZIpZXdIxOOZaCFcvHZ3elcSU5M+tndo1zY5ikDjQje3fUcGtPzEKYfPc0mDNEMydLy7bvasfrdP04MckPfvatcxY10A5-cdbxZ2gT2XYuZ0eHY4KUX3bA702OfXbsOzpg+rI5cyMhMWxfeahCw+3kZCcw9McTeCDvwOI6e7asePeRnXDLdg7EcvdMXHCXhraY6iZ6wIPSzLzA7CZ49WP49m0KTRjJwG2awAAUAAZDc71UvZXKT2Tt7A8qKeFuvJT-bCZx5jiUdCTmTu859RffT5T4CiDg1Ez-bD1yjImjTgzt7VDqVMclrTTt7ENk9Jtyzv7SXa6M65rVzuHUVsk1lPzgLqC9NwG7Slz+z1HDEkyRTmLwXGp9J66aYsL4Sg87tZM9L3S7T7af+YPHL8XXF7abIhL0zwXczwydjrz9i3Ry+vYPpuzir4SxzlAVe2rwXdzjkuN6Llr3StdpN1lWoorwWcjxXELzrnXZL7aFWXhvr2Xbxl+gRUXUbtKmFlsgIkbxLjy-y0C66VbnbkKvLxx66crlq4D6Br5qbkKqrs3ZRBbzq+DhXMbbb-rvytr5AONQro7vy7rlAJdUL37yKwbgHqL5rlq8bnqH797gfGbwyPeQ72HtKlj5a81x7iZMBOL1wjHhhPbyMryNL4H+6k7obn9CHtFchbHon5H9AvOhU0vIH2nga57uR59inzHpl5Rh+jnvH-75vc3Gnl66xrc5VG7lxcb++te3Utb9AujlLixXH2fVHlLpHJXnBXxRB709XqgvV2C0vN7hQ0n+IDicqnXtgy78PTmyw2X6Qu7kvABcXi36znnXr3nkJLnl+m2HI23zpbD6BpDJH3J3jGBJn9QoLr9fYJ33Q+HgCU3mH9QpF8GaPxwjb0-bKIPwZj7CFlPi2434Axrn34n1Pkrk9b9Q3rp+39yWmCv5Z1n-Q1YoXrpz7rIh793-5edpI1Ym34vi20H5AJA8cXPoOiPjme0pD83hlCLoA2N8S9vsFJboA4atYyf06CO66wn4f2Ok26XVfgVE1+bGcVlbL3vpO0vpNlWQCLfimKv-drW6-+rz0pDIv5npOz72Ns9+f4Vzv1NWfzP3O-vtqXqZf9moIrQmn9T37HF5erDRrrXyboq83IqtLmiAN1Zp9AajGBPhvTlpTEt+gJDjKeyb44k22wtICLgN7ZW9DUPfV-iIRd7N4F8hdFAYhHRIGZYEw3Mgf7y7LZQw+G9fvvclfgv9x6u7fsixX-7GtoBaCAymDDIEIC0E-5Y4NwLwxoDoGu6esowLQx69n41oaVCf2oEINq8dhUQUW2IHh41SenSAd-ir5x5YahgudvX2gaxtpeQCfzqf2-yfd7kImGwXW354R5x4mAxQbwMYbqcZeLgm0qP34qDw8a5g50oUk5J4c1B72GQekCBq95gh1AoRrWSex+CYs+PF7jAkIHZD8+V9BsDoKw7n94go5UlFvw0YsDlM+Q7LHYPDyt8AsVQz3vYLmhUCsO3g5EmMXMGEcDMyQzwa+yEEv1sK9lVIR2Vj5jwl4dQnRokOLj1wshMsQJj008gAdIBAnefHZSa7jC5yIneZuQS4gxhpQ8gWUJaGAC5BZAQAA)

image


And:
```python
import altair as alt

cars = data.cars.url

x_options = alt.binding_select(options=["Horsepower", "Acceleration"])
x_var = alt.variable(bind=x_options, value="Horsepower", name="var_x")

dots = alt.Chart(cars).mark_circle().encode(
    x=alt.field(x_var),  # <--- not existing yet on altair side and how to infer type?
    y='Horsepower:Q'
).add_variable(x_var)
```

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJghgLhIFygK4CcA29zQgegMYWQGcA6AK0IHsA7EAXwBoQAHAiAW0PgG1Qr2BTDADcCAfQAeIRiNSJBcEAAkKRfkwoB3fsikgARgEsqYeKCNNEUDIX6p+uK4wpMoB6pzhclKm+q07GAEFcXFttaDcaAF1aWijGNgIAawxcA2RcO11+KlwKMCMAc1MQSQQQADMDWxNy-nEmHQURZAk6RigATyZ5EABHRAgqVxhXIUEGEE6SqpqMAFkDO0JRHtaAcQhUVGpdLp6MAaGRiPG6WKA) (not possibe, as raised in https://github.com/vega/vega-lite/issues/7365)

jakevdp:
One piece of context: the reason we require `y=alt.value(5)` rather than `y=5` is because values may be strings, and otherwise `y='x'` would cause ambiguity between `'x'` as a value and `'x'` as an encoding name.

ChristopherDavisUCI:
Trying out a minimal version of @mattijn's example (not even using parameters), I noticed this works:
```
import altair as alt

cars = data.cars.url

dots= alt.Chart(cars).mark_circle().encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q'
)

xrule = alt.Chart(cars).mark_rule(strokeDash=[12, 6], size=2).encode(
    x=alt.datum(17)
)

dots + xrule
```
but it doesn't work without specifying `cars` as the data source in the `xrule` definition.  So this doesn't work:
```
import altair as alt

cars = data.cars.url

dots= alt.Chart(cars).mark_circle().encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q'
)

xrule = alt.Chart().mark_rule(strokeDash=[12, 6], size=2).encode(
    x=alt.datum(17)
)

dots + xrule
```
That seems okay to me, but I just wanted to check, since there were recent changes related to empty data scenarios.

jakevdp:
I think that second example should work... have you ported the changes from #2515 to your new `v5/api.py`?

ChristopherDavisUCI:
@jakevdp Thank you, you are right!

Do you think we should try to allow an assignment like `size=size_var` within `encode`?  That seems nice to me, on one hand, but I understand that each customization like this that we allow beyond Vega-Lite is a potential headache to maintain down the road.  So I have no strong opinion.

mattijn:
I'm a bit hesitant to allow for `size=size_var` within `encode`. It is not supported by Vega-Lite so it requires an action in Altair. This can be lifting it upwards into `mark(size=size_var)` or parse it within encode as `size=alt.value(size_var)`, `size=alt.datum(size_var)` or when possible one day `size=alt.field(size_var)`. 

Either decision leaves room for ambiguousness, but since `alt.value()` requires no infer of data type, while `datum` and `field` does, I'm leaning towards voting for `size=size_var` as a shorthand for `size=alt.value(size_var)`.
Since:
```json
{
  "data": {"url": "data/cars.json"},
  "params": [
    {
      "name": "var_x",
      "value": "red",
      "bind": {"input": "select", "options": ["red", "blue"]}
    }
  ],
  "mark": {"type": "circle", "color": {"expr": "var_x"}},
  "encoding": {
    "x": {"field": "Horsepower", "type": "quantitative"},
    "y": {"field": "Miles_per_Gallon", "type": "quantitative"}
  }
}
```

And
```json
{
  "data": {"url": "data/cars.json"},
  "params": [
    {
      "name": "var_x",
      "value": "red",
      "bind": {"input": "select", "options": ["red", "blue"]}
    }
  ],
  "mark": {"type": "circle"},
  "encoding": {
    "x": {"field": "Horsepower", "type": "quantitative"},
    "y": {"field": "Miles_per_Gallon", "type": "quantitative"},
    "color": {"value": {"expr": "var_x"}}
  }
}
```
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJghgLhIFygK4CcA29zQgegMYWQGcA6AK0IHsA7EAXwBoQAHAiAW0PgG1Qr2BTDADcCAfQAeIRiNSJBcEMn5gpIAEYBLKioQgtTRFAyF+qfriOMKTKBuqc4XRctVrZggLq1aHxmwIA1vCgUACeTPIguBrIuGZ0jPxUuBRgWgDmwSCSugBmGqY6IAASFET8TBQA7vzIqmERGACOiBBUtjC2QoIMIKFZ+YUYALIaZoSiEciiAOIQqKjU9eGRLW0d0BrdCVEUi3W6MnJZ-OJMByAi05Le3kA)

Next to this I'm wondering if there is any shorthand how to infer on data type in case of an expression within an DatumType, since VL requires the following :
```json
"encoding": {"x": {"datum": {"expr": "mpg_rule"}, "type": "nominal"}}
``` 
I can't get it shorter than:
```python
alt.X(alt.datum(mpg_rule), type='nominal')
```

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAFgAY2AacaYsiygAlMiRoRQBmDgF9p3AA4BDAE6KAtpRQBtUFHWYUINfIQB9ZaQA2B7gCMSdPCBLzS45CFVQENo4oAeKACsXEYkKKGUhJjyKACMciBoipakBsgATEHSALrclooAnpjK2qB0ioSKTqTKloZihPKUyAD0rWB0UAB0AFaUdJiWcGjK3VCYhK1Q8mqtaJgIigC0FVWUk5QAAmhx3RkAnN1srWuKHSqUfZTQIIlqKgDWhmBwymDWINyYUJB0JEhUCBAkD4ENHB4ALJwayUUzyEqmADiKUst24hEKCMMUAgahIKTu3EKTjBlghIEEEGUG3kEGwJS+IEx2I8AEdSIoiHAqsQFndZJxypVqkC9Gp0iBMMZMUS-MpnkCWZKLJ9uJQ4AAvdIZdWEZQQR6YAAiikozGQWjiuoAbDlEj8-gCnCDhYRSGonJh-PJSh5jGZVQZEsqcXiCfVZLluAASShgRjS0UgRrNNrzRaKboIHmMUi2bpwCCteOJh4ZpbLYbReYsbpxADsx2u6JAZw2hE0QOl8llluAuVkQA)

ChristopherDavisUCI:
Thanks @mattijn, maybe it's best for now to not try to add any of this extra functionality within `encode`.  I guess it can always be added later if it would be useful.

I should have some time later today to try to clean up the branch (like updating to match the most recent master branch for empty data and fixing some formatting that I had forgotten to run before).

ChristopherDavisUCI:
I tried an alternative route to getting operations to work with variables.  I added an `_as_ref = False` attribute to the `Expression` class, and I had `Variable` inherit from that `Expression` class but with `_as_ref = True`.  The point of `_as_ref` is to control whether `to_dict` turns the expression into an `ExprRef`.

https://github.com/ChristopherDavisUCI/altair/blob/e15cb8007628f9ef2ac0778e81b7beec053498b5/altair/expr/core.py#L51-L67

https://github.com/ChristopherDavisUCI/altair/blob/e15cb8007628f9ef2ac0778e81b7beec053498b5/altair/vegalite/v5/api.py#L239-L244

Does this seem like a reasonable approach?  In the few tests I've done it seems to be working well.  (Comment: I haven't updated the v5 api to allow empty data yet.)

mattijn:
Inheritance is much cleaner indeed! But now this:
```python
import altair as alt
year_slider = alt.binding_range(min=0, max=1, step=0.05)
size_var = alt.variable(bind=year_slider, value=0.2, name="sample")
size_var
```
Returns
> ```
> sample
> ```
But here I was expecting the `Variable` definition, which is now available as `size_var.variable` (or `size_var.param`):
```
VariableParameter({
  bind: BindRange({
    input: 'range',
    max: 1,
    min: 0,
    step: 0.05
  }),
  name: 'sample',
  value: 0.2
})
```
Can this also be turned around so that `size_var` returns a `Variable` as `repr` and the expression objectification is available as `size_var.expr`?

mattijn:
While reading #1431 and #1433, I think @jakevdp has probably a better opinion on this than me.

ChristopherDavisUCI:
> Can this also be turned around so that `size_var` returns a `Variable` as `repr`

@mattijn That should be possible.  Maybe we would then have to add some `try` block to `_js_repr` so that before returning `repr` at the end, it possibly returns `.name`.  Does that sound like a good solution?

One quick sales pitch for these new Vega-Lite parameters is I think [this example](https://altair-viz.github.io/user_guide/interactions.html#selection-values-in-expressions) in the documentation will be able to be simplified to something like

```
import altair as alt
import pandas as pd
import numpy as np

df = pd.DataFrame({
    'xval': range(100),
})

slider = alt.binding_range(min=0, max=100, step=1)
cutoff = alt.variable(bind=slider, init = 50)

alt.Chart(df).mark_point().encode(
    x='xval',
    y='yval',
    color=alt.condition(
        alt.datum.xval < cutoff,
        alt.value('red'), alt.value('blue')
    )
).add_variable(
    cutoff
)
```

cutoff

mattijn:
Another discussion point related to `params`. 
A nested `params` within `vconcat` works OK for a selection brush:
```python
import altair as alt

cars = data.cars.url

dots_top= alt.Chart(cars, width=200, height=200).mark_point().encode(
    x='Horsepower:Q',
    y='Miles_per_Gallon:Q'    
).add_parameter(brush)  # add_selection(brush) or add_variable(brush) works equal


dots_bottom = alt.Chart(cars, width=200, height=200).mark_point().encode(
    x=alt.X('Acceleration:Q', scale=alt.Scale(domain=[0,25])),
    y=alt.Y('Displacement:Q', scale=alt.Scale(domain=[0,500]))
).transform_filter(
    brush
)

(dots_top & dots_bottom).to_dict()
```

> ```python
> {'config': {'view': {'continuousWidth': 400, 'continuousHeight': 300}},
>  'vconcat': [{'mark': 'point',
>    'encoding': {'x': {'field': 'Horsepower', 'type': 'quantitative'},
>     'y': {'field': 'Miles_per_Gallon', 'type': 'quantitative'}},
>    'height': 200,
>    'params': [{'name': 'selector001', 'select': {'type': 'interval'}}],  # <---- params nested within vconcat
>    'width': 200},
>   {'mark': 'point',
>    'encoding': {'x': {'field': 'Acceleration',
>      'scale': {'domain': [0, 25]},
>      'type': 'quantitative'},
>     'y': {'field': 'Displacement',
>      'scale': {'domain': [0, 500]},
>      'type': 'quantitative'}},
>    'height': 200,
>    'transform': [{'filter': {'param': 'selector001', 'empty': True}}],
>    'width': 200}],
>  'data': {'url': 'https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/cars.json'},
>  '$schema': 'https://vega.github.io/schema/vega-lite/v5.1.0.json'}
> ```

But a nested `params` within `layer` does not work:
```python
import altair as alt

cars = data.cars.url

mpg_slider = alt.binding_range(min=0, max=50, step=1)
rule_mpg = alt.variable(bind=mpg_slider, value=0.2)

dots= alt.Chart(cars).mark_circle().encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q'
)

yrule = alt.Chart().mark_rule(strokeDash=[12, 6], size=2).encode(
    x = alt.datum(rule_mpg, type='quantitative')
).add_variable(rule_mpg)

(dots + yrule).to_dict()
```

> ```python
> {'config': {'view': {'continuousWidth': 400, 'continuousHeight': 300}},
>  'layer': [{'data': {'url': 'https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/cars.json'},
>    'mark': 'circle',
>    'encoding': {'x': {'field': 'Miles_per_Gallon', 'type': 'quantitative'},
>     'y': {'field': 'Horsepower', 'type': 'quantitative'}}},
>   {'data': {'name': 'empty'},
>    'mark': {'type': 'rule', 'size': 2, 'strokeDash': [12, 6]},
>    'encoding': {'x': {'type': 'quantitative',
>      'datum': {'expr': 'variable001'}}},
>    'params': [{'name': 'variable001',  # <---- params nested within layer
>      'bind': {'input': 'range', 'max': 50, 'min': 0, 'step': 1},
>      'value': 0.2}]}],
>  '$schema': 'https://vega.github.io/schema/vega-lite/v5.1.0.json',
>  'datasets': {'empty': [{}]}}
> ```
Resulting in:

> ```cmd
> Javascript Error: Unrecognized signal name: "variable001"
> This usually means there's a typo in your chart specification. See the javascript console for the full traceback.
> ```
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgNghgngpgTiAXAbVAEwgFwk0BXOMJEAC00wAcBnRAeloGM0A7AOgCsq0YwBLANzitmMTLWYUAtrX4wA5hAC0GbFVFUAAvwCMrAEwBOVgAZaKiIwhwqHKgHtmIAL4AaEJKsBrYg15wGYDAgbjDMDHZovMxyuCAAHrEAZrw8aMQAsryBVAD6FPA5AOIQYGAOwSCYUPnEAI54EMyYvNjNss5uUEkpYGmIIAASdtYwFHYA7vAVVTX99Y3NrQJBTqsuoBRWEJJUSKggzNtB-fxWvBAARoHGxtoVF1F9oFEUeJjEcI1yQW4eCYgAVmMvyiSGBIComFGSG0rhApzAeGOegBTgAum5zLFDpJjiAYJIKFUOu4vLEZni4HhAhUqLwAF7ItyQuB2TwwAAiECoJD22j0LgAbGi4aFwpForF-ugsHhJLEYHEKAgTmdLtdbiSKXUGk0Wlhls5VhiQAASKgMEgEnD9MiUGj0WQKVhyFokPAXVi8Oy0C1WjwyeRKPhQmQA1i6Yy2cqYrDc9QKwnElDAdGrIA)


But when this `params` is placed at the top-level it functions OK:
```python
{'config': {'view': {'continuousWidth': 400, 'continuousHeight': 300}},
 'params': [{'name': 'variable001',  # <---- params at top level
   'bind': {'input': 'range', 'max': 50, 'min': 0, 'step': 1},
   'value': 0.2}]}],
 'layer': [{'data': {'url': 'https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/cars.json'},
   'mark': 'circle',
   'encoding': {'x': {'field': 'Miles_per_Gallon', 'type': 'quantitative'},
    'y': {'field': 'Horsepower', 'type': 'quantitative'}}},
  {'data': {'name': 'empty'},
   'mark': {'type': 'rule', 'size': 2, 'strokeDash': [12, 6]},
   'encoding': {'x': {'type': 'quantitative',
     'datum': {'expr': 'variable001'}}},
 '$schema': 'https://vega.github.io/schema/vega-lite/v5.1.0.json',
 'datasets': {'empty': [{}]}}
```
[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgDghgThC2DOIBcBtUA7OBTZIBu0AlhAEYA2WADJQIwgA0IJh6AJsqC2AK4AuuMdAHMcjWBAAeyAKyUxLZHJDxeWMMhoBfRgTLccSAEzTNAXUZkIATyxRkaEKwi8IHENyhlcAC168w8EgA9EEAxqzoAHQAVvCsWGSEeFCR6Fi8QehgsEF4WEIQALROLvDp8AACeDSRhgCckZRBJRBh0PAx8AD26CDaIOJQANa4oYRQoRQMIFjooV2sLEJuUkigAGaECexIIACyhBTwAPpgtscA4hBkZD3TvFZnuACO3BDovIQun3l9jFZuTbbXAACS6UDKYC6AHdbPdHgYQK93p9vkkcJpMfRQC03JhYIisLAwA8-gNoCM1iAHk9dlBuFNGPBCAAvAyGJm8KBdIZYAAiEHg3nsNA5ADZTP1ZvNFsIVm4StxYG4sBIwHZdgQoMRyFRaGSaYjkR8vs50X1MeYQAASeChbxE1y7Xz+QIhPIFSJCL7ebgkSKELpBO0O8S5fJFRKqXLSSI1SidO6MFplXiIKlEkkA1DAMyYoA)

Is this something that Altair should accommodate for or is this a Vega-Lite issue?

ChristopherDavisUCI:
@mattijn I think I added back the `repr` functionality you wanted.  The solution felt a little ugly to me.  I added an `_is_variable` attribute to the Variable class, and then I use that in `_js_repr` to check if `val` is a Variable: 
```
    elif hasattr(val, "_is_variable") and val._is_variable:
        return val.name
```
I wanted to do something nicer like `isinstance(val,Variable)` but to me it seemed like the directory hierarchy does not allow that. 

I'm very open to a nicer way to do this... it feels like a very natural thing to want to test.

ChristopherDavisUCI:
@mattijn I don't think this is immediately related to your layer examples above, but I wanted to mention that so far the `add_variable` or `add_parameter` methods have only been added to `Chart` but not to things like `LayerChart`.  I can do that soon if it would be useful.

mattijn:
Good point though! If it is advised by VL-authors to collect all `params` on top-level than your suggested change is required. I noticed already that this was not possible, but thought upon first impression it was because of the schema.

ChristopherDavisUCI:
Currently there is an `add_selection` on those other top level charts, so I think it makes sense to also have that for `add_variable`.  I just kind of forgot to put `add_variable` there... I'll get it updated pretty soon assuming it doesn't raise any new issues.

ChristopherDavisUCI:
Here's a Jupyter notebook with some examples of parameters: [link to notebook](https://christopherdavisuci.github.io/UCI-Math-10/param.html)

Please ignore the unpolished course notes!  It just seemed like an easy way to share the notebook.

How are those examples looking?

I went through the [Vega-Lite parameter documentation](https://vega.github.io/vega-lite/docs/parameter.html) and was able to reproduce I think all of them except for the example of [varying bar heights](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUMSSCGcCGgDaoTHUFw0IAEpxSsvYpARmpzOcsBVAMpgFAAhpZP3o2BggkWWUDJQAnCLJ9IwBGMQAGBQAmDLEFAGZU1IBdJQQ4JGcYxNRDLPSMgHYikqQYgGs4GINq7KKAX0LepRwWpAQu4xBZUYt0eBoyLEdtXRmM9JAmQOU0UECcBkxLONkExwRAtCTmgA9LgqUITDgcS97B0CnSy0DZDoAhFpeGgALwsSjg1xwMUscwWmAIGRA-XBsigbGUgTIOxAt00IEwAE8cDMQABHBgRVxqGikRzQHQzD40KAzABmOggcEG+LMM3sgkEbyUgiQBI6RlAyDa2MJxMsTBajjRgjY0PQAGI4Fq6SDGSAIVCYXB5oshfrUejMdjcaBWTQ4IJtuhjgkIDVCkihZKWq0ZUSSQroUplarLOrlBGdaDsQa1RhjXDPSi0RiTtbsXaHU6QC79IYkh63u8QFLfXjZQHFcG2Cq4+qoA2o3rY0aTYczXALamsXibSBM46jvE80VPcXS365ehA0qa6H0MI4Y8tYImCsmzHIXGfv9Abqk+aU1bexn7YP0KVygxKm6C2OFN7peX-fKqyAQ3Hlw61w4Hvu8S26A7jEAIxEC0Ydl2x6gH2A7ZpeFQjoWXolj6k4kqYUC+tWtbyiKWEbgBW6tomkFHmmJ54nBlilh0bqjkWxTvnIdo9iYzJliYMAcb8EBdBkRZAA). I don't think there's any essential difficulty with that example; I just haven't sorted it out yet.

mattijn:
Looks great! What is the difference between `add_variable` and `add_parameter`. Should we distinguish between the two or pick one? I think in VL docs there is mainly a notion of `parameter`?

ChristopherDavisUCI:
No difference!  I just wanted some symmetry with the already existing 
add_selection.

ChristopherDavisUCI:
I added in that remaining example to the same [parameter notebook](https://christopherdavisuci.github.io/UCI-Math-10/param.html#variables-in-terms-of-other-variables).  (It was the data part, not the parameter part, that was giving me trouble.)

I think parameters are basically working now how I wanted, so I'm very happy to hear any changes you think should be made!

ChristopherDavisUCI:
By the way, I'm still getting the error `Error loading script: Script error for "vega-util", needed by: vega-lite http://requirejs.org/docs/errors.html#scripterror` if I try to run anything in Jupyter notebook with Vega-Lite 5.  (I did try clearing outputs, starting new notebooks, etc.)

I think there is some new dependency that is missing.  @mattijn pointed out this file, [utils/html.py](https://github.com/ChristopherDavisUCI/altair/blob/samplev51/altair/utils/html.py), which seems like a good clue, but I got a little lost trying to read the contents of that file.  I've tried doing some google searches for `vega-util`, but haven't really understood what I was looking at.

jakevdp:
Seems similar to this? https://github.com/altair-viz/altair/issues/2280

mattijn:
Gentle ping @domoritz, we have a question regarding https://github.com/altair-viz/altair/pull/2517#issuecomment-975689895, do you now if something happened or is missing related to vega-utils, vega-embed since VL5?

ChristopherDavisUCI:
Thanks @mattijn, I also asked @domoritz on Slack and he suggested it should be something more basic than what I thought.  Probably just a version needs to be updated.  I will investigate tonight.

domoritz:
Could you generate a standalone HTML page where you get the error with `vega-util`? I can take a look.

ChristopherDavisUCI:
Thank you!  Does this count as a standalone html page?  [link on Google drive](https://drive.google.com/file/d/184pko12jOUFRgTivOvBWfr2eQ9EC3B9X/view?usp=sharing)

I produced it by making a Jupyter notebook with the error and then downloading that notebook as an html file.

I can definitely do more debugging myself also, now that I know it should be an out-of-date reference.

mattijn:
I used this piece of code to debug. It works OK in JupyterLab (to start do from cmd `jupyter-lab`), but not in Jupyter Notebook (to start do from cmd `jupyter notebook`):

It is based on the [`spec_to_html()` ](https://github.com/altair-viz/altair/blob/master/altair/utils/html.py#L145) Altair function.
```python
from altair.utils.html import spec_to_html
from IPython.core.display import display, HTML
import json

dct = json.loads("""{
  "data": {
    "values": [
      {"a": "A", "b": 28},{"a": "B", "b": 55},{"a": "C", "b": 43},{"a": "D", "b": 91},
      {"a": "F", "b": 53},{"a": "G", "b": 19},{"a": "H", "b": 87},{"a": "I", "b": 52}
    ]
  },
  "mark": "bar",
  "encoding": {
    "x": {"field": "a", "type": "nominal"},
    "y": {"field": "b", "type": "quantitative"}
  },
  "height": 200,
  "width": 200
}
"""
)
# dct  # this seems ok

h = spec_to_html(
    spec=dct,
    mode='vega-lite',
    vega_version='5.21.0', 
    vegaembed_version='6.20.2', 
    vegalite_version='5.1.1',
    requirejs=False,
    base_url='https://cdn.jsdelivr.net/npm',
    fullhtml=True,
    template='standard'
)

# print(h)  # seems valid HTML

display(HTML(h))  # this doesn't work
```
For Jupyter Notebook I get this error in the console log:
```
VM48:17 
        
       Uncaught ReferenceError: vegaEmbed is not defined
    at <anonymous>:17:8
    at b (main.min.js?v=24786799f8766c957f062e6b8f0ea89d00055170b051b7d318e97aedfb60dea047c35f2dfd92a435f058e984b092f71bc1fc7d28aacf0c330637e94e2ac72e5d:2)
    at Pe (main.min.js?v=24786799f8766c957f062e6b8f0ea89d00055170b051b7d318e97aedfb60dea047c35f2dfd92a435f058e984b092f71bc1fc7d28aacf0c330637e94e2ac72e5d:2)
    at S.fn.init.append (main.min.js?v=24786799f8766c957f062e6b8f0ea89d00055170b051b7d318e97aedfb60dea047c35f2dfd92a435f058e984b092f71bc1fc7d28aacf0c330637e94e2ac72e5d:2)
    at OutputArea._safe_append (main.min.js?v=24786799f8766c957f062e6b8f0ea89d00055170b051b7d318e97aedfb60dea047c35f2dfd92a435f058e984b092f71bc1fc7d28aacf0c330637e94e2ac72e5d:59205)
    at OutputArea.append_display_data (main.min.js?v=24786799f8766c957f062e6b8f0ea89d00055170b051b7d318e97aedfb60dea047c35f2dfd92a435f058e984b092f71bc1fc7d28aacf0c330637e94e2ac72e5d:59412)
    at OutputArea.append_output (main.min.js?v=24786799f8766c957f062e6b8f0ea89d00055170b051b7d318e97aedfb60dea047c35f2dfd92a435f058e984b092f71bc1fc7d28aacf0c330637e94e2ac72e5d:59092)
    at OutputArea.handle_output (main.min.js?v=24786799f8766c957f062e6b8f0ea89d00055170b051b7d318e97aedfb60dea047c35f2dfd92a435f058e984b092f71bc1fc7d28aacf0c330637e94e2ac72e5d:59003)
    at output (main.min.js?v=24786799f8766c957f062e6b8f0ea89d00055170b051b7d318e97aedfb60dea047c35f2dfd92a435f058e984b092f71bc1fc7d28aacf0c330637e94e2ac72e5d:60836)
    at Kernel._handle_output_message (main.min.js?v=24786799f8766c957f062e6b8f0ea89d00055170b051b7d318e97aedfb60dea047c35f2dfd92a435f058e984b092f71bc1fc7d28aacf0c330637e94e2ac72e5d:62597)
```

jakevdp:
The error is coming from `requirejs`. JupyterLab does not use requirejs, while Jupyter Notebook does, and the `universal` template is the only HTML output that attempts to support `requirejs`. I suspect the issue is in one of the vega javascript sources, the requirejs mode specifies a dependency on `vega-util`. I'm not sure whether that's intended.

jakevdp:
You can see the requirement declared at the top of the vega-lite v5 source: https://cdn.jsdelivr.net/npm/vega-lite@5

Again, this problem will only arise in an environment where requirejs is loaded. If it's intended on the Vega-Lite side, it means that an extra library is necessary when creating vega-lite charts in a requireJS context. We could add this to the universal renderer:
```
const paths = {
      "vega": "https://cdn.jsdelivr.net/npm/vega@5.21.0?noext",
      "vega-lib": "https://cdn.jsdelivr.net/npm/vega-lib?noext",
      "vega-lite": "https://cdn.jsdelivr.net/npm/vega-lite@5.1.1?noext",
      "vega-embed": "https://cdn.jsdelivr.net/npm/vega-embed@6.20.2?noext",
      "vega-util": "https://cdn.jsdelivr.net/npm/vega-util?noext",
    };
```
But since it's not mentioned anywhere in the vega-embed documentation, I suspect this is a bug in the vega-lite.js build process. @domoritz might have a more definitive answer.

domoritz:
Good catch. Since Vega exports all the utilities from Vega-Util, we try to replace all imports from Vega-Util with imports from Vega in the bundle. It should happen with https://github.com/vega/vega-lite/blob/12bee8b450c1176434b5166ca5f42515c41a78db/rollup.config.js#L36 but evidently does not here. I will fix this now.

mattijn:
No typescript user, but here it seems mentioned as an external dependency? https://github.com/vega/vega-lite/blob/12bee8b450c1176434b5166ca5f42515c41a78db/rollup.config.js#L81

domoritz:
I think I fixed the issue in https://github.com/vega/vega-lite/pull/7823. It turns out rollup doesn't let us use both external and global. We have to alias the dependencies manually. The original goal was to reduce code duplication and import functionality from the Vega bundle instead of bundling it again.

domoritz:
Can you try Vega-Lite 5.2.0?

ChristopherDavisUCI:
I checked and it's working now in Jupyter Notebook!  Thanks @domoritz for the very fast work!

domoritz:
Thank y'all for isolating the issue. It helped tremendously.

ChristopherDavisUCI:
@jakevdp Can I ask, how do you think `selection`/`variable`/`parameter` should exist in the Vega-Lite 5 version of Altair?  Do you like the idea of all three existing, or should there only be `selection` and `variable` or maybe only `parameter`?

Currently as @mattijn pointed out above, there are `.add_selection`, `.add_variable`, and `.add_parameter` methods defined on Chart objects that all do the exact same thing.  (The first two are aliases for `.add_parameter`.)

I have some ad hoc code right now to keep old examples from breaking.  For example, if a `selection` has `init` specified, I change that to `value`.  And if `type='single'` or `type='multi'` is specified, I change that to `type='point'`.  And if a selection has `empty` specified, I have some code to move that into `alt.condition` or into `transform_filter`.  Does that all sound okay, or is it a longterm headache to have this sort of code to fix one-off problems?

jakevdp:
My feeling on this is we should stick to the language of the Vega-Lite schema; that is, use the name `parameter`, the method `add_parameter`, etc. For the 5.0 release, we could keep the previous `selection` functions around, but have them raise a `DeprecationWarning` mentioning the new syntax, and that the `selection` functionality will be removed in a future release. What do you think?

ChristopherDavisUCI:
That sounds good to me and it should be pretty easy to implement!

ChristopherDavisUCI:
@mattijn pointed out this old PR: https://github.com/altair-viz/altair/pull/1629

Is this a good time to try out adding the code from there?

jakevdp:
> Is this a good time to try out adding the code from there?

Maybe? I'm not sure about the approach in that PR. The fundamental issue is that currently for schema objects, attribute access returns the attribute. That is,
```python
import altair as alt
x = alt.X(field='x')
print(x.field)
# x
```
In order to have the semantics proposed there, attribute access would have to return a function that sets the attribute; that is, we'd have to change the API of Altair objects to be something like this:
```python
x = alt.X().field('x')
print(x['field'])
# x
```
That's certainly doable, but it's a fundamental departure from Altair's current object model. We could discuss whether the advantages of that outweight the disadvantages, but I think that's not something that should be done as part of this PR.

ChristopherDavisUCI:
Okay cool.  I'll ask no more questions until after Thanksgiving!

mattijn:
I raised it primarily because of the parameter definition. When having a chart with a few sliders, you have to define them now as follow:
```python
rad_slider = alt.binding_range(min=0, max=100, step=1)
rad_var = alt.variable(bind=rad_slider, value=0, name="radius")

rad2_slider = alt.binding_range(min=0, max=100, step=1)
rad_var2 = alt.variable(bind=rad_slider, value=50, name="radius2")

theta_slider = alt.binding_range(min=-2*np.pi, max=2*np.pi)
theta_var = alt.variable(bind=theta_slider, value=-0.73, name="theta_single_arc")

theta_slider2 = alt.binding_range(min=-2*np.pi, max=2*np.pi)
theta2_var = alt.variable(bind=theta_slider, value=0.73, name="theta2_single_arc")

corner_slider = alt.binding_range(min=0, max=50, step = 1)
corner_var = alt.variable(bind=corner_slider, value=0, name="cornerRadius")

pad_slider = alt.binding_range(min=0, max=np.pi/2)
pad_var = alt.variable(bind=pad_slider, value=0, name="padAngle")
```
I feel it is a bit verbose and wish there is a shorthand similar to #1629, so it can be done as an one-liner, eg.
```python
pad_var = alt.param(value=0, name='padAngle').bind(alt.range(min=0, max=np.pi/2))`
```
But maybe too much for now.

ChristopherDavisUCI:
If a parameter in Vega-Lite is defined by
```
{
  "name": "sel",
  "select": {"type": "point", "fields": ["Miles_per_Gallon"], "toggle": false}
}
```
what would be a good way to define that parameter in Altair?  Is something like the following okay, or is it too verbose?
```
s = alt.selection_point(fields=["Miles_per_Gallon"], toggle=False)
p = alt.parameter(select=s)
```

As far as I can tell, there's no "Parameter" defined in the Vega-Lite schema.  There is an
```
"anyOf": [
  {
    "$ref": "#/definitions/VariableParameter"
  },
  {
    "$ref": "#/definitions/SelectionParameter"
  }
]
```

jakevdp:
Having to wrap `alt.selection` in `alt.parameter` seems a bit cumbersome. But I haven't thought deeply about this question

jakevdp:
Maybe `alt.selection_single` by itself should return the full parameter definition?

ChristopherDavisUCI:
I started a new Pull Request #2528  so that I could clean up some of the commit history.  It uses the most recent suggestion of mostly following the terminology of Vega-Lite schema, but allowing `selection_point` etc to define parameters.

I'll probably close this PR in a day or two unless someone says I should leave it open.

Thanks and see you at #2528!

PR #2528, opened 2021-11-28
WIP: update to Vega-Lite 5.2

ChristopherDavisUCI:
The code here is pretty similar to #2517, but I wanted to redo some things and it seemed easiest to start with a new pull request.  Also the conversation on the other one was getting pretty long.

Main changes from #2517:
* In the previous version, an expression like `size_var + 3` created an Expression, but now it creates a Parameter.  That is the only major change.
* As briefly discussed in the other PR, `selection_point` etc now create the entire Parameter. 

Main changes overall (from the master branch):
* Include a `vegalite/v5` folder with schema 5.2.0.  The first commit is just a duplication of the `vegalite/v4` folder.  I hope that makes it easier to tell what changes were made.
* Move height/width/view from inner charts to the parent chart for LayerChart.
* Introduce parameters.  Here are some [examples](https://christopherdavisuci.github.io/UCI-Math-10/param.html) (please ignore the course notes!).
* Moved most code out of Expression and into a new class called `OperatorMixin`.  The goal is to use the same code for both something like `expr + 3` (which should produce an Expression) and something like `size_var + 3` (which should produce a Parameter).  I'm really not sure if the way I accomplished this is natural or not.
* Added a fair amount of ad hoc code to keep the old selection syntax mostly working.  I added many `warnings.warn(message, DeprecationWarning)`, but they seem to all be hidden by default, so I have never successfully seen one of these messages displayed.  The code would be a little cleaner if we didn't try to keep this old syntax.

If you see something that could be improved, please let me know, because I'm happy to learn about more efficient ways of doing things.

To do:
* When we are happy with the general syntax, I can write some documentation for parameters and add some examples.  I also think it would be good to see if some old examples can be simplified.  For example, this [US population over time](https://altair-viz.github.io/gallery/us_population_over_time.html) I think is more naturally made with a variable parameter as opposed to a selection parameter.
* Decide how forcefully we want to stop users from using the old selection syntax.  My `warnings.warn` approach seems worthless at the moment since the messages don't get displayed by default.
* The only example that is raising an error is [scatter plot with minimap](https://altair-viz.github.io/gallery/scatter_with_minimap.html).  It is failing because it provides an explicit dictionary using the keyword `selection`, which in the new schema should be `param`.  I think it's not worth the effort of writing code to try to make this example compile, and we should rewrite that example (only two words need to be changed).
* The other tests that I know of which are failing are some render examples (I haven't looked at those at all, nor have I updated Altair Saver) and `test_spec_to_vegalite_mimebundle`.

Thanks for any comments/requests!


mattijn:
This looks really good. I like the `OperatorMixin`. Left a few inline comments. I think once Altair 4.2 is released we can start working on this PR to get accepted in master including passing of tests on this branch.

ChristopherDavisUCI:
Thank you @mattijn!  I haven't succeeded in finding the comments yet.  I clicked on "Files Changed" above for this PR and scrolled through the `altair/vegalite/v5/api.py` and `altair/expr/core.py` files but didn't notice anything.  Am I looking in the wrong place?  Thanks again!

mattijn:
My bad, didn't submit review.

altair/vegalite/v5/api.py
```diff
@@ -0,0 +1,2791 @@
...
+        # TODO: move common data to top level?
+        # TODO: check for conflicting interaction
+        for spec in layer:
+            _check_if_valid_subspec(spec, "LayerChart")
+            _check_if_can_be_layered(spec)
+        super(LayerChart, self).__init__(data=data, layer=list(layer), **kwargs)
+        self.data, self.layer = _combine_subchart_data(self.data, self.layer)
+
+        # Some properties are not allowed within layer; we'll move to parent.
+        bad_props = ("height", "width", "view")
```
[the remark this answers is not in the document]
ChristopherDavisUCI: I changed to `layer_props`.

ChristopherDavisUCI:
Thanks @mattijn I made those and a few other small updates, including changing the `scatter plot with minimap` example.  (That's the only example of a "breaking change" in version 5 that I know of.)

I'll maybe try writing some drafts of documentation for parameters, but I'm open to suggestions if there's something else that would be more helpful.  (I don't think I can contribute anything useful on the JupyterLab front.)

ChristopherDavisUCI:
Here is a nice small error I'm getting with the tests for this `samplev52` branch. 

```
import altair as alt
from altair_saver import save

source = data.cars()
brush = alt.selection(type='interval')

    x='Horsepower:Q',
    y='Miles_per_Gallon:Q',
    color=alt.condition(brush, 'Cylinders:O', alt.value('grey')),
).add_selection(brush)

save(chart, 'chart.png')
```
If I instead just display the chart instead of trying to save it, it works fine.  Or if I comment out the `alt.condition` line, it works fine.

The error that displays for me is
```
/Users/christopherdavis/opt/anaconda3/envs/mattijn/share/vega-lite-cli/node_modules/vega-lite/build/vega-lite.js:672
    return !!op.not;
                ^

TypeError: Cannot read property 'not' of undefined
    at isLogicalNot (/Users/christopherdavis/opt/anaconda3/envs/mattijn/share/vega-lite-cli/node_modules/vega-lite/build/vega-lite.js:672:17)
    at logicalExpr (/Users/christopherdavis/opt/anaconda3/envs/mattijn/share/vega-lite-cli/node_modules/vega-lite/build/vega-lite.js:953:9)
    at expression (/Users/christopherdavis/opt/anaconda3/envs/mattijn/share/vega-lite-cli/node_modules/vega-lite/build/vega-lite.js:14035:12)
    at /Users/christopherdavis/opt/anaconda3/envs/mattijn/share/vega-lite-cli/node_modules/vega-lite/build/vega-lite.js:9302:11
    at Array.map (<anonymous>)
    at wrapCondition (/Users/christopherdavis/opt/anaconda3/envs/mattijn/share/vega-lite-cli/node_modules/vega-lite/build/vega-lite.js:9299:39)
    at nonPosition (/Users/christopherdavis/opt/anaconda3/envs/mattijn/share/vega-lite-cli/node_modules/vega-lite/build/vega-lite.js:9621:12)
    at color (/Users/christopherdavis/opt/anaconda3/envs/mattijn/share/vega-lite-cli/node_modules/vega-lite/build/vega-lite.js:9679:10)
    at Object.symbols (/Users/christopherdavis/opt/anaconda3/envs/mattijn/share/vega-lite-cli/node_modules/vega-lite/build/vega-lite.js:15252:10)
    at parseLegendForChannel (/Users/christopherdavis/opt/anaconda3/envs/mattijn/share/vega-lite-cli/node_modules/vega-lite/build/vega-lite.js:15820:72)
```
and the traceback is
```
CalledProcessError                        Traceback (most recent call last)
/var/folders/8j/gshrlmtn7dg4qtztj4d4t_w40000gn/T/ipykernel_96872/2813738607.py in <module>
----> 1 save(chart, f'chart.png')

~/opt/anaconda3/envs/mattijn/lib/python3.9/site-packages/altair_saver/_core.py in save(chart, fp, fmt, mode, embed_options, method, suppress_data_warning, **kwargs)
    170     saver = Saver(spec, mode=mode, embed_options=embed_options, **kwargs)
    171 
--> 172     return saver.save(fp=fp, fmt=fmt)
    173 
    174 

~/opt/anaconda3/envs/mattijn/lib/python3.9/site-packages/altair_saver/savers/_saver.py in save(self, fp, fmt)
    119             raise ValueError(f"Got fmt={fmt}; expected one of {self.valid_formats}")
    120 
--> 121         content = self._serialize(fmt, "save")
    122         if fp is None:
    123             if isinstance(content, dict):

~/opt/anaconda3/envs/mattijn/lib/python3.9/site-packages/altair_saver/savers/_node.py in _serialize(self, fmt, content_type)
    112 
    113         if self._mode == "vega-lite":
--> 114             spec = self._vl2vg(spec)
    115 
    116         if fmt == "vega":

~/opt/anaconda3/envs/mattijn/lib/python3.9/site-packages/altair_saver/savers/_node.py in _vl2vg(self, spec)
     63         vl2vg = exec_path("vl2vg")
     64         vl_json = json.dumps(spec).encode()
---> 65         vg_json = check_output_with_stderr(
     66             [vl2vg], input=vl_json, stderr_filter=self._stderr_filter
     67         )

~/opt/anaconda3/envs/mattijn/lib/python3.9/site-packages/altair_saver/_utils.py in check_output_with_stderr(cmd, shell, input, stderr_filter)
    196     """
    197     try:
--> 198         ps = subprocess.run(
    199             cmd,
    200             shell=shell,

~/opt/anaconda3/envs/mattijn/lib/python3.9/subprocess.py in run(input, capture_output, timeout, check, *popenargs, **kwargs)
    526         retcode = process.poll()
    527         if check and retcode:
--> 528             raise CalledProcessError(retcode, process.args,
    529                                      output=stdout, stderr=stderr)
    530     return CompletedProcess(process.args, retcode, stdout, stderr)

CalledProcessError: Command '['/Users/christopherdavis/opt/anaconda3/envs/mattijn/bin/vl2vg']' returned non-zero exit status 1.
```
Is it obvious what kind of error this is?  The possibilities I see:
* Something wrong with the way selections/conditions are currently defined in Altair.
* Something needs to be updated in Altair saver.
* Something wrong with how I have node packages installed.  (I admit I don't know what I'm doing.)

jakevdp:
Saving a chart requires `altair-saver` to have the correct Vega-Lite version available, which it doesn't yet.

ChristopherDavisUCI:
Great, thank you!  That has me down from about 35 tests failing to 8 tests failing.

I'm not sure how ill-advised this was, but I got the right versions installed by using `npm install -g --force vega-lite vega-cli canvas`.  (It wasn't working without the `-g` or the `--force`.)

ChristopherDavisUCI:
One observation about the current syntax in this PR is that you make a variable parameter using `alt.parameter`, but there is currently no "Pythonic" way to make a selection parameter using `alt.parameter(select=...)`, so there is some asymmetry with the Vega-Lite syntax.  Instead, currently the easy way to make a selection parameter is using `alt.selection` or `alt.selection_interval` or `alt.selection_point`.  Is that an issue?

Some alternatives I can think of:
* Have a keyword like `as_parameter` that defaults to `True` in the `alt.selection` functions.   One could then use `as_parameter=False` if they wanted to create the more low-level SelectionConfig, so it would look like `x = alt.selection_point(as_parameter=False)`, and then one could construct a selection parameter using `alt.parameter(select=x)`.
* Allow syntax like `alt.parameter(select=alt.selection_interval())`.  This feels a little inelegant to me, because `alt.selection_interval()` is already a parameter, but implementing it would be easy.  (We'd just check if the value passed for `select` was a selection parameter, and if it is, replace it with the corresponding selection config.)  But maybe I'm wrong and that's not so bad?

But maybe nobody would ever use those alternatives?

What do you think?  There's a balancing act between staying faithful to the Vega-Lite syntax on one hand, and having more concise code on the other hand (and not breaking the existing Altair syntax).

mattijn:
I would follow Vega Lite API in here, see https://vega.github.io/vega-lite-api/api/#parameters and https://github.com/vega/vega-lite-api/blob/master/api/api.js#L124-L130. Basically meaning that the low-level SelectionConfig syntax cannot be created anymore using Altair. Its still allowed within Vega-Lite, to remain backward compatibility, but its on the list to be deprecated in the future.

ChristopherDavisUCI:
Thanks @mattijn.  The vega-lite api approach seems very similar to what we currently have (e.g., `alt.parameter` makes a variable parameter but doesn't naturally make a selection parameter), so I don't see any changes that need to be made.

Please let me know if you come across anything you think should be changed!

ChristopherDavisUCI:
I've gotten myself a little confused about what's the right way to add parameters to a multiview chart.

Something like this works, where we add the parameter to the first chart and then refer to it in the second chart.  An analogue also appears in the [Vega-Lite documentation](https://vega.github.io/vega-lite/docs/parameter.html#as-predicates):

```
import altair as alt

cars = data.cars.url

c1 = alt.Chart(cars).mark_circle().encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q',
).add_parameter(
    brush
)

c2 = alt.Chart(cars).mark_circle().encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q',
).transform_filter(
    brush
)

c1|c2
```

But something like this does not work.  Again we add a parameter to the first chart and refer to it in the second:

```
year_slider = alt.binding_range(min=0, max=1000, step=5)
size_var = alt.parameter(bind=year_slider, value=200)

c1 = alt.Chart(cars).mark_circle(size=size_var).encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q',
).add_parameter(size_var)

c2 = alt.Chart(cars).mark_circle(opacity=size_var/1000).encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q',
)

c1|c2
```

If we instead add the parameter to the layer chart, then it works:

```
cars = data.cars.url

year_slider = alt.binding_range(min=0, max=1000, step=5)
size_var = alt.parameter(bind=year_slider, value=200)

c1 = alt.Chart(cars).mark_circle(size=size_var).encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q',
)

c2 = alt.Chart(cars).mark_circle(opacity=size_var/1000).encode(
    x='Miles_per_Gallon:Q',
    y='Horsepower:Q',
)

(c1|c2).add_parameter(size_var)
```

Do you see what the difference is between these two cases?  Does this seem to be working as expected?

mattijn:
It matters on which level the parameter are placed. I've experience in Vega-Lite that I had to add parameters in a nested concat chart, where it was not allowed to be defined on the top level. Will add an example soon. I'm not sure if this intended behaviour of VL5.

mattijn:
The example I got is a bit more complex (Google Colab URL [here](https://colab.research.google.com/drive/17pQMySdUBtn4Lf6pFqxq0CGpkTsOUlm4?usp=sharing)):
```python
import altair as alt

stocks = data.stocks.url
hover_1 = alt.selection_point(name="hover", nearest=True, on="mouseover", encodings=["x"])
hover_2 = alt.selection_point(name="hover", on="mouseover", encodings=["x"])
highlight = alt.selection_point(name="highlight", on="mouseover", clear="mouseout")
input_dropdown = alt.binding_select(options=[[0,300], [0,600]], name='fill domain')
param_domain = alt.parameter(value=[0, 600], bind=input_dropdown)

line = alt.Chart(height=150).mark_line(point=True).encode(
    x=alt.X("date:N", timeUnit="year"),
    y=alt.Y("mean(price):Q"),
    color="symbol:N"
)
rule = alt.Chart().mark_rule(tooltip=True).encode(
    x=alt.X("date:N", timeUnit="year"),
    color=alt.condition(hover_1, alt.value("black"), alt.value("transparent"), 
                        empty=False)

).add_parameter(hover_1)

rect = alt.Chart().mark_rect(strokeWidth=1.2, tooltip=True).encode(
    x=alt.X("date:T", timeUnit="year"),
    y=alt.Y("symbol:N"),
    fill=alt.Y("mean(price):Q", scale=alt.Scale(domain=param_domain)),
    stroke=alt.condition(highlight, alt.value("black"), alt.value(None), 
                        empty=False)
).add_parameter(hover_2, highlight) 

alt.concat(
    (line + rule), rect, data=stocks, columns=1
).resolve_scale(
    fill='independent', 
    color='independent', 
    x='shared'
).add_parameter(
  param_domain,
).configure_scale(bandPaddingInner=0.0, bandPaddingOuter=0.07)
```

If I move the `highlight` parameter from the `rect` to the `concat` it will give the following error: 
```cmd
Javascript Error: Duplicate signal name: "highlight_tuple"
```

But if Iremove the `param_domain` parameter from the chart and only add `highlight` to the `concat` parameters as an `add_selection` it will work (see colab for full examples). 

When looking to the Vega-Lite definitions I observe that the behaviour of `add_selection` is different than `add_parameter`. The `add_selection` on the `concat` add the `highlight` parameter to the correct `rect` parameters. How does it know it should do this?

Wrong VL-editor spec: [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAFgAY2AacaYsiygAlMiRoRQBmDgF9ulMAEMANpjwgARgqh0ACgrp0SCAJJQomAE4ouGrbv2GoCAPKlCl6wDo2AdmmyQAAcFCwUAW0oUAG1QKHDVZBBGCDQPOUwVMHFUEEIAT0CEoIgScW5MKEhHBEjkKJAADxAAXW5oFBAwgUwUj39OWPiO+CUlAH06CDCFEhBudRI6NRJAtw7KDMwsuZAIQOJoWqiorik2VpPOADYOZtaQOLCikaUAAknp2YC0ZVIEy5u52k9wIimyMRASgUeQ8dVA0wsAGs1PlCh0lCRVNxAiUiChCBY-gEKlUjGpIEoIFYcvAMktEpQ8mF1BAlDtUUUoFMSMoQAEmjSsEp6SA6Ap3Oy4E8AKpQODZEAwkLsgqc7lxNkBPJqBQIBAWTAIcVFJ5aHa04UdQIWOBgLG5VUdACOpC0xEI4rgqT5-XhIWROQ5HUJKnZEFZxEC+MJmGJlQg1XJrKpSe08rg7RyPyUfw66ihYGR2JC4Q6yVSVnKYX22uQMGUG2+vyKBK0lGCBrx-LUFpFYol3GIMrlCqVlYdaMSXLCPM1AQ74SOgyeZd6442mWyoCDiRxpR2JITRiOjRa3HMIUwlGyBL+bSgHS6VB6FZ9zWB3EYIgQYhQAEYAFY2FkP0kRRR1EgNbY5AJCBEUwWgGGYZA-08AAmQdwyUSNoyJcp40TQVRh1PUDSNCVElNB9uF7K0bTtHZ5GUBJQA+GYHxyTAGmtYY4FGCYpnYn1BwgkAXTdeVPW9AJrwsOCWJ4NMDg49BmzzAsiyCEswjLUQMR-MoQEwat8hQeslEbbhs1zZAyFGbtBTpDp+3tIdMFleUOjHFVJ1yYycVCTVuFrUBaIZJkWTZETfOnWdhK00IImiZcikYPTREMjcti3Ccij3PFuDAFRlUo7oKEMzNOjK19-HfEFWVIMIoFqP9uH7BQ1FICw2USMRCECShkAAeiGsA6CgTwACtKDoDIvQsTxzEIIaoECMIhtSI0AFp2o2QhKAAATQVC0IATm8Ib2qG68IELShPDASg0D5bgDUoVlvRyJjQxyCkU0SRZMEKbQKkMl4OkB4HZoK08GUYS8ln6EAABJ5C-aYy0IfrBpGzaFE8BB5UYUh1E8DNrrAdGFA2w0FC2jF3A2gD0O8Kb3ofaQgA)

Right VL-editor spec: [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAFgAY2AacaYsiygAlMiRoRQBmDgF9ulMAEMANpjwgARgqh0ACgrp0SCAJJQomAE4ouGrbv2GoCAPKlCl6wDo2AdmmyQAAcFCwUAW0oUAG1QKHDVZBB4JSUAfToIMIUSEG51Ejo1EkC3FBBKTBUwcW4IQOJoSOQoqK4pNgBdTlbOADYODq6QOLCEpLgUgAIMrJyAtGVSBJ7+zukhgkVxZtAlBQBPDx2QLIsAazVCfcCxpRJVbkCIEm3CCyWAzChIRyRUHiUECs-3glUKiUo+zC6ggSlyICuNzKUEyJGUIACAA81KClOCQHQFO54cRRgBVKBwbYgQ4hEnXMYosJouEBfZqBQIBAWTAIIljUZaeG4-GBCxwMAPBEMsoAR1IWmIhCJcDQqn8slApwu-0RY3eKhJEFhxECKDeH24Xx+RjUkEBwPw0EMDSgagWSiWZXUezAF0eIXCZUYEDVVitYXq7OQMGUFXmizGby0lGCPKIGO42JBWDxZUJxO4pMwFKpZVp4elSMSTJZGICafCTRiw3iwdDHjklUw1UuMsSTxe8OtEF+zZA2KG5hCmEor3emFqbsSYQEmA7Vn8HXW3EYIgQYhQAEYAKxsTUnEI60B6so83tyN4QM6YWgMZjII+eABMReNSlNc0F0+b5R1tHMUg5LkeT5YkV0wIVuBFMoxQlKV5GUBJQBmbJl1ATBMTFMpkjSHC5gCW9EnlRUqRVNVM3KJ8XztZ0qTgaB3UTb1fX9IJAzCYNRDuA8ahATBIyuFBYyUeNuA9L1kDIFIsRxXN8QLKVi1LakK3pasEXEp5QlZbho1AZCIShGE4SLfthlROJWQbfjm1iNtEhDMN4QqKpthvOzBwzK1QLHaIJxAIYOJXNcN3rTg3NGQSD2Ew8u18vt9MC0SwBUOloqoddSiXMpVwK2Ktx3AFSDCKAmiPbgCwUNRSAsOEPMIQhAkoZAAHoerAOgoE8AArSg6EqVULE8cxCB6qBAjCHq1T5ABaRqKkISgAAE0C-b8AE5vB6xqernCA-UoTwwEoNAGJ5ShYXo-4MMNf57SBMoCkwG5tC+USSM+36fvGoLwohRgZ0KfxuAAEnkPcsmDDqut6pbeQUTwECpRhSHUTx2NOsAEYUNHVrudwlpPH9vBGh63WkIA)

jakevdp:
I'm not sure, but it seems suspicious that you have two parameters with the same name (`"hover"`). I believe parameter names should be unique

mattijn:
Apologies, the VL-spec didn't reflect the text, I updated the spec ([Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAFgAY2AacaYsiygAlMiRoRQBmDgF9ulMAEMANpjwgARgqh0ACgrp0SCAJJQomAE4ouGrbv2GoCAPKlCl6wDo2AdmmyQAAcFCwUAW0oUAG1QKHDVZBBGUSVRcTlMFTBxVBBCAE9AhKCIEnTwFRCUEDCBTApy6Graqnq0D39OWPjq+CUlAH06CDCFEhBudRI6NRJAt2rKTMxsiZAIQOJoSOQoqK4pNgBdTn3OADYOI5OQOLDivqUAAmHR8YC0ZVIEs8vj6RuBEUORiICUCnyHl2oFGFgA1moCkVqqlzGtAqUiChCBZvgFMFBII4kLlIEoIFZcvBMjNEpR8mF1BAlGskcUoCMSMoQAEAB5qalKWkgOgKdysuD3ACqUDgORAkKq3DZ1Q5YS5LIC+TUCgQCAsmAQYuK9y0a0FwsCFjgYFUysKxQAjqQtMRCGK4O0eZ0YSEEbkVYlcSpWRBmcRAtjcZh8YSIMS1GSKYnoIYtlA1J8lN9qupwWAEdxgqEwtVGBB2lZuJgwpttcgYMolh8vsUcVpKMWCeI+QKsELqqLxcrJZgZXLqoqq3kHarOXFNQFi+EdqC7sVy5W1kssjlQIGSmU1gSiUZVyB+TdzCFMJQcjjvtwmokWksKx0AQDuIwRAgxCgAEYAFY2FkX14URWcgxWco7wsCA4UwWgGGYZAAM8AAmZUwyUCMozxas4wTKk4H6HU9QNI1xRfTAzW4C1qitG07RAeRlASUBXjGDNckwXkrV6UjBi494AgPZ1XTlD0vQCOCEI4nhtDlOBn3QVtc3zQsghCcIyxSNJj1rAoUEbJRm24LMc2QMh+l7EiaUHY0JWlWV5SnVkoLyGsMVCTVuHrUAGLpBkmRZe1kUSNUNW9IsdIiaJunuMt32nHcYMgiLDyxQjTycc9LyfHiajqFLvSOL8eGzMIoB2ADuCHBQ1FICwWUSMRCECShkAAem6sA6CgTwACtKDoTJPQsTxzEIbqoECMJuvaI0AFoGqWQhKAAATQdCMIATm8bqGu6u8IALShPDASg0B5bgDUoZkvVyNiQ1JZlk0SaZMCKbRu3NITqi+n6xuyi9FkYG8Zk6EAABJ5B-UYy0IDqut6paFE8BA5UYUh1E8FSTrABGFEWw0FGW1J3EWoDMO8YaHozaQgA)). There is just a single `highlight` parameter defined.

It seems that for concatenated views some of the parameters cannot be defined globally/top-level/low-level.

There are also two nested `hover` parameter defined on each concatenated views, by using the same name, they are connected to the same signal even though the behaviour for each view is slightly different (in top view, there is `nearest=True` and in the bottom view there is `clear=mouseout`).

All in all, I think I observe the following behaviour:
- Parameters used as [Expression strings](https://vega.github.io/vega-lite/docs/parameter.html#in-expression-strings) and for [data extents](https://vega.github.io/vega-lite/docs/parameter.html#data-extents) can be defined top-level, where usage is OK in nested views.
- Parameters used as [Predicates](https://vega.github.io/vega-lite/docs/parameter.html#as-predicates) need to be defined on each view, but by using the same parameter name one can define cross-view interaction.

mattijn:
I'm wrong. Predicates can be defined top-level and still be used in layers within concatenated views, [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEzjQATjRyZ289AGU4guFEwACBA0E6cxszjY1ZmCGaYBPM8vVIytWWRmSLJmTphw6sRIgpSKKkiYKKigDOqCaPGJBFBI6hBSMvIAvko4uUgIEGgA2qCyFXAZMGyk6nEQRiaYaKCY7jiN6A5hcbJwuWrdqJjqDHBKchkIbAwdLRFxUMa5SytrDN1FRQC6SlByOVO1IIJI7huo18jqANY9IH0DGVA06luNJRAcFk52UTjI7wAHu8AGY0IzKDIACTYeTgDgA7hslJ9BiAAI4MYI6RI6UggQHuWHwwSI9AAWRoxggAH0BuoWQBxaKCRY4-p4wnEugJGjkwHnXltZIgc6yME6RYyso+BBNdZtJRwhEZADymn8FKUUUEcwyZB8VKOJxKoGebxluO+v3+RqBILYYICUOpOvQKLRmOxHwFGSFzhFZIBSipMu1tIyjOZbIiXJ5fJDX3Q4ZJovFSggNAAXoNQHKFbp3iqKurWnETWbUGIAAzNwENwYAJjE1uORSAA). 
I will file an issue at the VL-repo.

mattijn:
@ChristopherDavisUCI, I filed a separate issue for your case: https://github.com/vega/vega-lite/issues/7855

ChristopherDavisUCI:
Thanks for investigating this @mattijn.

One observation on the Altair side is that for `add_selection` (which I think we decided should be deprecated), if it gets called by something like a ConcatChart, then we artificially include it on all the subcharts:
https://github.com/altair-viz/altair/blob/015b03616e926ebf832fef9badf75781052573ce/altair/vegalite/v5/api.py#L2430-L2436
This was inherited from Altair version 4 and is there in version 5 for backwards compatibility.

For `add_parameter` (which includes both variables and selections), we add it on the top level only:
https://github.com/altair-viz/altair/blob/015b03616e926ebf832fef9badf75781052573ce/altair/vegalite/v5/api.py#L2418-L2428

It will be good for us to think about whether this is the correct setup.

ChristopherDavisUCI:
Hi @jakevdp, is there anything that would be helpful for me to work on?  I have some free time in my schedule.

I've sat down a few times to try to get the remaining tests to pass in my local version, but it never feels like a great use of time, since it feels very specific (specific to those particular tests) and I'm not sure if there are bigger changes that should come first.  I haven't made any changes in a while, but am happy to make updates.

jakevdp:
Hi @ChristopherDavisUCI - thanks for all the work on this. I've been distracted with other issues, so I haven't had a chance to dig into the larger design questions here. I generally trust your judgment here: from what I've seen so far this is really great work! I agree that figuring out the parameter/selection issue with regard to both future-looking design and (some degree of) backward compatibility is probably the key here, but I don't have any concrete input on that at the moment.

ChristopherDavisUCI:
Thanks @jakevdp.  It's helpful to hear that you think things are generally looking good.  I'll try to think more carefully about this parameter/selection issue and will write again if something specific is causing trouble.  I'll also try to get the remaining tests working locally for me, and then I might ask you or @mattijn for help in getting them working here on GitHub.

ChristopherDavisUCI:
After making some updates to a few tests, the only test failing on my local machine is `test_vegalite_to_vega_mimebundle`
https://github.com/altair-viz/altair/blob/fb9e82eb378f68bad7a10c639f95e9a3991ac18d/altair/utils/tests/test_mimebundle.py#L160

@mattijn If I remember right, you fixed this test last time.  Is it obvious to you what should be changed with it?  I don't know what mimebundles are...  Also, do you see what changes should be made to get the `test_render_examples` tests working here on GitHub?  Last time @jakevdp wrote
> Regarding the testing: the way I've done it in the past is to add a branch to altair viewer with the necessary changes, and temporarily adjust the CI in this branch to point to that. Then when we're happy with everything, we cut a new altair-viewer release and then change this CI back to normal.

I don't quite see how to do that in practice...

On another note, I'm now less confused about the Selection/Variable Parameter discrepancy I mentioned a few comments ago.  These types of parameters seem to also be treated differently in Vega-Lite, for example:
https://github.com/vega/vega-lite/blob/0a0099f811b9846a18115d3c9d9088b66c5dc856/src/parameter.ts#L30-L34
Basically, I think if you have a multi-view chart consisting of two charts `c1` and `c2`, and you add a Selection Parameter to `c1`, then you can use that selection within `c2` (for example, via a `transform_filter`), but it does not get literally added to `c2` (because maybe selections should not be allowed in that chart).  That sort of situation does not arise with Variable Parameters, so if you want to refer to a variable parameter in both `c1` and `c2`, then you should add it at the top level.  So I think things are working as expected in Altair.

Please let me know if you have any requests or suggestions!

mattijn:
To make the `test_render_examples` to work in GitHub it is required that the `altair_viewer` contains support for VL5.2. This has been done by me and Jake last week, so now if you point the GitHub workflow files `docbuild.yml` and `build.yml` in the Altair repo then these tests should work (similar to previous time, here https://github.com/altair-viz/altair/commit/1eef3de7788f9e77655af59fb0278643d2cc8e1b, https://github.com/altair-viz/altair/commit/d594587fe2b7dcdf56ed47e519501c1acd0b3a0c).

Regarding `test_vegalite_to_vega_mimebundle`, I'll have a look once I find some time.

mattijn:
I was unable to find the error related to the test `test_vegalite_to_vega_mimebundle`, for me this test pass locally, but in GitHub Actions it does not pass.

ChristopherDavisUCI:
Thanks very much @mattijn!  It's exciting to see all but one test passing on here (if I understand the details correctly).  For me, `test_vegalite_to_vega_mimebundle` also fails locally.

mattijn:
Maybe something with my test environment. Let me explain what is happening in this test. 
1. The test compiles a Vega-Lite specification into a Vega specification using Vega-Embed. 
2. It compares this compiled Vega specification with the pre-specified Vega specification. If these contain the same key-value combinations than the test is successful (order should not be important).

I've prepared a Colab notebook, https://colab.research.google.com/drive/16yEYKAhN92NOAQS2gtsIWPDhYPdbFDxa?usp=sharing, using a DeepDiff on the dicts, you can see the following:
```python
{'dictionary_item_added': [root['marks'][0]['encode']['update']['y2']['field'], root['scales'][1]['domain']['fields']],
 'dictionary_item_removed': [root['marks'][0]['encode']['update']['y2']['value'], root['scales'][1]['domain']['field']],
 'iterable_item_added': {"root['data'][1]['transform'][0]": {'as': ['b_start',
    'b_end'],
   'field': 'b',
   'groupby': ['a'],
   'offset': 'zero',
   'sort': {'field': [], 'order': []},
   'type': 'stack'}},
 'values_changed': {"root['marks'][0]['encode']['update']['y']['field']": {'new_value': 'b_end',
   'old_value': 'b'}}}
```
Meaning the newly compiled Vega spec using the new Vega-Lite 5.2.0 bundle is slightly different compare to the Vega spec as defined in `test_mimebundle.py`. If you agree with the changes, than you can adjust the defined Vega where necessary in order to make the test pass or open a issue at Vega-lite if something seems not right (eg. https://github.com/vega/vega-lite/issues/7776)

ChristopherDavisUCI:
Thanks again @mattijn!  That notebook was very helpful.  Those differences seemed harmless to me at first glance, so I updated the test and everything seems to pass now!

mattijn:
Currently the following is confusing me:
image
For me it is not clear which method is preferred or deprecated. I don't mind changing `alt.selection_multi` and alt.selection_single` to `alt._selection_multi` and `alt._selection_single`.

ChristopherDavisUCI:
These types of comments are very appreciated!

The commands `selection_multi` and `selection_single` are meant to be deprecated, and are only around for backward compatibility.  (I don't think they even make sense any more, because there is no distinction on the [Vega-Lite side](https://vega.github.io/vega-lite/docs/selection.html#type).)

I've added some warnings, like this:
https://github.com/altair-viz/altair/blob/454bbfe61c030c54725156013320b131fa3597d5/altair/vegalite/v5/api.py#L372-L378
but I see I forgot to add any warning to the `selection_multi` and `selection_single` methods.

Does that look like the right way to issue warnings?  I've never successfully seen one of these warnings displayed to me.

I think @jakevdp suggested that we should try to break as little old code as possible, so a lot of the choices early on were made with that in mind.

mattijn:
I understand we should reduce breaking existing code, but I also wish there is a way they don't show up in the code-completion. When these methods live in co-existence I find myself looking often confused into the docs. I wish to reduce upfront confusion while writing code using new preferred methods instead of being dependent for warnings after rendering. One way to improve this is to mention it clearly in the docstrings of these methods, but this doesn't solve the code-completion issue.

jakevdp:
I think some systems (including IPython/Jupyter) draw code completion from the `__all__` attribute of modules, so we could probably address this by explicitly defining `__all__` and leaving the deprecated method names out.

ChristopherDavisUCI:
I'm being indecisive about what to work on.  Would any of these be more helpful than the others?
* Documentation for parameters
* Tests for parameters
* Examples for the example gallery

There is a little bit of documentation for parameters [here](https://github.com/altair-viz/altair/blob/454bbfe61c030c54725156013320b131fa3597d5/doc/user_guide/interactions2.rst), but I haven't done anything related to adding tests or examples.

mattijn:
I like the docs you wrote. Updating the docs and examples in the gallery should be updated indeed. Also a start with an overview of (breaking) changes. For example, here:
```python
import altair as alt

cars = data.cars.url
brush = alt.selection_interval(value={"x": [55, 160], "y": [13, 37]})
    x='Horsepower:Q',
    y='Miles_per_Gallon:Q',
    color=alt.condition(brush, 'Cylinders:O', alt.value('grey'))
).add_parameter(
    brush
)
```
The `value` parameter in `selection_interval()` is new, but the currently adopted approach of `init` parameter cannot be used anymore.

Currently this is written as:
```python
brush = alt.selection_interval(init={"x": [55, 160], "y": [13, 37]})
    x='Horsepower:Q',
    y='Miles_per_Gallon:Q',
    color=alt.condition(brush, 'Cylinders:O', alt.value('grey'))
).add_selection(
    brush
)
```

ChristopherDavisUCI:
Thank you @mattijn!  Good point, I had forgotten to keep track of breaking changes, but I should be able to find that info from the messages above and from the commit comments.  I hope to have some time to update things this weekend.

mattijn:
This is about it regarding changes to the core of altair for vl5. Do you like (modification of) examples for the example gallery in this PR as well @jakevdp?

ChristopherDavisUCI:
I was going to make some updates to the [Parameters](https://github.com/altair-viz/altair/blob/6e62f53766c2a1b531307f32a3bb006acf36233b/doc/user_guide/interactions2.rst) draft in the user guide.  What do I need to do to have ``:func:`add_parameter` `` link to the documentation the same way ``:func:`selection` `` does?  My best guess is that I'm supposed to update the lists in API.rst, but I'm not sure if that file is auto-generated or not.  Even just a hint of what kind of thing this `:func:` is would be helpful... is it a Sphinx domain??

jakevdp:
Yes, `:func:` here is short for `:py:func:` and is a sphinx *role*. To make ```:func:`add_parameter` ``` work correctly, the `add_parameter` function will have to appear somewhere in the documentation table of contents; probably the best place to add it would be here: https://github.com/altair-viz/altair/blob/master/doc/user_guide/API.rst

ChristopherDavisUCI:
Thanks @jakevdp, that was very helpful!  It also helps me to understand some of the warnings that the docs are generating, like `WARNING: [autosummary] failed to import 'altair.NormalizedSpec': no module named altair.NormalizedSpec`, which I think occurs because `NormalizedSpec` appears in that API.rst file, but `NormalizedSpec` seems to have been removed from the Vega-Lite schema.

Should I go through and delete these items from API.rst?  It wouldn't take more than 5 minutes, but I wanted to make sure that was a good thing to do.

jakevdp:
Yes, you can go ahead and remove obsolete schema classes

doc/user_guide/API.rst
```diff
@@ -550,7 +510,6 @@ Low-Level Schema Wrappers
    ValueDefWithConditionMarkPropFieldOrDatumDefstringnull
    ValueDefWithConditionStringFieldDefText
    ValueDefnumber
-   ValueDefnumberExprRef
    ValueDefnumberwidthheightExprRef
    Vector2DateTime
    Vector2Vector2number
```
jakevdp, line 515: It looks like some of the new classes are missing from the docs, in particular new encoding channels like `XOffset`. We'll also want to add these to the encodings tables elsewhere in the documentation, and perhaps add an example or two (maybe a jitter chart and/or a grouped bar chart?)

jakevdp: We could probably do that in a followup PR if you prefer

ChristopherDavisUCI: Thanks for the feedback!  I will try to get these added and will let you know if I get stuck.

Should I systematically go through the old examples and make sure they aren't using deprecated terminology?  I know that eventually needs to be done, but if you tell me it's smarter to wait, that also makes sense.

jakevdp: I think it's fine to leave those old examples for now, one benefit is that they act as tests that deprecated APIs still work correctly!

ChristopherDavisUCI: I tried to incorporate those changes (include missing classes, examples for XOffset) in the most recent few commits.  Please let me know if you notice anything else, or if any of these seem incorrect!

ChristopherDavisUCI:
I tried to get rid of `selection_single` and `selection_multi` from Jupyter's auto-complete, but I didn't succeed.  What I tried was to add the following lines to the `__init__.py` file.  It did create an `alt.__all__` attribute which contained `selection_point` (for example) and did not contain `selection_single`, but Jupyter Notebook's and Jupyter Lab's autocomplete both continue to show the deprecated options.

```
import sys

deprec_names = ['selection_multi', 'selection_single']
__all__ = [x for x in dir(sys.modules[__name__]) if x not in deprec_names]
```

Any ideas what's wrong or whether this approach is still likely to succeed with some changes?

joelostblom:
I tested a bit in a toy package, and I think what you are seeing happens because `__all__` only affects statement of the type `from mod import *`. So if you opened a notebook and typed `from altair import *` the namespace would be populated by everything except the names in `deprec_names`.

In order to change the autocompletion, I *believe* that we need to change `__all__` inside the module that defines `selection_single` and `selection_multi`, as this `__all__` variable would be read by the `__init__.py` files when they are importing `*`. So try sticking the following at the end of `altair/vegalite/v5/api.py`:

```python
__all__ = [x for x in dir(sys.modules[__name__]) if x not in ['selection_multi', 'selection_single']]
```

Unfortunately, I think it needs to be at the end since the function definitions are not added to `sys.modules` until they are defined in this file. If we want to define it at the top for better visibility, we would could write out all the names individually, but since there is a lot of them, this seems simpler. I would also avoid defining `deprec_names` since that variable would also be added to the namespace.

ChristopherDavisUCI:
Thanks very much @joelostblom.  It sounds like I should not do anything, unless someone writes in with a workaround for the issue you raised.

joelostblom:
I found something that seems to work. Part of the issue appears to be that [the IPython completion does not respect `__all__`](https://github.com/ipython/ipykernel/issues/129), however it seems like it respects [overriding `__dir__` which was introduced in Python 3.7](https://www.python.org/dev/peps/pep-0562/). Although the IPython issue I linked mentions that there is also custom logic in their completer, the following works for me for removing the deprecated functions from the completion in both Python and IPython while still having them accessible when the full name is typed out manually.

Note that this change goes in `__init__.py` rather than the module file as I suggested above:

```python
import sys


__all__ = [x for x in dir(sys.modules[__name__]) if x not in ['selection_multi', 'selection_single']]

def __dir__():
    return sorted(__all__)
```

This looks pretty benign to me, but I don't have experience with it and do not know if there can be unwanted side effects, so if anyone notices anything that might break, please comment.

ChristopherDavisUCI:
Thanks again @joelostblom!

@jakevdp What would your suggestion be?  If I understand right, options include:

1.  Do nothing.
2.  @joelostblom's suggestion of defining `__dir__()` in `__init__.py`

I was going to include:

3.  Populating `__all__` with literal strings (as opposed to programmatically generating it).

but if I understand @joelostblom's message and linked thread correctly, it seems like this third option won't work.

jakevdp:
I just looked through the change - it's a huge effort, thanks so much for tackling this! I don't think I have any further major comments, so I'd propose we merge this and then test & iterate further on the main branch. What do you think?

ChristopherDavisUCI:
Thanks @jakevdp, that all sounds good.  I'll watch for the merge!

joelostblom:
Excited to see this being on its way to merging! @jakevdp What do you think about including the autocomplete change I suggested in https://github.com/altair-viz/altair/pull/2528#issuecomment-1061075908 to hide the deprecated methods as @mattijn originally brought up in https://github.com/altair-viz/altair/pull/2528#issuecomment-1016173679 (and the two following comments). 

This is minor and I by no means want it to hold back a merge, but I thought I would mention it to double check since I agree that seeing the deprecated methods in the completion is somewhat confusing.

image

ChristopherDavisUCI:
Thanks @joelostblom for the reminder and @jakevdp for the comment.  My vote would be to not include this autocomplete change for now and maybe consider it again if things seem to be working.

joelostblom:
Thanks both! Erring on the side of caution sounds appropriate to me as well, so let's consider the modified tab completion for later if needed. I saw that you added the deprecation warnings @ChristopherDavisUCI so that still clearly communicates the deprecated state of these functions.

mattijn:
Hey @ChristopherDavisUCI, I'm stuck here. I've the feeling when an `expr` is defined within a `parameter` its not properly serialized into correct VL using this branch. 
Observe the following in Altair:
```python
import altair as alt

data_wind = [
    {'winddirection': 0, 'label': 'North'},
    {'winddirection': 90, 'label': 'East'},
    {'winddirection': 180, 'label': 'South'},
    {'winddirection': 270, 'label': 'West'},
]

# off-topic: I wish I could define names for data sources easier
source_wind = alt.DataSource(alt.InlineData(values=data_wind, name='source_wind'))

theta_shift = alt.parameter(expr="-PI/length(data('source_wind'))")

alt.Chart(source_wind).mark_arc(
    tooltip=True,
    thetaOffset=theta_shift
).encode(
    theta='winddirection:N',
    color='label:N'
).add_parameter(theta_shift)
```
image
Checking the `dict`:
```python
chart_dict = chart.to_dict()
chart_dict
```
image
Changing `epxr` to `param` name, is success:
```python
chart_dict['mark']['thetaOffset']['expr'] = 'parameter027'
alt.Chart.from_dict(chart_dict)
```
image

mattijn:
Perfect! Funny enough, if I do: `thetaOffset=theta_shift+0`, it does works. Its still not referenced by `param` name, but at least a correct `expr`. Will the serialization of `thetaOffset=theta_shift+theta_shift` also go well?

ChristopherDavisUCI:
@mattijn, can I ask, how do you decide what the "correct" serialization should be?  Do you plug it into the Vega-Lite editor and see if it gives any warnings or errors?  Do you look at the Vega-Lite docs?  Do you look at the typescript source code?  Do you look at the json schema?  Something else?  For most cases I use the Vega-Lite docs, but for edge cases (like how should a parameter be specified in a multi-view chart) it's very unclear to me where I should be looking.

mattijn:
I use the Vega editor as starting point. 
In this case, I would assume that the following Altair: 
1. `thetaOffset=theta_shift` is resolved in VL as `"thetaOffset": {"expr": "parameter027"}`
2. `thetaOffset=theta_shift + 0` is resolved in VL as `"thetaOffset": {"expr": ("parameter027 + 0)"}`
3. `thetaOffset=theta_shift + theta_shift` is resolved in VL as `"thetaOffset": {"expr": ("parameter027 + parameter027)"}`

But currently the following happens:
1. `thetaOffset=theta_shift` is resolved in VL as `"expr": "\"-PI/length(data('source_wind'))\""`
2. `thetaOffset=theta_shift + 0` is resolved in VL as `"expr": "(-PI/length(data('source_wind')) + 0)"`
3. `thetaOffset=theta_shift + theta_shift` is resolved in VL as `"(-PI/length(data('source_wind')) + -PI/length(data('source_wind')))"`

If I define the parameter as `value` and not as an `expr` (eg. `theta_shift = alt.parameter(value=-0.78)`) it is currently serialized with this logic already, namely the result is  `"expr": "(parameter027 + parameter027)"`

altair/vegalite/v5/api.py
```diff
@@ -0,0 +1,2790 @@
...
+        if name is None:
+            name = self._get_name()
+        self.name = name
+
+    def to_dict(self):
+        if self.param_type == "variable":
+            if self.param.expr is Undefined:
+                return {"expr": self.name}
+            else:
+                return {"expr": repr(self.param.expr)}
```
mattijn, line 177: I've the feeling the behavior seen is because of this line: https://github.com/altair-viz/altair/pull/2528/files#diff-bbe4e187b18d242a366c820d023afd89041759cb96e4ec66c3f34559a72c2f9dR177.

mattijn:
I just found out it is not necessary to use an `.add_parameter(theta_shift)`. I can just define an `alt.parameter` and use it in the spec. Is this intended? (open question).

```python
import altair as alt
from altair.expr import PI, length, data

data_wind = [
    {'winddirection': 0, 'label': 'North'},
    {'winddirection': 90, 'label': 'East'},
    {'winddirection': 180, 'label': 'South'},
    {'winddirection': 270, 'label': 'West'},
]

source_wind = alt.DataSource(alt.InlineData(values=data_wind, name='source_wind'))

theta_shift = alt.parameter(expr=(-PI/length(data('source_wind'))))

chart = alt.Chart(source_wind).mark_arc(
    tooltip=True,
    thetaOffset=theta_shift
).encode(
    theta='winddirection:N',
    color='label:N'
)
chart#.to_dict()
```

jakevdp:
Sorry for this falling off my radar. It looks like there are still some questions about corners of the new APIs... still, does anyone object to me clicking merge so we can more easily test and iterate on the new version?

mattijn:
No objections, let's iterate  further from main branch (or rc1 🤩)

ChristopherDavisUCI:
Hi @jakevdp, please go ahead!  I should have time to look over these issues @mattijn mentioned over the next few days.

ChristopherDavisUCI:
Two quick questions:

Is there any downside to deleting the Altair repository that I forked?  I find it difficult to keep these things organized.

When I respond to @mattijn's questions, should I do that in a new GitHub issue or PR, or here?

mattijn:
@ChristopherDavisUCI, you can respond at #2573, and if we find an answer, can create a PR afterwards. 
Also created #2574 for the auto-completion of deprecated methods.

mattijn:
Regarding:
> Is there any downside to deleting the Altair repository that I forked? I find it difficult to keep these things organized.
Up to you. Clean is nice. You'll have to create a new fork for a new PR.

joelostblom:
Wohoooo!!! It's so exciting to see this merged!! Thanks everyone for working on it, especially @ChristopherDavisUCI for championing this large undertaking :rocket: :star2:  :trophy:

PR #2530, opened 2021-12-03
ENH: track JS versions in default HTML display

jakevdp:
Tested in Chrome & Safari, with JupyterLab, Jupyter Notebook, and Colab frontends.

cc/ @mattijn

### DIFF
--- a/altair/utils/html.py
+++ b/altair/utils/html.py
@@ -85,6 +85,7 @@ HTML_TEMPLATE_UNIVERSAL = jinja2.Template(
     """
 <div id="{{ output_div }}"></div>
 <script type="text/javascript">
+  var VEGA_DEBUG = (typeof VEGA_DEBUG == "undefined") ? {} : VEGA_DEBUG;
   (function(spec, embedOpt){
     let outputDiv = document.currentScript.previousElementSibling;
     if (outputDiv.id !== "{{ output_div }}") {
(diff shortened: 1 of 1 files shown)


mattijn:
I don't see continuous script injection the html DOM, which is a good thing. 

Using MS Edge I can reproduce the effect that my mem/cpu goes high. This happens when a console log window is opened and holding ctrl+enter on an Altair charting cell. Eventually  the console log window crash and then my mem/cpu is stable again. This also can be reproduced with Altair v4.1.0 installed. This only happens in MS Edge; Chrome stays stable with the console log window open.

jakevdp:
Just to be clear - the console log crash on Edge happens with the current release, not just with this change?

mattijn:
Yes. Not related to this PR.

mattijn:
I thought maybe its tracking all unique added `divs` under the hood in the console log somewhere. But I cannot make it to crash anymore and now it runs stable in Edge too. Also I do not observe any tracking of previously rendered Altair charts. So probably an incident then.

jakevdp:
Seems like this is working as intended, and we haven't encountered any issues yet. I'm going to merge.

mattijn:
Good. Only one comment I want to add. It's not specific about this PR, but I was wondering for a certain scenario: In a Jupyter session, after upgrading Altair, the newest vega lite version is injected and used for rendering. If a person decide to switch back to a previous altair version, than the previous schema is used to validate against, but still the newest vega lite package is used for rendering. Can this result to some cases where the chart won't render due to breaking changes in higher versions of vega-lite?

jakevdp:
vega-lite.js tends to be mostly backward-compatible even for attributes that are no longer part of the schema (for example, vega-lite 5 still will correctly render most if not all vega-lite 4 specifications, though it will display console warnings about deprecated entries). So I expect if you are mixing Altair 4.1 & 4.2 charts in a single Jupyter JS session, after this change they should render fine no matter which version you display first.

mattijn:
Great, looking forward!

PR #2531, opened 2021-12-12
Update customization.rst

mattijn:
remove note about `rangeStep`


jakevdp:
The change looks fine.

Regarding the build failures: it looks like [mistune](https://pypi.org/project/mistune/#history), which is a transitive dependency of the doc build via `m2r`, just put out a new release that is incompatible with the current version of `m2r`. We'll have to pin that to the previous version (0.8.4, it appears) in our documentation requirements.

jakevdp:
If you sync to #2532, tests should pass

PR #2559, opened 2022-02-20
docs: correcting regression equations

robna:
I assume the addition in the exponential regression is a mistake and instead should be calculated as y = a * e^(b * x), which is also what can be found on the respective vega lite doc page: https://vega.github.io/vega-lite/docs/regression.html where it appears that the same correction was comitted due to a PR by @jakevdp [44123d888fff30c9d975b16f219d63c94e20ede2](https://github.com/vega/vega-lite/pull/5742/commits/44123d888fff30c9d975b16f219d63c94e20ede2)
Similarly, also in some of the other equations there were some typos to be fixed.


jakevdp:
Thanks!

PR #2568, opened 2022-03-22
ENH: Make the schema validation error for non-existing params more informative

### LINKED ISSUE #2565: Make `SchemaValidationError` more helpful by printing expected parameters
When a non-existing parameter name is used, I think it would be helpful to include the existing parameter names in the error message. For example, when misspelling a parameter name like in the example below it is not immediately clear whether I made a typo or something else went wrong, so the immediate feedback of seeing the correct parameter names would be helpful.

```python
import altair as alt
import pandas as pd

source = pd.DataFrame({
    'b': [28, 55, 43, 91, 81, 53, 19, 87, 52]
})

    x=alt.X("a", scale=alt.Scale(padingOuter=0.5)),
    y=alt.Y("b")
)
```

```

        altair.vegalite.v4.schema.core.Scale, validating 'additionalProperties'

        Additional properties are not allowed ('padingOuter' was unexpected)
```

I also belive the error message text could be clarified a little. Especially for novices, I don't think it is immediately clear what it means that "Additional properties are not allowed" and it could helpful to update this error message to something like:

```

        altair.vegalite.v4.schema.core.Scale, validating 'additionalProperties'

        No such parameter name exists: 'padingOuter'
        
        Existing parameter names are: 'align', 'base', 'bins', 'clamp', 'constant', 'domain',
            'domainMid', 'exponent', 'interpolate', 'nice', 'padding', 'paddingInner', 'paddingOuter',
            'range', 'reverse', 'round', 'scheme', 'type', 'zero', 'kwds'
        
        See the help for altair.Scale` for the full description of these parameters.
```

We can use the `inspect` module to retrieve the list of expected parameter names via `list(inspect.signature(altair.vegalite.v4.schema.core.Scale).parameters.keys())`. I can try to PR this if it sounds like a good idea (it looks like modifying the returned value of [altair/utils/schemapi.py](https://github.com/altair-viz/altair/blob/55bb4b7df07996abb4ff05e80a69cadae291482a/altair/utils/schemapi.py#L116-L123) is the right place?).

joelostblom:
@jakevdp What do you think of something like this as a solution to #2565?

I am still working on organizing the printed parameters in neatly formatted columns so that they are easier to read, but I thought I would check in with you whether you think this solution is robust enough. It has been working fine in my limited testing and I kept the old error message as a fallback to be safe.

close #2565


joelostblom:
I rebased on master so that the checks are passing (the many force pushed are for black and flake8, sorry about the noise). I also fixed the formatting of the parameters in the error message. It required more code than expected to make an even table with names sorted alphabetically in columns, but I think that the output is much easier to read now.

Instead of the unformated list of parameters:

```

altair.vegalite.v4.schema.core.Axis, validating 'additionalProperties'

altair.Axis has no parameter named 'here'

Existing parameter names are:
'aria', 'bandPosition', 'description', 'domain', 'domainCap', 'domainColor', 'domainDash',
'domainDashOffset', 'domainOpacity', 'domainWidth', 'format',
'formatType', 'grid', 'gridCap', 'gridColor', 'gridDash',
'gridDashOffset', 'gridOpacity', 'gridWidth', 'labelAlign', 'labelAngle',
'labelBaseline', 'labelBound', 'labelColor', 'labelExpr', 'labelFlush',
'labelFlushOffset', 'labelFont', 'labelFontSize', 'labelFontStyle',
'labelFontWeight', 'labelLimit', 'labelLineHeight', 'labelOffset',
'labelOpacity', 'labelOverlap', 'labelPadding', 'labelSeparation',
'labels', 'maxExtent', 'minExtent', 'offset', 'orient', 'position',
'style', 'tickBand', 'tickCap', 'tickColor', 'tickCount', 'tickDash',
'tickDashOffset', 'tickExtra', 'tickMinStep', 'tickOffset', 'tickOpacity',
'tickRound', 'tickSize', 'tickWidth', 'ticks', 'title', 'titleAlign',
'titleAnchor', 'titleAngle', 'titleBaseline', 'titleColor', 'titleFont',
'titleFontSize', 'titleFontStyle', 'titleFontWeight', 'titleLimit',
'titleLineHeight', 'titleOpacity', 'titlePadding', 'titleX', 'titleY',
'translate', 'values', 'zindex'

See the help for altair.Axis to read the full description of these parameters
```

The output is now a table of parameters:

```
altair.vegalite.v4.schema.core.Axis, validating 'additionalProperties'

altair.Axis has no parameter named 'here'

Existing parameter names are:
aria               gridDashOffset     labelLineHeight   tickCount        titleBaseline     
bandPosition       gridOpacity        labelOffset       tickDash         titleColor        
description        gridWidth          labelOpacity      tickDashOffset   titleFont         
domain             labelAlign         labelOverlap      tickExtra        titleFontSize     
domainCap          labelAngle         labelPadding      tickMinStep      titleFontStyle    
domainColor        labelBaseline      labelSeparation   tickOffset       titleFontWeight   
domainDash         labelBound         labels            tickOpacity      titleLimit        
domainDashOffset   labelColor         maxExtent         tickRound        titleLineHeight   
domainOpacity      labelExpr          minExtent         tickSize         titleOpacity      
domainWidth        labelFlush         offset            tickWidth        titlePadding      
format             labelFlushOffset   orient            ticks            titleX            
formatType         labelFont          position          title            titleY            
grid               labelFontSize      style             titleAlign       translate         
gridCap            labelFontStyle     tickBand          titleAnchor      values            
gridColor          labelFontWeight    tickCap           titleAngle       zindex            
gridDash           labelLimit         tickColor                                            

See the help for altair.Axis to read the full description of these parameters
```

I did some basic manual testing with all the encoding channels listed in https://altair-viz.github.io/user_guide/encoding.html#encoding-channels and it seems to work fine. Although I am not sure if there are edge cases, I hope that the fallback to the old method could guard again those, but let me know if you think of anything specific and I can try to add tests for that.

joelostblom:
I just noted that this fails when there is a condition in the spec. For example, the following spec:

```py
import altair as alt

source = data.barley()

    x='variety',
    y=alt.Y('sum(yield)', stack='asdf'),
    opacity=alt.condition('datum.yield > 0', alt.value(1), alt.value(0.2))
)
```

returns:

```

altair.vegalite.v5.schema.channels.OpacityValue, validating 'additionalProperties'

altair.OpacityValue has no parameter named 'test'

Existing parameter names are:
value   condition   

See the help for altair.OpacityValue to read the full description of these parameters
```

If we change to `opacity=alt.value(0.6)` we get the expected output:

```
            altair.vegalite.v5.schema.channels.Y->0, validating 'enum'
            'asdf' is not one of ['zero', 'center', 'normalize']
```

mattijn:
Hi @joelostblom can you synchronise your branch, https://github.com/joelostblom/altair/tree/schema-validation-error, with the main repo and maybe add a code-snippet how to this can be tested easily using eg. colab. This will help the review process. Hopefully @jakevdp is still willing to do this.

joelostblom:
@mattijn I rebased this on the latest main branch and also structured the code so that it is easier to review. Do you think you would be able to have a look at this? I think it could be quite useful to get feedback on in the RC. I have included the rationale and an example below:

When a non-existing parameter name is used, I think it would be helpful to include the existing parameter names in the error message. Currently, when misspelling a parameter name like in the example below, it is not clear whether I made a typo or the parameter doesn't exist at all, so the immediate feedback of seeing the correct parameter names would be helpful.

```python
import altair as alt
import pandas as pd

source = pd.DataFrame({
    'b': [28, 55, 43, 91, 81, 53, 19, 87, 52]
})

    x=alt.X("a", scale=alt.Scale(padingOuter=0.5)),
    y=alt.Y("b")
)
```

```

        altair.vegalite.v4.schema.core.Scale, validating 'additionalProperties'

        Additional properties are not allowed ('padingOuter' was unexpected)
```

I also belive the error message text could be clarified a little. Especially for novices, I don't think it is immediately clear what it means that "Additional properties are not allowed". This PR updates the error message to instead read:

```

altair.vegalite.v5.schema.core.Scale, validating 'additionalProperties'

altair.Scale has no parameter named 'padingOuter'

Existing parameter names are:
align      domain      interpolate    range      round    
base       domainMax   nice           rangeMax   scheme   
bins       domainMid   padding        rangeMin   type     
clamp      domainMin   paddingInner   reverse    zero     
constant   exponent    paddingOuter                       

See the help for altair.Scale to read the full description of these parameters
```

It is possible that we could even remove the two first lines and start with `altair.Scale has no...` as I don't think the two first lines add any info that helps the user solve the error, what do you think?

This PR also includes a fallback to the old method of displaying error messages, so that in case the assumption in the if-statement does not hold true, things will not break and the user will just see a less helpful error message. In my testing, I have never seen this actually happen, but it is there just in case.

binste:
This looks super helpful! Haven't reviewed the code but just a quick reminder that the changes need to be moved to `tools/schemapi/schemapi.py`.

joelostblom:
Thanks @binste, updated!

binste:
Just a note that it might make sense to merge this PR after #2842 and test it again. I'd expect that your changes work in even more cases as the correct `additionalProperties` error is raised more often, but just to be sure.

tools/schemapi/schemapi.py
```diff
@@ -170,14 +225,41 @@ def __str__(self):
...
             )
-        return """Invalid specification
-
-        {}, validating {!r}
 
-        {}
-        """.format(
-            schema_path, self.validator, message
-        )
+        if hasattr(vegalite, schema_path.split(".")[-1]):
```
joelostblom, line 229: @binste I noticed an issue here after #2842 was merged. Previously, error messages would include a reference to the class where the error was raised, as in this case with the Y channel:

```
altair.vegalite.v5.schema.channels.Y, validating 'additionalProperties'
```

After #2842 the first part of that error message changes to no longer point to the particular class where the error was encountered, but just the chart in general:

```
altair.vegalite.v5.api.Chart->0, validating 'additionalProperties'
```

In addition to being less informative in general, this becomes an issue here because this PR was relying on that first part of the error message to find all the correct parameters names that the Y channel accept. There might be ways to work around that in this PR, but I wanted to ask you first if you think it is possible to restore the more precise error message that points to the class where the error was encountered or if that information is lost with the new approach to raising errors.

joelostblom: To add some more info, previously with deep validation, what I think was happening was that the iterative finding of the error meant that the class of the SchemaBase (I think) would change in each iteration to reflect which part of the spec was checked for errors. For example if I print out `type(self)` of the `self` that is passed to the `SchemaValidationError` it would look like this with deep validation (for a spec with a typo in a param in the Y channel):

```
<class 'altair.vegalite.v5.api.Chart'>
<class 'altair.vegalite.v5.schema.channels.Y'>
```
Without deep validation, all the errors are found without iteration and therefor the `self` param remains as the chart itself without changing to the Y channel where the error is occurring:

```
<class 'altair.vegalite.v5.api.Chart'>
```

I tried to look briefly if there was a way to not only return all the error messages, but also all the classes so that we could still find out that in this case it was the Y class that the error pertained to.

I looked briefly and it seems like some of this information is stored in the `absolute_path` attribute of the `self` object inside the `SchemaValidationError` class, and it might be possible to reconstruct it from there, but I couldn't find the full class path anywhere and I wanted to hear your thoughts before going down that path.

binste: Thanks for the detailed write-up! Indeed, I agree that it would be more informative if that original behaviour is restored and of course helpful for your proposed changes here, which look very helpful.

I think `absolute_path` is the correct place to take this from. From that name, we can try to use the usual conversion to a Python name that is used in the propertysetter https://github.com/altair-viz/altair/blob/master/altair/utils/schemapi.py#L680 and also in the code generation https://github.com/altair-viz/altair/blob/master/tools/generate_schema_wrapper.py#L454
```python
classname = prop[0].upper() + prop[1:]
```
and then fetch that class from `vegalite`. So in `SchemaValidationError.__str__` instead of `cls = self.obj.__class__` we could do something like

```python
class_name_lower = self.absolute_path[-1]
class_name = f"{class_name_lower}"[0].upper() + f"{class_name_lower}"[1:]
cls = getattr(vegalite, class_name)
cls
```
and before that in `SchemaValidationError.__init__`:

```python
self.absolute_path = err.absolute_path
```

I haven't tested this yet but most classes should be importable from the top-level like this right? For others we could still fall back on `self.obj` (the `self` that is passed into `SchemaValidationError`).

If I find more time in the next days I can make a PR with this to restore the original behaviour of showing the class in which the error occured.

binste: See #2883

joelostblom: Thank you! I merged in your PR and this seems to be working as intended now. There are a few failing tests which needs the error message updated; I think I can get to those on Friday or Sunday this week. There seems to be some larger error also where my solution here still needs some change to work with the new multiple returned error messages.

altair/utils/schemapi.py
```diff
@@ -181,19 +240,53 @@ def __str__(self):
             for val in schema_path[:-1]
             if val not in (0, "properties", "additionalProperties", "patternProperties")
         )
-        message = self.message
-        if self._additional_errors:
-            message += "\n        " + "\n        ".join(
-                [e.message for e in self._additional_errors]
-            )
-        return """Invalid specification
```
joelostblom, line 189: I moved this down because I don't think there will ever be multiple lines in the error message if it contains an unknown parmater name which is what the next if condition tests.

altair/utils/schemapi.py
```diff
@@ -181,19 +240,53 @@ def __str__(self):
...
 
-        {}, validating {!r}
+        # Output all existing parameters when an unknown parameter is specified
+        if (
+            hasattr(vegalite, schema_path.split(".")[-1])
+            and self.validator == "additionalProperties"
```
joelostblom, line 247: I had to add this line to restrict this to only activate for unknown parameter names and not also unknown values to existing parameter names.

altair/utils/schemapi.py
```diff
@@ -181,19 +240,53 @@ def __str__(self):
...
+        ):
+            altair_class = schema_path.split(".")[-1]
```
joelostblom, line 249: I shortened this to remove the `altair` prefix in the error message and just say e.g. <code>\`X\`</code> instead of `altair.X`

tests/utils/tests/test_schemapi.py
```diff
@@ -377,9 +378,7 @@ def test_schema_validation_error():
     assert isinstance(the_err, SchemaValidationError)
     message = str(the_err)
 
-    assert message.startswith("Invalid specification")
-    assert "test_schemapi.MySchema->a" in message
-    assert "validating {!r}".format(the_err.validator) in message
```
joelostblom, line 382: This was all part of the redundant info that is not removed from the error message. Maybe it it redundant to have both the lines below actually.

joelostblom (commented):
@binste I have updated this PR so that it works with the new error handling and also only shows the names of the existing parameters when an unknown parameter was specified and not for other types of errors. The tests you added previously was very helpful so thanks for that!

## Redundant info removed

I have also removed some redundant info from the error message that I don't think is helpful for the user. As an example, take the following spec:

```python
import altair as alt
import pandas as pd

source = pd.DataFrame({
    'b': [28, 55, 43, 91, 81, 53, 19, 87, 52]
})

    x=alt.X("a", scale=alt.Scale(padingOuter=0.5)),
    y=alt.Y("b")
)
```

It now outputs an error that get straight to the point of what is incorrect:

```
SchemaValidationError: `Scale` has no parameter named 'padingOuter'

Existing parameter names are:
align      domain      interpolate    range      round    
base       domainMax   nice           rangeMax   scheme   
bins       domainMid   padding        rangeMin   type     
clamp      domainMin   paddingInner   reverse    zero     
constant   exponent    paddingOuter                       

See the help for `Scale` to read the full description of these parameters
```

Previously it included "invalid specification" and a mention of attributes in the json schema, neither of which I think adds any info that makes it easier to troubleshoot this error, just more text to read:

```

altair.vegalite.v5.schema.core.Scale, validating 'additionalProperties'

altair.Scale has no parameter named 'padingOuter'

Existing parameter names are:
align      domain      interpolate    range      round    
base       domainMax   nice           rangeMax   scheme   
bins       domainMid   padding        rangeMin   type     
clamp      domainMin   paddingInner   reverse    zero     
constant   exponent    paddingOuter                       

See the help for altair.Scale to read the full description of these parameters
```

## Summary line added

In addition to removing some redundant info, I have added a summary line when an invalid value is passed to a correct parameter name as can be seen in the following spec:

```python
source = data.cars()
(
    alt.Chart(source)
    .mark_text(align="right")
    .encode(alt.Text("Horsepower:N", title=dict(text="Horsepower", align="right")))
)
```

which now outputs this error:

```
SchemaValidationError: '{'text': 'Horsepower', 'align': 'right'}' is an invalid value for `title`:

{'text': 'Horsepower', 'align': 'right'} is not of type 'string'
{'text': 'Horsepower', 'align': 'right'} is not of type 'array'
```

instead of this more verbose, but less informative error:

```

        altair.vegalite.v5.schema.core.TitleParams, validating 'type'

        {'text': 'Horsepower', 'align': 'right'} is not of type 'string'
        {'text': 'Horsepower', 'align': 'right'} is not of type 'array'
```

I find that this is also helpful when there are multiple errors spit out as in the example above and this example (I know you were also thinking about another solution here @binste so let me know if you prefer that instead):

```python
alt.Chart(data.barley()).mark_bar().encode(
    x=alt.X('variety'),
    y=alt.Y('sum(yield)', stack='asdf'),
)
```

which now shows the error:

```
SchemaValidationError: 'asdf' is an invalid value for `stack`:

'asdf' is not one of ['zero', 'center', 'normalize']
'asdf' is not of type 'null'
'asdf' is not of type 'boolean'
```

instead of:

```

        altair.vegalite.v5.api.Chart->0, validating 'enum'

        'asdf' is not one of ['zero', 'center', 'normalize']
        'asdf' is not of type 'null'
        'asdf' is not of type 'boolean'
```

---

## Invalid encoding channels

For specs that include an invalid encoding channel, there is no summary line but the error message is still shortened to remove redundant info:


```python
selection = alt.selection_point()
(
    alt.Chart(data.barley())
    .mark_circle()
    .add_params(selection)
    .encode(
        color=alt.condition(selection, alt.value("red"), alt.value("green")),
        invalidChannel=None,
    )
)
```

which now shows:

```
SchemaValidationError: Additional properties are not allowed ('invalidChannel' was unexpected)
```

instead of:


```

        altair.vegalite.v5.schema.core.Encoding->encoding, validating 'additionalProperties'

        Additional properties are not allowed ('invalidChannel' was unexpected)
```

altair/utils/schemapi.py
binste, line 249:
```suggestion
```

altair/utils/schemapi.py
```diff
@@ -181,19 +240,53 @@ def __str__(self):
...
+            vegalite_core_class = getattr(vegalite, schema_path.split(".")[-1])
```
binste, line 250:
```suggestion
```

altair/utils/schemapi.py
```diff
@@ -181,19 +240,53 @@ def __str__(self):
...
+            param_dict_keys = inspect.signature(vegalite_core_class).parameters.keys()
```
binste, line 251:
```suggestion
            param_dict_keys = inspect.signature(cls).parameters.keys()
```

altair/utils/schemapi.py
```diff
@@ -181,19 +240,53 @@ def __str__(self):
...
+
+            # `cleandoc` removes multiline string indentation in the output
+            return inspect.cleandoc(
+                """`{}` has no parameter named {!r}
+
+                Existing parameter names are:
+                {}
+                See the help for `{}` to read the full description of these parameters
+                """.format(
+                    altair_class,
```
binste, line 262:
```suggestion
                    cls.__name__,
```

altair/utils/schemapi.py
```diff
@@ -181,19 +240,53 @@ def __str__(self):
...
+                    self.message.split("('")[-1].split("'")[0],
+                    param_names_table,
+                    altair_class,
```
binste, line 265:
```suggestion
                    cls.__name__,
```

altair/utils/schemapi.py
```diff
@@ -181,19 +240,53 @@ def __str__(self):
...
+                )
+            )
+        # Use the default error message for all other cases than unknown parameter errors
+        else:
+            message = self.message
+            # Add a summary line when parameters are passed an invalid value
+            # For example: "'asdf' is an invalid value for `stack`
+            if hasattr(vegalite, schema_path.split(".")[-1]) and self.absolute_path:
```
binste, line 273: Same reasoning as above
```suggestion
            if self.absolute_path:
```

tests/utils/tests/test_schemapi.py
```diff
@@ -467,36 +466,64 @@ def chart_example_invalid_y_option_value_with_condition():
...
+
+                {'text': 'Horsepower', 'align': 'right'} is not of type 'string'
+                {'text': 'Horsepower', 'align': 'right'} is not of type 'array'"""
+            ),
         ),
         (
             chart_example_invalid_channel_and_condition,
-            r"schema.core.Encoding->encoding.*"
-            + r"Additional properties are not allowed \('invalidChannel' was unexpected\)",
+            r"Additional properties are not allowed \('invalidChannel' was unexpected\)\n",
```
binste, line 526: With my suggested changes above, this error now also shows the table. I think previously the `if` statement in `schemapi` failed to add a table for this as `schema_path` did not end with the class name but instead with `->encoding...`.


```python
`Encoding` has no parameter named 'invalidChannel'

Existing parameter names are:
angle         key          order     strokeDash      tooltip   xOffset   
color         latitude     radius    strokeOpacity   url       y         
description   latitude2    radius2   strokeWidth     x         y2        
detail        longitude    shape     text            x2        yError    
fill          longitude2   size      theta           xError    yError2   
fillOpacity   opacity      stroke    theta2          xError2   yOffset   
href                                                                     

See the help for `Encoding` to read the full description of these parameters
```
This looks very helpful to me! Maybe it makes sense to replace `Encoding` in the first and last line with something a bit more informative on encoding channels but I think already the default as shown above is ok.

binste (changes requested):
I really like the proposed changes to the error messages! They are much more succinct and informative and the tables are awesome 🚀 Thank you also for the detailed comment, this made the review much easier for me. I only have some small suggestions.

altair/utils/schemapi.py
```diff
@@ -181,19 +240,53 @@ def __str__(self):
...
+                # The indentation here must match that of `cleandoc` below
+                message = f"""'{self.instance}' is an invalid value for `{self.absolute_path[-1]}`:
```
binste, line 275: Any idea why flake8 does not raise an error for this line? It is too long. Same for a few lines in the tests.

altair/utils/schemapi.py
mattijn: I've no idea if this is possible or desirable, but as this becomes something like this:
```
See the help for `Scale` to read the full description of these parameters
```
Could this also become a hyperlink to the docs? See also this [snippet](https://github.com/altair-viz/altair/pull/2610/files#diff-d110e8643c6feff37c22c1a5c7d213b8384177718b8e26edd8e1b8f2ed027a0aR545-R548) that I saw in https://github.com/altair-viz/altair/pull/2610:
```python
# Patch in a reference to the class this docstring is copied from,
# to generate a hyperlink
doclines = Obj.__doc__.splitlines()
doclines[0] = f"Refer to :class:`{Obj.__name__}`"
```

tests/utils/tests/test_schemapi.py
joelostblom: Sweet!

altair/utils/schemapi.py
joelostblom: Hmm it is possible to put in the full link, but there is not way to hide it behind a friendlier text since the error message only outputs raw unformatted text (I believe). We would have to dump out the full URL and it does get pretty long, e.g.  `https://altair-viz.github.io/user_guide/generated/core/altair.Encoding.html#altair.Encoding`.

I actually considered removing this line completely and I am still not sure how helpful it is. I added it to encourage people to read the docstring of the class inside their IDE, but it is not always helpful as the name someone might have used in their spec does not always match with the one in the error message. For example, if I do `.mark_point(asdf='fdsa')` the error message will point the user to `MarkDef` rather than `mark_point` which might be unclear for people new to altair. I am not sure if it is worth improving that though.

cc @binste for thoughts here as well

tests/utils/tests/test_schemapi.py
joelostblom: Yup these two lines were testing the same message so removing one.

altair/utils/schemapi.py
```diff
@@ -181,19 +240,53 @@ def __str__(self):
             for val in schema_path[:-1]
             if val not in (0, "properties", "additionalProperties", "patternProperties")
         )
```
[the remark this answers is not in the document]
joelostblom: Done!

altair/utils/schemapi.py
joelostblom: I htink both black and flake8 allow multiline strings are allowed to be longer than regular strings:

image

I would rather not break this line up to facilitate the formatting of the output.

altair/utils/schemapi.py
binste: SQLAlchemy somewhat solved the issue with long URLs with a URL shortener which they host at `sqlalche.me`. See https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/exc.py#L60 how they then craft the URL in the exception message. Don't know what service is behind it and how we could keep something like that up-to-date. Maybe there is a good Sphinx extension which helps? I'm also in favour of not including the long URLs and maybe exploring the URL shortener or other ideas in a future release.

No strong preference if that line should be dropped. I see the point that references to classes such as `MarkDef` might be more confusing then they help and implementing some special logic is maybe also one for a future release.

altair/utils/schemapi.py
binste: Good to know, thanks! Agree, let's keep it like it is.

binste (changes requested):
The changes in `test_schemapi.py` need to be moved into the respective `tools` script. This happens to me way too often! I opened #2906 with an idea on how we can test this.

joelostblom:
Good catch, thank you! I have updated to added the changes to the tools script as well

altair/utils/schemapi.py
joelostblom: I vote for punting this to the future and opened https://github.com/altair-viz/altair/issues/2910 with a reference to this discussion.

mattijn:
I think it is a really nice PR. Thanks @joelostblom. Really helpful, nice that it gives suggestions. It makes me want to make mistakes on purpose to read the suggestions..

I found one spec that gives an incorrect hint:
```python
import altair as alt
cars = data.cars.url

    x='Acceleration:Q',
    y='Horsepower:Q',
    color=alt.value(1)  # should be eg. alt.value('red')
)
```
```cmd
SchemaValidationError: `Color` has no parameter named 'value'

Existing parameter names are:

See the help for `Color` to read the full description of these parameters
```
It is allowed to use `alt.value()` in this context, but in this case it should be a string and not a number.

Also the context description is based on the jsonschema. So I'm still linking this comment https://github.com/altair-viz/altair/pull/2842#issuecomment-1400938604 here as well.
```python
alt.Chart(data.barley()).mark_bar().encode(
    x=alt.X('variety'),
    y=alt.Y('sum(yield)', stack='null'),  # should be eg. stack=None
)
```
```cmd
SchemaValidationError: 'null' is an invalid value for `stack`:

'null' is not one of ['zero', 'center', 'normalize']
'null' is not of type 'null'
'null' is not of type 'boolean'
```
I tried to follow the suggestion, but it was not directly clear that `type 'null'` actually is `None`.

No further comments. Congrats again on this PR!

mattijn (approved):

joelostblom:
Thank you @mattijn :smile: I will go ahead and merge this since you approved and the change that @binste suggested has also been implemented.

We can follow up on the remaining two issues separately. I noticed what you mentioned regarding alt.datum/value and it was introduced before this PR so I opened https://github.com/altair-viz/altair/issues/2913 to track it. I agree that it would be helpful to show the Python types as well and opened https://github.com/altair-viz/altair/issues/2914 for that.

PR #2575, opened 2022-03-25
MAINT: Update actions' links to use https

joelostblom:
These pip installs were added in #2528 for installing the dev version of altair viewer. As per https://stackoverflow.com/questions/70663523/the-unauthenticated-git-protocol-on-port-9418-is-no-longer-supported, git will only support secure links in its build files since recently


jakevdp (approved):
Thanks!

PR #2576, opened 2022-03-25
MAINT: Update examples and tests to VL5 syntax

joelostblom:
This fixes [all the warnings seen in the CI currently](https://github.com/altair-viz/altair/runs/5686331424?check_suite_focus=true). I will go through the docs and do the same in my other PR, but I thought it was better to keep this separate for clarity.

One note is that once these are updated they will no longer work as tests for the older APIs. Maybe this is OK since the older APIs won't see any changes (although it might need to stay compatible with changes in other packages), but if it is deemed that we still need the old examples around, we could create a `gallery-vl4.17` folder or something like that to use for test of older version.


joelostblom:
@mattijn Since we have been updating the examples quite a bit in the last few PRs, I rebased this one and think it is a good idea if we merge it soon so that it doesn't go out of sync too much with the main branch (even if some things might change later such as `add_parameter` -> `add_params`).

mattijn (approved):

PR #2582, opened 2022-03-28
MAINT: Update from v4 to v5 in some additional files

joelostblom:
I grepped the repo for `v4` and noticed a few places where I think it should be replaced with `v5`. I ran this locally with the updated tests in https://github.com/altair-viz/altair/pull/2576 and there are no failures introduced.

### DIFF
--- a/altair/_magics.py
+++ b/altair/_magics.py
@@ -13,6 +13,7 @@ from toolz import curried
 
 from altair.vegalite import v3 as vegalite_v3
 from altair.vegalite import v4 as vegalite_v4
+from altair.vegalite import v5 as vegalite_v5
 from altair.vega import v5 as vega_v5
 
 try:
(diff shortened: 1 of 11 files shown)


altair/_magics.py
```diff
@@ -76,12 +82,12 @@ def _get_variable(name):
     nargs="*",
     help="local variable name of a pandas DataFrame to be used as the dataset",
 )
-@magic_arguments.argument("-v", "--version", dest="version", default="5")
+@magic_arguments.argument("-v", "--version", dest="version", default="v5")
```
joelostblom, line 85: I made this change and the one below so that this will be easier to detect in the future. It was giving me issues initially and it is hard to grep the repo for just the version number without the `v` prefix.

joelostblom:
@mattijn Does this look OK to you? I think it is mostly cleanup from changing from v4 to v5.

mattijn (approved):

mattijn:
I only checked the code-diff and all looks fine.

PR #2586, opened 2022-04-02
FEAT: Include only referenced data columns in chart specs

### LINKED ISSUE #2428: Include only referenced columns in the data of each chart specification
The main goal of this suggestion is to reduce chart sizes, by including only the necessary data.

I read in https://github.com/altair-viz/altair/issues/1250#issuecomment-457318979 that similar initiatives were abandon previously because it would be too complex to determine which columns VegaLite needs from the Python side, especially for complex charts.

However, I noticed that [altair-transformer](https://github.com/altair-viz/altair-transform#example-pre-aggregating-large-datasets) already drops unused columns when preaggregating data. Maybe it is the same limitation that prevents it from working with compound charts etc, but the current functionality is already quite useful. Could it be added as a data transformer, so that it was not needed to be called on each chart separately?

If not, do you think there would room to add something less elegant that is not compatible with arbitrarily complex chart layouts, but still practical in many cases? I am naively imagining parsing the chart dictionary `'field'` keys and any dictionary string values containing `'datum.'` for column names, and filter the data with those.

---

Please follow these steps to make it more efficient to respond to your feature request.

- [x] Since Altair is a Python wrapper around the Vega-Lite visualization grammar, [most feature requests should be reported directly to Vega-Lite](https://github.com/vega/vega-lite/issues). You can click the Action Button of your Altair chart and "Open in Vega Editor" to create a reproducible Vega-Lite example.
- [x] Search for duplicate issues.
- [x] Describe the feature's goal, motivating use cases, and its expected behavior.

joelostblom:
This PR takes an inelegant and conservative approach to fix #2428. It converts the chart spec (without the data) to a string and removes each of the data field names that is not found in that string (i.e. the data fields not referenced anywhere in the spec). It will not work for short columns names such as 'x', or those that have the same name as a key in the VegaLite spec, e.g. 'field', but it is still likely useful for the vast majority of cases and can reduce the chart size significantly specification with many unused columns. For example, this chart

```python
import altair as alt

alt.data_transformers.disable_max_rows()
source = data.movies()

import sys
sys.getsizeof(alt.Chart(source).mark_rect().encode(
    alt.X('IMDB_Rating:Q', bin=alt.Bin(maxbins=60)),
    alt.Y('Rotten_Tomatoes_Rating:Q', bin=alt.Bin(maxbins=40)),
    alt.Color('count():Q', scale=alt.Scale(scheme='greenblue'))
).to_json())
```
has a size of 1873345 bytes because there is a lot of unused columns included in the data that are not in the spec:

```
{'data-d01e86a42596d3c55cbbce3d21a64616': [
   {'Title': 'The Land Girls',
   'US_Gross': 146083.0,
   'Worldwide_Gross': 146083.0,
   'US_DVD_Sales': None,
   'Production_Budget': 8000000.0,
   'Release_Date': 'Jun 12 1998',
   'MPAA_Rating': 'R',
   'Running_Time_min': None,
   'Distributor': 'Gramercy',
   'Source': None,
   'Major_Genre': None,
   'Creative_Type': None,
   'Director': None,
   'Rotten_Tomatoes_Rating': None,
   'IMDB_Rating': 6.1,
   'IMDB_Votes': 1071.0},
   ...
```

Creating the same chart with the changes in this PR reduces the chart size ~7x to 269838 bytes and successfully remove all the irrelevant columns since they have unique names not found anywhere in the spec:

```
{'data-d01e86a42596d3c55cbbce3d21a64616': [
  {'Rotten_Tomatoes_Rating': None, 'IMDB_Rating': 6.1},
  ...
```

This PR seems to pass all the tests, but I still want to run some benchmarks to see if the loops introduce any significant performance regressions for larger data sets.

I also need to move the change to the schema generation file instead of the automatically generated vegalite api file and this PR is not important as some of the other open PRs, but I started playing around with it and thought I would put this up in case anyone else wants to try or have better ideas of how to achieve the same results.

### DIFF
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -21,7 +21,9 @@ DataTransformerType = Callable
 
 
 class DataTransformerRegistry(PluginRegistry[DataTransformerType]):
-    _global_settings = {"consolidate_datasets": True}
+    _global_settings = {
+        "consolidate_datasets": True,
+    }
 
     @property
     def consolidate_datasets(self):
@@ -30,6 +32,40 @@ class DataTransformerRegistry(PluginRegistry[DataTransformerType]):
     @consolidate_datasets.setter
     def consolidate_datasets(self, value):
         self._global_settings["consolidate_datasets"] = value
+
+    def compress_datasets(self, enable=True, transformed_charts=False):
+        """Whether to remove data fields that do not occurr elsewhere in the chart spec
+
+        The approach taken here is conservative as it will
+        leave some of the data fields that should be removed because they
+        have a common or short name that happens to occur elsewhere in the
+        spec by coincidence.
+
+        Parameters
+        ----------
+
+        enable : bool
+            Whether compression is enabled.
+
+        transformed_charts : bool
+            Whether to enable compression for transformed charts.
+            Data compression can break charts with transforms where data
+            fields are referenced without the field name occurring
+            literally in the chart spec.
+        """
+        if transformed_charts:
+            warnings.warn(
+                "Data compression can break charts with transforms where data"
+                " fields are referenced without the field name occurring"
+                " literally in the chart spec. Please revert this option if you"
+                " are observing unexpected behavior with transformed charts."
+            )
+        self._compress_datasets = enable
+        self._compress_transformed_charts = transformed_charts
+        return
+
+    _compress_datasets = True
+    _compress_transformed_charts = False
 
 
 # ==============================================================================
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -568,6 +568,28 @@ class TopLevelMixin(mixins.ConfigMethodMixin):
             if context["datasets"]:
                 dct.setdefault("datasets", {}).update(context["datasets"])
 
+        # Remove every data field that does not occur anywhere else in the spec
+        # The approach taken here is inelegant and conservative as it will
+        # leave some of the data fields that should be removed because they
+        # have a common or short name that happens to occur elsewhere in the
+        # spec by coincidence.
+        # Removing data fields properly is difficult as mentioned here
+        # https://github.com/altair-viz/altair/issues/2428#issuecomment-798917267.
+        # This inelegant approach is still helpful and will reudce the chart
+        # size in many cases
+        if "datasets" in dct:
+            spec_as_string = [
+                val
+                for key, val in dct.items()
+                if key not in ["data", "datasets", "$schema"]
+            ].__str__()
+            for dataset_name in dct["datasets"]:
+                for dict_row in dct["datasets"][dataset_name]:
+                    # Must iterate over a list since the dict will change size
+                    for field in list(dict_row.keys()):
+                        if field not in spec_as_string:
+                            del dict_row[field]
+
         return dct
 
     def to_html(
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -103,6 +103,34 @@ def _prepare_data(data, context=None):
     return data
 
 
+def _compress_datasets(dct):
+    """Remove data fields that do not occurr elsewhere in the chart spec
+
+    The approach taken here is conservative as it will
+    leave some of the data fields that should be removed because they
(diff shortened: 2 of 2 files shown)


jakevdp:
I've been worried about adding this sort of thing because the failure mode in which we remove a column that's needed by the chart is something that can't easily be recovered from. If we land something like this, I think it needs an "off switch" so that Altair does not become completely unusable in the event of a corner case we haven't thought of.

joelostblom:
Do you mean adding something like `alt.data_transformers.disable_data_compression()`? But still having the "compression" in this PR (or similar) as the default behavior?

I was trying to think of a scenario where a column could be used in Vega-Lite without the column name being explicity referenced at least somewhere in the spec. Maybe @domoritz knows if this could happen (let me know if you prefer not to be pinged like this @domoritz)?

jakevdp:
One situation I can think of is column names with special characters that may or may not be escaped when referenced in the chart spec. You could also construct column names dynamically using a Vega expression, which is probably not common but would break with this approach.

jakevdp:
Here's an example that would break with this PR, I believe:
```python
import altair as alt
import pandas as pd

df = pd.DataFrame({'somecolumn': [1, 2, 3, 4]})

alt.Chart(df).transform_calculate(
    data='datum["some" + "column"]'
).mark_point().encode(
    x='data:Q'
)
```

domoritz:
Another case could be expressions that reference the keys of a data tuple. Then we should not project either. 

In general, I think projecting data to only what's really needed is great but we need to be careful to provide an escape hatch and that we consider common hard cases like the ones we already identified.

joelostblom:
Thanks both! I will add a toggle switch and try to account for the
scenarios raised here.

 @domoritz do you have an example of what a spec would look like for the
issue you brought up? I can't quite visualize it myself.

jakevdp:
Even with an escape hatch, I’m not sure we’d want to land this change. Imagine the headache of trying to write some non-trivial Vega expression and finding after an hour that Altair is silently removing the column you’re trying to reference. Unless we can make this account for all cases, we probably shouldn’t land it. And accounting for all cases would essentially require re-implementing Vega in Python.

joelostblom:
Hmmm yeah that would indeed be a frustrating situation... But I also have a
hard time letting go of that the vast majority of use cases would benefit
from an addition like this. What do you think of making it opt-in instead
of opt-out? We could add the option
`alt.data_transformers.enable_data_compression()` to *enable* this behavior
and when doing so a warning/info message about edge cases could be
displayed?

jakevdp:
Sure, opt-in would be better, but then 99% of users would never know the feature exists and never benefit from it. I think better would be to avoid the easy broken approach and try for the full correct approach... it would basically require rolling the `altair_transform` package into the main `altair` library, expanding its completeness, and utilizing that.

A simpler option might be to only enable this only in the case that the chart contains no transformations, which I think are the only cases that create problems.

joelostblom:
I agree about the problems with discoverability. I like your suggestion of a conservative default option to only enable this compression/projection for specs that do not include transforms. This will benefit many users and we could add an optional flag to include charts with transforms and raise a warning message if this is enabled:

```python
alt.data_transformers.compress_datasets(transformed_charts=True)
```

> ```
> UserWarning: Data compression can break charts with transforms where data fields are referenced
> without the field name occurring literally in the chart spec. Please revert this option if you are
> observing unexpected behavior with transformed charts.
> ```

I updated the draft PR with a suggestion of what this could look like. I also agree that the altair_transform options sounds like the best long term solution, and this PR could be valuable in the meantime without breaking changes when/if there is a switch to the altair_transform option in the backend.

binste:
For reference, removing unused columns is soon possible in a new version of VegaFusion, see discussions in #2738.

joelostblom:
Ah that is great, thanks for mentioning that! We can probably close this PR once we have played around with VegaFusion a bit more then. There would probably still be value in having a simple approach in altair by default, but unless we can solve the issues mentioned above, it might be unnecessarily complicated to have something in altair that works only if there are no transforms instead of just referring to the VegaFusion option for all chart size reduction needs.

mattijn:
Let's leave this to VegaFusion if this a concern for anyone.

PR #2592, opened 2022-04-16
Add method based attribute setting to encoding classes for Altair 5

joelostblom:
I have been busy lately and not made progress beyond moving Jake's changes from #1629 into the right places of the current development branch. They seem to still be working fine from my local testing, but I will not have much time to work on this until June/July so anyone who wants to take over before then please do! I would love to see this syntax in Altair!

I think the biggest outstanding item is the one mention in https://github.com/altair-viz/altair/pull/1629#issuecomment-1100770312. So either figuring out a workaround for that or making progress on these smaller outstanding changes would be helpful?

1. [x] Add docstrings.
1. [ ] (started) Add keyword argument completion for these methods and their parameter names.
2. [ ] (started) Add tests ensuring equivalence of old and new syntax.
    - Maybe we can can create all relevant the gallery plots with both the old and new syntax?
3. [ ] Add useful documentation
    - For this I think it would be great with a dedicated section mentioning the method syntax explicitly and then [use Spinx Tabs](https://sphinx-tabs.readthedocs.io/en/latest/#code-tabs) to display both the old and new syntax throughout the docs and for the example gallery.
    - After more thought, I think maybe a dedicated section is enough and then promoting the new more terse syntax everywhere else (unless we discover some clear downside to using the new syntax)

Working example:

```python
import altair as alt


source = data.cars.url

    alt.X('Horsepower:Q').scale(zero=False),
    alt.Y('Miles_per_Gallon:Q').scale(zero=False).title('Miles / Gallon'),
    alt.Color('Origin:N').title('').legend(orient='bottom').scale(scheme='Set2')
)
```

Maybe it's nicer to format it like this:

```python
    alt.X('Horsepower:Q')
        .scale(zero=False),
    alt.Y('Miles_per_Gallon:Q')
        .scale(zero=False)
        .title('Miles / Gallon'),
    alt.Color('Origin:N')
        .title('')
        .legend(orient='bottom')
        .scale(scheme='Set2')
)
```

image


mattijn:
I merely think people will combine the methods as such

```python
!python -m pip install git+https://github.com/joelostblom/altair.git@method-based-attr-setting

import altair as alt

source = data.cars.url

s1 = alt.Scale(zero=False)
s2 = alt.Scale(scheme='viridis')

    x=alt.X('Horsepower:Q').axis(labels=False).scale(s1).title('foo'),
    color=alt.Color('Origin:N', scale=s2).legend(orient='bottom')
)
chart
```
Observe on the `x` encoding, I use the `.scale(s1)` and on the `color` encoding I use the `scale=s2`. It still works though.

Personally I think this deep nested setting and getting of attributes is neither a show-stopper nor a big issue. I do not think it is common usage by normal users. What I did do in the past is when a newer version of VL can render something which cannot yet pass validation in altair that I compile the chart into a VL-dict and then change the value of the corresponding key. Something as such:

```python
ctd = chart.to_dict()
ctd['encoding']['x']['title'] = {'signal': 'bah'}
```
But it felt like a hack and not following the zen of altair, but again, this will still works.

I think the interesting outstanding issue is the part how we can solve this keyword argument completion for these methods.

joelostblom:
I made some progress on the completions in my latest commit! The help now shows the correct signature and docstring:

image
(there are some unnecessary fields here that should be removed somehow, `type`, `file`, etc)
image

The tab completion works for the class (the suboptimal ordering is an issue with jlab/ipython, but the `maxbins` param shows up):

image

but not for instances of the class:

image

[I asked about this on SO,](https://stackoverflow.com/questions/72223285/method-parameter-auto-completion-for-a-class-instance) but if anyone here has ideas, please chime in.

One last issue I am running into is that completion for the method name also only works for the class, but not for instances of the class. So typing `alt.X.b` and then pressing <kbd>Tab</kbd> completes to `bin` but there is no completion on `alt.X().b` (this works in a toy example so I think it is an altair specific issue)

ChristopherDavisUCI:
Hi @joelostblom, is there anything I can do to help with this PR?  I think it would be great if this syntax can be used after the next major release.

joelostblom:
Thank you! It would be awesome if you want to look into getting the tab completion to work properly! Both the method name and the parameters inside the method are currently not showing up in the tab completion. I gave a detailed example on exactly what works and what doesn't in my comment just above and I have not worked on it more since then. I think that is the only technical solution left to fix (then there is this potential conceptual challenge https://github.com/altair-viz/altair/pull/1629#issuecomment-1100770312 and we need to add some tests and docs, but that's later). Let me know if it is unclear and I will try to clarify.

ChristopherDavisUCI:
Sorry for my git ignorance @joelostblom... should I be able to switch between my own branches of Altair and yours here: https://github.com/joelostblom/altair/tree/method-based-attr-setting
I haven't been able to figure out how to reach your `method-based-attr-setting` on my own computer.  Should I just create a whole new folder on my computer for your branch?

All but the basics of git tend to leave me confused...

joelostblom:
No worries Chris, you should be able to add to my branch if you follow the steps in this article https://tighten.com/blog/adding-commits-to-a-pull-request/. You could also do what you suggested and create a new folder and clone my fork of the repo there. Since you don't have the "maintainer" permissions on Altair, I went ahead and invited you to my fork separately, so you should have no issues pushing directly to this branch.

If you run into issues with this, I am also more than happy for you to copy over these changes manually and open a new PR, I don't want you to be held up by git issues

ChristopherDavisUCI:
Thanks a lot Joel, that worked!  Essentially everything related to git takes two more steps than I expect.  In this case, I think I was trying to go directly to the `git checkout` step, and was missing the `git remote add` step and was also missing the `git fetch` step.

ChristopherDavisUCI:
Hi @joelostblom, I'm trying to get a sense for the issues you mentioned https://github.com/altair-viz/altair/pull/2592#issuecomment-1125544599

Does the following count as tab-completion working for an instance?  (I didn't make any changes to the code.)  I don't really see how this is substantively different from the non-working example `alt.X().bin(ma...)`.

tab

joelostblom:
Interesting example, I didn't know that would work. Maybe it has something to do with the function only being called when the cell is executed and before that tab completion can't inspect further? But it must be be possible to bypass that because tab completion is working for eg `alt.Chart().<tab>` and I think also for `alt.Chart().mark_point(col<tab>` (on my phone right now and can't double check). 

We would ideally want it working for `alt.X(). bin(ma<tab>` and `alt.X().b<tab>`.

ChristopherDavisUCI:
Hi @joelostblom, I've browsed around a little but haven't made any real progress.  I tried using the Python console, and also there `alt.X.bi` will tab-complete to `alt.X.bin` but `alt.X().bi` will not tab-complete.  So it's at least not some special functionality related to Jupyter Lab or VS Code.

On the other hand, in Jupyter Lab, `alt.Chart().mark_point(si` will tab-complete to `alt.Chart().mark_point(size=`, but `alt.Chart().mark_point(si` won't tab-complete in the Python console.

One non-Altair example, in Jupyter Lab, `pd.DataFrame().max(ax` will tab-complete to `pd.DataFrame().max(axis=`, but `pd.DataFrame().max(ax` won't tab-complete in the Python console.

I'm not sure what the upshot is to that, just thought I would write it down!

joelostblom:
Thanks again for working on this Chris! I believe that the special tab completion  functionality you are referring to is inherent to IPython, so if you try in the IPython console rather than the regular one you should see similar completion to that of vs code and juputerlab (although the order of the suggestions might differ).

ChristopherDavisUCI:
Hi @joelostblom, thanks for the reply!  Just to clarify, the tab-completions in the console I was talking about above are just in the regular Python console (like what I get by typing "Python" into terminal).  In that Python console, `alt.X.bi` will tab-complete, but `alt.X().bi` won't tab-complete.

I will keep looking around, but please keep me updated on any ideas!  I haven't thought about this kind of issue before.

ChristopherDavisUCI:
Hi @joelostblom, do you think the tab-completion issues are related to `__getattr__` and `__dir__` being defined explicitly in https://github.com/altair-viz/altair/blob/bff48d5a235363a74df881c0d167206a9bbb45e1/altair/utils/schemapi.py#L255-L260 and in https://github.com/altair-viz/altair/blob/bff48d5a235363a74df881c0d167206a9bbb45e1/altair/utils/schemapi.py#L474-L475

I tried making a toy example with (as far as I can tell) similar logic and the behavior seemed similar.  I put the following in a separate file named `sample.py`:
```
class A:
    def __init__(self):
        self.x = 2
        
    def __getattr__(self, attr):
        if attr in ["y", "z"]:
            return 5
    
    def __dir__(self):
        return ["y", "z"]
```

and then tried tab-completion in Jupyter Lab:
comp1
and
comp2

To me this feels similar to the behavior we saw above.

In this toy example, if we wanted `A().` to have `y` and `z` offered as completions, is it clear how we should proceed?

joelostblom:
Hmm, one thing that seems different is that in your example `A.<tab>` does not complete whereas in our original example, we can see that `alt.X.bi<tab>` does complete. So I am not sure if it is exactly the same, and I am not sure what would be the best way to proceed on the top of my head unfortunately.

ChristopherDavisUCI:
Hi @joelostblom,

I haven't had any success getting the auto-completion to work so far.

Do you see any potential in taking a slightly different approach like the following?

I tried adding a few sample methods to the `FieldChannelMixin` class:
```
    def set_bin(self, *args, **kwargs):
        copy = self.copy()
        copy.bin = core.BinParams(*args, **kwargs)
        return copy

    def set_scale(self, *args, **kwargs):
        copy = self.copy()
        copy.scale = core.Scale(*args, **kwargs)
        return copy

    def set_title(self, *args, **kwargs):
        copy = self.copy()
        copy.title = core.Text(*args, **kwargs)
        return copy

    def set_legend(self, *args, **kwargs):
        copy = self.copy()
        copy.legend = core.Legend(*args, **kwargs)
        return copy
```

In my simple tests that seemed to work well.  For example, the following code works (based on your code above):
ss1

And auto-completion seems to behave better:
ss2
and
ss3

It seems to me like our lives would be a lot easier if we used a syntax like `bin` for the attribute and `set_bin` for the attribute setter method.

If it doesn't make sense to use the same definitions for everything that is a `FieldChannelMixin`, the definitions could be made more locally.

I also wasn't sure if specifying something like `title = core.Text` was too restrictive for this use.

Do you have any immediate comments on that type of approach?  I wanted to check in before trying to work on it.  @jakevdp and @mattijn I'd also be very happy to hear your thoughts.

joelostblom:
Hmmm I don't have a good sense of what a solution here might be, but in general I would prefer working with a `.bin` method rather than a `.set_bin` method. [Jake's older comment here](https://github.com/altair-viz/altair/pull/1629#issuecomment-524894409) seems to suggest that he is/was of similar opinion:

> I suppose `set_scale()` makes sense, but I find explicit getters & setters like this to be a bit cluttery.

Could a way forward be to break this down to a minimal reproducible example outside of altair and try to troubleshoot it that way? We could even post it to stackoverflow and get some input. [I tried this earlier](https://stackoverflow.com/questions/72223285/method-parameter-auto-completion-for-a-class-instance) but ran into a separate issue that was specific to JupyterLab and didn't have time to rework the example. I don't know if we can include the specifics of this problem while still making it small enough for SO, but it might still be a useful way forward for us to troubleshoot?

I will also say that personally I believe that even if we don't have tab completion, the fact that the docstring help popup is now working properly makes this PR worth merging, but it would still be very nice to get the tab completion working as well.

mattijn:
@joelostblom can you synchronise this PR with the main repo? 
I feel the same, the current setter methods using decorators make more sense over explicit `set_` functions. Also agreed that we can work on tab completion in a follow-up PR if we feel this PR is worth merging already. Tests and documentation are still missing I think?

ChristopherDavisUCI:
Thanks for the comments!  I am happy with whatever approach you think is best.  I did come across this [comment by Jake](https://github.com/altair-viz/altair/pull/2517#issuecomment-978455614) from almost exactly one year ago that made me wonder about alternatives to redefining `bin`, `scale`, etc as functions.

I'm happy to work on producing a minimal example of the failure of tab-completion, although my initial attempts make me think that whatever we produce will be too complicated and will get downvoted on Stack Overflow ;)

joelostblom:
> can you synchronise this PR with the main repo?

Done! Tests seem to be failing because of an issue installing chromedriver on the test image?

> Tests and documentation are still missing I think?

Yup, do you think it would be enough to create a version of each of the gallery examples with this new syntax and use those as tests? Or should we create some new types of tests?

> I did come across this https://github.com/altair-viz/altair/pull/2517#issuecomment-978455614 from almost exactly one year ago that made me wonder about alternatives to redefining bin, scale, etc as functions.

I was thinking about that too, but it does indeed seem like a bigger change. Maybe it is better though because it will make it really clear when we are trying to access and attributes value versus trying to call a function?

> By the way, I don't think the title portion of the very top example above is working, or at least it didn't work on my end when I tried

I can confirm that this didn't work for me either. There are other parameters such as `stack` that don't work either so maybe this is expected? But if I recall correctly it was working before to do `.title()`, maybe I'm misremembering.

mattijn:
I re-run the tests and these are fine now based on current defined tests.

Basically currently only variables that can be declared as `={}` in the encoding channel will function:
```python
foo = alt.Chart().mark_point().encode(
    alt.X(scale={})
).to_dict()

bah = alt.Chart().mark_point().encode(
    alt.X().scale()
).to_dict()

foo == bah  # True
```

But variables that only can be declared as `str` will not work:
```python
    alt.X().shorthand('field:Q')
).to_dict()
```
Gives:
```cmd
TypeError: 'UndefinedType' object is not callable
```

I would suggest to have a test suite that shows which variables work, but also which variables are (as expected) not working.
```python
with pytest.raises(TypeError):
        alt.X().shorthand('field:Q')
    ).to_dict()
```

 Maybe it can be done in combination with `@pytest.mark.parametrize()`

I can understand why this results in a `TypeError` (not clear error though) but it can be confusing too. Maybe its sensible for `shorthand`, but for variables such as `title` and `stacked` its not strange thinking to define it as the examples that in earlier stages of this PR did seem to work (wrongly?).

ChristopherDavisUCI:
Hi @mattijn, I didn't totally understand the `str` vs `={}` distinction you mentioned.  Does that explain which of these functions are working and which don't?  Do you see what the issue is?  Here are the ones that seem to be working/not working for `alt.X()`:

```
import altair as alt

schema = alt.X().resolve_references()

for k in schema['properties'].keys():
    print(k)
    print(getattr(alt.X(),k))
    print()
```
produces the following, so I guess `bandPosition`, `impute`, etc are not working at the moment.
```
aggregate
<altair.utils.schemapi._PropertySetter object at 0x163541000>

axis
<altair.utils.schemapi._PropertySetter object at 0x16358cf70>

bandPosition
Undefined

bin
<altair.utils.schemapi._PropertySetter object at 0x16358cc40>

field
<altair.utils.schemapi._PropertySetter object at 0x16358ee90>

impute
Undefined

scale
<altair.utils.schemapi._PropertySetter object at 0x16358ea10>

sort
<altair.utils.schemapi._PropertySetter object at 0x16358f2b0>

stack
Undefined

timeUnit
Undefined

title
Undefined

type
<altair.utils.schemapi._PropertySetter object at 0x1635a87c0>
```

mattijn:
It seems that currently the attribute setter is enabled only for properties in the encoding channel that can contain an object. Strange though, that `impute` is `Undefined` since it clearly is possible to define an object as well for impute in VL: https://vega.github.io/vega-lite/docs/impute.html#encoding.

ChristopherDavisUCI:
I'm still a little apprehensive about having something like `alt.X().axis` evaluate to something different from the method `alt.X().axis()`.  Overall that just seems kind of confusing to me.  What would you think about eventually removing this second `alt.X().axis` value, so that `alt.X().axis` returns the function that's called when we use `alt.X().axis()`?

But I'm not at all confident in my opinion and am happy to hear that having both is good and not unusual.

joelostblom:
Yeah, I agree that it can be confusing. I am not against having something like what is suggested in https://github.com/altair-viz/altair/pull/2517#issuecomment-978455614. Not sure how much breakage this would cause for people's workflows though. I mostly use `chart.to_dict()` currently myself because of https://github.com/altair-viz/altair/issues/2497

ChristopherDavisUCI:
Thanks @joelostblom!  It's interesting reading [comment](https://github.com/altair-viz/altair/pull/2517#issuecomment-978455614) again, because that comment seems to almost be the opposite of what I thought was going on in this PR.  I had thought that in the end, this PR retained the `x.field` returns `"x"` functionality.  I guess I should keep staring at it.

ChristopherDavisUCI:
Hi @joelostblom, I think I tracked down what the issue is with `impute` and some of the other methods we're trying to define.  I think the problem is this line: https://github.com/altair-viz/altair/blob/9fa0c387fdcf31900015ac87d52ede68e30832b6/altair/utils/schemapi.py#L605
There's no class named exactly `Impute` (rather there is `ImputeParams` and a few others).

How about this for a temporary fix.  We replace `_PropertySetter.__get__` with the following:
```
    def __get__(self, obj, cls):
        self.obj = obj
        self.cls = cls
        try:
            altair_prop = getattr(vegalite, f"{self.prop}".capitalize())
            self.__doc__ = altair_prop.__doc__
            self.__signature__ = inspect.signature(altair_prop)
            self.__wrapped__ = inspect.getfullargspec(altair_prop)
            self.__name__ = altair_prop.__name__
        except AttributeError:
            pass
        return self
```

What do you think?

joelostblom:
Ah ok, that makes sense, great find! What do you think about making it a conditional statement instead of try/except? The we can be more explicit about what we're checking for, and we could have a three scenarios:

1. Check if the name exists then use as is.
2. Check if the name exists with "params" appended, then use that.
3. Pass through in case there are additional cases. Or maybe raise a warning for now so that we can detect them easier?

ChristopherDavisUCI:
Sounds great to me!  I feel like this code runs when the class is instantiated, not when the method is called, so raising a warning doesn't seem like the right thing to do to me.  But maybe I'm misremembering the tests I did.  I would just pass through in case 3.

There is this code which makes the `bin` method already work:
https://github.com/joelostblom/altair/blob/9fa0c387fdcf31900015ac87d52ede68e30832b6/altair/vegalite/v5/api.py#L108-L109
I don't know if that makes you rethink the strategy at all, but to me your strategy sounds good.

joelostblom:
Oh cool that alias is a nice find. I
To be honest, it always bothers me that I have to write out TitleParams instead of just Title... So I would be in favor of adding aliases for all or at least the most common Params-classes, there instead of the fix I mentioned earlier. 

Since this seems relatively simple, it makes me wonder if there is a reason this isn't already there though. Maybe it was just added later to VegaLite?

ChristopherDavisUCI:
Maybe we could stick to your original `if`-`elif` for now since that seems like a change that's less likely to inadvertently break anything else, and make a separate PR with the extra aliases?  Just an idea!  If you think the alias idea is equally safe, I'm fine with that.

Just let me know if you want me to do anything.

joelostblom:
Yes we should probably separate that out in a separate PR. Do you want to create one at least including impute and title? I'm not sure how many others there are but we can see if any test breaks to start with. I can do it later this week otherwise.

Let's only use a single if statement here for the time being then. So check if the property exists, otherwise pass through (or raise / print just to catch any errors in our own testing? I agree with you that it should be pass through in production).

ChristopherDavisUCI:
Did you know that `"timeUnit".capitalize()` does not equal `"TimeUnit"`?  I think that explains the last remaining mystery from the methods that weren't working.

joelostblom:
Ah, that makes sense, we can use something like `x[:i].upper() + x[i:]` instead of `capitalize` to leave the case of all other letters untouched.

ChristopherDavisUCI:
That sounds good!  Would you like me to do anything?

My vote (as we discussed a little above) is to merge this if we're ready and then later think more carefully about how to get the autocompletion to work.

joelostblom:
I updated this PR based on what we discussed @ChristopherDavisUCI . It now deals with the upper casing properly and passes through parameters that don't have a corresponding helper class. I tested this with the changes in #2732

I also improved the docstring so instead of just taking the info from the helpclass (e.g. `BinParams`), it now also includes the info of the parameter from the encoding class docstring (e.g. the general snipper on `bin` in X, Y etc). In pictures, that means we have gone from this:

image

to this:

image

I find the latter much more informative. The only doc component still missing is the types, which is only includes in some helper classes it seems, but not all. For `bin` that would be something like this:

```
bin : anyOf(boolean, :class:`BinParams`, string, None)
```

I don't think this is necessary and might even be confusing since you don't need to pass a `BinParams` class (you can use the parameters of that class inside `bin()` directly), but it would be helpful to somehow convey that you could also pass a string or `None` . Not super important though.

For you suggestion regarding merging this, I am a bit hesitant since we might still do some bigger change such as the `.x` to `['x']` if we can't figure out tab completion? Or are you also on board with having this without tab completion? I feel like this PR is a big decision where ideally Jake would way in, but I also don't want to hold things up to long. I will flag this as reviewable at least for now.

ChristopherDavisUCI:
(This comment is more related to down-the-road changes.  It's not meant to delay this PR.)

One random thing I learned this morning is that Jedi does not seem to find the completions for either `alt.X.` or `alt.X().` (unlike in Jupyter where `alt.X.` does find the correct completions), so whatever makes completions work for the class but not for the instance does not seem to be related to Jedi.

I don't know anything about type hinting in Python, but I tried adding this to the channels.py file (I know that file should not be modified directly):

```
from typing import Any, Protocol

class BinCallable(Protocol):
    def __call__(self, anchor=Undefined, base=Undefined, binned=Undefined, divide=Undefined,
                 extent=Undefined, maxbins=Undefined, minstep=Undefined, nice=Undefined, step=Undefined,
                 steps=Undefined, **kwds):
        return None
```

Then I added `bin: BinCallable` to the definition of `class X` (just on its own line, as an example).

This seems to get the `bin` method working as we want:

maxbins

Do you see any potential there or does this give you any ideas?

mattijn (commented):
I think this PR is still work in progress I'm not sure what is expected from this review. The tests and documentation are still missing and the autocomplete is still being investigated. Am I wrong to assume that this will lead to more incoming changes in this PR?

In general, I think this PR is a semantic change leading to a positive enhancement, without regression in any of the examples in the documentation.

People who auto-generate altair code might need to change their workflows, but personally I'm not aware of these type of workflows. Moreover, I think the changes are only for getting and not for setting (see `x3` below), so that will even more reduce the number of people affected?

Potential tests:
```python
import altair as alt
x1 = alt.X(field='foo')
x2 = alt.X().field('foo')
x3 = alt.X()
x3.field = 'foo'

assert x1 == x2 == x3
assert x1['field'] == x2['field'] == x3['field']

x1['field'] = 'bah'
x2 = x2.field('bah')
x3.field = 'bah'
assert x1 == x2 == x3

assert x1.field != x1['field']
```

joelostblom:
> Am I wrong to assume that this will lead to more incoming changes in this PR?

No, you're correct in assuming that there will be more changes. Sorry, I should have been more specific and asked if the approach we are taking still makes sense given the latest discussion here and the updates. Does what we have so far look good enough to invest more time into this and start writing documentation and tests even without the tab autocompletion working yet, or is that a show stopper that should be fixed first because we think the PR is not mergeable without it? If we are planning to add documentation changes to all current examples that use the old syntax (`bin=alt.Bin()` instead of `.bin()` etc) approach, that will be a substantial amount of work and I just want to check that we are pretty convinced that this PR could be merged in the future before doing so. Currently, I believe we have the functionality we want and the docstrings are showing all the correct info.

> In general, I think this PR is a semantic change leading to a positive enhancement, without regression in any of the examples in the documentation.

This is how I view this PR too.

---

@ChristopherDavisUCI That looks interesting! I am not familiar enough with type hinting to say what is going on there, but maybe worth exploring further. I am surprised that Jedi does find the completions for those classes and that is something we probably want working (and maybe could lift tab completion our of this PR, but we would need to look into it more first). I wonder what provides the completion then, I think it works in both JupyterLab, Notebook, IPython, and VScode (but I didn't double check right now)

mattijn:
No need to be sorry. Let's just push a bit forward. It seems @ChristopherDavisUCI is on to something. 
Start of including tests will definitely help in understanding what will work and what will not work, so that can't hurt to include (and are not depending on these docstrings).

joelostblom:
I added the tests you mentioned @mattijn and also converted a few gallery examples to the new syntax (we probably want to keep the old gallery examples around to test the previous api, but to make it easy to see what has changed I kept the file name the same here). The tests all seem to work well from what I can see and I can keep converting them one by one, but let me know if you have any comments regarding syntax etc.

One note is that I don't think the `bandPosition` and `timeUnit` parameters look good as method names, e.g.

```py
alt.X("date:O").timeUnit("yearmonth").title("date")
```

An underscore is more pythonic and more in line with how other methods are named in Altair:

```py
alt.X("date:O").time_unit("yearmonth").title("date")
```

However, I am not sure it is worth the effort to try to fix this and if we change the name of the method, we should ideally also change the name of the parameter, which means we should change the name of all the parameters that use lower camel case, which would be a breaking change so I'm unsure if it is worth it.

ChristopherDavisUCI:
Is it possible this line is not correct?
https://github.com/altair-viz/altair/blob/adddf46af5f3018e0681c4e5d2e33d4e5a3c3a9c/tools/schemapi/schemapi.py#L12

I can't get `python tools/generate_schema_wrapper.py` to work at the moment.  It's very possible I'm just confused, but the command `python tools/generate_schema_wrapper.py` seems to work as expected if I'm on the master branch.

ChristopherDavisUCI:
I think as things are currently written, `python tools/generate_schema_wrapper.py` will add the `property_setters` to v3 and v4.  Is that intentional?  It seems like it would be better to only add them to v5.  (But it's possible I broke something while making edits.)

joelostblom:
> I can't get python tools/generate_schema_wrapper.py to work at the moment. It's very possible I'm just confused, but the command python tools/generate_schema_wrapper.py seems to work as expected if I'm on the master branch.

I pushed an update that uses absolute imports, it should work as long as you have an editable install (`pip -e`) locally. Other options we could consider is to have a startup file (https://stackoverflow.com/a/64310290/2166823 and https://stackoverflow.com/a/57292232/2166823)

> I think as things are currently written, python tools/generate_schema_wrapper.py will add the property_setters to v3 and v4. Is that intentional? It seems like it would be better to only add them to v5. (But it's possible I broke something while making edits.)

I think this change should not be backported to older version. I made a change in the schema to reflect this. I think we should probably regenerate the 4.17 vegalite api with the latest versions that actually used 4.17 and then freeze that folder and have all following changes only be present in the v5 folder.

ChristopherDavisUCI:
I've been thinking more about the experiments from https://github.com/altair-viz/altair/pull/2592#issuecomment-1335331116

Does adding something like 
```
axis: axisCallable
bandPosition: bandPositionCallable
bin: binCallable
...
```
to every class with `@with_property_setters` seem reasonable?  (I don't think this has any impact other than helping auto-completion.)  Those would appear in many classes (so there would be a lot of repetition, but then the definitions of `axisCallable`, etc, would only happen once.

My understanding is this general strategy is endorsed by Jedi: https://jedi.readthedocs.io/en/latest/docs/usage.html?highlight=annotations#type-hinting but it's also very possible I misunderstand, or that there is a simpler solution (or that this won't actually work when we try to implement it).

One thing that seems a little strange to me is, the `with_property_setters` is happening after the class is defined.  For example, it uses `cls.resolve_references()`.  But these lines `axis: axisCallable`, etc. are getting inserted as part of the class definition.  That feels a little asymmetric to me.  

The hardest part seems to be deciding which methods we need (like `axis`, `bin`, etc.), and it seems wrong to do this work in two different places, once inside of `with_property_setters` and once in whatever code would write the `axis: axisCallable`, etc. lines.

@joelostblom and @mattijn, do you see any way to have these `axis: axisCallable` lines happening as part of the `with_property_setters` definition?

Not sure if that makes any sense or not!

mattijn:
With documentation I merely meant explanation of this approach within the documentation. How is this different vs current approach. I don't think we should already update all the examples in the example gallery in this PR. 

The current codeblock of defining the `self.__doc__` in the class `_PropertySetter(object)` feels a bit.. custom.  Not sure if that is very future proof with automatic changes in the API. I also don't see it [elsewhere](https://github.com/altair-viz/altair/search?q=self.__doc__&type=code) in the altair codebase that the `.__doc__` is defined like this.

I still feel still a bit hesitant regarding the non-functioning tab completion for the `alt.X().bin` structure. @ChristopherDavisUCI did you get one step further with the type hinting?

Edit: I just noticed you post something

mattijn:
So for my understanding, your potential approach will define more classes in the [channel.py](https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/schema/channels.py) file. Currently the classes defined in this file are the different Encoding Channels and you are going to extend these with the [Encoding Channel Options GitHub page](https://github.com/altair-viz/altair/blob/master/doc/user_guide/encoding.rst#encoding-channel-options) [Altair-doc](https://altair-viz.github.io/user_guide/encoding.html#encoding-channel-options).

These options will be different for different encoding channels so these have to be defined using a kind of lookup in the `with_property_setters`?

When reading this linked to the GitHub page of these encoding channel options, I noticed that it defines the class `altair.PositionFieldDef` as being valid for the channels `X` and `Y` and when looking to this class [here](https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/schema/core.py#L11235) in the core.py file, it seems it has defined all the appropriate encoding channel options. 

Can we use these directly instead of defining new ones in the channel.py?

And then there is also no many repetition in the `with_property_setters`, apart from writing a certain logic that connects the channel to the correct option class:
| Channel                 | Class options                                                      |
|-------------------------|--------------------------------------------------------------------|
| X and Y                 | PositionFieldDef                                                   |
| Color, Fill, and Stroke | FieldOrDatumDefWithConditionMarkPropFieldDefGradientstringnull     |
| Shape                   | FieldOrDatumDefWithConditionMarkPropFieldDefTypeForShapestringnull |
| Order                   | OrderFieldDef                                                      |
| ..                      | ..                                                                 |

mattijn:
Ah yes. I forgot that the example gallery also functions as test suite.

But the side effect is that the _enhancement_ in this PR now becomes the main approach to specify Altair charts.

joelostblom:
> But the side effect is that the enhancement in this PR now becomes the main approach to specify Altair charts.

I agree that this is something we should discuss (although I think it is probably a good idea since the new syntax is more convenient to both type and read). We should probably make a new `v5` folder for these update examples so that we still keep the old ones for testing the old syntax. I can do that before merging this, I just wanted to make it easy to see what was changed in each example so I am keeping the names the same for now. Later, we can decide if we want to show both the new and old syntax in the gallery in a tabbed interface or keep the old examples in a hidden folder just for testing purposes.

mattijn:
Good. For me personally, I'm in favour of the method based approach if the tab completion works. Let's wait patiently until the succesful findings of Chris. No pressure;).

@ChristopherDavisUCI: I mentioned [this](https://github.com/altair-viz/altair/pull/2592#issuecomment-1340064239) before: 

> When reading this linked to the GitHub page of these encoding channel options, I noticed that it defines the class altair.PositionFieldDef as being valid for the channels X and Y and when looking to this class [here](https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/schema/core.py#L11235) in the core.py file, it seems it has defined all the appropriate encoding channel options.
> 
> Can we use these directly instead of defining new ones in the channel.py?

But I think I was wrong here isn't it? You need classes of the variables _within_ these classes if I understand correctly?

ChristopherDavisUCI:
Thanks @mattijn!  It does seem like things might be able to be streamlined a little by adding to the `PositionFieldDef` definition, for example, but I think the overall difficulty is going to be similar.  I'm hoping to get something that works, and then maybe we can look at it and see if there's a way to make it more elegant.

joelostblom:
A note that there is a small inconsistency in that this new method based syntax does not apply to the `datum` function. We can do `alt.datum(4, scale=None)` but not `alt.datum(4).scale(None)`. This doesn't seem like a big deal to me but thought I would record it since `datum` can be used instead of the encoding class `alt.X()` etc so it might be expected that the syntax is similar. Relevant gallery example here https://joelostblom.github.io/altair-docs/gallery/pacman_chart.html

ChristopherDavisUCI:
Hi @joelostblom, am I right that it does still work if you use `alt.XDatum` or `alt.ThetaDatum` rather than `alt.datum`?  I think maybe `alt.datum` didn't exist when this PropertySetters code was originally written, so it might be possible to add.

joelostblom:
That does work! Then I think we should probably add it to `alt.datum` too to avoid that being the odd one out which could be confusing.

ChristopherDavisUCI:
Hi @joelostblom and @mattijn, I just wanted to say it might be another week before I'm able to get back to looking at this.  Just wanted to pass along that quick update.

ChristopherDavisUCI:
Hi @joelostblom and @mattijn, do you know what produces the `Type` column in these [docs](https://joelostblom.github.io/altair-docs/user_guide/encoding.html#encoding-channel-options)?  I think I want to have access to those same values.

I've stared at this code for a few minutes:
https://github.com/altair-viz/altair/blob/09988e9df19578c398dd24c55dd9b1c00c7d400b/altair/sphinxext/schematable.py#L96
but don't really get the picture of what's happening in there.  Do you know if that's the right place to look?

mattijn:
I think I found two approaches. Here for eg the class channel options of `alt.PositionFieldDef` for channel `X` and `Y`:

1. First is to use the `type_description` from the sphinx extension (this is how the table in the docs is generated):

```python
import altair as alt
from altair.sphinxext.schematable import type_description

schema = alt.PositionFieldDef.resolve_references()

for prop, propschema in schema.get("properties", {}).items():
    print(prop, type_description(propschema))
```
```cmd
aggregate :class:`Aggregate`
axis anyOf(:class:`Axis`, `null`)
bandPosition `number`
bin anyOf(`boolean`, :class:`BinParams`, `string`, `null`)
field :class:`Field`
impute anyOf(:class:`ImputeParams`, `null`)
scale anyOf(:class:`Scale`, `null`)
sort :class:`Sort`
stack anyOf(:class:`StackOffset`, `null`, `boolean`)
timeUnit anyOf(:class:`TimeUnit`, :class:`TimeUnitParams`)
title anyOf(:class:`Text`, `null`)
type :class:`StandardType`
```

2. Another approach is to use the functions within `utils` part of the tools that are used to generate the altair codebase.
```python
from tools.schemapi import utils
import altair as alt

si = utils.SchemaInfo(alt.PositionFieldDef)
for p in si.properties:
    print(p, si.properties[p].medium_description)
```
```cmd
aggregate anyOf(:class:`NonArgAggregateOp`, :class:`ArgmaxDef`, :class:`ArgminDef`)
axis anyOf(:class:`Axis`, None)
bandPosition float
bin anyOf(boolean, :class:`BinParams`, string, None)
field anyOf(:class:`FieldName`, :class:`RepeatRef`)
impute anyOf(:class:`ImputeParams`, None)
scale anyOf(:class:`Scale`, None)
sort anyOf(:class:`SortArray`, :class:`AllSortString`, :class:`EncodingSortField`, :class:`SortByEncoding`, None)
stack anyOf(:class:`StackOffset`, None, boolean)
timeUnit anyOf(:class:`TimeUnit`, :class:`TimeUnitParams`)
title anyOf(:class:`Text`, None)
type enum('quantitative', 'ordinal', 'temporal', 'nominal')
```

ChristopherDavisUCI:
I made a new branch adding some type hints in #2769 (I didn't want to make edits to this branch directly because I am sure there are some issues in what I've produced).  I'm very happy for any comments!

joelostblom:
Implemented in #2795

PR #2566, opened 2022-03-21
DOCS: Revamp docs for the 5.0 release

joelostblom:
This is a suggestion for what the Altair docs could look like based on the discussion in #1994. I tried to bring the aesthetics a bit closer to the appearance of the docs for other Python data science libraries and Vega-Lite.

You can browse the documentation here https://joelostblom.github.io/altair-docs/index.html.

I think it would be good to merge #2415 before this PR, so that I can take into account those new doc sections, but I can also do that later if we want to merge this first. I hope to write some more doc pages, such as working with images, geodata, and more tutorials/case studies at some point later this year.

Screenshot of the front page:
image

Things I am still working on:
1. [x] 1. The organization of the user guide. I currently included "Advanced Usage", but moved "Case Studies" and "Getting Help" to the topbar, very open to feedback. **I am happy with my current suggestion, but still open to feedback!**
2. [x] 2. The sidebar ToC is not persistent in the User Guide for some reason, although it works fine on the Getting Started page...
3. [x] 3. The `...` action button is far to the right of each chart
4. [x] 4. The project title doesn't align with the logo when the page is made smaller.
5. [x] 5. The gallery thumbnails didn't show up initially (maybe the same issue as https://github.com/altair-viz/altair/issues/2092). I fixed this in https://github.com/altair-viz/altair/pull/2566/commits/1bd8b69f1922f5f0dc40a12fcf814d4c5981eb91  by making the paths relative, but if someone else can confirm that this works for the way the documentation is built before being put on altair-viz.github.io that would be great.
6. [x]  6. Headings don't show up in the gallery

Please have a look at the page and add additional things you see that are broken or could be improved.

Note that I have removed the gallery items from their toc trees to better display the examples page and not clutter its ToC. This raises warnings during the sphinx build but they still show up fine and are easy to navigate via the right hand headings menu. There is also a warning about using the same "Vega-Lite" target for different links, but everything works as expected.

@mattijn  Regarding [your comment on reorganization and breaking existing links across the web](https://github.com/altair-viz/altair/issues/1994#issuecomment-970672429): ~~For now I only moved one page, the Seattle weather case studies, and used the sphinx extension `rediraffe` to redirect the old link (https://joelostblom.github.io/altair-docs/case_studies/exploring-weather.html) to https://joelostblom.github.io/altair-docs/user_guide/case_studies/exploring-weather.html~~. Edit: I returned this page to its original location for other reasons, but rediraffe seems to work so it could be an option if we want to reorganize.

close #1994

PR #2593, opened 2022-04-20
Add interactive gallery via JupyterLite

joelostblom:
I re-discovered JupyterLite today and find it remarkably cool! We could use it to add interaction to the example gallery and add a "Try online" menu entry in the navbar similar to what is available on the VegaLite home page. You can try this live at https://joelostblom.github.io/altair-docs/ and here is a short video of what it looks like 

https://user-images.githubusercontent.com/4560057/164151778-e0c3cd3e-c476-4208-b971-28922399e524.mp4

As you can see in the demo, there is still some boilerplate setup code required, and it is worth noting that JupyterLite is still in beta. I also played around with having it inline in the documentation, I think it could look good with with [the replite directive](https://jupyterlite-sphinx.readthedocs.io/en/latest/directives/replite.html) on the bottom of each gallery example once it matures a bit and adds the ability to load on demand instead of on each page load. Currently I used jupytext to convert all the examples to notebook files and add them to JupyterLite; we could sort them into folders and make other tweaks if this is a direction we want to pursue.

This PR doesn't change anything in #2566 (the demo is just served on the same URL) so that it still ready to merge as is and the diff from this PR will be easier to understand afterwards. The only new commits here are the following:

- Add minimal jupyterlite setup
- Add extension gallery notebooks to jupyterlite


jakevdp:
That's pretty cool! Great idea to have this available via the docs

Issue #2578, opened 2022-03-25
Include documentation for each mark type


mattijn:
The docs for marks are currently still missing in the Altair documentation. It would be nice to have a similar structure as the [Data Transformations](https://joelostblom.github.io/altair-docs/user_guide/transform/index.html#) for Marks.

Raised before in https://github.com/altair-viz/altair/pull/2566#issuecomment-1079289236

hebarton5:
Moving conversation from #2580 here! I would be willing to contribute to documentation on each mark, especially now that you have provided and example with mark_geoshape(). Since each mark already has a tagged example associated with it, what additional information/examples should be included in the documentation?

joelostblom:
Hello @hebarton5 ! Thank you so much for your interest in contributing to Altair. I think it would be really nice to follow the design of the [`Mark` page in Vega-Lite](https://vega.github.io/vega-lite/docs/mark.html) when building it for Altair. I also like the reorganization that @mattijn started in https://github.com/joelostblom/altair/pull/1 ([preview](https://deploy-preview-1--spontaneous-sorbet-49ed10.netlify.app/user_guide/marks/index.html)) so we can draw inspiration from there as well. Here is what I would suggest (1 and 2 are the most important, and I'm open to feedback):

1. I think the landing page should include a short intro to what marks are (similar to what is there now already), the table of all available marks (same format as in @mattijn PR), and then a heading "Mark Properties" which contains the [current mark properties table](https://joelostblom.github.io/altair-docs/user_guide/marks.html#mark-properties), but [divided into subheadings as can be seen on the VegaLite page](https://vega.github.io/vega-lite/docs/mark.html#mark-def) (general, position and offset, color, stroke style, and finally hyperlink properties) (i don't think the last "style" section is as important).
2. I like the format of individual mark pages being subpages under the "Mark" headings (as in @mattijn 's PR) rather than links to the gallery. Let's include a few examples for each mark directly on the page, ideally translating the ones showed at the Vega-Lite pages to Altair code (+ mentioning any differences in Altair where appropriate, I can help with that part). Some of these are relative brief, e.g. the [arc examples](https://vega.github.io/vega-lite/docs/arc.html#examples) whereas others are really long, e.g. the [boxplot example](https://vega.github.io/vega-lite/docs/boxplot.html#boxplot-types). Feel free to shorten the really long pages (e.g. we don't need to talk about "The Parts of Box Plots") and only include a few examples to start with (we can always add more later). 
[...]

PR #2603, opened 2022-05-17
Marks subpages 

hebarton5:
This is a draft of work being done to address issue #2578. I have included sample mark subpages for mark_arc and mark_area, as well as changes to the marks landing page as suggested - I didn't separate the Mark Properties table by section because I'm not sure if that should be done manually or in a way that will pull information automatically by filtering properties from MarkDef.


joelostblom:
This looks like a great start @hebarton5! Thanks again for working on adding this material, it will be very helpful to have these marks examples in the documentation.

PR #2621, opened 2022-06-21
Add waterfall chart example

yanghung:


altair/examples/waterfall_chart.py
```diff
@@ -0,0 +1,96 @@
...
+    , y='calc_sum_dec:Q'
+)
+text_bar_values_mid_of_bar=base_chart.mark_text(baseline='middle').encode(
+    text=alt.Text('calc_text_amount:N')
+    , y='calc_center:Q'
+     , color=alt.condition("datum.label ==='Begin'||datum.label === 'End'", alt.value("white"), alt.value("white"))
+)
+
+## layer everything together
+(bar+rule+text_pos_values_top_of_bar+text_neg_values_bot_of_bar+text_bar_values_mid_of_bar).properties(width=800, height=450)
```
joelostblom, line 96: Since this line is so long I would write it like this instead:

```py
alt.layer(
    bar,
    rule,
    text_pos_values_top_of_bar,
    text_neg_values_bot_of_bar,
    text_bar_values_mid_of_bar
).properties(
    width=800,
    height=450
)
```

altair/examples/waterfall_chart.py
```diff
@@ -0,0 +1,96 @@
...
+)
+
+bar=base_chart.mark_bar().encode(
+    y=alt.Y('calc_prev_sum:Q',title='Amount')
+    , y2=alt.Y2('window_sum_amount:Q')
+    , color=color_coding
+)
+
+## the "rule" chart is for the lines that connect bars each month
+rule=base_chart.mark_rule(color='#404040', opacity=0.9, strokeWidth=2, xOffset=-25, x2Offset=25, strokeDash=[3,3]).encode(
```
joelostblom, line 74:
```suggestion
rule = base_chart.mark_rule(
    color='#404040',
    opacity=0.9,
    strokeWidth=2,
    xOffset=-25,
    x2Offset=25,
    strokeDash=[3,3]
).encode(
```

I would break this up since the line is really long

joelostblom (changes requested):
Thanks for creating this PR @yanghung ! I made a few minor comments on coding style , but other than that this looks good to me! Could you address them and ping me after so that I can review again?

joelostblom:
@yanghung When you are addressing the comments I made, could you also rebase this on the latest main branch so that the CI tests above pass (I recently committed a fix)?

binste:
If you merge the PR in its current state I can create a follow-up one in the next few days to address the open points. If it doesn't work due to the failed checks I can also copy over the code from here and add @yanghung as an author.

mattijn:
Thanks @binste! Which category is this supposed to be? It's currently in `other charts` but it is not there anymore.

binste:
As it contains quite a few transformations I'd suggest to place it into Advanced Calculations `# category: advanced calculations`.

mattijn:
thanks again @yanghung, and the floor is yours @binste;)

PR #2623, opened 2022-06-29
Example of a line chart with a label annotating the final value

### LINKED ISSUE #2622: Example gallery should show how to mark and label the final value in a timeseries
Charts like this have become commonplace in the COVID pandemic. It would be great to include an example showing how to do it. I'd submit a pull request, but I don't know how to do it! If someone can show me how, I'll submit an example for the docs.

Screenshot from 2022-06-29 14-29-55

palewire:
Fixes #2622

It should look like this:
visualization

The example was drawn from [the Vega-Lite docs](https://vega.github.io/vega-lite/docs/aggregate.html#example-labeling-line-chart)


joelostblom:
Thanks for this contribution! I recently fixed the CI tests, could you rebase this on the main branch so that we can see that all the tests are passing here before merging?

palewire:
I believe I've done that.

mattijn:
Interesting approach! I was not aware of `argmax` in this context. 

While I like your approach, there exist a shorthand for 
```python
x=alt.X('date:T', aggregate='max')
```
namely 
```python
x=alt.X('max(date):T')
```

And, while shorter in your example, this line: 
```python
y=alt.Y('price:Q', aggregate={'argmax': 'date'})
``` 
could be written in the Altair approach as 
```python
y=alt.Y('price:Q', aggregate=alt.ArgmaxDef(argmax='date'))
```
since there is no shorthand for this (that I'm aware of).

All in all, the possible changes won't make it more readable.. 
What you think @joelostblom?

Maybe we could create an issue as well to expand the Altair docs for [aggregation ](https://altair-viz.github.io/user_guide/transform/aggregate.html) as well with more info from [vl-aggregate-ops](https://vega.github.io/vega-lite/docs/aggregate.html#ops) and [vl-aggregate-argmax](https://vega.github.io/vega-lite/docs/aggregate.html#argmax).

palewire:
I'm fine with those edits if it's what you'd prefer to showcase.

palewire:
I made the changes for you.

mattijn:
Thanks!

PR #2624, opened 2022-06-30
Pin selenium to avoid doctest breakage from deprecation

joelostblom:
Altair's doctests are currently failing due to a deprecation in Selenium 4.3.0 which is a dependency of `altair-saver` (as reported in https://github.com/altair-viz/altair_saver/issues/104). Until there is a new release of `altair-saver` that addresses this (maybe with https://github.com/altair-viz/altair_saver/pull/105), this PR make sure we can run doctests by pinning the version of `selenium`. 

There will likely still be confusion among people who use the altair-saver library directly until it sees a new release since selenium 4.3.0 seems to be the default version installed with `altair-saver` currently.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install .[dev]
+        pip install 'selenium<4.3.0'
         pip install altair_saver
         pip install git+https://github.com/altair-viz/altair_viewer.git
     - name: Test with pytest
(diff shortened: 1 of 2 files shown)


mattijn (commented):
If I try locally I need double quotes. 
```python
pip install "selenium<4.3.0"
``` 
Also documented here: https://docs.python.org/3/installing/index.html#basic-usage:

> When using comparator operators such as >, < or some other special character which get interpreted by shell, the package name and the version should be enclosed within _double quotes_

joelostblom:
Good catch! This is unexpected for me since most Unix shells expand variables within double quotes but not single quotes, and I would expect us to want the string to be passed literally to `pip` without any possible shell expansion taking place. I wonder if it is an OS-specific issue but as you pointed out, let's follow what the Python docs say. I updated to double quotes and will merge if all the tests pass.

Issue #2424, opened 2021-03-02
Replace the broken example "Cumulative Wikipedia Donations"


harabat:
## Issue 1

The example ["Cumulative Wikipedia Donations"](https://altair-viz.github.io/gallery/cumulative_wiki_donations.html) is broken, and goes against the Contributing guidelines, namely _Example code should not require downloading external datasets_, so I'd suggest replacing it with another chart. 

## Issue 2

I spent more than an hour yesterday reading through a number of StackOverflow questions trying to style a facet plot (anchoring of titles and order of plots). Believing that my issue was not unique and that finding the solution was not straightforward, I wanted to post a question with my own answer on StackOverflow, to consolidate the information in one place, but maybe I can include it into the new chart?

## Suggested solution

The following code provides an illustration of a cumulative sum (to replace the broken chart), but also clarifies the styling of titles, headers, order, and marks of a facet plot, and gives an example of combining predicates with the filter transform. 

```python
import altair as alt
import vega_datasets as data

source = data.disasters()
columns_sorted = ['Drought', 'Epidemic', 'Earthquake', 'Flood']

alt.Chart(source)\
.mark_line(interpolate='basis')\
.encode(
    alt.X('Year:O', title=None),
    alt.Y('cumulative:Q', title=None),
    alt.Color('Entity:N', legend=None)
)\
.properties(width=300, height=150)\
.facet(
    facet=alt.Facet(
        'Entity:N',
        title=None,
        sort=columns_sorted,
        header=alt.Header(labelAnchor='start', labelFontStyle='italic')
    ),
    title={
        'text': ['Cumulative casualties by type of disaster', 'in the 20th century'],
        'anchor': 'middle'
    },
    columns=2
)\
.resolve_scale(y='independent')\
.transform_filter(
    {'and': [
        alt.FieldOneOfPredicate(field='Entity', oneOf=columns_sorted), 
        alt.FieldRangePredicate(field='Year', range=[1900, 2000])
    ]}
)\
.transform_window(cumulative='sum(Deaths)', groupby=['Entity'])
```

![faceted_plot_with_cumsums](https://user-images.githubusercontent.com/45575009/109661701-2c966f80-7b62-11eb-9fcd-4d5f139077f6.png)

Can I make a PR, including this into _Other Charts_ maybe, with the title `Faceted plot of cumulative sums with custom order and title styling`?

jakevdp:
Looks great! A PR in the *case studies* category would be perfect.

[...]

PR #2625, opened 2022-06-30
Remove broken Wikipedia donations chart

palewire:
This chart doesn't work. #2424 proposed a replacement, but it hasn't been merged. In the meantime, I suggest we delete the chart. 
Screenshot from 2022-06-30 14-07-45


mattijn:
Make sense. Thanks

PR #2626, opened 2022-06-30
Tidy Falkensee case study

palewire:
This adds a title and cleans up the axis labels. IMHO, our case studies should all strive to be fully polished charts that are ready to for publication.


mattijn:
Thanks!

PR #2628, opened 2022-06-30
Tidy up U.S. Population by Age and Sex case study

palewire:
This proposal tidies up one of our case studies so it is polished and better ready for publication.

Before:
visualization

After:
visualization2


mattijn:
Thanks!

PR #2634, opened 2022-07-01
Clean up airport maps

### LINKED ISSUE #2629: There are two U.S. Airports maps. There probably shouldn't be.
https://altair-viz.github.io/gallery/airports_count.html

https://altair-viz.github.io/gallery/airports.html

Screenshot from 2022-06-30 14-42-01
Screenshot from 2022-06-30 14-42-04

palewire:
Fixes #2629 by cutting a duplicate map and reframing the airports map more clearly as our best, most direct point map example.


mattijn:
I agree on the lack of diversity on the maps! But I think the techniques behind these approaches are quite different. Especially when reading the related PR: https://github.com/altair-viz/altair/pull/1407 and issue: https://github.com/altair-viz/altair/issues/1366 of https://altair-viz.github.io/gallery/airports_count.html.

Instead of deleting, we should improve the title and description to reflect more the goal and technique of the chart.

palewire:
I've added it back in with a more specific name

mattijn:
Great!

PR #2640, opened 2022-07-01
Delete multiple_marks.py

palewire:
This chart doesn't have a clear purpose, and it is duplication of https://altair-viz.github.io/gallery/line_chart_with_points.html

So I propose we cut it.


joelostblom:
I agree that there is overlap between these two from a documentation perspective. However, the gallery examples are also used as a test suite for Altair, so there is an advantage from that angle to include both since they test `point=True` and `point=alt.OverlayMarkDef()`, respectively. But maybe we could make them slightly more different so that it make sense from a documentation perspective as well?

palewire:
Does `point=True` translate to `point=alt.OverlayMarkDef()` under the hood inside of Altair internals? If that's the case, substituting it in for the current example would result in the code being "covered" in both instances, which would address your testing concern without clinging to a cruft example. It also would result in a simpler example for newbies. What do you think?

palewire:
I've revised this proposal to satisfy your request and bring out examples into harmony with what's in the Vega-Lite documentation. It conflicts a bit with #2644 but that's okay by me. We can always revised that request after we settle this one.

joelostblom (approved):
LGTM! Thank you for the updates!

PR #2644, opened 2022-07-03
Tidy line chart examples

palewire:
This proposal is a revival of #1284, which lost its downstream fork and had conflicts requiring resolution. I believe it is ready now for reconsideration.

I've added two new line chart examples lifted from the ones upstream in the Vega Lite docs. In the interest of uniformity, I also modified two existing examples to ensure that they all are drawn from the same datasets as complimentary examples, which will make things quicker to grok for newbies.


joelostblom:
I think these additional examples are useful! Could we change the dataset in all these examples to the stocks data instead so that the axis can be temporal instead of ordinal? It looks like this is what they have done in the Vega-Lite repo as well https://vega.github.io/vega-lite/examples/#line-charts for the corresponding line charts.

palewire:
Alright. I think a good chunk of what we had here was handled upstream. But, FWIW, I think what remains is still worth merging.

joelostblom (approved):
Looks good, thank you!

PR #2646, opened 2022-07-03
Create a new tables section

palewire:
The other charts in the "Simple Charts" section at the top of the page have their own sections lower down. If that's our rule of thumb, we should add a heatmaps section too.


joelostblom:
What do you think about calling this section "Table-based" plots like in the Vega-Lite gallery https://vega.github.io/vega-lite/examples/#table-based-plots? Then you could also move over the GitHub punchcard and wind vector charts.

palewire:
Sure. I've made the change as requested.

joelostblom (approved):
LGTM!

PR #2647, opened 2022-07-04
Create a new advanced calculations section of the example gallery

palewire:


joelostblom:
I am leaning towards accepting the move of this and the violin plot PR #2648 , but I am not 100% certain since the current "Other" section is similar to Vega-Lite's "[Advanced](https://vega.github.io/vega-lite/examples/#advanced-calculations)" section. I guess the question is whether we want to only classify charts by the type of mark they use, or how they are thought of conceptually (i.e. violins are not not just a type of area chart, but a distinct category on its own). 

We currently have density charts under "Area charts" so with that logic it would make sense to move violins as well. If we move these two, I think we should also move the pyramid pie to the "circular plots" section, "Multiple marks" the Lineplots, Ridgeplot to Area, Wilkinson to Scatter, and Highlighted segment to Barplots.

What do you think @mattijn

mattijn:
Hm, good one. The VL counterpart is 'Advanced Calculations'. This _calculations_ part is not something to ignore. If I look the altair specifications of these parallel coordinate plots, I would say the pith of the matter is on the _calculations_ part and not on the _line_ part.
So I would suggest to introduce a section on _advanced calculations_ or maybe even better, _advanced transformations_. 
If this means the altair _other_ section will be renamed or recategorized, that's OK.

palewire:
That's done. All of the other charts have been moved into a new Advanced Calculations section. IMHO, we could find better homes for a few of these. But I figured let's do this one step at a time.

altair/sphinxext/altairgallery.py
```diff
@@ -271,8 +271,8 @@ def main(app):
             "Histograms": [],
             "Maps": [],
             "Interactive Charts": [],
+            "Advanced Calculations": [],
             "Case Studies": [],
-            "Other Charts": [],
         }
```
joelostblom, line 276:
```suggestion
            "Other Charts": [],
        }
```

The docbuild is failing without this and we still have some logic to assign uncategorized charts as "Other Charts", so let's keep it here as a backup.

joelostblom: I am still running into another issue locally with `multiple_interactions.py` which raises `Error: Unrecognized signal name: "Big_Budget_Films_tuple_fields"` after the suggested. But let's commit this suggestion first and see if it fixes the CI since you haven't modified the `multiple_interactions.py` file in this PR.

joelostblom (changes requested):
Thanks for changing, I think this looks good! One minor thing to get the CI to pass

palewire:
I resolved conflicts from upstream merges and our tests are now passing.

PR #2652, opened 2022-07-07
formatted rst list correctly

tempdata73:
Fixed a formatting error within the documentation (user_guide -> data.rst).
Deleted the point after "Pandas" so as to make it consistent with the writing style of the other lists inside the same document.

Although I did not make any code changes, this change passes all tests, and I also followed the black formatting step.


mattijn:
Thanks!

Issue #2645, opened 2022-07-03
improve documentation on aggregation


mattijn:
It would be good to expand the Altair docs for [aggregation ](https://altair-viz.github.io/user_guide/transform/aggregate.html) with more info from [vl-aggregate-ops](https://vega.github.io/vega-lite/docs/aggregate.html#ops) and [vl-aggregate-argmax](https://vega.github.io/vega-lite/docs/aggregate.html#argmax).

tempdata73:
I'm gonna be tackling this in the next couple of days. Since I'm new to this I do have a couple of questions:
1. Do I replicate the ops table and argmax example that are provided in the vega-lite site? Or should I create distinct examples? Or may I reuse the ones that are provided in this repo?
2. Should the documentation be more technical oriented or example oriented?

joelostblom:
Thanks for offering to help with this! My vote is as follows:

1. Replicate/reuse the ones from the vega-lite site
2. I like how it is done on the vega-lite page for this example. In general I think it is fine to add some additional explanatory text compared to the vega-lite docs because they are sometimes sparse with the explanations.
3. I'm OK with moving it to the aggregate transform section. I think it is slightly better there as you say, but I don't feel too strongly about it either way.

betaigeuze:
I think this can be closed since #2655 resolved this

PR #2654, opened 2022-07-10
Improve docs on Aggregate Transforms #2645

tempdata73:
So this is what I've worked on. A couple of notes:
- I messed up my commits and branches, so the first two are just noise and won't affect the repo in any way whatsoever.
- Similar to the previous point, I merged the add-argmax-example branch with this one, and so it appears that I want to add an example that's already been added in another pull request I made. My guess: it's because that hasn't been merged yet.
- I've intentionally left out the  argmax example given in vl's site because it shows the same thing that can be done using the max function. Instead of that, I created another example which I think is more illustrative.

I hope this is what you're looking for, but if it isn't please tell me and I'll try to fix it right away.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@ __pycache__/
 
 # Distribution / packaging
 .Python
-env/
+venv/
 build/
 develop-eggs/
 dist/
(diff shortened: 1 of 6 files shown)


joelostblom:
@tempdata73 I think the changes you suggested here look good and I am not sure why the PR was closed. I can reopen it but maybe you want to create a new one based on the latest main branch and remove the unnecessary commits you mentioned?

Issue #2598, opened 2022-04-26
Naming of `parameter` and `add_parameter`


mattijn:
It's not directly a problem that blindside me at 4pm on some idle Tuesday, but I prefer `alt.param()` over `alt.parameter()` and `alt.Chart().add_params()` over `alt.Chart().add_parameter()`...

ChristopherDavisUCI:
I'm happy to change that if others agree, but doesn't `add_param` match `add_selection` better than `add_params`? 

Naming in Python has always been mysterious to me, so I don't claim to have any good intuition (like `isna` vs `isnan` vs `isnull` ...).

joelostblom:
I also don't think this is a big issue, but I'm in favor for the shorter form since it is both consistent with what parameters are called in VegaLite specs and a bit easier to type. Notably `selection` is different in Altair versus VegaLite (`select`), but I don't think that should stop using the most appropriate term for new additions.

mattijn:
Another strong argument: on my qwerty keyboard, the `e`, `r` and `t` are next to each other, so when typing `parameter` the `param` part comes automatically, but I always have to think how to type the `eter` part🙈

ChristopherDavisUCI:
It's a compelling argument @mattijn!  I'll plan to make this change next time I'm making updates.

joelostblom:
Closed in #2656

PR #2656, opened 2022-07-11
Change naming to alt.param and alt.add_params

ChristopherDavisUCI:
I believe this pull request incorporates the naming changes suggested by @mattijn in #2598.  I didn't come across anything subtle in making these changes; I mostly just used Find and Replace.  Please let me know if there are aspects that need to be thought through more carefully!

Needless to say, this will break the code of anyone currently using `alt.parameter` or `alt.add_parameter`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ points = alt.Chart(source).mark_point().encode(
     x='Horsepower',
     y='Miles_per_Gallon',
     color=alt.condition(brush, 'Origin', alt.value('lightgray'))
-).add_parameter(
+).add_params(
     brush
 )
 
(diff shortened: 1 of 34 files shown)


mattijn:
Nice! The linting check presents one issue to be resolved: https://github.com/altair-viz/altair/runs/7285808845?check_suite_focus=true

The conflict is around here: https://github.com/altair-viz/altair/pull/2656/files#diff-bbe4e187b18d242a366c820d023afd89041759cb96e4ec66c3f34559a72c2f9dR2229.
Can you solve this issue? 

It's understood that this will affect code in the wild of people already trying out the main branch. Therefore thanks again for working on this!

ChristopherDavisUCI:
Thanks @mattijn.  Weird that the search-and-replace led to wrong formatting.  (But then there was also a mistaken parameter still around that the flake8 check found.)  I think those issues are corrected now.

mattijn:
Nice!

PR #2657, opened 2022-07-14
Same HTML output when plot is remade

### LINKED ISSUE #2616: does HTML rendering need to depend on `outputDiv.id`
I don't really understand how the HTML rendering of `altair` plots work, but one thing that is annoying at least for my workflows is that even if I make an identical plot twice, it has different HTML output. This makes `git` tracking annoying.

The responsible line is due to the `outputDiv.id` as shown below. I don't understand what this is doing, but am just curious if there is a way that it could not need to depend on this ID which changes each time an identical plot is remade:

```
if (outputDiv.id !== "altair-viz-24d195b1a15a4248b89adf5c26955a13") {
   outputDiv = document.getElementById("altair-viz-24d195b1a15a4248b89adf5c26955a13");
}
```

---

Please follow these steps to make it more efficient to respond to your feature request.

- [ ] Since Altair is a Python wrapper around the Vega-Lite visualization grammar, [most feature requests should be reported directly to Vega-Lite](https://github.com/vega/vega-lite/issues). You can click the Action Button of your Altair chart and "Open in Vega Editor" to create a reproducible Vega-Lite example.
- [ ] Search for duplicate issues.
- [ ] Describe the feature's goal, motivating use cases, and its expected behavior.

tempdata73:
Fixes #2616.

Instead of dynamically changing the chart's id depending on the other charts, keep a history of the charts made and assign each one their respective id.

The id is a function of a "template" (in this case, "altair-viz"), the md5 hash of its respective spec (the "signature"), and the number of times a chart has been created.

**Caveat:** Since the `counts` object only registers the number of times a chart is created, if the same cell is run twice, then the id will change. **In order to fix this just restart the notebook and run all cells again** at the end of your workflow.  Personally, I don't think this is much of a problem.

This caveat is due to the fact that two identical charts are, well... identical, so there is no way to differentiate between the two apart from having a counter, or a random uuid, or adding the time of creation, etcetera. The counter, however, maintains states between kernel sessions.

### DIFF
--- a/altair/utils/display.py
+++ b/altair/utils/display.py
@@ -1,9 +1,10 @@
+import hashlib
 import json
 import pkgutil
 import textwrap
 from typing import Callable, Dict
-import uuid
 
+from collections import defaultdict
 from jsonschema import validate
 
 from .plugin_registry import PluginRegistry
@@ -166,17 +167,19 @@ def json_renderer_base(spec, str_repr, **options):
 class HTMLRenderer(object):
     """Object to render charts as HTML, with a unique output div each time"""
 
-    def __init__(self, output_div="altair-viz-{}", **kwargs):
+    def __init__(self, output_div="altair-viz-{}-{}", **kwargs):
         self._output_div = output_div
         self.kwargs = kwargs
+        self.counts = defaultdict(int)
 
-    @property
-    def output_div(self):
-        return self._output_div.format(uuid.uuid4().hex)
+    def get_output_div(self, spec):
+        template = json.dumps(spec).encode()
+        signature = hashlib.md5(template).hexdigest()
+        self.counts[signature] += 1
+        return self._output_div.format(signature, self.counts[signature])
 
     def __call__(self, spec, **metadata):
+        output_div = self.get_output_div(spec)
         kwargs = self.kwargs.copy()
         kwargs.update(metadata)
-        return spec_to_mimebundle(
-            spec, format="html", output_div=self.output_div, **kwargs
-        )
+        return spec_to_mimebundle(spec, format="html", output_div=output_div, **kwargs)


jakevdp:
Thanks for looking into this - unfortunately I don't think this is something we can merge. When using Jupyter Notebook, this will break in very strange and confusing ways if someone displays the same chart twice in the same notebook: the second time you attempt to display the same chart, it will update the first output cell where the chart is displayed. In JupyterLab it's even weirder: because all open notebooks exist in the same HTML browser window, if you attempt to display a chart that already exists in *any notebook tab open in your jupyterlab session*, the updated chart will display in that distant output cell.

I understand there are issues with the current model, but a solution that causes that kind of strange action-at-a-distance is not one we can merge. Unfortunately, given the frontend architectures of Jupyter and JupyterLab, I can't think of any way to solve the problems mentioned in the linked issue.

tempdata73:
I don't think I follow. I tried replicating the problems you've mentioned but everything seems to work fine.

With respect to the first problem: the second time I tried to display the chart, the only thing that changes is the chart's id, but apparently Jupyter does not seem to have a problem with it. In order to solve the issue of changing id's using this PR, the user would have to restart the kernel and rerun everything, then do `git add`.

[In jupyter notebook (video)](https://user-images.githubusercontent.com/29875488/178884409-ac663182-a34f-4a8a-8758-364cbbddc7fa.webm)


With respect to the second problem: The chart can be displayed in both notebooks without any problem. **However** --and this is a problem I didn't notice before-- both charts have the same id. It seems that JL does not have any problem with this, but it still feels sketchy.

[In jupyter lab (video)](https://user-images.githubusercontent.com/29875488/178883714-bcdfad87-2777-4dcd-83b4-8793009da9b3.webm)

I understand if both of these caveats (having to restart the kernel and shared charts within a jl session having the same id) make it impossible for this PR to be merged.

mattijn:
Clear review from @jakevdp. Closing for now.

PR #2676, opened 2022-09-05
added 2 examples of interactive boxplot and histogram

rahul-raoniar:
Added two new interactive plot examples
- The first one contains an interactive scatterplot with dual axis interactive histograms
- The second example contains an interactive scatterplot with dual axis interactive boxplots
- The plots are inspired by Plotly's interactive examples
- Here, I have used an interval selection mechanism

### DIFF
--- a/altair/examples/scatterplot_boxplots_interactive.py
+++ b/altair/examples/scatterplot_boxplots_interactive.py
@@ -1,0 +1,51 @@
+"""
+Title: Scatterplot and boxplot with interval selection    
+
+Description: 
+1. This example shows how to link a scatter plot and boxplots
+together such that an interval selection in the scatterplot will plot the selected
+values in the boxplots. 
+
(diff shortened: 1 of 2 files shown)


joelostblom:
Thanks for your contribution @rahul-raoniar ! I was looking through the gallery page and I can see that there is already an [example of adding marginal histograms](https://altair-viz.github.io/gallery/scatter_marginal_hist.html), as well as [one example on how to create interactions between marginal plots and a scatter plot](https://altair-viz.github.io/gallery/dot_dash_plot.html). Since there is already quite a few examples in the gallery overall, my understanding is that we're trying to be selective with adding new ones if the functionality is already present in existing examples. Do you think there is anything in particular that is different with your examples that is not easy to create/understand based on the two existing example I linked above?

mattijn:
Thanks for this PR @rahul-raoniar, but I agree with @joelostblom. Let me close this PR.

Issue #2650, opened 2022-07-06
Incompatible with selenium 4.3.0


cgbahk:
## TL;DR;
`chart.save(...)` doesn't work with `selenium 4.3.0` (released 2022/6/24), due to API deprecation

## Detail
When I try to save chart in png or svg using `chart.save(...)`, it fails like this
```
AttributeError: 'WebDriver' object has no attribute 'find_element_by_id'
```
<details>
<summary>
Full error trace
</summary>

```
Traceback (most recent call last):
  ...
  File "<REDACTED>.py", line 62, in main
    chart.save("image.png")
  File "<REDACTED>/lib/python3.8/site-packages/altair/vegalite/v4/api.py", line 488, in save
    result = save(**kwds)
  File "<REDACTED>/lib/python3.8/site-packages/altair/utils/save.py", line 116, in save
    mimebundle = spec_to_mimebundle(
  File "<REDACTED>/lib/python3.8/site-packages/altair/utils/mimebundle.py", line 60, in spec_to_mimebundle
    return altair_saver.render(spec, format, mode=mode, **kwargs)
  File "<REDACTED>/lib/python3.8/site-packages/altair_saver/_core.py", line 257, in render
    mimebundle.update(saver.mimebundle(fmt))
  File "<REDACTED>/lib/python3.8/site-packages/altair_saver/savers/_saver.py", line 90, in mimebundle
    bundle[mimetype] = self._serialize(fmt, "mimebundle")
  File "<REDACTED>/lib/python3.8/site-packages/altair_saver/savers/_selenium.py", line 284, in _serialize
    out = self._extract(fmt)
  File "<REDACTED>/lib/python3.8/site-packages/altair_saver/savers/_selenium.py", line 267, in _extract
    driver.find_element_by_id("vis")
AttributeError: 'WebDriver' object has no attribute 'find_element_by_id'
```

</details>

This stackoverflow answer was helpful to find why it fails: 
> Selenium 4.3.0
> * Deprecated `find_element_by_*` and `find_elements_by_*` are now removed (# 10712)
>
> (from https://stackoverflow.com/a/72773269/4214521)

`selenium 4.3.0` released on 2022/6/24 https://github.com/SeleniumHQ/selenium/releases/tag/selenium-4.3.0

## Reproduction

I guess any `save()` call can reproduce it. For example, I can reproduce this issue with most simple example: [:link:](https://altair-viz.github.io/gallery/simple_bar_chart.html)

`save()` successful case:
```
[...]

PR #2681, opened 2022-09-20
use 'UndefinedLike = Any' as the type hint

brahn:
Improvement on #2670 -- see summary and discussion there

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -4,6 +4,7 @@ import collections
 import contextlib
 import inspect
 import json
+from typing import Any
 
 import jsonschema
 import numpy as np
@@ -137,7 +138,7 @@ class UndefinedType(object):
         return "Undefined"
 
 
-Undefined = UndefinedType()
+Undefined: Any = UndefinedType()
 
 
 class SchemaBase(object):
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -4,6 +4,7 @@ import collections
 import contextlib
 import inspect
 import json
+from typing import Any
 
 import jsonschema
 import numpy as np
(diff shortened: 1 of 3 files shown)


mattijn:
As is mentioned in this comment: https://github.com/altair-viz/altair/pull/2614, which counts here as well:

> Thanks! Note the line at the top of the file: https://github.com/altair-viz/altair/blob/c4248a953c8518efb3cf5bef8d9755bd0fc7391e/altair/utils/schemapi.py#L1-L2
> 
> You should make these changes in https://github.com/altair-viz/altair/blob/master/tools/schemapi/schemapi.py instead, and then run [generate_schema_wrapper.py](https://github.com/altair-viz/altair/blob/master/tools/generate_schema_wrapper.py) to sync the results to this file.

brahn:
Thanks @mattijn !  I believe I've proceeded as you described.  I don't know the origin of the (small) changes to the .rst file since they don't seem related to my schema change, but I'm guessing that they are either the result of some prior changes for which `generate_schema_wrapper.py` or a slight version change in some of the dependencies.

Also I should mention that when I first tried to run `generate_schema_wrapper.py` I encountered a bug in the `m2r` library described [here](https://bugs.archlinux.org/task/75433), which I resolved by using the patch helpfully offered in that post.

brahn:
@mattijn @jakevdp just checking back in, please let me know if there's anything I can do to help!

mattijn:
I've no opinion on this. 

I hope others can review this. 

If there will not appear any reviewers nor disapprovals I will merge this in two weeks from now.

joelostblom:
I don't have enough expertise to review this properly unfortunately.

brahn:
thanks @mattijn !

brahn:
@mattijn quick follow-up -- forgive my ignorance, how does this factor into the version-release cycle?

joelostblom:
It is currently not clear when the next release will be. We are doing work to have Altair support the latest features in VegaLite 5 and this will likely results in a major release eventually. Although I could see an argument for making a minor release that fixes some package compatibility issues and deprecations (e.g. #2650 #2683 ), I think we also want to focus our efforts so I am not sure that will happen.

Issue #2675, opened 2022-09-02
entrypoints is in maintenance mode


EdgyNortal:
![image](https://user-images.githubusercontent.com/108864217/188197431-6f23baa1-9643-4a49-b1b5-937fddbc5d58.png)
Ran into an issue building a Docker image that involved altair, and a conflicting entrypoints version. The work around was to hard pin `entrypoints` in our code. But as that project is in maintenance mode, it might be nice to decouple altair from it.

mattijn:
Closed [#2675](https://github.com/altair-viz/altair/issues/2675) as completed via [#2686](https://github.com/altair-viz/altair/pull/2686).

PR #2686, opened 2022-10-01
fix(#2675): replace entrypoints with importlib.metadata

daylinmorgan:
As mentioned in #2675, `entrypoints` is now in maintenance mode. We can replace it with `importlib.metadata` and remove the extra dependency for `python < 3.8`.


daylinmorgan:
Ah! Tested locally with 3.10 and 3.6. Missed that the selection interface was added in 3.10. I'll update this to parse the dict from `entry_points()` instead.

mattijn:
Thanks for the PR @daylinmorgan! Can you synchronize this PR with the latest changes on master? Hopefully @joelostblom can do a review here. 
A bit of explanation of the suggested changes would help as well.

daylinmorgan:
This has been rebased on master.

PR drops unnecessary (and the now in maintenance mode) `entrypoints` module for a standard library equivalent where possible and otherwise relies on the backport (python <3.8).

mattijn:
Thanks @daylinmorgan! 
Honestly, I'm not very familiar with the `entrypoints` concept. Reading https://amir.rachum.com/amp/blog/2017/07/28/python-entry-points.html was a fun introduction. 
I am interested in approaches where and how this is being used in corporation with Altair. Are you aware of any examples of this usage?

daylinmorgan:
The entrypoints I'm aware of are in [`altair_saver`](https://github.com/altair-viz/altair_saver/blob/dbbc093203a2456e294ae4f296477a7ef252a339/setup.py#L70-L72) and [`altair_viewer`](https://github.com/altair-viz/altair_viewer/blob/33b6d52b9bff6d9d864a660e217758f844ca699b/setup.py#L73-L75)

mattijn:
Thanks! It is out of scope of this PR, but it seems the entry points in these packages also require updating once Altair is going to vegalite 5. The entry points should become `"altair.vegalite.v5.renderer"` I think. 

Again out of scope, but I'm wondering if it already breaks on the main repo of Altair if a specific VL5 spec is parsed through these entry points.

I'm fine with this PR, you also @joelostblom?

mattijn:
Thank you @daylinmorgan!

PR #2694, opened 2022-10-14
Create selection_regression.py

zcpisme:


zcpisme:
I add an example which may demonstrate the regression line for selected sample.
visualization

Issue #2696, opened 2022-10-16
Create a section for distribution and/or statistical plots in the gallery


joelostblom:
Currently there is a separate section for histograms. I think we should expand this to contain all distribution plots, including densities, violins, cumulative ones, etc, and rename the section to "distributions".

There are also some other "statistical charts" that I think would be useful to lift out of their respective sections. These relate to visualizing uncertainty/errors and trends in data. The motivation for lifting these out of their respective "mark" section is that the mark used here is often secondary to something else in the spec, e.g. a transform and someone looking for how to visualize a confidence interval would likely benefit from seeing this for both lines and points in the same section about uncertainty rather than spread out over different marks.

Below I have indicated what I think should be added to these two sections and where those charts are currently located.

## Distributions

OTHER
- binned heat
- violin
- ridge
- box
- faceted scatter?

SCATTER
- strip?
- scatter with marginal dashes?
- qqplot
- binned scatter

AREA
- Cumulative count
- Faceted density
- stacked density

## Uncertainties and trends

SCATTER
- loess
- polynomial regression
- points with error bars

LINE
- confidence interval

This is a follow up on https://github.com/altair-viz/altair/issues/2661 and cc @binste in case you have any thoughts on this

joelostblom:
Great! Thanks for offering to work on this, that would be very welcome. I think these changes will only make it easier to navigate the gallery, and if anyone disagrees they can always chime in here, so please go ahead and get started when you have time. Note that when you build the example pages locally for testing, it will look quite different from the current online docs, since we had a style revamp to be launched online later (it will look like this https://joelostblom.github.io/altair-docs/gallery/index.html). Let me know if there is anything you think should be clarified regarding the instructions for contributing to the docs and setting things up.

Regarding your comment on the stripplots, there is already  an example of one without jitter in the "Simple plots" sections (which I think can stay there), so maybe you can just change the current stripplot implementation to use the offset encoding instead of the transform and call it "Stripplot with jitter" (and flip x and y so that the labels are easier to read)? I don't know if there is some some scenario where the old transform would be more helpful, but if you can think of one feel free to make a separate example for that (otherwise I believe it is fine to remove this chart, and I don't see an equivalent in the vega library either although I personally find it interesting).

joelostblom:
Closed in #2706

PR #2706, opened 2022-11-04
Create new sections in the gallery for distributions and uncertainties and trend charts

binste:
@joelostblom I created the two new sections in the example gallery which you proposed in #2696. I hosted the documentation based on this pull request at https://binste.github.io/altair-docs/index.html so it's easier to review. I also took the liberty to harmonize the titles of the examples in a separate commit, e.g. title case, consistently use "Scatter Plot" instead of mix with "Scatterplot", etc.

Open questions from my side:

- [x] All examples in the "Tables" section, except for maybe the Wind Vector Map, could also be moved into the newly created "Distributions" section (in your issue you mentioned the [binned histogram](https://binste.github.io/altair-docs/gallery/binned_heatmap.html)). I'd suggest to do this, what do you think?
- [x] As discussed, I rewrote the [Strip Plot with Jitter](https://binste.github.io/altair-docs/gallery/strip_plot_jitter.html) example using the new offset encoding. However, it's now almost the same as [Jitter Chart](https://binste.github.io/altair-docs/gallery/jitter_chart.html) which has a better description. What about removing Jitter Chart and adding the reasoning of why using jitter to the "Strip Plot with Jitter" chart?
- [x] The category "Other Charts" is currently empty but shows up at the bottom of the gallery. Ok if I remove it? I saw that it is used as a fallback category for unclassified examples but we could also just raise an error or warning if this happens.


mattijn:
The table section is based on the table-based plots from vega-lite. I like this section https://vega.github.io/vega-lite/examples/#table-based-plots and maybe we can do both? Expand the table section a bit more and move some to the distribution section? 
The wind vector map is a bit odd there. That one could be added to the Maps section if we add a base layer*.

I favour the strip plot with jitter. So from me +1 to take the description and from the jitter chart and remove the jitter chart.

OK, for error/warning when removing empty other charts.

/* eg like this:
```python
import altair as alt

df_wind = data.windvectors()
data_world = alt.topo_feature(data.world_110m.url, "countries")

wedge = (
    alt.Chart(df_wind)
    .mark_point(shape="wedge", filled=True)
    .encode(
        latitude="latitude",
        longitude="longitude",
        color=alt.Color(
            "dir", scale=alt.Scale(domain=[0, 360], scheme="rainbow"), legend=None
        ),
        angle=alt.Angle("dir", scale=alt.Scale(domain=[0, 360], range=[180, 540])),
        size=alt.Size("speed", scale=alt.Scale(rangeMax=500)),
    )
    .project("equalEarth")
)

xmin, xmax, ymin, ymax = (
    df_wind.longitude.min(),
    df_wind.longitude.max(),
    df_wind.latitude.min(),
    df_wind.latitude.max(),
)

# clockwise, left-hand-rule
extent = [
    {
        "type": "Polygon",
        "coordinates": (
            (
                (xmax, ymax),
                (xmax, ymin),
                (xmin, ymin),
                (xmin, ymax),
                (xmax, ymax),
            ),
        ),
    }
]

# use fit combined with clip=True
base = (
    alt.Chart(data_world)
    .mark_geoshape(clip=True, fill="lightgray", stroke="black", strokeWidth=0.5)
    .project(type="equalEarth", fit=extent)
)

base + wedge
```
image

joelostblom (commented):
Awesome! Thanks for also harmonizing the titles, very satisfying =)

> All examples in the "Tables" section, except for maybe the Wind Vector Map, could also be moved into the newly created "Distributions" section (in your issue you mentioned the binned histogram). I'd suggest to do this, what do you think?

Yeah, I see your point. I think we should move the one called "Binned heatmap" and rename it to "2D histogram heatmap". To me the Github Punch card is also a distribution chart (a categorical 2D histogram). I also agree with Mattijn that the wind vector one is better in the maps section after his modification.

That doesn't leave to many table based plots, but for the heatmap with text, we could keep it in the table section but make it based on this VL example instead https://vega.github.io/vega-lite/examples/rect_heatmap.html, so that it does not show a count but the mean instead. And maybe we could also add the weather and lasagna heatmaps from the VegaLite gallery (the mosaic chart also looks cool, but time consuming to make)?

- https://vega.github.io/vega-lite/examples/rect_lasagna.html
- https://vega.github.io/vega-lite/examples/rect_heatmap_weather.html

> As discussed, I rewrote the Strip Plot with Jitter example using the new offset encoding. However, it's now almost the same as Jitter Chart which has a better description. What about removing Jitter Chart and adding the reasoning of why using jitter to the "Strip Plot with Jitter" chart?

Yes, I am good with this too. I also added a suggestion for including both uniform and gaussian jitter.

> The category "Other Charts" is currently empty but shows up at the bottom of the gallery. Ok if I remove it? I saw that it is used as a fallback category for unclassified examples but we could also just raise an error or warning if this happens.

Yes, I think this is a better approach so feel free to add a warning or even raise an error if an example does not include a category so that we notice it right away.

binste:
Thank you both for the great inputs! I addressed them all, including the Mosaic plot :), in the latest push and updated the documentation at https://binste.github.io/altair-docs/gallery/index.html

joelostblom (approved):
This looks fantastic! Thanks for all the additional work in adding the new examples, really neat =)

PR #2716, opened 2022-11-15
Update docs styling based on the new pydata template

joelostblom:
In recent PRs I have noticed a few style things that was changed in the latest updates to the docs pydata template. These make the site look a bit odd so I have tried to fix those here as well as a small additional beautification.

Most notable changes:

1. I moved the logo to the top corner since this seems to be what most sites have (seaborn, pandas, numpy, mpl, etc). We could make the text more stylized and part of the logo in the future.
2. I removed the sidebars of the landing page so that it would feel more like a landing page and the gallery could take up more space.
3. Changed the default color to white again instead of system specific. Also added a symbol to toggle it (but it seems to only work per page rather than for the entire docs)
4. Added GitHub and SO symbols (maybe these could be bigger, unsure on how)
5. Removed left hand search in favor of the one in the navbar

Before:
image

After:
image

TODO in separate follow up PRs in the future:

- Figure out this warning
    ```
    WARNING: [autosummary] failed to import altair.TopLevelParameter.
    Possible hints:
    * AttributeError: module 'altair' has no attribute 'TopLevelParameter'
    * ModuleNotFoundError: No module named 'altair.TopLevelParameter'
    * ImportError:
    ```
- Go through all examples and update to new syntax (e.g. using `values` instead of `init`)
- Maybe including a left navbar for examples with links to everything, or remove the left navbar region if it is not used.
- Make gallery examples interactive
- Maybe add back the copyright info and sphinx credit somewhere


doc/user_guide/encoding.rst
```diff
@@ -50,16 +50,16 @@ x           :class:`X`           The x-axis value                   :ref:`galler
 y           :class:`Y`           The y-axis value                   :ref:`gallery_scatter_tooltips`
 x2          :class:`X2`          Second x value for ranges          :ref:`gallery_gantt_chart`
 y2          :class:`Y2`          Second y value for ranges          :ref:`gallery_candlestick_chart`
-longitude   :class:`Longitude`   Longitude for geo charts           :ref:`gallery_airports`
-latitude    :class:`Latitude`    Latitude for geo charts            :ref:`gallery_airports`
+longitude   :class:`Longitude`   Longitude for geo charts           :ref:`gallery_airport_connections`
+latitude    :class:`Latitude`    Latitude for geo charts            :ref:`gallery_airport_connections`
```
mattijn, line 54: Should be changed to `gallery_point_map`. This was previous the `gallery_airports` example.

mattijn (changes requested):

doc/user_guide/encoding.rst
joelostblom: Thanks! Fixed in my latest commit

mattijn (approved):

PR #2719, opened 2022-11-16
Fix missing 's' from filename in quick start

dwootton:
Previously, contributing md asked to install "requirement_dev.txt"
image
but the repo actually contains a requirements_dev.txt file:
image


joelostblom (approved):
Thanks!

PR #2721, opened 2022-11-18
Improve documentation titles

binste:
I'd suggest to remove 'Transform' from the titles of the transform pages. The navigation bar looks much cleaner this way, especially now that we also have a page for every mark. This is also in line with the vega-lite docs. There are also two other smaller commits in this PR.

Screenshot 2022-11-18 at 23 26 35

### DIFF
--- a/doc/user_guide/interactions.rst
+++ b/doc/user_guide/interactions.rst
@@ -55,7 +55,7 @@ Here is a simple scatter-plot created from the ``cars`` dataset:
         color='Origin:N'
     )
 
-We can create a variable parameter using :func:`alt.param`, and assign that parameter a default value of 0.1 using the ``value`` property, as follows:
+We can create a variable parameter using :func:`param`, and assign that parameter a default value of 0.1 using the ``value`` property, as follows:
 
 .. altair-plot::
     :output: none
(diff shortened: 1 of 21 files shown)


doc/user_guide/interactions.rst
```diff
@@ -55,7 +55,7 @@ Here is a simple scatter-plot created from the ``cars`` dataset:
         color='Origin:N'
     )
 
-We can create a variable parameter using :func:`parameter`, and assign that parameter a default value of 0.1 using the ``value`` property, as follows:
+We can create a variable parameter using :func:`alt.param`, and assign that parameter a default value of 0.1 using the ``value`` property, as follows:
```
joelostblom, line 58: I can see that there is a mix of `alt.param` and `parameter` used currently on this page. I think the approach most consistent with the rest of the docs would be to change all these to just say `param`. Could you update all the occurrences of this on the page?

doc/user_guide/interactions.rst
```diff
@@ -562,11 +562,11 @@ For instance, using our example from above a dropdown can be used to highlight c
     )
 
 
-    
 
-The above example shows all three elements at work. The :input_dropdown: is :bind: to the :selection: which is called from the :condition: encoded through the data. 
 
-The following are the input elements supported in vega-lite: 
+The above example shows all three elements at work. The :input_dropdown: is :bind: to the :selection: which is called from the :condition: encoded through the data.
```
joelostblom, line 567: It seems like this is not rendering properly in the docs 
image

joelostblom (changes requested):
Yup I agree that your suggestions makes the navbar cleaner. A couple of minor comments and then we can merge.

binste:
Thank you for the feedback. I addressed both of your comments.

joelostblom (approved):
Looks great now, thanks again!

PR #2722, opened 2022-11-18
Fix some formatting issues

binste:
Fixes two small formatting issues on the "Saving charts" doc page

### DIFF
--- a/doc/user_guide/saving_charts.rst
+++ b/doc/user_guide/saving_charts.rst
@@ -159,7 +159,7 @@ To save an Altair chart object as a PNG, SVG, or PDF image, you can use
 However, saving these images requires some additional extensions to run the
 javascript code necessary to interpret the Vega-Lite specification and output
 it in the form of an image. There are two packages that can be used to enable
-image export: vl-convert-python_ or altair_saver_.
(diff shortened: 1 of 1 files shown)


joelostblom (approved):
Thanks!

mattijn:
I thought it is `vl-convert-python`, see https://pypi.org/project/vl-convert-python/.

This line maybe explains the confusion:
> The `vl-convert-python` package provides a series of conversion functions under the `vl_convert` module.

There is variation in dash and underscore and package name and import name..

joelostblom:
Yeah, that is technically the package name on PyPI but I think it makes sense to refer to it as "vl-convert" everywhere except where we note how to install it via `pip install vl-convert-python`. I don't care about it much though as long as the installation command is clear and the links are correct, so feel free to change it if you think that is more accurate

ChristopherDavisUCI:
I've also found myself confused by that in the past... when to use dash and when to use underscore.  Right now vega-datasets is the only example I can think of, but it seems to trip me up regularly.

mattijn:
> I think it makes sense to refer to it as "vl-convert" everywhere except where we note how to install it

I agree. It's good like this 👍.

PR #2724, opened 2022-11-26
Make vl-convert saving work when the data server is enabled

joelostblom:
As per the discussion in https://github.com/vega/vl-convert/issues/31, vl-convert is not currently working when the altair data server is enabled. As it seems like it is not straightforward to fix this on the vl-convert side of things, the workaround suggested here makes sure that the default transformer is enabled when using vl-convert for saving. cc @jonmmease


jonmmease:
Looks good, thanks @joelostblom!

mattijn (approved):

PR #2720, opened 2022-11-18
Improve mark type sections

binste:
This PR addresses the points raised by @joelostblom in #2607 and #2578 (the later can be closed after this PR). In addition, I went through all mark pages to fix any bugs or vega-lite specific parts I could find. I think the pages are now in pretty good shape but it would certainly be nice if someone else can take a thorough look as well. CC @mattijn in case you want to take a look as well.

I have not yet included the sections with the interactive sliders from the vega-lite documentation. Still need to figure out how to best accomplish this. I'll try the suggestions from comment 4 in #2578 and add it to this PR later on or then in a new one.

You can view the updated documentation at https://binste.github.io/altair-docs/ with the exception of the charts in the geoshape section as I did not have geopandas installed. As a sidenote, where would be a good place to add a note that this is now necessary to build the documentation? Maybe we should add it in `requirements_dev.txt` so it is also included in the docbuild workflow?

PR #2725, opened 2022-11-27
Add toggle to hidden code snippets

joelostblom:
This adds a toggle to hidden code snippets as discussed in #2720. The main reason we want this is to be able to hide the verbose parameter code by default but show it on demand for interested readers. Live docs here https://joelostblom.github.io/altair-docs/user_guide/times_and_dates.html

 There are currently only two other pages that use the `:hide-code:` tag and neither of them suffer from having the toggle added in my opinion:

image

image

### DIFF
--- a/altair/sphinxext/altairplot.py
+++ b/altair/sphinxext/altairplot.py
@@ -205,7 +205,7 @@ class AltairPlotDirective(Directive):
             result += [plot_node]
 
         if hide_code:
-            html = "<details><summary><a>Show code</a></summary>"
+            html = "<details><summary><a>Click to show code</a></summary>"
             raw_html = nodes.raw("", html, format="html")
             result += [raw_html]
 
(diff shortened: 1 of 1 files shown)


mattijn:
It was not directly clear for me that I had to click on the text "show code" to activate the toggle. 
Changing the text into something more descriptive could help "click here to .." or changing the style to a button?

Another thing, this toggle now appear above the chart, can this be below the chart? Maybe using `:code-below` in the directive?

joelostblom:
Thanks for the feedback, I changed it to "Click to show code". I am not sure if it can easily be made into a button, maybe in the future we can switch to that extension I linked in the other issue where we discussed this.

Regarding have it below the plot, it is possible to combine with the `code-below` tag as you guessed, but I think it makes sense to have the default to be over as with the visible code snippets. For some reason it also seems to render more nicely in terms of spacing to the chart when it is above.

mattijn (approved):

PR #2734, opened 2022-12-01
Fix syntax for with statement in _spec_to_mimebundle_with_engine

binste:
@joelostblom I fixed the syntax of the with statement which got added in PR #2724.

With the "and" between the context managers, it executed the first command (`enable("default"`) and then returned the second one `disable_max_rows()` to be used as the only context manager in this statement. Therefore, after using this function the data_transformers pipeline was always reset to "default" and any previous settings were lost. By using "," between the context managers, both __exit__ methods are called and original settings are restored.


joelostblom (approved):
Great catch! Thank you for noticing this.

Issue #2572, opened 2022-03-24
Document the differences between `value` and `datum`?


joelostblom:
I struggle to find a use case for `alt.value` since `alt.datum` was introduced in 4.2.0. It seems like the latter completely covers the function of the former + more. ~~If that is the case, should `value` be deprecated to encourage the usage of the more powerful `datum`? Deprecation or not,~~ I can add something to the documentation once I get more clarity on the relationship between the two.

I just found one difference that I don't know if it is significant. I seems like `datum` constructs an axis when used with a primitive value / constant, whereas as `value` does not (this could possibly be more convenient when trying to place in a single label in  a chart since you don't have to worry about hiding axes?):

```python
alt.Chart().mark_text().encode(
    x=alt.datum(0),
    y=alt.value(0),
)
```

![image](https://user-images.githubusercontent.com/4560057/159953181-cc22b1ab-f6c1-4760-853a-aab46fcfda29.png)

jakevdp:
This would be good to document. `alt.value()` specifies a constant range value, while `alt.datum` specifies a domain value. So `x=alt.value(100)` would place a mark 100 pixels from the left edge of the plot, while `x=alt.datum(100)` would place a mark corresponding to the location of x=100 on the x axis.

mattijn:
Fixed by https://github.com/altair-viz/altair/pull/2735

PR #2735, opened 2022-12-01
Improve encoding section

binste:
* Add explanation on difference between `datum` and `value`. Addresses #2572 
* Add explanation and description for the `detail` channel
* Remove any mentions of `key` and `description` channels as they are not very relevant for Altair and not properly documented anyway

I also considered creating separate encoding pages as in the vega-lite documentation. However, it felt like the most important content is already covered in the existing encoding section (with the additions mentioned above) as well as in [Customizing Visualization](https://binste.github.io/altair-docs/user_guide/customization.html), [Scale and Guide Resolution](https://binste.github.io/altair-docs/user_guide/scale_resolve.html), and [Top-Level Chart Configuration](https://binste.github.io/altair-docs/user_guide/configuration.html)


doc/user_guide/encoding.rst
```diff
@@ -677,3 +701,84 @@ While the above examples show sorting of axes by specifying ``sort`` in the
...
+
+Datum and Value
+~~~~~~~~~~~~~~~
+
+So far we always mapped an encoding channel to a column in our dataset. However, sometimes
+it is also useful to map to a single constant value. In Altair, you can do this with
+
+* ``datum``, which describes a constant data value encoded via a scale
+* ``value``, which describes an encoded constant visual value
+
```
joelostblom, line 713:
```suggestion
* ``datum``, which encodes a constant domain value via a scale using the same units as the underlying data
* ``value``, which encodes a constant visual value, using absolute units such as an exact position in pixels, the name or RGB value of a color, the name of shape,  etc

```

doc/user_guide/encoding.rst
```diff
@@ -677,3 +701,84 @@ While the above examples show sorting of axes by specifying ``sort`` in the
...
+``datum`` is particularly useful for annotating a specific data value. 
+For example, you can use it with a rule mark to highlight a 
+threshold value (e.g., 200 dollars stock price).
```
joelostblom, line 716:
```suggestion
threshold value (e.g., 300 dollars stock price).
```

doc/user_guide/encoding.rst
```diff
@@ -677,3 +701,84 @@ While the above examples show sorting of axes by specifying ``sort`` in the
...
+    lines = base.mark_line().encode(
+        x="date:T",
+        y="price:Q",
+        color="symbol:N"
+    )
+    rule = base.mark_rule(strokeDash=[2, 2]).encode(
+        y=alt.datum(300)
+    )
+
+    lines + rule
```
joelostblom, line 734:
```suggestion
    rule = base.mark_rule(strokeDash=[2, 2]).encode(
        y=alt.datum(300)
    )

    lines + rule

If we instead used `alt.value` in this example, we would position the rule 300 pixels from the left of the chart border rather than at the 300 dollars position:

.. altair-plot::

    rule = base.mark_rule(strokeDash=[2, 2]).encode(
        y=alt.datum(300)
    )

    lines + rule
```

joelostblom (changes requested):
This mostly looks good to me, I suggested a few clarifications. Regarding whether we need a separate encoding page for, I am not sure. Adding the full extent of what is in the VegaLite encoding pages does seem useful but also like a lot of work. I think it would be nice to have the general intro text for each encoding and maybe one simple example for each of them (could even be the same one from the example gallery that is currently linked in the overview table)? Could that be a good middle ground (it would also move and break up the lengthy options table)

If we don't create separate pages for each encoding, I think the headings "Encoding Channels" and "Encoding channel options" should be moved to last on the page. These are more like reference and they don't feel like narrative documentation that I am expected to read. I think the presence of long tables is a general problem with the Altair docs that I remember confused me when starting out. It is not clear that there will be more narrative content under these tables so we should move them last on each page or wrap them in a `details` tag so that they can be expanded on demand and it is easy to see the narrative content. Another notable case of this is the "Top-Level Chart Configuration" page.

binste:
What specifically do you mean with “encoding”? I’m a bit confused with the naming around the encoding concepts. An encoding channel such as “color” or an encoding channel **option** such as “axis”?

Initially, this also confused me in the vega-lite docs although I don’t have a better suggestion as I think it makes sense content-wise. Under transforms, we have sub pages about actual transforms, under marks the same, but for encoding the vega-lite documentation has sub pages for encoding channel **options** and not the channels themselves.

Creating separate pages per options would probably involve rewriting some of the existing pages I linked in the first comment. 

Good point about the tables. I’ll look into this next and will also consider the “Top-Level Chart Configuration” page.

binste:
Collapsing longer tables looks nice but I find it a bit less convenient as I can no longer search for a keyword such as a parameter name on the page without first expanding the tables.

What do you think about
* moving "Encoding Channels" and "Encoding channel options" to the end of the page (as suggested by you)
* and move "Altair Themes" out of "Top-Level Chart Configuration" to "Customizing Visualizations" as I think that section is a bit lost there. Also, in the beginning of that page it does not mention the themes currently but only the "configure_*" methods.

joelostblom:
> What specifically do you mean with “encoding”? I’m a bit confused with the naming around the encoding concepts. An encoding channel such as “color” or an encoding channel option such as “axis”? Initially, this also confused me in the vega-lite docs although I don’t have a better suggestion as I think it makes sense content-wise. Under transforms, we have sub pages about actual transforms, under marks the same, but for encoding the vega-lite documentation has sub pages for encoding channel options and not the channels themselves.

Great point, I think the VegaLite doc structure in this regard has been secretly confusing for me as well, but I have gotten used to it. I think explicit mention of the encoding channels with an example of each would be great and clearly separate them from the channel options. What do you think of this structure?

- Encodings - The main page
    - The main page contains the introductory example with a high level explanation of channels and channel options, and also the content from the Data Types and Shorthands headings.
        - The shorthand section contains a brief explanation of binning and aggregation and then links to the corresponding channel option pages
    - Channels - A subpage with little content, mostly a title for all the channel subpages (These could mostly be pretty simple and just show the examples linked in the current table? Maybe with a short intro statement?):
        - X
        - Y
        - Order (move the content from "ordering marks" here)
        - ...
    - Channel Options - A subpage with little content, mostly just a title for all the channel options subpages (These could eventually be more elaborate as per the VegaLite docs, but we can start out simple):
        - Aggregate
        - Bin
        - Sort (move the content from "sorting" here)
        - ...


> Collapsing longer tables looks nice but I find it a bit less convenient as I can no longer search for a keyword such as a parameter name on the page without first expanding the tables.

I was surprised that these are not searchable so I looked around a bit. It seems like expanding detail tags on search was made part of the official web spec last year https://github.com/whatwg/html/pull/6466 and that it works in [chrome already while there is an open issue for it in firefox](https://github.com/haskell/haddock/issues/1451#issuecomment-1056063675). With that background, the fact that there is still a search button for the docs in general, and that the "Top level chart config" page would benefit a lot from it (and I can't think of another workaround there), I think it could still be beneficial to have tables collapsed inside details tag by default, what do you think?

> and move "Altair Themes" out of "Top-Level Chart Configuration" to "Customizing Visualizations" as I think that section is a bit lost there. Also, in the beginning of that page it does not mention the themes currently but only the "configure_*" methods.

Agree with this move


close #2572

binste:
- Moved Altair Themes
- All object tables are now collapsed by default. I agree it does look much better and it's easier to navigate the pages! Good to know that browsers start to support the searchability

Regarding the restructuring of the encoding page. I like your suggestion of clearly separating channels and channel options. I pushed a first suggestion, you can preview them at https://binste.github.io/altair-docs/user_guide/encoding.html 

I for now sticked to the existing content and did not rewrite the tables for the channels into sections for each channel including the example, same for the options. For the channels I see that this could be nice but I'd prefer if this can be tackled in a separate PR. Happy to create an issue to keep track of this enhancement. For the options, I'm not sure how to approach this. How it's currently set up is great if you want to see which options are supported by a channel. This information is lost if we have one section for each option. Of course we could then include in that section which channels support it but I'd find that somewhat less convenient to navigate. Does that make sense for you?

doc/user_guide/encodings/index.rst
```diff
@@ -679,3 +411,101 @@ While the above examples show sorting of axes by specifying ``sort`` in the
...
+        y="price:Q",
+        color="symbol:N"
+    )
+    rule = base.mark_rule(strokeDash=[2, 2]).encode(
+        y=alt.datum(300)
+    )
+
+    lines + rule
+
+If we instead used ``alt.value`` in this example, we would position the rule 300 pixels from the bottom of the chart border rather than at the 300 dollars position:
```
joelostblom, line 446:
```suggestion
If we instead used ``alt.value`` in this example, we would position the rule 300 pixels from the top of the chart border rather than at the 300 dollars position. Since the default charts height is 300 pixels, this will show the dotted line just on top of the x-axis -line:
```

doc/user_guide/encodings/index.rst
```diff
@@ -679,3 +411,101 @@ While the above examples show sorting of axes by specifying ``sort`` in the
...
+
+.. altair-plot::
+
+    rule = base.mark_rule(strokeDash=[2, 2]).encode(
+        y=alt.value(300)
+    )
+
+    lines + rule
+
+You can also use ``datum`` with :class:`DateTime`, for example, to highlight a certain year.
```
joelostblom, line 456:
```suggestion
If we want to use ``datum``  to highlight a certain year on the x-axis,
we can't simply type in the year as an integer,
but instead need to use ``datum`` together with :class:`DateTime`.
```

doc/user_guide/encodings/index.rst
```diff
@@ -679,3 +411,101 @@ While the above examples show sorting of axes by specifying ``sort`` in the
...
+In addition, we will set the color for the rule to the same one as the line for the symbol ``MSFT``
```
joelostblom, line 457:
```suggestion
Here we also set the color for the rule to the same one as the line for the symbol ``MSFT``
```

doc/user_guide/encodings/channels.rst
```diff
@@ -0,0 +1,187 @@
...
+    from vega_datasets import data
+
+    barley = data.barley()
+
+    alt.Chart(barley).mark_bar().encode(
+        x='variety:N',
+        y='sum(yield):Q',
+        color='site:N',
+        order=alt.Order("site", sort="descending")
+    )
```
joelostblom, line 145:
```suggestion
    )

If we want to sort stacked segments in a custom order, we can `follow the approach in this issue comment <https://github.com/altair-viz/altair/issues/245#issuecomment-748443434>`_, although there might be edge cases where this is not fully supported. This also makes the order of the segments align with the order colors shows up in a legend that uses custom sorting for the color domain.

```

What do you think of including this?

doc/user_guide/customization.rst
```diff
@@ -566,3 +566,145 @@ it is rendererd, you can set ``width`` or ``height`` to the string ``"container"
 Note that this will only scale with the container if its parent element has a size determined
 outside the chart itself; For example, the container may be a ``<div>`` element that has style
 ``width: 100%; height: 300px``. 
+
```
joelostblom, line 569: This page is so nice to read with the collapsed tables! I noticed there were just a few sections that were missing a sentence introducing the tables, could we add these:

Can we use the same wording for projection and selection as for scale ranges ("Projections can be configured using..."):

image

The table in the Header section is missing a leading sentence, could we add something like "Additional properties are summarized in the following table:"

image

doc/user_guide/encodings/channels.rst
binste: Cool, didn't know about this created field! I'll add your suggestion

binste:
Great, thanks for the suggestions and for setting up the separate issue! I incorporated all changes and agree with you that separate subpages could be nice but maybe not the most value-adding contribution right now.

Ready to merge from my side :)

joelostblom (approved):
Looks great! Thanks for improving this section of the docs, I think it will be a big help to many people!

PR #2737, opened 2022-12-01
Various smaller updates to documentation (harmonizing title cases, vega version numbers, browser compatibility, ...)

binste:
Most of the changes are smaller updates. Most noteworthy is the move of the "Note on browser compatibility" section in "Times and Dates" to the bottom of that page. I still think it's great that its documented but for most users this is not relevant anymore as all modern browsers are now es6 compatible for quite a while, see [https://caniuse.com/\?search\=es6](https://caniuse.com/%5C?search%5C=es6)


joelostblom (approved):
LGTM! Thank you!

Issue #2736, opened 2022-12-01
Some tables in documentation are cut off


binste:
Some tables in the new documentation design are too wide to be displayed. For example https://binste.github.io/altair-docs/user_guide/encoding.html#color-fill-and-stroke

![image](https://user-images.githubusercontent.com/23366411/205069201-f1dee255-70d8-4a19-af4d-18a7a2ff2544.png)

This happens because some types in the second column are too big and so they push the table beyond 100% width:

![image](https://user-images.githubusercontent.com/23366411/205069358-e79ab463-6193-426d-92fb-34482e4be008.png)

I'm not very familiar with the layout and how these tables are generated so I don't know how to fix this. Some ideas:

* Is it possible to shorten the names of these types? The long ones are not very useful anyway in my opinion. I'm aware that the tables are autogenerated so it would probably require some adjustment in that Sphinx directive. 
* Make the table scrollable horizontally. This is still not very convenient as the scrollbar will be at the bottom and some of the tables are very long.
* Shorten section titles in left navigation bar ("Interactive Charts: Parameters, Conditions, Bindings" -> "Interactive Charts", I prefer this anyway) and then make it narrower. This won't solve it completely but helps.
* If the layout is responsive, we could make the tables wider on bigger screens. On a 13'' laptop screen the documentation has a good size but on a 27'' 2560x1440 monitor it does look rather narrow:

![image](https://user-images.githubusercontent.com/23366411/205070651-a33d4993-4de1-4a08-9106-a71b9c1bd995.png)

---

Please follow these steps to make it more efficient to solve your issue:

- [x] Since Altair is a Python wrapper around the Vega-Lite visualization grammar, [most bugs should be reported directly to Vega-Lite](https://github.com/vega/vega-lite/issues). You can click the Action Button of your Altair chart and "Open in Vega Editor" to create a reproducible Vega-Lite example and see if you get the same error in the Vega Editor.
- [x] Search for duplicate issues.
- [x] Use the latest version of Altair.

joelostblom:
This caused an issue in the old documentation as well where it was difficult (but still possible) to read the rightmost column as it became very narrow:

![image](https://user-images.githubusercontent.com/4560057/205127401-657abc40-6188-4dbb-a998-f2e438a5a4eb.png)

I don't know if we can make the names shorter, but what do think of making the left column scrollable horizontally and limiting the max width?

[...]

PR #2746, opened 2022-12-06
Fix cut off property tables

binste:
PR addressing #2736. @joelostblom I restricted the type column to max. 200px after which the table looks pretty good:

image

However, I did not find a good way to make that column scrollable so I had to cut off overflowing values. All examples I found so far online had to do some fiddling with margins and paddings so that columns don’t fold into each other. E.g. https://www.w3docs.com/snippets/html/how-to-create-an-html-table-with-a-fixed-left-column-and-scrollable-body.html and https://stackoverflow.com/questions/3402295/html-table-with-horizontal-scrolling-first-column-fixed

I’m not very experienced with CSS so there might be a way to accomplish this after all. But I also don't think it's bad if the type is cut off as the very long values are not informative for most users anyway.

As mentioned in the issue, I would also be in favour of making the left sidebar a bit narrower:
image (`max_width: 20%` instead of `25%`)
To prevent some titles from folding, they could be shortened. I did not yet include this change in the PR but happy to add.


doc/_static/custom.css
```diff
@@ -25,6 +25,11 @@ table.docutils td li {
 
 }
 
+table td.vl-type-def {
+  max-width: 200px;
+  overflow-x: hidden;
```
joelostblom, line 30:
```suggestion
  overflow-x: clip;
```

For me, "hidden" shows overlapping text.

binste:
I played a bit around with the widths and I think 170px for the type column is a good value (at least based on the tables on the encoding page) as below a few types are cut off and going higher doesn't seem to be necessary as longer parameters are cut off anyway.

Regarding the first column, I'm not sure if we want to cut off the names of parameters which can happen by reducing max-width? Now it dynamically sizes the width based on the content and some parameter names need the space. 

We could of course reduce the padding (green) around the name although I like the current version:

image

With 20% navbar as of latest commit:

image

Thoughts on shortening the 3 titles which take up two lines? Could remove the parts after the semicolon. Looks cleaner to me but maybe that's just personal taste.

joelostblom:
> Regarding the first column, I'm not sure if we want to cut off the names of parameters which can happen by reducing max-width? Now it dynamically sizes the width based on the content and some parameter names need the space.

I agree we don't want to cut these off. It looked to me like there was plenty of white space beyond even the longest parameter name which is why I thought there might be room to reduce it, but if you looked into it and it is actually adapting to the longest name then let's leave it as is.

> Thoughts on shortening the 3 titles which take up two lines? Could remove the parts after the semicolon. Looks cleaner to me but maybe that's just personal taste.

I mostly agree with this an almost made a PR for it myself the other day. I think `Interactive Charts` and `Altair Internals` give a good sense of what those sections are about. However I am not sure if it is intuitive to find faceting under `Compound Charts`. Yes, it is technically correct, but faceting is so common that I would expect people to be looking for the name directly or something like "Small Multiples". I couldn't think of a better name though; e.g. anything mentioning "Subplots" would not be correct for the layering charts. Maybe "Compound Charts & Facets" just to make it easy to find even if it is a bit redundant? Overall I am probably OK just cutting down the names as you suggested, but if you have any thoughts of improving the "Compound Charts" title, that would be welcome.

binste:
Good point. I like "Layered and Multi-View Charts" (similar to [Composing Layered & Multi-view Plots](https://vega.github.io/vega-lite/docs/composition.html) in the vega-lite docs), combines both the "subplots" and "layering". And I'd associate multi-view with faceting more quickly then with compound.

joelostblom (approved):
Nice, I like that title too and agree it is easier to associate with faceting. Thank you!

PR #2747, opened 2022-12-06
Update _magics.py

johnmarkpittman:
switched yaml.loader to use SafeLoader instead of FullLoader as FullLoader is slated for deprecation. https://github.com/yaml/pyyaml/issues/420#issuecomment-696752389


mattijn:
Thanks for your PR @johnmarkpittman.

Issue #2738, opened 2022-12-01
DOC: Proposal to remove 'Help' pages and gather and update content on 'Large Datasets'


binste:
I have two suggestions how we could improve the documentation even further. I'm happy to create a PR for both of them but I first wanted to get some input if I'm on the right track and if this is of general interest.

### Remove the "Help" pages
These pages are currently nested under "More" -> "Help" in the top navigation bar. There content is either duplicated or could be reorganised which would lead to a better structure in my opinion:

* [Getting help](https://binste.github.io/altair-docs/user_guide/getting_help.html#) is already fully contained in the [start page](https://binste.github.io/altair-docs/index.html) (last two paragraphs)
* [Display Troubleshooting](https://binste.github.io/altair-docs/user_guide/troubleshooting.html) could be added as a "Troubleshooting" section at the end of [Displaying Altair Charts](https://binste.github.io/altair-docs/user_guide/display_frontends.html). As a user, this is where I would look for it in the first place. Thanks to the great "table of contents" on the right side of a page, I think the navigation on "Displaying Altair Charts" would still be good despite the increased length
* [FAQ](https://binste.github.io/altair-docs/user_guide/faq.html):
  * "Does Altair work with IPython Terminal/PyCharm/Spyder/<my favorite IDE>": all of the information here is already contained in [Displaying Altair Charts](https://binste.github.io/altair-docs/user_guide/display_frontends.html)
  * "I tried to make a plot but it doesn’t show up": All information is contained in "Display Troubleshooting"
  * The remaining questions deal with large datasets. See next section on how this could be handled

### Gather content on "Large Datasets" in one place
There is already some content in the documentation on how to deal with larger datasets. There have also been some new developments in third-party packages such as vl-convert and vegafusion which open up new possibilities which are not yet documented. I think it would make sense to gather this information in one place, especially as I don't think it is obvious and there are many options by now with various tradeoffs. I'd suggest to do this in a new page called "Large Datasets" under "Advanced Usage" and link to it from [Specifying Data](https://binste.github.io/altair-docs/user_guide/data.html#)

Existing pages from which some content could be consolidated:
* Last 2 questions in [FAQ](https://binste.github.io/altair-docs/user_guide/faq.html): Explain problem with large notebooks and  mention disabling of maxrowserror, json transformer, etc.
[...]

PR #2755, opened 2022-12-15
Consolidate docs and add section on Large Datasets

binste:
Builds on discussion in #2738

@joelostblom I would not merge this yet but I already wanted to get your feedback, especially in case I don't get to the items below in the coming days. Outstanding from my side:

- [x] Find a good way to keep a Help indicator but deduplicate information which is currently on start page and on "Getting Help" -> Opted for moving "Getting Help" to "Getting Started" and removing some of the duplicated content on the start page.
- [x] Proofread the new content in "Large Datasets" and potentially extend
- [x] Provide examples on how to precalculate boxplots in Python and then plot in Altair. Other charts are more straightforward but I think this one would be helpful to have a reference implementation in the docs. For boxplot, see [Box Plot with Pre-Calculated Summaries](https://vega.github.io/vega-lite/docs/boxplot.html#box-plot-with-pre-calculated-summaries) on how to plot it


joelostblom:
Thanks for starting this PR! I have a busy few days with work and traveling coming up so I am not sure exactly when I will get to this, but one thing I wanted to mention is that there is an Altair example for a box plot with pre-calculated summaries here https://stackoverflow.com/questions/59194197/how-to-change-altair-boxplot-infobox-to-display-mean-rather-than-median

binste:
Sure, thanks for the heads-up and no worries! The PR is ready to review and the docs are built at https://binste.github.io/altair-docs/index.html

mattijn:
I'd had a look to this as well and I think this looks great! Really helpful. Merging for now.

mattijn:
Hi @binste, I noticed an issue with the Altair documentation site when viewed on a mobile phone. It appears that there may be a problem with the CSS styling, although I'm not sure if this is a new issue or if it has been happening for a while. I've included a screenshot for reference. This issue is not related to the current pull request.

image

PR #2756, opened 2022-12-17
Improve waterfall example

binste:
Addresses open points from #2621. I also further simplified the code, hopefully making the example a bit easier to understand.

image


joelostblom (approved):
LGTM! Thank you!

PR #2757, opened 2022-12-17
Run tests with Python 3.11

binste:
Altair works without issues for me under 3.11 so I added 3.11 in the build GH workflow and the classifier for then Python package.


joelostblom:
Thanks! I think it makes sense to add this to the workflow so that we can see if there are any errors coming up before officially adding support in a future release.

PR #2415, opened 2021-02-22
Update resource section

joelostblom:
Somehow I missed that this section already existed when opening #2414. It was quite out of date, so I removed packages that haven't seen updates in a few years and added new ones as well a section on the learning resources. I also attempted to make this more discoverable by adding links from the Overview and Gallery pages.

close #2414

PR #2762, opened 2022-12-19
Add docs on displaying Altair charts in Streamlit apps

snehankekre:
Adds a new "Streamlit apps" section under Displaying Altair Charts -> Working in non-Notebook Environments. 

Streamlit is a  popular non-Notebook environment for Python web apps. It has a [`st.altair_chart`](https://docs.streamlit.io/library/api-reference/charts/st.altair_chart) command to display Altair charts within Streamlit apps. 

With Streamlit's growing popularity among data enthusiasts as a UI layer on top of Python, and it's popularity among Altair users, I think it makes a good addition to the Altair docs.  Full disclosure: I'm the maintainer of the [Streamlit docs](https://github.com/streamlit/docs).

A similar case can be made about Gradio (another open-source Python library to build web apps), Dash, etc, but those can come as follow-up, tool-specific PRs like this one.

### DIFF
--- a/doc/user_guide/display_frontends.rst
+++ b/doc/user_guide/display_frontends.rst
@@ -191,6 +191,30 @@ Manual ``save()`` and display
 If you would prefer, you can save your chart to a file (html, png, etc.) first and then display it.
 See :ref:`user-guide-saving` for more information.
 
+
+Streamlit apps
+~~~~~~~~~~~~~~
+Streamlit_ is an open-source Python library for creating web apps. To display Altair charts in a
+Streamlit app, pass the chart object to the ``st.altair_chart`` command::
+
+    import streamlit as st
+    import altair as alt
+
+    # load a simple dataset as a pandas DataFrame
+    from vega_datasets import data
+    cars = data.cars()
+
+    chart = alt.Chart(cars).mark_point().encode(
+        x='Horsepower',
+        y='Miles_per_Gallon',
+        color='Origin',
+    ).interactive()
+
+    st.altair_chart(chart)
+
+To learn more about Altair in Streamlit, see the 
+`st.altair_chart documentation <https://docs.streamlit.io/library/api-reference/charts/st.altair_chart>`_.
+
 .. _renderer-api:
 
 Renderer API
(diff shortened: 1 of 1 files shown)


joelostblom:
For reference, there is some discussion of this in https://github.com/altair-viz/altair/pull/2415. We likely don't want to add code examples of all the dashboarding libraries to the Altair docs since that would mean making all those packages dependencies of building the docs (or if we just include the code but don't execute, we would still need to keep it up to date and I am not sure that adds anything over what is already in the docs of the dashboarding library).

I do think it would be very helpful to mention which dashboarding packages work well with Altair and link to Altair examples in the dashboarding libraries documentation. That's what I suggested in the other PR where I also included a link to a short example I made comparing the most common ones https://stackoverflow.com/a/67428170/2166823. Thoughts on this approach?

snehankekre:
Thanks for that context, @joelostblom! 

> We likely don't want to add code examples of all the dashboarding libraries to the Altair docs since that would mean making all those packages dependencies of building the docs...

This PR only includes the code and doesn't execute, so Altair doesn't have to include streamlit in its dependencies. 

> (or if we just include the code but don't execute, we would still need to keep it up to date and I am not sure that adds anything over what is already in the docs of the dashboarding library).

I resonate with the issue of keeping the code up to date. I'd like to note though that Streamlit does not make breaking API changes, especially after our 1.0.0 milestone in 2021. The code snippet from this PR works regardless of the version of Streamlit used. As such, there would be no reason to update the code snippet.

The value I see this PR adding is in letting Altair users know that Streamlit is one such non-notebook environment where one can display Altair charts. Just as VSCode is included on the page out of all other IDEs, I assumed Streamlit could similarly be included.

> I do think it would be very helpful to mention which dashboarding packages work well with Altair and link to Altair examples in the dashboarding libraries documentation. 

That said, mentioning which dashboarding packages work well with Altair and linking to Altair examples in the dashboarding libraries documentation as part of #2415 would be a great common ground! That approach seems very reasonable. Streamlit could be in a section similar to [`Altair in R`_](https://github.com/altair-viz/altair/blob/60fe05b5cabe2d9602ac4181e4877a6d1df54e35/doc/user_guide/ecosystem.rst#altair-in-r). 

Do you recommend I close this PR in favor of #2415, and submit a commit to that PR?

joelostblom:
I think we can keep this PR open for now and I will either close it when implementing the other one, but it is a good reminder to not forget to include the dashboard section there. I think we will probably just add them as a section, but maybe we will decide actually including an example and then I can merge this or ping you. Thanks again for contributing!

binste:
Closing this in favour of #3299

PR #2769, opened 2022-12-21
WIP: type hints for tab completion

ChristopherDavisUCI:
This is an attempt to have tab completion for the property setters by using type hints, as in [this comment](https://github.com/altair-viz/altair/pull/2592#issuecomment-1335331116).  I've never worked with type hints before, so I don't know if there is any issue with this approach.

@joelostblom and @mattijn could you please take a look and let me know your impressions?


mattijn:
Hi @ChristopherDavisUCI, seems you've made great progress! 

I tried the following in jupyterlab:
```python
import altair as alt

    alt.X().axis(None).title(None)
)
```
I was only able to tab-complete the first encoding channel option, here `axis()`, and not for the following chained encoding channel options, here `title()`. Can you confirm if this also happens on your side?

joelostblom:
@ChristopherDavisUCI Great! It might take some time for me before looking closer at this, sorry about that.

@mattijn What you describe sounds like it could be related to this issue in jupyterlab https://github.com/jupyterlab/jupyterlab/issues/12570

ChristopherDavisUCI:
@mattijn @joelostblom Thanks for checking it out!  @mattijn yes I can confirm that doesn't work on either Jupyter notebook or Jupyter lab for me.  I wasn't careful about specifying the return type of these methods, so maybe that will help.

mattijn:
Each encoding channel option should return the encoding channel class in which it is defined, isn't it? Is this possible with these generic protocol class objects?

ChristopherDavisUCI:
Closing in favor of #2770 which addresses the issues raised above.

PR #2770, opened 2022-12-23
WIP: Tab completion 2

ChristopherDavisUCI:
Branch off of #2592 by @jakevdp and @joelostblom.  Adds type hints for the setter methods to help with tab completion.

@mattijn Can you please take another look and see if this new version addresses your concern from https://github.com/altair-viz/altair/pull/2769#issuecomment-1363187638?

I've never used type hints before, so I'm very happy for any comments or corrections!


mattijn:
Different approach I see! 

Tab-completion works for the chained options👏:
image
For the first chained method the docstring/signaturea appear (using shift+tab in jlab)
image
But signature won't appear within the second chained method🫣:
image

Almost there it seems!

mattijn:
When looking to the code change I notice the following:
```python
    @overload
    def aggregate(self, _: str, **kwds) -> 'Angle':
        ...

    @overload
    def aggregate(self, argmax=Undefined, **kwds) -> 'Angle':
        ...

    @overload
    def aggregate(self, argmin=Undefined, **kwds) -> 'Angle':
```

It seems it defines somehow multiple identical named methods with different arguments combinations.

ChristopherDavisUCI:
Thanks @mattijn!  I didn't know about this `@overload` decorator until a few hours ago, but my understanding is that it is specific to type hinting (it is defined in the `typing` module) and it is a way of offering different possible signatures for the same function.  For example, calling `bin(True)` is valid, as is calling `bin(extent=[-4, 4], step=0.5)`, and so this is a way to tell the (linter? type checker?) that both are okay.  In the previous PR, I had used `Union` (also from the `typing` module), but I think this `@overload` approach feels less hacky.

Notice that none of these lines you posted are defining anything; they're just indicating what types of arguments are valid.

Does that make sense?

I will think about the signature question (but it might be more a question related to @joelostblom's earlier commits).

mattijn:
Thanks for the explanation @ChristopherDavisUCI! Will give these changes a few more trials, but nice👍.

joelostblom:
> Or is this because of a limitation within ipython? See https://github.com/jupyterlab/jupyterlab/issues/12570#issuecomment-1363540055.

Yes, I think that is the issue and not something we can easily fix on the Altair side of things I believe =/ I'm really quite surprised that this isn't considered a more important issue as method chaining is used so often but maybe it is a really complex problems to solve?

mattijn:
Thanks @joelostblom, are you aware of a GitHub issue raising this somewhere higher up, like in ipython? 
While playing with this PR I missed the docstrings:/.

joelostblom:
I'm not unfortunately. I wasn't sure what would be a good way to test it in Ipython since the shift+tab shortcut is for JupyterLab only and the tab completion works as expected in both jupyterlab and ipython

jcmkk3:
> I'm not unfortunately. I wasn't sure what would be a good way to test it in Ipython since the shift+tab shortcut is for JupyterLab only and the tab completion works as expected in both jupyterlab and ipython

@joelostblom I tested your example from the jupyterlab issue linked above with **xeus-python**. I get the same behavior with the `ipykernel` and `XPython` kernels, but if I use the `XPython Raw` kernel (doesn't use ipython in any way), then it works. I think that confirms that it is a ipython/ipykernel issue and not a jupyter issue.

mattijn:
Thanks @jcmkk3 for helping. Appreciated! I opened the following issue at the ipython/ipykernel repo: https://github.com/ipython/ipykernel/issues/1065

mattijn:
I also noticed that the docstrings are serialised differently when using different kernels. 
This is the IPython kernel:
drawing

And this is the XPython Raw kernel:
drawing

As this PR is about completion. I tried to complete with `stack`. The `stack` encoding channel option is valid for the `X` encoding channel, but not for the `Color` encoding channel.
image
I could not complete `stack` using the `Color` channel, but I could with the `X` channel. So that works as expected.

I tried to chain the same method and define it within the channel to check what will return:
image

It seems the last defined method is used. That is fine.

One question @ChristopherDavisUCI. I noticed this:
```python
"boolean": "bool",
"array": "list",
"null": "Type[None]"
```
Why this `null` needs to be defined as `Type[None]` and not just `None`?

ChristopherDavisUCI:
Hi @mattijn, thanks for looking at it as always!

I asked chatgpt and it seems like you're right that it is more clear to use `None`.  This is counter-intuitive to me, since to me `None` does not seem to be in the same family as `bool` or `list`.  (I had originally tried `NoneType` and was surprised to learn that `NoneType` does not seem to be defined.  Then I had tried using `type(None)`, but apparently you are not supposed to use the function `type` when writing type hints.  That was how I landed on `Type[None]`.)

I'll try to make this change to `None` later.

mattijn:
Thanks chatgpt, this feels reassuring🤗

ChristopherDavisUCI:
I made that small change just now @mattijn.  Is there anything else I can do on this PR?

mattijn:
Hi @ChristopherDavisUCI, I collected the content of three comments from @joelostblom that still contain useful information for a few changes to this PR. Would this be possible for you to work on?

1. https://github.com/altair-viz/altair/pull/2592#issuecomment-1340219884

>     With documentation I merely meant explanation of this approach within the documentation. How is this different vs current approach.
> 
> Yup I will add a section in the docs about this. Before releasing the next version, I think we should have a discussion of whether the method based approach should be the default (which I think sounds reasonable if we don't come across any drawbacks), but for now I will write the docs page as if it isn't.
> 
>     I don't think we should already update all the examples in the example gallery in this PR.
> 
> I think it is a good idea to do all the conversions now just so that we can see that the entire test suite is passing with the new syntax, and we are not overlooking anything. We don't need to merge them in this PR (or could merge them into a separate `v5` folder).

* * *

2. https://github.com/altair-viz/altair/pull/2592#issuecomment-1341321041

>     But the side effect is that the enhancement in this PR now becomes the main approach to specify Altair charts.
> 
> I agree that this is something we should discuss (although I think it is probably a good idea since the new syntax is more convenient to both type and read). We should probably make a new `v5` folder for these update examples so that we still keep the old ones for testing the old syntax. I can do that before merging this, I just wanted to make it easy to see what was changed in each example so I am keeping the names the same for now. Later, we can decide if we want to show both the new and old syntax in the gallery in a tabbed interface or keep the old examples in a hidden folder just for testing purposes.

* * *

3. https://github.com/altair-viz/altair/pull/2592#issuecomment-1341965280

> A note that there is a small inconsistency in that this new method based syntax does not apply to the `datum` function. We can do `alt.datum(4, scale=None)` but not `alt.datum(4).scale(None)`. This doesn't seem like a big deal to me but thought I would record it since `datum` can be used instead of the encoding class `alt.X()` etc so it might be expected that the syntax is similar. Relevant gallery example here https://joelostblom.github.io/altair-docs/gallery/pacman_chart.html

* * *

For me, I would favour a `5.0rc1` version where the possibility of this PR is explained in a section in the docs without merging all the changes of the examples in this PR. 

Subsequently we can gather feedback of users how to proceed best with the trade-off between not having docstrings and function signatures when chaining methods and the reduced character typing of specs.

ChristopherDavisUCI:
Thanks @mattijn, yes, it should be very possible for me to work on this.  At least the third one should be no problem, and I'll let you know if I get stuck on 1 or 2.

binste:
Very exciting to see the progress here, thanks everyone involved!

I like the idea of explaining the new syntax in a separate section in the docs but not modifying the rest of the docs and the examples just yet. What do you think about making this the actual `5.0` release instead of only a release candidate? Once there is enough positive feedback from users the approach could be made the standard one by updating the documentation everywhere and publishing a minor release (there will probably be a few bug fixes and new features until then anyway), e.g. `5.1`. 

As it's not a backwards-incompatible change, this would be fine following [semantic versioning](https://semver.org) and it would allow us to already publish the new documentation (don't think we would update it for an rc version?). Furthermore, it would be great to get a proper `5.0` release out soon as it will contain quite a few important bug fixes (e.g. #2771 <- fresh Altair v4 installation with recent jsonschema versions will break some charts) and new features (vl-convert-python support, vega-lite 5, etc.).

ChristopherDavisUCI:
Is "update from upstream/master" a good thing to do before working on the docs?  Or is bringing in those 19(?) commits too messy?  Should I instead work on the docs in a different branch off of the master branch?

mattijn:
No is fine, you can update from master. But if you have to undo all commits from this PR regarding the changes to the examples, it might be easier to open a new PR in a new branch?

ChristopherDavisUCI:
@mattijn I thought your [comment 3](https://github.com/altair-viz/altair/pull/2770#issuecomment-1366124482) about `alt.datum` would be the easiest to implement, but I think I was over-confident.  This `alt.datum` is defined pretty minimally https://github.com/altair-viz/altair/blob/f288e2df580b8a2059f086dc8f67e2b11e340598/altair/expr/core.py#L4-L23 as something that returns a dictionary when called.  For example, `type(alt.datum(5))` is `dict`.

To me, at this point it doesn't seem worth rewriting this `alt.datum` in a way that the property setters work on it.  (In the best case scenario, we would need `x=alt.datum` to have different methods than `color=alt.datum`.)  If someone is trying to use `x=alt.datum(5).scale(...)`, we can always instruct them to use `x=alt.XDatum(5).scale(...)` instead.

Does that seem reasonable?

I should have some time today to think about your other comments 1 and 2.

mattijn:
That is unfortunate. I agree with you that it is not worth too much effort at the moment. Just mentioning this limitation is enough for now!

ChristopherDavisUCI:
Thanks @mattijn!  I've put the changed examples into an `attribute_syntax` folder and replaced them with whatever was in the main branch.  Is this along the lines of what you had in mind in [comment 2](https://github.com/altair-viz/altair/pull/2770#issuecomment-1366124482)?

mattijn:
👍 perfect

ChristopherDavisUCI:
Hi @mattijn and @joelostblom, I added some basic documentation in this last commit (briefly explaining the setter methods), and at the end mentioning that it does not work with `alt.datum` (nor with `alt.value`).  I'm happy for any comments!

@binste I don't have a great overview of the current layout of the docs, so feel free to move this to another location or make changes to make it more inline with what you have.

doc/user_guide/encodings/index.rst
```diff
@@ -44,6 +44,44 @@ the chart. For example, below we adjust the y-axis title and increase the step b
     )
 
 
+.. _method-based-attribute-setting:
+
+Alternative Syntax
+~~~~~~~~~~~~~~~~~~
+
+Altair 5.0 introduced an alternative method-based syntax for setting channel attributes.  In the example above, the ``axis`` attribute of the x channel encoding is set using the ``axis`` keyword argument: ``x=alt.X('Horsepower', axis=alt.Axis(tickMinStep=50))``.  To define the same :class:`X` object using the method-based syntax (new as of Altair 5.0), we can use ``x=alt.X('Horsepower').axis(tickMinStep=50)``.  In other words, the use of the ``axis`` keyword argument is replaced by the use of the ``axis`` method.
```
binste, line 52:
```suggestion
Altair 5.0 introduced an alternative method-based syntax for setting channel options.  In the example above, the ``axis`` option of the ``x`` channel encoding is set using the ``axis`` keyword argument: ``x=alt.X('Horsepower', axis=alt.Axis(tickMinStep=50))``.  To define the same :class:`X` object using the alternative method-based syntax, we can use ``x=alt.X('Horsepower').axis(tickMinStep=50)``.  In other words, the use of the ``axis`` keyword argument is replaced by the use of the ``axis`` method.
```

doc/user_guide/encodings/index.rst
```diff
@@ -44,6 +44,44 @@ the chart. For example, below we adjust the y-axis title and increase the step b
...
+
+The same technique works with all encoding channels and all channel attributes.  For example, notice how we make the analogous change with respect to the ``title`` attribute of the y channel.  The following produces the same chart as the previous example.
```
binste, line 54:
```suggestion
The same technique works with all encoding channels and all channel options.  For example, notice how we make the analogous change with respect to the ``title`` option of the ``y`` channel.  The following produces the same chart as the previous example.
```

doc/user_guide/encodings/index.rst
```diff
@@ -44,6 +44,44 @@ the chart. For example, below we adjust the y-axis title and increase the step b
...
+
+.. altair-plot::
+    import altair as alt
+    from vega_datasets import data
+    cars = data.cars()
+
+    alt.Chart(cars).mark_point().encode(
+        x=alt.X('Horsepower').axis(tickMinStep=50),
+        y=alt.Y('Miles_per_Gallon').title("Miles per Gallon"),
```
binste, line 63:
```suggestion
        y=alt.Y('Miles_per_Gallon').title('Miles per Gallon'),
```

doc/user_guide/encodings/index.rst
```diff
@@ -44,6 +44,44 @@ the chart. For example, below we adjust the y-axis title and increase the step b
...
+        color='Origin',
+        shape='Origin'
+    )
+
+These attribute-setter methods can also be chained together, as in the following, in which we set the ``axis``, ``bin``, and ``scale`` attributes of the x channel by using the corresponding methods (``axis``, ``bin``, and ``scale``).  We break the ``x`` definition over multiple lines to improve readability.  (This is valid syntax because of the enclosing parentheses from ``encode``.)
```
binste, line 68:
```suggestion
These option-setter methods can also be chained together, as in the following, in which we set the ``axis``, ``bin``, and ``scale`` options of the ``x`` channel by using the corresponding methods (``axis``, ``bin``, and ``scale``).  We can break the ``x`` definition over multiple lines to improve readability.  (This is valid syntax because of the enclosing parentheses from ``encode``.)
```

doc/user_guide/encodings/index.rst
```diff
@@ -44,6 +44,44 @@ the chart. For example, below we adjust the y-axis title and increase the step b
...
+    import altair as alt
+    from vega_datasets import data
+    cars = data.cars()
+
+    alt.Chart(cars).mark_point().encode(
+        x=alt.X('Horsepower')
+                .axis(ticks=False)
+                .bin(maxbins=10)
+                .scale(domain=(30,300), reverse=True),
+        y=alt.Y('Miles_per_Gallon').title("Miles per Gallon"),
```
binste, line 80:
```suggestion
        y=alt.Y('Miles_per_Gallon').title('Miles per Gallon'),
```

doc/user_guide/encodings/index.rst
```diff
@@ -504,6 +542,20 @@ the color scale used for the lines, you can use ``value``, e.g. ``alt.value("red
 
     lines + rule
 
+One caution is that ``alt.datum`` and ``alt.value`` do not possess the (newly introduced as of Altair 5.0) method-based attribute setter syntax described in :ref:`method-based-attribute-setting`.  If you are using ``alt.datum`` for the y channel encoding (for example) and you wish to use an attribute setter method (e.g., ``scale``), then you should probably be using :class:`YDatum` instead.  Here is a simple example.
```
binste, line 545:
```suggestion
One caution is that ``alt.datum`` and ``alt.value`` do not possess the (newly introduced as of Altair 5.0) method-based syntax to set channel options described in :ref:`method-based-attribute-setting`. For example, if you are using ``alt.datum`` for the ``y`` channel encoding and you wish to use an option setter method (e.g., ``scale``), then you can use :class:`YDatum` instead.  Here is a simple example.
```

doc/user_guide/encodings/index.rst
```diff
@@ -504,6 +542,20 @@ the color scale used for the lines, you can use ``value``, e.g. ``alt.value("red
...
+    import altair as alt
+
+    bar = alt.Chart().mark_bar().encode(
+        y=alt.YDatum(220).scale(domain=(0,500)),
+        color=alt.value("darkkhaki")
+    )
+
+    bar
+
+If you were to instead use ``y=alt.datum(220).scale(domain=(0,500))``, an ``AttributeError`` would be raised, due to the fact that ``alt.datum(220)`` is simply a Python dictionary and does not possess a ``scale`` attribute.  If you insisted on producing the preceding example using ``alt.datum``, one option would be to use ``y=alt.datum(220, scale={"domain": (0,500)})``.  Nevertheless, the ``alt.YDatum`` approach is strongly preferred to this "by-hand" approach of supplying a dictionary to ``scale``.  As one benefit, tab-completions are available using the ``alt.YDatum`` approach.  For example, typing ``alt.YDatum(220).scale(do`` and hitting ``tab`` in an environment such as JupyterLab will offer ``domain``, ``domainMax``, ``domainMid``, and ``domainMin`` as possible completions.
```
binste, line 558:
```suggestion
If you were to instead use ``y=alt.datum(220).scale(domain=(0,500))``, an ``AttributeError`` would be raised, due to the fact that ``alt.datum(220)`` simply returns a Python dictionary and does not possess a ``scale`` attribute.  If you insisted on producing the preceding example using ``alt.datum``, one option would be to use ``y=alt.datum(220, scale={"domain": (0,500)})``.  Nevertheless, the ``alt.YDatum`` approach is strongly preferred to this "by-hand" approach of supplying a dictionary to ``scale``.  As one benefit, tab-completions are available using the ``alt.YDatum`` approach.  For example, typing ``alt.YDatum(220).scale(do`` and hitting ``tab`` in an environment such as JupyterLab will offer ``domain``, ``domainMax``, ``domainMid``, and ``domainMin`` as possible completions.
```

doc/user_guide/encodings/index.rst
```diff
@@ -44,6 +44,44 @@ the chart. For example, below we adjust the y-axis title and increase the step b
     )
 
 
+.. _method-based-attribute-setting:
+
+Alternative Syntax
+~~~~~~~~~~~~~~~~~~
```
binste, line 50:
```suggestion
Alternative Syntax for Channel Options
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

binste (changes requested):
I like the placement in the docs and how it builds on the previously introduced example.

ChristopherDavisUCI:
Thanks for all the feedback @binste!

@mattijn please let me know if there is anything else here you'd like me to add or change!

mattijn:
I've modified the tests so the examples using the attribute syntax method are integrated within the tests. All are passing, so that is good. But what I don't like, which becomes much clear in this PR, is that all tests(-files) are included as part of the Altair package. A few test files are fine, but now, there are many. Maybe this is a good moment to extract the test-files from the Altair folder into a specific test folder? Maybe good to do this in a separate PR, but I'm not sure how interwoven the tests are with the Altair package.

mattijn:
Hi @ChristopherDavisUCI! First, best wishes for 2023! 
I tried to rebase this PR with master, but I ran into some conflicts. I tried to solve these conflicts, but I was not able to pass commit `0dbbd260d18d229dc349142fc2ef6da6619dcf77` with the error message `could not apply 0dbbd260... Lifting parameters to the top level`. Which is a bit strange for me, since I there is no lifting of parameters in this PR as far I'm aware. Maybe that part was not yet included when this PR started?

ChristopherDavisUCI:
Thanks @mattijn, Happy New Year to you too!

Are you able to tell what file is causing the problem?  I tried going to Github Desktop and "Update from upstream/master", and about 400 "conflicts" show up, but most of them seem to be related to the examples.

I think you're right that @joelostblom's version of this PR did start before the parameter work.

Worst case scenario I should be able to manually make the changes in this PR starting with the main branch.

I'm very happy for any suggestions of what you think is the best way to proceed!

mattijn:
Maybe we should create a new branch and `git cherry pick` the commits we want from here into this new branch. Then the commit-credits are still tracked to the right contributors. There still will be a bit of refactoring necessary, but at least the history is as much as possible intact.

One thing that I don't know of is the following: how will this upcoming PR: https://github.com/altair-viz/altair/pull/2732 work with this commit: https://github.com/altair-viz/altair/pull/2770/commits/cc5ca7c12b659bc206cc04546700b5ec806780e0? Would it still be necessary to get the docstring of the helper class? Or by setting up aliases the signature will be treated as general descriptors of these parameters? Probably not and is better to test in #2732 once we can merge this first.

ChristopherDavisUCI:
Thanks @mattijn, I'll take a look at this over the next few days and let you know if I get confused.

ChristopherDavisUCI:
Hi again @mattijn, I created a new PR at #2795 if you want to keep an eye on it.  I'll keep you updated on how it's going.  I've only `cherry-pick`ed one commit so far.

ChristopherDavisUCI:
Closing in favor of #2795.

Issue #2775, opened 2022-12-28
Fix documentation page for mobile devices


binste:
See https://github.com/altair-viz/altair/pull/2755#issuecomment-1366574385

@mattijn I'll look into it. Probably need to make the `max-width` of the sidebar in `custom.css` dependent on the device. It's currently set to a constant `20%`.

```css
.bd-sidebar-primary {
  max-width: 20%
}
```

PR #2778, opened 2022-12-28
Fix docs for mobile devices

binste:
Fix for #2775.

image

Unrelated to this but I also hid the `rtd-footer-container` element which created a vertical scrollbar in the primary sidebar on the left side even if it's not needed, i.e. on larger screens.

Before:
image

After:
image


mattijn:
Thanks @binste!

PR #2781, opened 2022-12-29
Merge data pages again

### LINKED ISSUE #2761: Split out data documentation section into separate subpages
Follow up on what was done for spatial data here https://github.com/altair-viz/altair/pull/2750#issuecomment-1355751420.

binste:
Closes #2761. I added another layer of headers to compensate for its lengths. With this, I think it's still easy to navigate:

image


mattijn:
That is a good idea 👍

PR #2784, opened 2022-12-30
Test vl-convert engine in chart.save test

jonmmease:
Follow on to https://github.com/altair-viz/altair/pull/2701 that addresses https://github.com/altair-viz/altair/pull/2701#issuecomment-1334413985 by updating the chart.save test to test both altair_saver and vl-convert.

I also updated a few function signatures to include the engine argument explicitly along with the engine docstring.

PR #2786, opened 2022-12-30
Fix altair test save if vl-convert-python is installed and altair_saver is not

binste:
If vl-convert-python is installed and altair_saver is not, `test_save` failed. For v3 and v4 this was probably introduced with the addition of vl-convert-python, for v5 it is related to #2784. Probably didn't pop up previously as all contributors had altair_saver installed anyway.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ install:
 	python setup.py install
 
 test :
-	black .
+	black . --check
 	flake8 . --statistics
 	python -m pytest --pyargs --doctest-modules altair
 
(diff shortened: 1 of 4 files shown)


mattijn:
Good catch! Currently we used a pytest parametrize to test both engines `@pytest.mark.parametrize("engine", ["altair_saver", "vl-convert"])` see here:

https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/tests/test_api.py#L262

Do you think it would be an improved to install each (and uninstall the other) explicitly on workflow level in the [build.yml](https://github.com/altair-viz/altair/blob/master/.github/workflows/build.yml) and run the pytest command two times?

PR #2787, opened 2022-12-30
Remove update_subtraits as not used anywhere

binste:
Presumably from an older implementation based on traitlets?


mattijn:
Pff, I don't know about this one. I noticed it was introduced in this PR: https://github.com/altair-viz/altair/pull/358, but I cannot see or say with confidence what is the status of this code.

binste:
Understandable, I did some further digging which might give some confidence. I searched through the commit history for all commits which either added or removed a line of code mentioning the function `update_subtraits` using `git log -S update_subtraits` ("git pickaxe"):

* The function itself was indeed introduced in [this commit](https://github.com/altair-viz/altair/commit/aca842cc78503cbdf90b11a1b4b5e26f9f452da1#diff-1b8d7a29b9a8ebd8c1e286ef9e3443ec1f7cf9b6e7ec308d718d094c061361b4R85) in 2017 (as mentioned by you) and referenced in `vegalite/v1` files (btw, other functions from that commit e.g. `update_traits` are also removed in the current master branch).
* [This commit](https://github.com/altair-viz/altair/commit/6bd6562ef9bda32735bad8cccec13558d88995da#diff-bc113654dd96e82276d6fcfa8c53a9a213af695742458932507c6bebe0e42fa5R114) moved the function to `utils.py`
* Then there are some commits in between which are less relevant for us
* In March 2019, [this commit](https://github.com/altair-viz/altair/commit/162b48bff41e23a1676d26bad35fb02a911ec9e0) removed all references to the function by removing the `vegalite/v1` folder which was the only one containing any calls to this function in the first place. 
  * This can be verified by checking out the commit prior to this one with `git checkout 01e9e0f28da86638176` and searching for `update_substring` in the whole codebase -> shows that it's only referenced in `vegalite/v1`. Then checkout the above mentioned commit which removed this folder (`git checkout 162b48bff41e23a167`) and search again -> No remaining references
  * According to the git commit history, this was the last commit which ever introduced or removed the mention of `update_substring`

mattijn:
That is thorough investigation.

PR #2790, opened 2022-12-31
Remove altair_viewer from GitHub build workflow as not needed

binste:
I don't see a need to install the package in the workflows. Maybe previously there was a test which used it?


binste:
Never mind, altair_viewer is required for altair_saver and altair_viewer currently needs to be installed from git for vega-lite 5.2 compatibility :)

PR #2791, opened 2022-12-31
Run tests for both save engines altair_saver and vl-convert-python

binste:
Extended the build workflow as suggested by @mattijn here https://github.com/altair-viz/altair/pull/2786#issuecomment-1368090897. Fixed another bug in the process when vl-convert-python was not installed.

I did not mark the rendering of the gallery examples to png for rerunning as that test takes quite some time and I don't think it is necessary to rerun it as this is about converting the examples, not about the behaviour of Altair if one of the engines is not installed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,3 +30,15 @@ jobs:
     - name: Test with pytest
       run: |
         pytest --doctest-modules altair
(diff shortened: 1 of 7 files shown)


mattijn:
Best wishes for 2023! Thanks for this PR.

binste:
Thanks, same to you!

PR #2804, opened 2023-01-04
Modify generator to move autogenerated test to tests folder

mattijn:
Fix https://github.com/altair-viz/altair/issues/2802.

This PR modifies the `generate_schema_wrapper.py` so the generated test file is placed in the `tests/utils/` folder. 
I had to change the test file to use absolute imports instead of relative imports as it is moved.
I also ran the `generate_schema_wrapper.py` and `generate_api_docs.py` thad leads to the update of two auto-generated files (utils/schemapi.py and channels.py)


mattijn:
All tests pass, merging.

PR #2807, opened 2023-01-05
Add inline argument to chart.save() for html export

jonmmease:
This includes the Vega/Vega-Lite/Vega-Embed JavaScript source in the exported html file so that it works offline.  The logic is ported from altair_saver.

See https://github.com/altair-viz/altair/issues/2765

**Note:** Like altair_saver, this pulls the JavaScript source from the `altair_viewer` package. In the future we could talk about whether to include the JavaScript bundles in Altair itself to avoid this dependency.

### DIFF
--- a/altair/utils/html.py
+++ b/altair/utils/html.py
@@ -141,9 +141,42 @@ HTML_TEMPLATE_UNIVERSAL = jinja2.Template(
 )
 
 
+# This is like the HTML_TEMPLATE template, but includes vega javascript inline
+# so that the resulting file is not dependent on external resources. This was
+# ported over from altair_saver.
+#
+# implies requirejs=False and full_html=True
+INLINE_HTML_TEMPLATE = jinja2.Template(
+    """\
+<!DOCTYPE html>
+<html>
+<head>
+  <script type="text/javascript">
+    // vega.js v{{ vega_version }}
+    {{ vega_script }}
+    // vega-lite.js v{{ vegalite_version }}
+    {{ vegalite_script }}
+    // vega-embed.js v{{ vegaembed_version }}
+    {{ vegaembed_script }}
+  </script>
+</head>
+<body>
+<div class="vega-visualization" id="{{ output_div }}"></div>
+<script type="text/javascript">
+  const spec = {{ spec }};
+  const embedOpt = {{ embed_options }};
+  vegaEmbed('#{{ output_div }}', spec, embedOpt).catch(console.error);
+</script>
+</body>
+</html>
+"""
+)
+
+
 TEMPLATES = {
     "standard": HTML_TEMPLATE,
     "universal": HTML_TEMPLATE_UNIVERSAL,
+    "inline": INLINE_HTML_TEMPLATE,
 }
 
 
@@ -218,6 +251,22 @@ def spec_to_html(
     if mode == "vega-lite" and vegalite_version is None:
         raise ValueError("must specify vega-lite version for mode='vega-lite'")
 
+    render_kwargs = dict()
+    if template == "inline":
+        try:
+            from altair_viewer import get_bundled_script
+        except ImportError:
+            raise ImportError(
+                "The altair_viewer package is required to convert to HTML with inline=True"
+            )
+        render_kwargs["vega_script"] = get_bundled_script("vega", vega_version)
+        render_kwargs["vegalite_script"] = get_bundled_script(
+            "vega-lite", vegalite_version
+        )
+        render_kwargs["vegaembed_script"] = get_bundled_script(
+            "vega-embed", vegaembed_version
+        )
+
     template = TEMPLATES.get(template, template)
     if not hasattr(template, "render"):
         raise ValueError("Invalid template: {0}".format(template))
@@ -233,4 +282,5 @@ def spec_to_html(
         output_div=output_div,
         fullhtml=fullhtml,
         requirejs=requirejs,
+        **render_kwargs,
     )
--- a/altair/utils/save.py
+++ b/altair/utils/save.py
@@ -33,7 +33,7 @@ def save(
 ):
     """Save a chart to file in a variety of formats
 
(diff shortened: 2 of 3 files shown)


jonmmease:
Seeing a couple of CI errors with altair_saver and Python 3.11. Does this look familiar to anyone? Should be try kicking off the tests again?

altair/utils/html.py
```diff
@@ -218,6 +251,22 @@ def spec_to_html(
...
         raise ValueError("must specify vega-lite version for mode='vega-lite'")
 
+    render_kwargs = dict()
+    if template == "inline":
+        try:
+            from altair_viewer import get_bundled_script
+        except ImportError:
+            raise ImportError(
+                "The altair_viewer package is required to convert to HTML with inline=True"
+            )
```
joelostblom, line 261: I like this approach versus adding `altair_viewer` as a required dependency for altair

altair/utils/save.py
```diff
@@ -27,6 +28,7 @@ def save(
     webdriver=None,
     scale_factor=1,
     engine=None,
+    inline=False,
     **kwargs,
 ):
     """Save a chart to file in a variety of formats
```
joelostblom, line 34: Should the line below that lists the supported formats also include `pdf`?

altair/utils/save.py
```diff
@@ -111,6 +122,7 @@ def save(
             vegalite_version=vegalite_version,
             vegaembed_version=vegaembed_version,
             embed_options=embed_options,
+            template="inline" if inline else "standard",
```
joelostblom, line 125: Would this prevent the use of `'universal'` as a value for the `template` parameter? [It seems like that is used in a the `display.py`](https://github.com/altair-viz/altair/search?q=universal) files in both the `vega` and `vega-lite` folders.

joelostblom (commented):
I reran the test and it is passing now. I think this looks good overall, just one or two questions in the comments. @mattijn does this approach makes sense to you too?

altair/utils/save.py
jonmmease: Yeah, I suppose it would. But I don't think `universal` would work anyway because it doesn't write a full HTML document (only an outer `<div>` element).

But your point still stands that this would prevent the use of a custom Jinja template object. I'll update it to only overwrite the template if `inline=True`

jonmmease: Done in 90baae7b2876232bd9e9224249a685dc5fcfe9a6

jonmmease:
Also, I tested this manually. But it would be good for someone else to try it out manually as well since the test case for checking that it's working just checks for an expected substring that's different that what's expected with `inline=False`.

altair/utils/save.py
joelostblom: Thanks! Could you do it on line 36 also which says "Supported formats are [json, html, png, svg]"? Or maybe preferably just remove that line since it is redundant.

jonmmease: whoops, yeah. I added it since it's probably a good idea to have the supported formats featured prominently in the docstring like this.

tests/vegalite/v5/tests/test_api.py
```diff
@@ -335,6 +335,23 @@ def test_save(format, engine, basic_chart):
...
+    out.seek(0)
+    content = out.read()
+
+    assert content.startswith("<!DOCTYPE html>")
+
+    if inline:
+        assert '<script type="text/javascript">' in content
+    else:
+        assert 'src="https://cdn.jsdelivr.net/npm/vega@5"' in content
+        assert 'src="https://cdn.jsdelivr.net/npm/vega-lite@5.2.0"' in content
```
mattijn, line 351:
```suggestion
        assert 'src="https://cdn.jsdelivr.net/npm/vega-lite@5"' in content
```

mattijn:
@jonmmease: out of curiosity: where does vl-convert get the javascript bundles of vega and vega-lite from?

jonmmease:
> where does vl-convert get the javascript bundles of vega and vega-lite from?

It pulls them from the [Skypack CDN](https://www.skypack.dev/), which [integrates well with Deno](https://deno.land/manual@v1.14.2/npm_nodejs/cdns#skypack).  The kind of interesting thing is that there's no global bundler, VlConvert inlines a bunch of individual bundles (for each dependency) and Deno looks them up dynamically on import.

This is the directory of source files that are inlined into into VlConvert on build: https://github.com/vega/vl-convert/tree/main/vl-convert-rs/vendor.

I was thinking about whether there would be a good way for vl-convert to inline these into a standalone HTML file, but I'm not sure how to manually bundle them so that imports work.

I was thinking about whether there would be a good way to use these

mattijn:
or how altair could use these if vl-convert-python is installed..

mattijn:
Regarding this PR, I can understand the intention, but I trust you here @joelostblom. I never have used this functionality in `altair_saver`. If the tests don’t cover what should be tested, maybe we can add some tests that do cover what is needed to check this? 
I had one question, after this PR, can you still use the same functionality from `altair_saver`?

joelostblom (approved):
I tested manually (using the 4.17.0 vega-lite version in altair viewer) and this works as expected.

>  I had one question, after this PR, can you still use the same functionality from altair_saver?

I tested both with and without altair_saver installed and the changes in this PR takes precedence so if you are using a version of altair from after this PR was merged, then you will not use the altair_saver functionality. However, I think it makes sense to keep that functionality around in altair saver to support older version of altair.

Merging, thanks @jonmmease !

mattijn:
Note: once Altair v5 is released, this feature will only work correctly after a new release of [altair_viewer](https://github.com/altair-viz/altair_viewer) as well. Since the current release of altair_viewer (v0.4.0) does not include the JavaScript source of vega-lite v5, but just v4 (v4.17.0). 
The main branch of the repository on Github of altair_viewer already contains VL5 javascript sources (currently v5.2).

PR #2809, opened 2023-01-06
docs: Use SVG thumbnail for emoji example and use VlConvert to build docs on CI

jonmmease:
## Background
VlConvert PNG export does not support the [Isotype Emoji](https://altair-viz.github.io/gallery/isotype_emoji.html), so it wasn't suitable as a full drop-in replacement for altair_saver when building docs on CI.

## Overview
This PR is trying out an idea I had this morning. Interested to hear what folks think.

This PR updates the docs pipeline slightly to support using SVG image thumbnails for select examples (only the isotype_emoji example for now).  Since VlConvert's SVG export of the isotype_emoji example works fine, this makes it possible to use VlConvert to build all of the images in the docs.

I also updated the docs CI job to use VlConvert.

### DIFF
--- a/.github/workflows/docbuild.yml
+++ b/.github/workflows/docbuild.yml
@@ -20,9 +20,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install .[dev]
         pip install "selenium<4.3.0"
-        pip install altair_saver
         pip install -r doc/requirements.txt
-        pip uninstall vl-convert-python --yes
         pip install git+https://github.com/altair-viz/altair_viewer.git
     - name: Run docbuild
       run: |
--- a/.github/workflows/docbuild.yml
+++ b/.github/workflows/docbuild.yml
@@ -19,7 +19,6 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install .[dev]
-        pip install "selenium<4.3.0"
         pip install -r doc/requirements.txt
         pip install git+https://github.com/altair-viz/altair_viewer.git
     - name: Run docbuild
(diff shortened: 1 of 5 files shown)


binste:
I really like the idea and would be in favour of merging this, thanks @jonmmease! I think it's great if we can simplify the development experience and remove the potential for errors by using the wrong package when generating the docs. Tested it locally and docs built as expected and thumbnail showed in Chrome and Safari.

To make sure this configuration possibility does not get lost, what do you think about adding a sentence in contributing.md in the [Adding examples](https://github.com/altair-viz/altair/blob/master/CONTRIBUTING.md#adding-examples) section? Another way to configure it could be to add the flag in the example file itself, similar to `#category: advanced calculations`, e.g. a new line in `isotype_emoji.py`: `#use_svg`. Implementation would happen in `get_docstring_and_rest` which is called https://github.com/altair-viz/altair/blob/master/sphinxext/altairgallery.py#L171. Just as an alternative, not sure if it's better and worth the effort.

.github/workflows/docbuild.yml
```diff
@@ -20,9 +20,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install .[dev]
         pip install "selenium<4.3.0"
```
binste, line 22: Can't test it right now but maybe we can also remove the installation of selenium

jonmmease: Oh, that's a good point. Yeah, selenium shouldn't be needed any more.

jonmmease:
Thanks for taking a look and for the suggestions on where to add docs. I'll do that shortly. Since it won't be needed commonly, I probably won't put the effort into making it a custom `#use_svg` flag in the example for now.

jonmmease:
Added some comments in [2f12382](https://github.com/altair-viz/altair/pull/2809/commits/2f1238259e90b375dbae5cd5ffda598ed092b2b2)

mattijn:
Thanks for the PR @jonmmease! Thanks for checking @binste! Good alternative.

PR #2811, opened 2023-01-07
Add test for layer properties

ChristopherDavisUCI:
Reading through the old discussions, these properties in subcharts for a LayerChart were the cause of one of the biggest issues in updating to Vega-Lite version 5.


mattijn:
Thanks @ChristopherDavisUCI! This PR fix https://github.com/altair-viz/altair/issues/2810.

Quote from this [comment](https://github.com/altair-viz/altair/issues/2810#issuecomment-1374491326)
> (The problem was that I was using `is` to check that the `height` values were consistent, which works in the cases I tried like `height=200` but did not work in your case.)

PR #2812, opened 2023-01-07
Dynamically determine jsonschema validator

### LINKED ISSUE #2800: Use dialect identifier from vega-lite-schema to set jsonschema validator
As was pointed out in https://github.com/altair-viz/altair/pull/2771#issuecomment-1369219098, and summarised here:

* vega specifies the `$schema` property in the schema (see for example L3 [here](https://raw.githubusercontent.com/altair-viz/altair/master/altair/vegalite/v5/schema/vega-lite-schema.json)), which is a "dialect identifier", it basically is a URI that identifies which version of the JSON Schema spec the schema is written for -- an example of one is `https://json-schema.org/draft-07/schema` which is the dialect ID for draft 7 of the spec
* we can then use [`jsonschema.validators.validator_for`](https://python-jsonschema.readthedocs.io/en/latest/api/jsonschema/validators/#jsonschema.validators.validator_for) to get the right validator class given the schema, which will return you `Draft7Validator` if the schema use draft 7, but if/when Vega-lite repo upgrade to draft2020 say, then you'd get `Draft202012Validator` instead and things will work properly still.

Currently (or at least after PR https://github.com/altair-viz/altair/pull/2771 is merged) the validator version is defined as a constant (draft 7). Altair currently uses this validator for the schema definitions for Vega-lite v3/v4/v5 (and that is still fine as they all use the draft 7 schema). So in my understanding the required change to get this working properly is to make it dynamic and set and define the validator on a version by version basis.

binste:
Closes #2800

Tested with jsonschema 4.17.3 (newest) and 3.0.0 (minimum requirement). I had to introduce `$schema` in the test schemas as well so that it takes the same validator for the tests as when validating against the latest vega-lite schema.

### DIFF
--- a/altair/utils/display.py
+++ b/altair/utils/display.py
@@ -125,7 +125,10 @@ class Displayable(object):
         # type: () -> None
         """Validate the spec against the schema."""
         schema_dict = json.loads(pkgutil.get_data(*self.schema_path).decode("utf-8"))
-        validate_jsonschema(self.spec, schema_dict)
+        validate_jsonschema(
+            self.spec,
+            schema_dict,
+        )
 
     def _repr_mimebundle_(self, include=None, exclude=None):
         """Return a MIME bundle for display in Jupyter frontends."""
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -62,9 +62,10 @@ def validate_jsonschema(spec, schema, rootschema=None):
         # No resolver is necessary if the schema is already the full schema
         resolver = None
 
-    validator = validator_cls(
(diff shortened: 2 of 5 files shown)


mattijn:
Thanks for this PR @binste! Could we add a test with two dummy schemas where one uses a draft 7 `$schema` specification and the other a draft 2020 `$schema`  property and assert if it this is correct?

binste:
Sure! Is this what you had in mind? The test would fail in case the validation would no longer use a different validator class based on the $schema property.

mattijn:
Yes! Looks good. Thank you!

PR #2813, opened 2023-01-07
Remove side-effects of calling EncodingChannelMixin.to_dict

### LINKED ISSUE #2497: Chart representation changes after the chart is displayed
The representation of the chart object seems to change after the chart is displayed. Take this example

```py
import altair as alt


cars = data.cars()

scatter = alt.Chart(cars).mark_point().encode(
    x='Weight_in_lbs:Q',
    y='Horsepower',
    color='Origin')
```

It looks like this initially 

```py
scatter.encoding
```

```py
FacetedEncoding({
  color: Color({
    shorthand: 'Origin'
  }),
  x: X({
    shorthand: 'Weight_in_lbs:Q'
  }),
  y: Y({
    shorthand: 'Horsepower'
  })
})
```

But after displaying the chart in JuptyerLab (by just typing `scatter` in a cell and executing it), the encoding representation changes to this longer format:

```py
FacetedEncoding({
  color: Color({
    field: 'Origin',
    type: 'nominal'
  }),
  x: X({
    field: 'Weight_in_lbs',
    type: 'quantitative'
  }),
  y: Y({
    field: 'Horsepower',
    type: 'quantitative'
  })
})
```

There is probably some good reason for this that I don't understand, but for testing/autograding purposes it would be convenient if the representation was always the same and did not change after the chart is displayed.

Edit: It seems like `scatter.to_dict()` will always return the long format so maybe that could be relied on instead.

binste:
Fixes #2497. CC @joelostblom. It can be verified by using the example code in the mentioned issue.

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -350,8 +350,24 @@ class SchemaBase(object):
         if self._args and not self._kwds:
             result = _todict(self._args[0], validate=sub_validate, context=context)
         elif not self._args:
+            kwds = self._kwds.copy()
+            # parsed_shorthand is added by FieldChannelMixin.
+            # It's used below to replace shorthand with its long form equivalent
+            # parsed_shorthand is removed from context if it exists so that it is
+            # not passed to child to_dict function calls
+            parsed_shorthand = context.pop("parsed_shorthand", {})
+            kwds.update(
+                {
+                    k: v
+                    for k, v in parsed_shorthand.items()
+                    if kwds.get(k, Undefined) is Undefined
+                }
+            )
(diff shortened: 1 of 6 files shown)


joelostblom (commented):
Looks great, thank you for tackling this issue! To prevent this regressing in the future, could we add a test that checks that the chart spec contains the shorthand syntax and that it is expanded to the long form syntax when using `to.dict()`, both before and after displaying (or equivalent) a chart?

binste:
Sure! Done :)

joelostblom (approved):
Looks great, thank you!

PR #2774, opened 2022-12-28
Update and simplify README

binste:
I updated the README.md. Furthermore, I really like READMEs which are concise, similar to https://github.com/pandas-dev/pandas, i.e. a README which summarizes what the library does, a fully contained example that can be copy-pasted, simple installation instructions, and a reference to the documentation. This is where users can then go to learn more. Afterwards, some additional content for developers. In my opinion, all the rest should be in the documentation so it's easier to find.

Reasoning for the changes is usually in the commit messages but let me know if anything is unclear!

Included content from #1122 in CONTRIBUTING.md

Is the Google Altair Group still something that should be actively recommended for asking questions or only StackOverflow? I think ideally there is one preferred place to ask questions (StackOverflow) and one place to report feature requests and bugs (GH issues). But as the Google Group already has quite some content, maybe we want to keep the reference to it and I could also add it in the [Getting Help](https://binste.github.io/altair-docs/getting_started/getting_help.html) page in the documentation?

PR #2815, opened 2023-01-07
Add --check to black test command

binste:
Prior to changing this command in #2774, it already had the `--check` argument. This argument makes sure that the status is returned and therefore that the command fails if files would be formatted. I think removing it was unintentional (the reasoning in https://github.com/altair-viz/altair/pull/2774#discussion_r1060521819 was around adding `--diff --color` and not removing `--check`). I prefer if the command fails, else you need to scroll through the console output of pytest to get to the output of black to see the outcome.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ install:
 	python setup.py install
 
 test :
-	black --diff --color .
+	black --diff --color --check .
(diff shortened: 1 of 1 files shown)


mattijn:
Thanks! Looks good.

PR #2822, opened 2023-01-08
Remove vega-v5 wrappers

mattijn:
This PR removes the Vega V5 wrappers from the Altair repository.

It removes them from the generate_schema_wrapper, magics, mimebundle, tests and docs.

Not everything related to vega is removed since the saving utilities requires vega bundles to be included.


mattijn:
Can you check @joelostblom? I think I removed everything related to the Vega wrapper, but I might have missed some parts in the docs/docstring.

tests/utils/tests/test_mimebundle.py
```diff
@@ -40,170 +40,6 @@ def vegalite_spec():
...
-        vegaembed_version=alt.VEGAEMBED_VERSION,
-        engine=engine,
-    )
-
-    bundle["application/vnd.vega.v5+json"]["axes"] = delete_none(
-        bundle["application/vnd.vega.v5+json"]["axes"]
-    )
-    assert bundle == {"application/vnd.vega.v5+json": vega_spec}
-
-
```
joelostblom, line 206: It looks like this vl-convert test was accidentally deleted

joelostblom (changes requested):
Nice, thanks for doing this. Apart from one of the deleted tests I think this looks good. I grepped the repo for "vega" and didn't notice any other places needing an update.

I noticed the name of the branch and just wanted to say that I think it makes sense to have a separate PR for removing the old Vega-Lite versions and limiting this PR to Vega as you have done now :+1:

tests/utils/tests/test_mimebundle.py
joelostblom: Nevermind, I can see that this is only for conversion to the Vega mimebundle now, so it should be deleted

joelostblom (approved):
I misread the test I thought was accidentally deleted, this looks good.

joelostblom:
One thing I was thinking here is that maybe we should add an explicit note in the docs that Altair specs can programmatically be turned into VegaLite with `to_json`, but not Vega specs, and then include the way to create Vega specs directly in Jupyter via IPython.display that you mentioned in https://github.com/altair-viz/altair/issues/2817?

PR #2824, opened 2023-01-09
Make error messages on typos and missing/incorrect data types more informative

joelostblom:
When trying to reference a column from a dataframe and a typo is made, the error message suggests that there is an issue with the type specification rather than the column name itself:

```py
import altair as alt

alt.Chart(data.cars()).mark_point().encode(x='Horsepowers')
```

**Old error**
> ValueError: Horsepowers encoding field is specified without a type; the type cannot be inferred because it does not match any column in the data.

I think that message is incorrect and unhelpful, and I often see learners struggle with understanding that they have made a typo. This PR suggests replacing the error message with this more informative alternative:

**New error**
> ValueError: "Horsepowers" cannot be found.
> - If you are trying to reference a column in the dataframe, there is likely a typo in the column name.
> - If you are trying to reference a field from a transform, there is either a typo in the field name or the data type specification is missing (Altair can only infer data types from dataframe columns).

---

Similarly, when making a typo in the encoding data type, the error message could also be made more informative:

```py
alt.Chart(data.cars()).mark_point().encode(x='Horsepower:Qi')
```

>**Old error**
> ValueError: Horsepower:Qi encoding field is specified without a type; the type cannot be inferred because it does not match any column in the data.

This PR changes that to the following:

**New error**
> ValueError: "Qi" is not one of the valid encoding data types: O, N, Q, T, G.
> For more details, see https://altair-viz.github.io/altair-docs/user_guide/encodings/index.html#encoding-data-types. (If you are trying to use a column name that contains a colon, prefix it with a backslash, \\:)

**Note** that this second change means that it is now required to escape colons in column names and although this is is a breaking change I think it makes it more consistent instead of having colon in column names sometimes work depending on whether the part after the colon happens to equal an Altair data type or not.


altair/vegalite/v5/schema/channels.py
```diff
@@ -38,9 +38,14 @@ def to_dict(self, validate=True, ignore=(), context=None):
                 parsed.pop('type', None)
             elif not (type_in_shorthand or type_defined_explicitly):
                 if isinstance(context.get('data', None), pd.DataFrame):
-                    raise ValueError("{} encoding field is specified without a type; "
-                                     "the type cannot be inferred because it does not "
-                                     "match any column in the data.".format(shorthand))
+                    raise ValueError(
+                        '"{}" cannot be found.'.format(shorthand)
+                        + "\n- If you are trying to reference a column in the dataframe,"
+                        + " there is likely a typo in the column name."
```
ChristopherDavisUCI, line 44: What about when a URL is being used, like `alt.Chart(data.cars.url)`?  I think that is an important example where the old error message is helpful, because the type is necessary.  Do you see what I mean?

joelostblom: That will activate the `else` section below, which still raises the old error message about the missing data type. This new modified error message only shows up if the data is a pandas data frame.

I could not find a way to raise an informative error when there is a typo in the column name in a non-pandas data source and opened https://github.com/altair-viz/altair/issues/2825 to track that.

joelostblom: The old message does not contain any reference to "columns", so if you use a URL without a type specified you will see:

> ValueError: Horsepowers encoding field is specified without a type; the type cannot be inferred because it does not match any column in the data.

Is that what you wanted or do you think we should modify the message in this case too?

ChristopherDavisUCI: Thanks @joelostblom I misunderstood again.  Everything here looks good to me!

altair/vegalite/v5/schema/channels.py
```diff
@@ -38,9 +38,14 @@ def to_dict(self, validate=True, ignore=(), context=None):
                 parsed.pop('type', None)
             elif not (type_in_shorthand or type_defined_explicitly):
                 if isinstance(context.get('data', None), pd.DataFrame):
```
mattijn, line 40: As you correctly observed, the `isinstance` condition on line 40 is insufficient as it will not filter out errors related to defined transforms. 
I think it is better to catch this separately or remove this check completely. If you decide to go for the latter than you would have a single `raise ValueError` within `elif not (type_in_shorthand or type_defined_explicitly):`, maybe something as such:

```python
raise ValueError(
    "Error: Unable to determine data type for field '{}'. Please verify that the column"
    " or field name is spelled correctly and that the data type is specified correctly,"
    " particularly if the field is not from a pandas.DataFrame.".format(shorthand)
)
```

altair/utils/core.py
```diff
@@ -527,6 +527,22 @@ def parse_shorthand(
...
+        and attrs["field"][attrs["field"].rfind(":") - 1] != "\\"
+    ):
+        raise ValueError(
+            '"{}" '.format(attrs["field"].split(":")[-1])
+            + "is not one of the valid encoding data types: {}.".format(
+                ", ".join(TYPECODE_MAP.values())
+            )
+            + "\nFor more details, see https://altair-viz.github.io/altair-docs/user_guide/encodings/index.html#encoding-data-types. "
+            + "(If you are trying to use a column name that contains a colon, prefix it with a backslash, \\:)"
+        )
```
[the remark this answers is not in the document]
joelostblom: I think it does help improve the error message since there will now be a clear error message when the type is incorrect. Otherwise, this would be a potential issue to mention in the other error message as well which makes it more vague. If you prefer I can break this out into its own PR to discuss more?

altair/vegalite/v5/schema/channels.py
joelostblom: I like the way you phrased the error message and updated the PR with a similar phrasing, thanks! I would still prefer to have a separate error message for data that is not a pandas data frame since we in those cases know that the issue is a missing data type since the data can't be inferred (although if we remove the other change you commented on than it could also be a misspelled data type I guess, but never a typo due to #2825).

altair/vegalite/v5/schema/channels.py
```diff
@@ -38,9 +38,12 @@ def to_dict(self, validate=True, ignore=(), context=None):
...
-                    raise ValueError("{} encoding field is specified without a type; "
-                                     "the type cannot be inferred because it does not "
-                                     "match any column in the data.".format(shorthand))
+                    raise ValueError(
+                        'Unable to determine data type for the field "{}";'
+                        " verify that the field name is not misspelled."
+                        " If you are referencing a field from a transform,"
```
mattijn, line 44: But this phrasing is incomplete. Since datasets defined as url/json/csv (not pd.DataFrame) requires a type definition too.

joelostblom: Those data sources are all handled on line 48 below. The error message here is only if we know that the data is a pandas data frame as per the check on line 40.

mattijn: But that error message is also incomplete, since those fields can also be misspelled instead of the missing type definition.

mattijn: No, that is inaccurate what I say, since that becomes https://github.com/altair-viz/altair/issues/2825..

joelostblom: Actually they can't. that's what #2825 is about. If the type is correct, it doesn't matter if those fields are misspelled, the chart will still appear (but be empty). This error is only raised if there is no type detected.

joelostblom:
@mattijn How do you think we should proceed here? I think the improved error messages would be nice to have in the release candidate.

altair/utils/core.py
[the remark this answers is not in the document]
mattijn: Regarding escaping the colon, I don't really like the breaking change, but I also agree with you that the current behaviour can lead to confusing situations. 

Can you include an example in the `ValueError` how this is supposed to look like with a real example? Maybe for a column that is named as such `"release_date:theater"`, how is this supposed to look like when escaped? Would this change also be required if a person does not use the shorthand `:T`, but defines it as `"release_date:theater", type="temporal"` instead?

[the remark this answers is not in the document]
joelostblom: Yeah, I agree that a breaking change is not ideal, but at the same time we are planning a bump in the major version number with the 5.0 release and that is an indication that there might be breaking changes. I updated the error message to include a column name and it now reads:

> If you are trying to use a column name that contains a colon, prefix it with a backslash; for example "column\:name" instead of "column:name".

---

Essentially what this PR does is treating `:` the same as other special characters such as `.`, `[`, `-`, etc. The one difference being that `:` is a special character at the altair level, whereas the others are special characters at the VL level. For all these special characters, we need to both escape them and specify the type (although that could be improved on in a later PR). Here are three examples:

## `:`

```py
import pandas as pd
import altair as alt

source = pd.DataFrame({
    'release_year:national': [2000, 2001, 2004],
    'release_year:international': [2001, 2001, 2005],
    'color-': [2001, 2001, 2005]
})

    x='release_year\:national:Q',
    y=alt.Y('release_year\:international', type='quantitative'),
)
```

## `[`

```py
import altair as alt
import numpy as np
import pandas as pd

x = np.arange(100)

xname = "x [mV"
xname_escaped = "x \[mV"

source = pd.DataFrame({xname: x})

alt.Chart(source).mark_line().encode(
    x=alt.X(field=xname_escaped, type="quantitative"),
    y=xname_escaped + ':Q',
)
```

## `.`

```py
import altair as alt
import pandas as pd
import numpy as np

source = pd.DataFrame({"x.x": np.arange(10)})

alt.Chart(source).mark_line().encode(
    x="x\.x:Q",
    y=alt.Y("x\.x", type='quantitative')
)
```

And some old issues discussing special characters (posting as a reference for myself later too):

- https://github.com/altair-viz/altair/issues/284
- https://github.com/altair-viz/altair/issues/2494
- https://github.com/altair-viz/altair/issues/2600

altair/utils/core.py
```diff
@@ -532,6 +532,23 @@ def parse_shorthand(
...
+        and attrs["field"][attrs["field"].rfind(":") - 1] != "\\"
+    ):
+        raise ValueError(
+            '"{}" '.format(attrs["field"].split(":")[-1])
+            + "is not one of the valid encoding data types: {}.".format(
+                ", ".join(TYPECODE_MAP.values())
+            )
+            + "\nFor more details, see https://altair-viz.github.io/altair-docs/user_guide/encodings/index.html#encoding-data-types. "
+            + "If you are trying to use a column name that contains a colon, "
+            + 'prefix it with a backslash; for example "column\\:name" instead of "column:name".'
```
mattijn, line 550:
```suggestion
            + 'prefix it with a backslash; for example "column\:name" instead of "column:name".'
```
singe `\` colon isn't it?

joelostblom: The additional backslash is needed to make sure a single backslash is shown in the outputted warning message (the first backslash makes sure the second one is printed instead of being interpreted as an escape character for the following colon):

image

mattijn (approved):

joelostblom:
Merging in this too, yay better error messages everywhere!

PR #2828, opened 2023-01-12
WIP Parameter tests

ChristopherDavisUCI:
@joelostblom @mattijn Do these look like the sort of tests we would like for testing the new parameter syntax?  If this looks like the right spirit, I will add more, probably mostly based off the Vega-Lite documentation and the Altair documentation.

By the way @mattijn, I played around with adding property setter syntax like `alt.param().bind(slider)`, but I don't think it's a good idea at this stage.  There's already a lot of custom code related to these parameters (for example `__getattr__` defined [here](https://github.com/altair-viz/altair/blob/af756e9a9549b2bf0f88685d2f93e6224665bf18/altair/vegalite/v5/api.py#L216-L224)) and adding another layer of customization on top of it felt like it was getting overly difficult to maintain.

### DIFF
--- a/tests/vegalite/v5/tests/test_params.py
+++ b/tests/vegalite/v5/tests/test_params.py
@@ -2,6 +2,11 @@
 
 import pandas as pd
 
+import warnings
+import pytest
+
+import re
+
 import altair.vegalite.v5 as alt
 
 
@@ -41,11 +46,11 @@ def test_param_expr():
     inner_var = height_var / 2
     tick_var = (3 * height_var) / 4
 
-    inner_str = inner_var._to_expr().replace(" ", "")
-    inner_ans = "(height/2)"
+    inner_str = inner_var._to_expr()
+    inner_ans = "(height / 2)"
 
-    tick_str = tick_var._to_expr().replace(" ", "")
-    tick_ans = "((3*height)/4)"
+    tick_str = tick_var._to_expr()
+    tick_ans = "((3 * height) / 4)"
 
     assert height_var._to_expr() == "height"
     assert inner_str == inner_ans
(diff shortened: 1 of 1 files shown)


tests/vegalite/v5/tests/test_params.py
```diff
@@ -0,0 +1,69 @@
...
+    c2 = alt.Chart(data).mark_bar(size=inner_var).encode(x="r2")
+
+    c3 = alt.Chart(data).mark_tick(size=tick_var).encode(x="r3")
+
+    c = (c1 + c2 + c3).add_params(height_var)
+
+    dct = c.to_dict()
+
+    ans_set = {v._to_expr() for v in [height_var, inner_var, tick_var]}
+    expr_set = {d["mark"]["size"]["expr"] for d in dct["layer"]}
```
mattijn, line 67: The assert in this test is a bit too complex. Eg. `ans_set` and `expr_set` are both computed in the test, so once the test fails it’s not directly clear which side has (wrongly) changed. Can you declare the `expr_set` in full as in previous test? If it becomes too verbose, maybe change the test slightly.

mattijn (commented):
Hopefully this weekend more time. Now just some suggestions from phone.

tests/vegalite/v5/tests/test_params.py
```diff
@@ -0,0 +1,69 @@
...
+    assert dct["mark"] == mark_dict
+    assert dct["params"] == param_list
+
+
+def test_param_expr():
+    height_var = alt.param(name="height")
+    inner_var = height_var / 2
+    tick_var = (3 * height_var) / 4
+
+    inner_str = inner_var._to_expr().replace(" ", "")
```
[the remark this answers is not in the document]
ChristopherDavisUCI: Thanks for these suggestions @mattijn they're very helpful!  I will keep them in mind for the later tests.  For this `replace` in particular, my reasoning was just that the white space in these expressions isn't significant (as far as I know).

ChristopherDavisUCI:
Thanks for the comments @mattijn, I made some changes and added some more tests.  I think this covers the main parameter functionality, but please let me know if you see anything missing or would like any changes!

tests/vegalite/v5/tests/test_params.py
```diff
@@ -0,0 +1,167 @@
...
+
+    inner_str = inner_var._to_expr()
+    inner_ans = "(height / 2)"
+
+    tick_str = tick_var._to_expr()
+    tick_ans = "((3 * height) / 4)"
+
+    assert height_var._to_expr() == "height"
+    assert inner_str == inner_ans
+    assert tick_str == tick_ans
```
mattijn, line 57:
```suggestion
    height_str = height_var._to_expr()
    inner_str = inner_var._to_expr()
    tick_str = tick_var._to_expr()

    assert height_str == "height"
    assert inner_str == "(height / 2)"
    assert tick_str == "((3 * height) / 4)"
```
Looks great! Improved readability a little bit here. To make it directly clear what is expected in the assertion. Furthermore all fine.

ChristopherDavisUCI: Thank you!

Issue #2817, opened 2023-01-08
refactoring: remove vega and vegalite wrappers


mattijn:
Continuation of the discussion that started from this [comment ](https://github.com/altair-viz/altair/pull/2785#issuecomment-1371339432) and onwards.

Vega 5 specifications can currently be rendered using Altair, with support of interactivity:

```python
import altair.vega.v5 as alt
import json
vg_spec = r"""
{
  "$schema": "https://vega.github.io/schema/vega/v5.json",
  "description": "A basic bar chart example, with value labels shown upon mouse hover.",
  "width": 400,
  "height": 200,
  "padding": 5,
  "data": [
    {
      "name": "table",
      "values": [
        {"category": "A", "amount": 28},
        {"category": "B", "amount": 55},
        {"category": "C", "amount": 43},
        {"category": "D", "amount": 91},
        {"category": "E", "amount": 81},
        {"category": "F", "amount": 53},
        {"category": "G", "amount": 19},
        {"category": "H", "amount": 87}
      ]
    }
  ],
  "scales": [
    {
      "name": "xscale",
      "type": "band",
      "domain": {"data": "table", "field": "category"},
      "range": "width",
      "padding": 0.05,
      "round": true
    },
    {
      "name": "yscale",
      "domain": {"data": "table", "field": "amount"},
      "nice": true,
      "range": "height"
    }
  ],
  "axes": [
    {"orient": "bottom", "scale": "xscale"},
    {"orient": "left", "scale": "yscale"}
  ],
  "marks": [
    {
      "type": "rect",
      "from": {"data": "table"},
      "encode": {
        "enter": {
          "x": {"scale": "xscale", "field": "category"},
          "width": {"scale": "xscale", "band": 1},
          "y": {"scale": "yscale", "field": "amount"},
          "y2": {"scale": "yscale", "value": 0}
        },
        "update": {"fill": {"value": "steelblue"}},
        "hover": {"fill": {"value": "red"}}
      }
    }
  ]
}
"""
alt.vega(json.loads(vg_spec))
```
![image](https://user-images.githubusercontent.com/5186265/211169422-95103317-6acb-457e-a17a-d0bf021b9384.png)

Now, one can also use `vl-convert` (`pip/conda install vl-convert-python`) to render these as png or svg in a notebook:
```python
import vl_convert as vlc
from IPython.display import Image
from IPython.display import SVG, display

#vg_png = vlc.vega_to_png(vg_spec)  # Image(vg_png)
[...]

PR #2829, opened 2023-01-13
Remove vega v5 wrappers

mattijn:
Same as #2822, but renamed the brach, which closed the previous PR.

As mentioned in #2822:
> This PR removes the Vega V5 wrappers from the Altair repository.
> 
> It removes them from the generate_schema_wrapper, magics, mimebundle, tests and docs.
> 
> Not everything related to vega is removed since the saving utilities requires vega bundles to be included.

Thanks for the review @joelostblom, I will update soon. Will it be better to raise deprecation warnings when vega options are called, instead of all the deletions as is proposed currently?


joelostblom:
> Thanks for the review @joelostblom, I will update soon. Will it be better to raise deprecation warnings when vega options are called, instead of all the deletions as is proposed currently?

I was considering that too, and I am OK either way. I don't think we *need* it since this is a major release and there is an easy other way to get the same functionality via ipython.display, and we would then have this code lying around and need to delete it at a later point. But if you think it is sounder/safer to have it lying around and marked as deprecatedfor a few release, then I'm OK with that too.

mattijn:
I changed it a bit. Now it is allowed to define `vega` as format in the mimebundle, so you can export a Vega-lite specification to Vega. But I deprecate the rendering of Vega specifications (defined by `mode`) as we removed the Vega wrappers.

Now I get a C901 error on flake8 that the `save()` function is too complex. Not sure how to pass that one, any suggestions?

altair/utils/mimebundle.py
```diff
@@ -12,18 +14,18 @@ def spec_to_mimebundle(
     engine=None,
     **kwargs,
 ):
-    """Convert a vega/vega-lite specification to a mimebundle
+    """Convert a vega-lite specification to a mimebundle
 
     The mimebundle type is controlled by the ``format`` argument, which can be
-    one of the following ['html', 'json', 'png', 'svg', 'pdf', 'vega', 'vega-lite']
+    one of the following ['html', 'json', 'png', 'svg', 'pdf',  'vega', 'vega-lite']
```
binste, line 20:
```suggestion
    one of the following ['html', 'json', 'png', 'svg', 'pdf', 'vega', 'vega-lite']
```

altair/utils/mimebundle.py
```diff
@@ -12,18 +14,18 @@ def spec_to_mimebundle(
...
 
     Parameters
     ----------
     spec : dict
         a dictionary representing a vega-lite plot spec
-    format : string {'html', 'json', 'png', 'svg', 'pdf', 'vega', 'vega-lite'}
+    format : string {'html', 'json', 'png', 'svg', 'pdf',  'vega', 'vega-lite'}
```
binste, line 26:
```suggestion
    format : string {'html', 'json', 'png', 'svg', 'pdf', 'vega', 'vega-lite'}
```

altair/utils/mimebundle.py
```diff
@@ -43,16 +45,16 @@ def spec_to_mimebundle(
...
-    The png, svg, pdf, and vega outputs require the altair_saver package
+    The png, svg, vega and pdf outputs require the vl-convert or altair_saver package
     to be installed.
     """
-    if mode not in ["vega", "vega-lite"]:
-        raise ValueError("mode must be either 'vega' or 'vega-lite'")
+    if mode != "vega-lite":
+        if mode == "vega":
+            warnings.warn(
+                "mode 'vega' is deprecated, use 'vega-lite'", AltairDeprecationWarning
```
binste, line 54: I like that this informs users of the change. Should it raise a value error within the if statement with this message instaed of a deprecation warning? I usually encounter deprecation warnings as "something will be deprecated in a **future** release but right now it still works so this does not throw an exception", not "something is already removed".

altair/utils/mimebundle.py
```diff
@@ -147,7 +145,7 @@ def _validate_normalize_engine(engine, format):
 
     engine : {None, 'vl-convert', 'altair_saver'}
         the user-provided engine string
-    format : string {'png', 'svg', 'pdf', 'vega'}
+    format : string {'png', 'svg', 'pdf'}
```
binste, line 148:
```suggestion
    format : string {'png', 'svg', 'pdf', 'vega'}
```
After your latest changes, format can be `'vega'`, right?

altair/utils/save.py
```diff
@@ -45,7 +88,7 @@ def save(
         the format to write: one of ['json', 'html', 'png', 'svg', 'pdf'].
         If not specified, the format will be determined from the filename.
     mode : string (optional)
-        Either 'vega' or 'vegalite'. If not specified, then infer the mode from
+        Must be 'vegalite'. If not specified, then infer the mode from
         the '$schema' property of the spec, or the ``opt`` dictionary.
         If it's not specified in either of those places, then use 'vegalite'.
```
binste, line 93:
```suggestion
        Must be 'vega-lite'. If not specified, then infer the mode from
        the '$schema' property of the spec, or the ``opt`` dictionary.
        If it's not specified in either of those places, then use 'vega-lite'.
```

tests/utils/tests/test_mimebundle.py
```diff
@@ -215,10 +216,16 @@ def test_spec_to_vegalite_mimebundle(vegalite_spec):
...
 
 def test_spec_to_vega_mimebundle(vega_spec):
-    bundle = spec_to_mimebundle(
-        spec=vega_spec, mode="vega", format="vega", vega_version=alt.VEGA_VERSION
-    )
-    assert bundle == {"application/vnd.vega.v5+json": vega_spec}
+    with warnings.catch_warnings():
+        warnings.simplefilter("error")
+        with pytest.raises(ValueError):
+            with pytest.warns(alt.utils.deprecation.AltairDeprecationWarning):
```
binste, line 222: See my other comment on `AltairDeprecationWarning`

altair/utils/save.py
```diff
@@ -15,6 +16,48 @@ def write_file_or_filename(fp, content, mode="w"):
...
+        if "mode" in embed_options:
+            mode = embed_options["mode"]
+        elif "$schema" in spec:
+            mode = spec["$schema"].split("/")[-2]
+        else:
+            mode = "vega-lite"
+
+    if mode != "vega-lite":
+        if mode == "vega":
+            warnings.warn(
```
binste, line 50: Same remark here about the deprecation warning

altair/utils/save.py
```diff
@@ -45,7 +88,7 @@ def save(
         the format to write: one of ['json', 'html', 'png', 'svg', 'pdf'].
```
binste, line 88:
```suggestion
        the format to write: one of ['json', 'html', 'png', 'svg', 'pdf', 'vega', 'vega-lite'].
```

Only add the `vega-lite` suggestion if it is also supported in the rest of `save`

altair/utils/save.py
```diff
@@ -15,6 +16,48 @@ def write_file_or_filename(fp, content, mode="w"):
...
+    """Inspect the format argument in the save function"""
+    if format is None:
+        if isinstance(fp, str):
+            format = fp.split(".")[-1]
+        elif isinstance(fp, pathlib.PurePath):
+            format = fp.suffix.lstrip(".")
+        else:
+            raise ValueError(
+                "must specify file format: "
+                "['png', 'svg', 'pdf', 'html', 'json', 'vega']"
```
binste, line 29:
```suggestion
                "['png', 'svg', 'pdf', 'html', 'json', 'vega', 'vega-lite']"
```
Does it make sense to factor out this list into a variable and use it in the strings everywhere? Feels like this is repeated quite often.

This list should also be updated in the docstring of `v5/api.py:TopLevelMixin.save`

altair/utils/save.py
```diff
@@ -128,7 +148,7 @@ def save(
             **kwargs,
         )
         write_file_or_filename(fp, mimebundle["text/html"], mode="w")
-    elif format in ["png", "svg", "pdf"]:
+    elif format in ["png", "svg", "pdf", "vega"]:
```
binste, line 151: I can't comment on the correct line but `vega` does not work here. It will try to access the mimebundle with key `image/svg+xml` on line 169. 

Could also add support for "vega-lite" here?

altair/utils/save.py
```diff
@@ -15,6 +16,48 @@ def write_file_or_filename(fp, content, mode="w"):
         fp.write(content)
 
 
+def set_inspect_format_argument(format, fp, inline):
+    """Inspect the format argument in the save function"""
+    if format is None:
+        if isinstance(fp, str):
+            format = fp.split(".")[-1]
+        elif isinstance(fp, pathlib.PurePath):
+            format = fp.suffix.lstrip(".")
```
binste, line 25: We could associate the file endings `.vg` and `.vg.json` with `vega` and `.vl`, `.vl.json` with `vega-lite` so the files can be opened in JupyterLab https://jupyterlab.readthedocs.io/en/2.2.x/user/file_formats.html#vega-vega-lite

binste (changes requested):
Thanks @mattijn! Suggested some changes.

altair/utils/save.py
mattijn: Opend https://github.com/altair-viz/altair/issues/2852 to track this

altair/utils/save.py
mattijn: `vega-lite` cannot be supported as `format`. That would be `json`. `format` functions a bit like export to, or save as in other programs. If it is defined as `vega` it should compile the vega-lite specification into vega json.
Apparently that never has functioned.

binste: Of course, you're right, that would be `json`!

mattijn: Hm, I just noticed the following variations in the code:
https://github.com/altair-viz/altair/blob/1ed3f6f2fd3647bcf1f2e5e81dd328a0f9cf623d/altair/utils/mimebundle.py#L72-L77

It seems there is different type of json, namely, `application/json` and `application/vnd.vegalite.v5+json` (for altair 5). So I'll include `vega-lite` also as `format`, it probably would be better if it was renamed to something as `vl-json`.

mattijn:
Thanks for the continuous review @binste! 

I've reverted the introduction of warnings and also the enhancements to get an improved support for `vega`/`vega-lite` as output save format.
These potential enhancements are registered as feature requests in https://github.com/altair-viz/altair/issues/2852. Now this PR is only on removing the Vega 5 wrappers/rendering options.

This PR is even with main & all tests are passing. So I'm merging.

binste:
Awesome, great to see that we can already close #2817 One step closer towards a release candidate!

Issue #2744, opened 2022-12-05
Incorrect error message when an unknown param value is used


joelostblom:
I just noted that in a spec that contains a condition, using an unknown parameter value anywhere in the spec will always return an error pointing to the condition even when it has nothing to do with actual error. For example, the following spec:

```py
import altair as alt
from vega_datasets import data

source = data.barley()

alt.Chart(source).mark_bar().encode(
    x='variety',
    y=alt.Y('sum(yield)', stack='asdf'),
    opacity=alt.condition('datum.yield > 0', alt.value(1), alt.value(0.2))
)
```

returns:

```
SchemaValidationError: Invalid specification

        altair.vegalite.v4.schema.channels.OpacityValue, validating 'additionalProperties'

        Additional properties are not allowed ('test' was unexpected)
```

If we change to `opacity=alt.value(0.6)` we get the expected output:

```
SchemaValidationError: Invalid specification

        altair.vegalite.v4.schema.channels.Y->0, validating 'enum'

        '' is not one of ['zero', 'center', 'normalize']
```

Note that there is nothing wrong the opacity condition here and removing the incorrect `stack` value leads to zero errors. Looking closer at the error raising sequence, it seems like the stack error is always raised first, but if there is a conditional in the spec there will also be a second error raised. Not sure how raising one error leads to the other or why the sequence doesn't stop at the first raised error.

It seems like `{'test': 'datum.yield > 0', 'value': 1}` is evaluating against a schema that only contains the `condition` key, so it should have been `{'condition': {'test': 'datum.yield > 0', 'value': 1}}` instead and then `{'test': 'datum.yield > 0', 'value': 1}` in the next validation.

joelostblom:
This is not unique to `condition`, it also happens for concat charts. In the example below the error is with the title data type, but this gets swallowed and leads to another error being raised:

```
import altair as alt
from vega_datasets import data

source = data.cars()

points = alt.Chart(source).mark_point().encode(
    x='Horsepower',
    y='Miles_per_Gallon',
)

text = alt.Chart(source).mark_text(align='right').encode(
    alt.Text('Horsepower:N', title=dict(text='Horsepower', align='right'))
)

points | text
```

The real error is the first one printed below, but the one that actually gets raised is the last one which is really confusing

![image](https://user-images.githubusercontent.com/4560057/207788941-4ea79f92-5577-419b-88a6-88f7fb728e58.png)

[...]

Issue #2495, opened 2021-09-24
Cannot edit properties in layer / FacetChart 


srush:
Following the examples here: 

https://altair-viz.github.io/gallery/grouped_bar_chart_with_error_bars.html

I try to change the size of the graph. 

This gives the error

`SchemaValidationError: Invalid specification altair.vegalite.v4.api.Chart, validating 'required' 'data' is a required property`

My code is exactly the same as the example, with one line to edit the properties. 

```
from vega_datasets import data                                                                                                                
                                                                                                                                              
source = data.barley()                                                                                                                        
                                                                                                                                              
bars = alt.Chart().mark_bar().encode(                                                                                                         
    x='year:O',                                                                                                                               
    y=alt.Y('mean(yield):Q', title='Mean Yield'),                                                                                             
    color='year:N',                                                                                                                           
)                                                                                                                                             
                                                                                                                                              
error_bars = alt.Chart().mark_errorbar(extent='ci').encode(                                                                                   
    x='year:O',                                                                                                                               
    y='yield:Q'                                                                                                                               
)                                                                                                                                             
                                                                                                                                              
o = alt.layer(bars, error_bars, data=source).facet(                                                                                           
    column='site:N'                                                                                                                           
).properties(height=100)   # only new part 
```

jakevdp:
I'm not sure why that is happening (`height` is a valid property for a top-level layer chart according to the schema) but a workaround for the issue is to specify the height on one of the layers rather than on the top-level chart.

jakevdp:
Wait, I see: the top level chart is a `FacetChart`. The issue is that the schema does not allow the `height` attribute on the facet chart. Altair should be updated to respect this. With that in mind, the correct code would be:
```python
alt.layer(bars, error_bars).properties(
    height=100
).facet(                                                                                           
    column='site:N'                                                                                                                           
)
```

srush:
oh wow, this fixed it. thanks!

My issue was that I also wanted `configure_axis` which needs to be outside. But ordering these correctly works. 

```python
chart = (alt.layer(bars, error_bars)
   .properties(height=100, width=100)
   .facet(column="site:N")
   .configure_axis(grid=False))
```

PR #2835, opened 2023-01-15
Check all specs in the context of top level data

### LINKED ISSUE #2609: Improve the error message when trying to use `.properties` on concatenated and facet charts
I think it would be great to provide a better error message when `properties` are used on a concatenated or layered chart. Currently this results in `Data is a required property` which is not really informative since the error is that the chart dimensions need to be specified on the individual charts. I think this is especially confusing since it works for layered charts.

So this works:

```python
import altair as alt

source = data.cars.url

    x='Horsepower:Q',
    y='Miles_per_Gallon:Q',
)
(points + points).properties(width=400)
```

But this raises the error:

```python
(points | points).properties(width=400)
```

A better error message for this was already suggested in https://github.com/altair-viz/altair/issues/2495#issuecomment-926839178. An alternative solution would be to allow `properties` on concatenated/faceted charts to maintain syntactical parity with layered charts and then silently set it on each chart under the hood, but there might be caveats with this so the improved error message seems like the easiest way forward.

Is `properties` ever relevant for concatenated/faceted charts or could we just add a message such as ".properties are not allowed for concatenated or faceted charts, please set it on each individual chart instead." similar to the messages for layering of concatenated/faceted charts?

joelostblom:
This PR fixes #2609 and half of #2744 (and would likely have prevented #2495). There is more detail in #2609, but in brief this allows layer/concat specs to be validated correctly with `validate='deep'` and take top level data into account for each layer, instead of failing because the mark in each layer is missing data.

Example:

```py
points = alt.Chart(data.cars.url).mark_point().encode(
    x='Horsepower:Q',
    y='Miles_per_Gallon:Q',
)
(points & points).properties(width=400)
```

The spec above would previously raise the error:

```
'data' is a required property
```

which should not be raised here since there is data in the spec (see https://github.com/altair-viz/altair/issues/2744#issuecomment-1378704707 for more info). After this PR, the correct error is raised instead, pointing to that the issue is that concat charts do not take `width` (it has to be specified on each individual chart):

```
Additional properties are not allowed ('width' was unexpected)
```

The change here feels a bit ineloquent to me, and I would still like to see if I can figure out a way to only add the top level data if the spec contains a layer or concat. However, I am not super hopeful that I will succeed with that and it took my quite a long time to understand what was going on here and what changes could be made (there are many functions with similar names, exceptions, and functions calling themselves which made this section tricky for me to comprehend and thus slow to work on).

It seems like the suggested solution doesn't break any tests or cause any regressions, although I haven't checked if it adds performance overhead. If there are any regressions you can think of where this would hide a correctly raised `'data' is a required property`-error, please let me know (I tried with empty data and `None` and they both pass both before and after this PR).

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -384,9 +384,13 @@ class SchemaBase(object):
                 "{} instance has both a value and properties : "
                 "cannot serialize to dict".format(self.__class__)
             )
+
         if validate:
             try:
-                self.validate(result)
+                if "top_level_data" in context and "mark" in result:
+                    self.validate({**result, "data": context["top_level_data"]})
+                else:
+                    self.validate(result)
             except jsonschema.ValidationError as err:
                 raise SchemaValidationError(self, err)
         return result
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -551,6 +551,12 @@ class TopLevelMixin(mixins.ConfigMethodMixin):
 
         if original_data is not Undefined:
             context["data"] = original_data
+            # Set top level data so that layered/concat specs don't fail
+            # because data is not present in each layer/concat
+            if isinstance(copy["data"], dict) or copy["data"] is None:
+                context["top_level_data"] = copy["data"]
+            else:
+                context["top_level_data"] = copy["data"].to_dict()
 
         # remaining to_dict calls are not at top level
         context["top_level"] = False


joelostblom:
> I think your solution can work but I'm wondering if we could have a more elaborate logic when the deep validation should be run and when the original error message is the correct one.

Good point! I also noticed that the first error message tends to be correct for the concat/layer specs, but I didn't investigate if that is always the case. I am also not sure if that will sometimes fail to display the more detailed error message in a layer/concat spec. It does for sure in a non-layer/concat, spec as in [the example in this comment](https://github.com/altair-viz/altair/issues/2744#issuecomment-1382864998). Interestingly (and mysteriously), wrapping that spec in a layer or concat find the correct error right away, even after [commenting out deep validation in the source code](https://github.com/altair-viz/altair/blob/ecf726d816d77c759b824b273a5d5ad9e9b81641/altair/vegalite/v5/api.py#L568)...

```py
alt.layer(alt.Chart(data.barley()).mark_bar().encode(
    x='variety',
    y=alt.Y('sum(yield)', stack='asdf'),
))
```

```

        altair.vegalite.v5.api.LayerChart->0, validating 'enum'

        'asdf' is not one of ['zero', 'center', 'normalize']
```

binste:
I suspect that the jsonschema validation already surfaces all those errors and the deeper root causes but I'm not yet sure which one should be shown. I probably have some time tomorrow to read up on the documentation and try to figure out if we can get the same or a better logic as done by the deep validation but without actually doing it. That would be great as validating the indiviudal parts of the schema can fail as other parts are missing as we are experiencing here with the missing data.

Taking your example with `stack='asdf'`, the correct error message can be found with:

```python
import altair as alt
import jsonschema
import jsonschema.validators

# This function is very similar to validate_jsonschema but it gives back the validator
# so we can explore the errors
def create_validator(schema, rootschema=None):
    if rootschema is not None:
        validator_cls = jsonschema.validators.validator_for(rootschema)
        resolver = jsonschema.RefResolver.from_schema(rootschema)
    else:
        validator_cls = jsonschema.validators.validator_for(schema)
        # No resolver is necessary if the schema is already the full schema
        resolver = None

    validator_kwargs = {"resolver": resolver}
    if hasattr(validator_cls, "FORMAT_CHECKER"):
        validator_kwargs["format_checker"] = validator_cls.FORMAT_CHECKER
    validator = validator_cls(schema, **validator_kwargs)
    return validator

chart = alt.Chart(data.barley()).mark_bar().encode(
    x='variety',
    y=alt.Y('sum(yield)', stack='asdf'),
)
chart_dct = chart.to_dict(validate=False)

validator = create_validator(chart._schema, chart._rootschema)
errors = list(validator.iter_errors(chart_dct))
errors[0].context[0].context
```

```
[<ValidationError: "'asdf' is not one of ['zero', 'center', 'normalize']">,
 <ValidationError: "'asdf' is not of type 'null'">,
 <ValidationError: "'asdf' is not of type 'boolean'">]
```
This is even more informative as `stack` can also be `null` or a boolean.

joelostblom:
Great, that looks promising! I think it would be great if deep validation was not necessary and if there is a better approach to fix this than what I suggested here. I will be afk the next few days, but can likely review early next week.

joelostblom:
Closed in favor of #2842

PR #2837, opened 2023-01-16
WIP: Parse shorthand when creating the condition

ChristopherDavisUCI:
This is an attempt to fix the issue from the first example in #2744.

I think this works well for the "standard" cases, but I haven't thought through more advanced cases.

Previously code like `size = alt.condition('true', alt.value(100), alt.value(10))` would be converted by Altair into
```
SizeValue({
  condition: SizeValue({
    test: 'true',
    value: 100
  }),
  value: 10
})
```
but I think the inner of the two `SizeValue`s here is never technically the correct thing to use (because of the `test` property).  As far as I can tell, the main benefit of the inner wrapping was to get shorthands like `"Cylinders:O"` to parse correctly; we address that by doing the parsing as soon as the condition is made.

This PR will convert the above into
```
SizeValue({
  condition: ConditionalPredicateValueDefnumberExprRef({
    test: Predicate('true'),
    value: 100
  }),
  value: 10
})
```

Some tests are failing but at first glance they seem expected.  (One issue is that I have only updated v5, not v4 or v3.)

@joelostblom Could you please try this out and see if it addresses the issue you raised in #2744?

@mattijn @binste I'm also very happy if you can spot any uses of `condition` which will be broken by this PR.

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -14,7 +14,7 @@ import jsonschema
 import pandas as pd
 import numpy as np
 
-from altair.utils.schemapi import SchemaBase, Undefined
+from altair.utils.schemapi import SchemaBase
 
 try:
     from pandas.api.types import infer_dtype as _infer_dtype
@@ -664,10 +664,6 @@ def infer_encoding_types(args, kwargs, channels):
         kwargs[encoding] = arg
 
     def _wrap_in_channel_class(obj, encoding):
-        try:
-            condition = obj["condition"]
-        except (KeyError, TypeError):
-            pass
 
         if isinstance(obj, SchemaBase):
             return obj
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -668,10 +668,6 @@ def infer_encoding_types(args, kwargs, channels):
             condition = obj["condition"]
         except (KeyError, TypeError):
             pass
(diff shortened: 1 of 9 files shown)


altair/utils/core.py
```diff
@@ -668,10 +668,6 @@ def _wrap_in_channel_class(obj, encoding):
             condition = obj["condition"]
         except (KeyError, TypeError):
             pass
```
joelostblom, line 670: I think we can delete this entire try/except block, `condition` does not seem to be used elsewhere.

joelostblom (commented):
Thank you! This does indeed fix the issue I raised and the approach seems to make to me at first glance. I think figuring out why some of the tests are failing and see if that can be fixed easily is a good next step.

joelostblom:
I am not sure why the tests didn't run actually, but usually when that happens I either rebase or make an additional commit (even if it doesn't have anything valuable in it) and push again just to trigger the tests, so try one of those solutions.

You don't need to add this to v3 and v4 for the time being. We are discussing deprecating them and there is a bunch of functionality that would need to be backported if we don't so if we decide to keep them around we can add it all at once.

ChristopherDavisUCI:
Closing this.  An updated version is #2841

PR #2843, opened 2023-01-20
deprecate Vega wrappers and VegaLite 3 wrappers

mattijn:
This PR provides deprecation warnings for `altair.vega.v5` and `altair.vegalite.v3` including mentioning that the modules will be removed in Altair v5.

So in practice it means that the alternative for:
```python
import altair.vega.v5 as alt
```
and
```python
import altair.vegalite.v3 as alt
```
is:
```python
import altair as alt
```

I did not add `AltairDeprecationWarnings` for `altair.vegalite.v4`, since it is imported when doing ```import altair as alt```. I was not able to distunguish
```python
import altair.vegalite.v4 as alt
```
from
```python
import altair as alt
```


binste:
I can't quite follow the logic here. Given that Altair 5 is the next release, who would ever see this deprecation warning and could react to it?

binste:
Great to have these warnings. For the reason you stated, I think it's better to not mention the alternative of `import altair as alt` for the vega wrappers.

mattijn:
Thanks for the review @binste! Merging this

PR #2847, opened 2023-01-22
Remove vegalite v3 and v4 wrappers

binste:
See #2817 for the discussion. Removal of vega v5 wrapper happens in #2829


mattijn:
Nice @binste, `+7 -106.915`👍

PR #2853, opened 2023-01-25
Add documentation for remaining config methods

binste:
The configuration page in the documentation was missing a few `configure_` methods which were added in more recent vega-lite releases. With this PR, all `configure_` methods are documented.


mattijn:
Improvements on the documentation are always great. Thanks!

PR #2870, opened 2023-02-03
Update from vega-lite 5.2.0 to 5.6.0

ChristopherDavisUCI:
Based on the [release notes](https://github.com/vega/vega-lite/releases/tag/v5.6.0) for Vega-Lite 5.6.0, I tried setting the `zero=False` as the scale configuration default.

```
import altair as alt

cars = data.cars.url

c = alt.Chart(cars).mark_circle().encode(
    x = "Weight_in_lbs:Q",
    y = "Acceleration:Q"
).properties(
    height=100,
    width=100
).configure_scale(
    zero=False,
)
```

No errors are raised (even when using `validate="deep"`), whereas errors are raised if I use this code with the current version on GitHub.  I take that as a good sign.

On the other hand, the displayed chart doesn't reflect this change:
chart

Maybe vega-embed is not matching vega-lite 5.6 yet?  That's my first guess.


ChristopherDavisUCI:
I seem to have something wrong in my commit history.  I will close this and try again.

PR #2874, opened 2023-02-06
Make layer warnings clearer

### LINKED ISSUE #2863: doc: ValueError: Faceted charts cannot be layered.
The error: `ValueError: Faceted charts cannot be layered.` is an often reoccurring error:
- https://github.com/altair-viz/altair/issues/2862 
- https://github.com/altair-viz/altair/issues/1570
- https://github.com/altair-viz/altair/issues/1329
- https://github.com/altair-viz/altair/issues/1147
- https://github.com/altair-viz/altair/issues/785

How can we improve the error message and/or improve the documentation, so that this is explained a bit better?

joelostblom:
This is a suggestion to fix #2863

### DIFF
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -2272,24 +2272,38 @@ def _check_if_can_be_layered(spec):
     if encoding is not Undefined:
         for channel in ["row", "column", "facet"]:
             if _get(encoding, channel) is not Undefined:
-                raise ValueError("Faceted charts cannot be layered.")
+                raise ValueError(
+                    "Faceted charts cannot be layered. Instead, layer the charts before faceting."
+                )
     if isinstance(spec, (Chart, LayerChart)):
         return
 
     if not isinstance(spec, (core.SchemaBase, dict)):
         raise ValueError("Only chart objects can be layered.")
     if _get(spec, "facet") is not Undefined:
-        raise ValueError("Faceted charts cannot be layered.")
+        raise ValueError(
+            "Faceted charts cannot be layered. Instead, layer the charts before faceting."
+        )
     if isinstance(spec, FacetChart) or _get(spec, "facet") is not Undefined:
-        raise ValueError("Faceted charts cannot be layered.")
+        raise ValueError(
+            "Faceted charts cannot be layered. Instead, layer the charts before faceting."
+        )
     if isinstance(spec, RepeatChart) or _get(spec, "repeat") is not Undefined:
-        raise ValueError("Repeat charts cannot be layered.")
+        raise ValueError(
+            "Repeat charts cannot be layered. Instead, layer the charts before repeating."
+        )
     if isinstance(spec, ConcatChart) or _get(spec, "concat") is not Undefined:
-        raise ValueError("Concatenated charts cannot be layered.")
+        raise ValueError(
+            "Concatenated charts cannot be layered. Instead, layer the charts before concatenating."
+        )
     if isinstance(spec, HConcatChart) or _get(spec, "hconcat") is not Undefined:
-        raise ValueError("Concatenated charts cannot be layered.")
+        raise ValueError(
+            "Concatenated charts cannot be layered. Instead, layer the charts before concatenating."
+        )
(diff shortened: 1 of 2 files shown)


joelostblom:
Thanks for the comments @mattijn! Here are my thoughts on what you mentioned:

> Maybe the error message should be extended even more stating that usage of row and columns encoding channels are also considered facet?

Although the `column` and `row` channel don't include the name "facet" anywhere, we do specify in [the documentation that they create a "faceted chart"](https://joelostblom.github.io/altair-docs/user_guide/compound_charts.html#faceted-charts) (although only briefly). So I'm leaning towards that it is up to the documentation to contain this information on not include it in the error message directly.

> Addendum: and in this example one can relieve the user, or annoy him/her even more when saying that one can layer points to a chart including a column encoding channel without using a facet by using the point=True on the mark_line().

Similarly here, I think it is up to the user to read the docs and understand that `mark_point` + `mark_line` creates a layered chart, but `mark_line(point=True)` does not create a layered chart (I also think it could make the error message a bit complicated to include this).

> And as a bonus in the edit-modus... 

That's odd, are you running the latest version of altair after the merge of #2842? For me it looks good:

image

mattijn:
> Similarly here, I think it is up to the user to read the docs and understand that mark_point + mark_line creates a layered chart, but mark_line(point=True) does not create a layered chart (I also think it could make the error message a bit complicated to include this).

In the Vega-Lite specification it is just `mark_line(point=True)`, but if you observe the Extended Vega-Lite spec in the Vega-Editor it does creates a layered chart and shifts the `column` encoding channel to a top-level `facet`, see [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAZgAY2AacaYsiygAlMiRoVYcAvpO50AhoTl4QpAE4AbFCDGEADpWQB6Q2DpQAdACtKdTOrhpV5qJkKGougLaG0mBHIBaeUVKV0oAATQARnMAJgBOczZDYLlDSkIIMABrSnMwSjQQGRBPOVVs5UIAT11MLXsXEG5dCBJxZEJVUkwSzChIOhIkVB51Uk8oZXg7Oi1Kas8AIwhNbhq6rSgITxI5TRKAD2msdTnkEGD69dr6i8JMT1bVfeLuapPZrV1VODBrkAbO4gACOpDkRDgimIvmKJUYIgQYhQABYuCBsPQmCgABzogAklDACLKWh0+iMPj8cnMCChjFIS3McAg6WJjzSvn8AXsDx8AFY4kkrJRoMUgA), click bottom-left next to Compiled Vega.

> That's odd, are you running the latest version of altair after the merge of https://github.com/altair-viz/altair/pull/2842? For me it looks good:

After updating I do get the right error:
I meant this:
```python
import altair as alt

stocks = data.stocks.url
c4 = alt.Chart(stocks, height=40, width=80).encode(x="date:T", y="price:Q")
(c4.mark_line() + c4.mark_point()).facet(alt.Facet('symbol:N', columns=3))
```

> ```cmd
> SchemaValidationError: Invalid specification
>         altair.vegalite.v5.api.FacetChart->0, validating 'additionalProperties'
>         Additional properties are not allowed ('columns' was unexpected')
> ```

It is not directly obvious that one needs to move the `columns` in `alt.Facet(..., columns=3)` out of this `alt.Facet()` and into `configure_facet(columns=3)`. But that is a bit off-topic. 

On-topic: for me the difference between `row`, `column` and `facet` as encoding channels vs the `facet` as operator is still a bit vague from the error. What about:
"ValueError: Faceted charts using the `column`, `row` or `facet` as encoding channels cannot be layered. Instead, layer the charts before faceting. To facet a layered chart use the `facet` operator."

I'm open to just improving the docs as well. For me this discussion is already worthy documentation.

mattijn (approved):
My last suggestion to improve the error message is inaccurate as a faceted chart using the `facet` operator also cannot be layered. It's hard to provide concise and accurate error messages including context. 
I'm fine with the PR as is.

PR #2882, opened 2023-02-12
Add Heat Lane example

palewire:
This code attempts to port the VegaLite example found here: https://vega.github.io/vega-lite/examples/bar_heatlane.html

Here's how it turns out in my notebook: 
visualization


mattijn:
Thanks for the PR @palewire! I was about to merge this, but then realised we now have two example directories. One named:
- `examples_arguments_syntax` and one named: 
- `examples_methods_syntax`

The examples in both directories are the same, but they are created using two different approaches.

The approach you used is based on the arguments syntax (current altair v4 approach), but you have placed the file in the methods syntax example directory.

So what is the difference between each other?
For example you have currently defined the X encoding channel options using arguments as such:
```python
x=alt.X(
     "bin_Horsepower_start:Q",
     title="Horsepower",
     axis=alt.Axis(grid=False)
 ),
```
But the method based syntax will allow this to be:
```python
x = alt.X("bin_Horsepower_start:Q").title("Horsepower").axis(grid=False)
```
Three things you can do now:
1. Copy this file to the `examples_arguments_syntax` directory. 
2. Adapt the current file to follow the methods syntax
3. Both

Currently the examples on the altair documentation page are based _only_ on the example files in the `examples_arguments_syntax` directory, but I noticed today that I already miss a location where I can see examples using the methods syntax.

palewire:
Thanks for pointing that out. I didn't know all the details. I can revise this patch and resubmit. Stand by.

mattijn:
Thanks for the continuous stream of new examples @palewire. Much appreciated. All tests pass and I checked the code-diff as well. Merging!

palewire:
Thank you for the feedback and prompt response.

PR #2883, opened 2023-02-12
Use most specific class possible in schema validation errors

binste:
Addresses https://github.com/altair-viz/altair/pull/2568#discussion_r1096814378

Works already well but I'll still want to

- [x] Extend existing and add new tests to check if correct class name appears in error message
- [x] Test against multiple `jsonschema` versions (3, 4, 4.5, 4.16, 4.17)

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -159,15 +159,18 @@ class SchemaValidationError(jsonschema.ValidationError):
         self._additional_errors = getattr(err, "_additional_errors", [])
 
     def __str__(self):
-        try:
-            class_name_raw = self.absolute_path[-1]
-            class_name = class_name_raw[0].upper() + class_name_raw[1:]
-            cls = getattr(vegalite, class_name)
-        except (IndexError, AttributeError):
-            # IndexError if self.absolute_path has 0 length
-            # AttributeError if vegalite does not have attribute class_name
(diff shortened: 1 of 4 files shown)


binste:
@joelostblom This already works if a wrong argument was passed to an encoding channel. Your example https://github.com/altair-viz/altair/pull/2568#issuecomment-1373564808 works anyway even without this PR because `alt.Scale` is immediately validated when it's instantiated. Do you have any other examples from #2568  I should test it against?

joelostblom:
This is looking great! Thanks for getting to it so quickly @binste. I tried a few different examples and they all seem to work fine. One thing that I did notice is that if we use to datum encoding then their message is incorrect:

```py
alt.Chart().mark_point().encode(x=alt.datum(1, wrong_name=1))
```

```

        altair.vegalite.v5.schema.channels.X, validating 'additionalProperties'

        Additional properties are not allowed ('datum', 'wrong_name' were unexpected)
```

`'datum'` should not be included above as an incorrect property to the X channel, instead `wrong_name` should be mentioned as an incorrect parameter to `datum`. The way to handle this is probably with special error checking inside the datum function, and it is not introduced by this pull request, but I thought I would mention it here anyways. The tests here also look reasonable to me.

binste:
Good to go from my side

joelostblom (approved):

PR #2889, opened 2023-02-15
Report filename of failing test

ChristopherDavisUCI:
In another PR, a change was causing the `test_render_examples_to_chart` test to fail, but I had a hard time locating which specific example was failing.  This PR is an attempt to report the filename.

I tested this PR by temporarily adding an error to one of the example files (I used `:ON` rather than `:O` for an encoding type shorthand).  This is what was reported to me by VS Code:

Screenshot 2023-02-15 at 9 48 11 AM


mattijn:
Thanks @ChristopherDavisUCI!

PR #2892, opened 2023-02-18
Add missing attribute descriptions

binste:
# Background
Some attributes on classes are missing descriptions in the docstrings, see for example [`columnPadding` for `alt.Legend`](https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/schema/core.py#L6808):

```python
...
    clipHeight : anyOf(float, :class:`ExprRef`)

    columnPadding : anyOf(float, :class:`ExprRef`)

    columns : anyOf(float, :class:`ExprRef`)
...
```
These descriptions are then also missing in the documentation as the API references are generated based on these docstrings, see [altair.Legend](https://binste.github.io/altair-docs/user_guide/generated/core/altair.Legend.html):

image

Due to this, I find myself often having to go to the [vega-lite docs](https://vega.github.io/vega-lite/docs/legend.html#symbol-layout) to find out what an attribute means. However, the description shown there would also be available in the VL json schema so we just need to adapt the code and docs generation for Altair to properly include it. 

Currently, descriptions for attributes/properties are only parsed during code generation if `description` is defined on the top-level of the attribute, see [`Legend.description`](https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/schema/vega-lite-schema.json#L12681). For some attributes such as `Legend.columnPadding` the [description is nested in `anyOf`](https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/schema/vega-lite-schema.json#L12638) and is currently ignored. Still figuring out when it is useful to take this nested description and what are all the possible cases.

# Current status
image
It already looks much better :) Next up is figuring out if this makes sense for all attributes and if there are still missing descriptions. I think some descriptions are also cut off in the docs but properly included in the Python docstrings. There are also some other formatting issues to take care of. I'm hoping to get this done for the release candidate.

### DIFF
--- a/altair/vegalite/v5/schema/channels.py
+++ b/altair/vegalite/v5/schema/channels.py
@@ -94,8 +94,6 @@ class Angle(FieldChannelMixin, core.FieldOrDatumDefWithConditionMarkPropFieldDef
     """Angle schema wrapper
 
     Mapping(required=[shorthand])
-    A FieldDef with Condition :raw-html:`<ValueDef>` {   condition: {value: ...},   field: ...,
-    ... }
 
     Attributes
     ----------
@@ -433,8 +431,6 @@ class AngleDatum(DatumChannelMixin, core.FieldOrDatumDefWithConditionDatumDefnum
     """AngleDatum schema wrapper
 
     Mapping(required=[])
-    A FieldDef with Condition :raw-html:`<ValueDef>` {   condition: {value: ...},   field: ...,
-    ... }
 
     Attributes
     ----------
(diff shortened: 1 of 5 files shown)


mattijn:
Thanks @binste, the `:class:`ExprRef`` support for attributes is a recent addition to VL. Maybe that is the reason why this was not an issue before.

joelostblom:
This change is awesome! I also often find myself going to the vega-lite docs to read the docstrings and having them in the altair docstrings will be much more helpful :tada:

binste:
Thanks for the context, wasn't aware that `ExprRef` is a new thing! Did some last changes and formatting fixes, it's now good to go from my side. You can preview the docs here https://binste.github.io/altair-docs/user_guide/generated/core/altair.Legend.html


There is another issue with docstrings in the documentation for encoding channels. It seems to have been introduced due to the property setter methods. I will try to open another issue/PR later today for this.

mattijn:
I've to say this looks really great now! Didn't expect we could ever reach VL-level on this🙌. I'm fine with the PR as is, but will let @joelostblom merge it in if he agrees.

joelostblom (approved):
Looks great to me!

PR #2899, opened 2023-02-19
Apply minor change from #2841 in correct tools script

binste:
#2841 Introduced a change here https://github.com/altair-viz/altair/commit/c3f93f072266dc885174124c010422842d223e4e#diff-ff5209c42c54727d16bcb580e10a0b8973304e32333df10d4362b7d4deaefcb6R73 which needs to be applied in the tools script, else it gets overwritten during schema generation. CC @ChristopherDavisUCI, maybe you can quickly double-check if this looks good.


mattijn:
Auto-generated code is great, but it continues to be hard to change it at the right location. Thanks!

PR #2901, opened 2023-02-19
Rename 'Attributes' to 'Parameters' to fix documentation formatting

binste:
With the addition of method-based attribute setting in #2795, Sphinx started generate the documentation differently for the encoding channels. The documentation for attributes is now the first sentence of the documentation which is defined by `_PropertySetter`:

image

This is much less informative then what is in the docstring of `altair.X` and which would be shown otherwise:
```python
    Attributes
    ----------

    shorthand : string
        shorthand for field, aggregate, and type
    aggregate : :class:`Aggregate`
        Aggregation function for the field (e.g., ``"mean"``, ``"sum"``, ``"median"``,
        ``"min"``, ``"max"``, ``"count"`` ).

        **Default value:** ``undefined`` (None)

        **See also:** `aggregate <https://vega.github.io/vega-lite/docs/aggregate.html>`__
        documentation.
    axis : anyOf(:class:`Axis`, None)
        An object defining properties of axis's gridlines, ticks and labels. If ``null``,
        the axis for the encoding channel will be removed.

        **Default value:** If undefined, default `axis properties
        <https://vega.github.io/vega-lite/docs/axis.html>`__ are applied.

        **See also:** `axis <https://vega.github.io/vega-lite/docs/axis.html>`__
        documentation.
```
I tried to disable this behavior but didn't find a way. However, I'd suggest to tackle the root cause of this directly instead of trying to reconfigure Sphinx. Following [the numpydoc style for docstrings](https://numpydoc.readthedocs.io/en/latest/format.html#parameters) (see also [`ExampleError` on this page](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html)), what is called `Attributes` in the Altair class docstrings should, in my understanding, be called `Parameters` as we document what parameters can be passed to e.g. `altair.X` and not what attributes it has after instantiation. As all `Parameters` usually also appear as `Attributes` (with the exception of the ones which are now replaced by the property-based methods) this didn't matter much so far. I'd suggest to now introduce this distinction by replacing `Attributes` with `Parameters` in the docstrings:

```python
    Parameters
    ----------

    shorthand : string
        shorthand for field, aggregate, and type
```
After this change, the documentation renders again how it's supposed to:

image

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -197,7 +197,7 @@ class SchemaValidationError(jsonschema.ValidationError):
 
 
 class UndefinedType(object):
-    """A singleton object for marking undefined attributes"""
+    """A singleton object for marking undefined parameters"""
 
     __instance = None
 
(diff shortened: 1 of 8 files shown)


mattijn:
Good find!

joelostblom (approved):
Another great addition, thank you!

PR #2902, opened 2023-02-19
Add instructions on how to install release candidate

binste:
I think we need some instructions in the docs on how to install the release candidate as `pip install altair` will still install version 4, see https://jugmac00.github.io/til/how-to-install-release-candidates/. CC @mattijn What do you think of this? Had to shorten the existing text to make room for the pip install command.

image

I did not modify the pip install command on the Installation page as I maybe the announcement is enough and easy to spot and remove once version 5 is out but not sure.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -155,7 +155,7 @@ html_theme_options = {
         },
     ],
     "header_links_before_dropdown": 6,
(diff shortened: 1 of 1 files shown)


mattijn:
Good idea! I like it. If @joelostblom agrees as well, then he can merge it in.

joelostblom:
Great idea! When we publish the release candidate we can also create a pinned issue in the altair repo with installation instructions and ask for feedback there. We can link that issue in this banner as well, but let's merge this in first.

joelostblom (approved):

PR #2905, opened 2023-02-20
Expand and document support for column names with special characters

joelostblom:
Continuing on some of the discussion in https://github.com/altair-viz/altair/pull/2824. This issue has been raised a few times in the past so thought it was worth documenting and supporting slightly better:

- https://github.com/altair-viz/altair/issues/284
- https://github.com/altair-viz/altair/issues/2494
- https://github.com/altair-viz/altair/issues/2600

Updated docs:

image


mattijn:
Its good to document this! I keep having the same question that I do not yet see answered. I know why special characters are not allowed when using:
 ```python
shorthand='column_cannot_have_special_characters:Q'`
```
but are special characters allowed if one define it using `field`/`type` combination? 
```python
field='column_with_special_character', type='quantitative'
```
Eg. I found out that `vl-convert` cannot deal with the shorthand approach using a colon-datatype.

joelostblom:
Good point, currently the long form with explicit `type` works for fields with a colon, but not with a period or bracket. Do you think I should update this PR so that colons do not work in long form with explicit type to make things consistent? Even if that limitation is "artifical"

mattijn:
I think the escaping of colons is a feature of using shorthand and should not become a feature of field.

Also I'm fine with the possibility of finding nested data using the brackets. I use this approach myself sometimes and is also used in VL here: https://vega.github.io/vega-lite/examples/facet_bullet.html and in Altair https://stackoverflow.com/a/55214346/2459096

doc/user_guide/encodings/index.rst
```diff
@@ -269,6 +269,11 @@ indicating the data type is optional
 just as for columns without escaped characters.
 Note that the axes titles include the backslashes by default
 and you will need to manually set the title strings to remove them.
+If you are using the long form syntax for encodings,
+you do not need to escape colons as the type is explicit,
+e.g. ``alt.X('col:colon', type='quantitative')``
```
mattijn, line 274:
```suggestion
e.g. ``alt.X(field='col:colon', type='quantitative')``
```

joelostblom:
> I think the escaping of colons is a feature of using shorthand and should not become a feature of field.

Agree. I added a note in the docs about them not having to be escaped in the long form.

> Also I'm fine with the possibility of finding nested data using the brackets.

Yes, I also like having this, just wish it was immediately clear when this is happening, e.g. instead of showing an empty plot it would be great if VL outputed an error message along the lines of "No nested attribute found in the data" or similar. But nothing we can do about that here; this is good to merge from my side.

Issue #2692, opened 2022-10-13
docstrings for new introduced methods regarding params and selection_*


mattijn:
Currently the docstrings for some new introduced parameter methods are rather limited:
<img width="429" alt="image" src="https://user-images.githubusercontent.com/5186265/195705014-5aaa1553-c5b9-4d57-936c-287157440860.png">

Where in the above example I've defined it as such:
```python
highlight = alt.selection_point(
    name="hover",
    on="mouseover",
    nearest=True,
    encodings=["x"],
    empty=False,
    clear="mouseout",
)
```
How can one know about this without docstrings?

ChristopherDavisUCI:
Is it possible some of these issues have been incidentally fixed along the way?  For example, I seem to have docstrings showing up okay.

<img width="365" alt="Screenshot 2023-02-15 at 11 17 57 AM" src="https://user-images.githubusercontent.com/82750240/219130605-0c6dbf3e-d7fb-4d48-922f-b3e015d93251.png">

shift+tab also seems to reveal a docstring (I can't get a screenshot at the moment).

A different issue is that these are the docstrings for the lower-level objects, such as `core.IntervalSelectionConfig`, so for example, `name` is not mentioned, even though `name` is an expected keyword argument for our higher-level Altair function.

Before looking into this further, I wanted to check @mattijn to what extent the issue you reported in October has been resolved.  Thanks!

ChristopherDavisUCI:
@joelostblom I agree it could make sense to hide `alt.section`, so that `alt.selection_point` and `alt.selection_interval` are encouraged.  Do you remember how to do that?

mattijn:
You are right! This is fixed along the way. Great.
<img width="638" alt="image" src="https://user-images.githubusercontent.com/5186265/219148704-1650f5f6-822e-43fd-a3ce-bc1fb039c1a1.png">

If you want to hide the `alt.selection` from the autocomplete, you'll have to assign `alt.selection` a deprecated status around here https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/api.py#L372 by adding:
```python
@utils.deprecation.deprecated(
    message="'selection' is deprecated.  Use xxx"
)
def selection(type=Undefined, **kwds):
```

mattijn:
Hm, not sure. This is the current docstring for `alt.param`, https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/api.py#L303-L320.
```python
def param(name=None, select=None, **kwds):
[...]

PR #2908, opened 2023-02-20
Update parameter docstrings and signatures

ChristopherDavisUCI:
Addresses #2692.

Adds more complete docstrings and signatures for parameter creation functions.  Also deprecates `selection` (in favor of `selection_point` and `selection_interval`).

Although this is primarily about documentation, the signatures were changed, so it is worth checking this carefully.  I did go through the examples [here](https://joelostblom.github.io/altair-docs/user_guide/interactions.html) and they seem to still be working.

This parameters portion of Altair has more customization and divergence from Vega-Lite (I gave one example in https://github.com/altair-viz/altair/issues/2692#issuecomment-1436026066), so these changes will not automatically be updated if changes happen in the Vega-Lite schema.  If anyone sees a more robust way to accomplish this, please let me know or feel free to make changes!

In the doc strings, I just approximated how long of lines we typically used in docstrings and inserted line breaks at corresponding locations.  I'm not sure if there is a way to have VS Code or something else automate that process, or if it's a problem that there is probably some inconsistency in terms of the exact lengths of the lines.

### DIFF
--- a/altair/__init__.py
+++ b/altair/__init__.py
@@ -561,7 +561,6 @@ __all__ = [
     "repeat",
     "sample",
     "schema",
-    "selection",
     "selection_interval",
     "selection_point",
     "sequence",
--- a/altair/__init__.py
+++ b/altair/__init__.py
@@ -561,6 +561,7 @@ __all__ = [
     "repeat",
     "sample",
     "schema",
+    "selection",
     "selection_interval",
     "selection_point",
     "sequence",
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -311,6 +311,37 @@ def param(
     expr=Undefined,
     **kwds,
 ):
+    """Create a named parameter.  See https://altair-viz.github.io/user_guide/interactions.html for examples.  Although both variable parameters and selection parameters can be created using this 'param' function, to create a selection parameter, it is recommended to use either 'selection_point' or 'selection_interval' instead.
+
+    Parameters
+    ----------
+    name : string (optional)
+        The name of the parameter. If not specified, a unique name will be
+        created.
+    bind : :class:`Binding` (optional)
+        Binds the parameter to an external input element such as a slider,
+        selection list or radio button group.
+    value : any (optional)
+        The default value of the parameter. If not specified, the parameter
+        will be created without a default value.
+    empty : boolean (optional)
+        For selection parameters, the predicate of empty selections returns
+        True by default. Override this behavior, by setting this property
+        'empty=False'.
+    expr : :class:`Expr` (optional)
+        An expression for the value of the parameter. This expression may
+        include other parameters, in which case the parameter will
+        automatically update in response to upstream parameter changes.
+    **kwds :
+        additional keywords will be used to construct a parameter.  If 'select'
+        is among the keywords, then a selection parameter will be created.
+        Otherwise, a variable parameter will be created.
+
+    Returns
+    -------
+    parameter: Parameter
+        The parameter object that can be used in chart creation.
+    """
     parameter = Parameter(name)
 
     if empty is not Undefined:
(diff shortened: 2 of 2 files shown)


mattijn:
Thanks @ChristopherDavisUCI! One short comment, I miss the dictating value `expr` in `param` (`alt.param(expr=...)`).

ChristopherDavisUCI:
Thanks @mattijn, I added `expr` just now.  There aren't any `expr` examples here https://joelostblom.github.io/altair-docs/user_guide/interactions.html, do you think one should be included?

There were a few uncertainties like the following.  In https://vega.github.io/vega-lite/docs/parameter.html#expr the type of `expr` is listed as `Expr`, so I put expr : :class:`Expr` (optional) for consistency.  In practice, I think most people will use a string.  Would it be better to have expr : string or :class:`Expr` or maybe some other version?

joelostblom:
Thanks for working on this! I think improved docstrings would be helpful here. Regarding expanding the docs for expressions I also agree that this is need and opened https://github.com/altair-viz/altair/issues/2904 yesterday with what I think would be ideal to have in an `expr` section in the docs.

binste:
Great to have better docstrings and signatures! Just briefly, I saw that there are some __doc__ assignments which I'm not sure if they work in vs code for the same reasoning as we are discussing in the issue about ".encode". I'm on my phone now but could test this until Friday.

mattijn:
> Also deprecates selection (in favor of selection_point and selection_interval).

Can you do this in a separate PR? Then we can also use that PR to replace the occurrences where `alt.selection(type='interval')` currently is used in the docs.

binste:
I see why it would be convenient to define more of the docstrings using assingments to `__doc__`. However, they won't show up in IDEs such as VS Code. See #2920 for some context on why it doesn't work. With your change:

image

Without it:

image

I'd be in favour of accepting the duplication of text across docstrings (something that is done quite often in other libraries as well) so that we have the broadest possible coverage across editors, notebook environments, etc.

mattijn:
Thanks @ChristopherDavisUCI! Do you know if we have covered somewhere the interval selection properties? Such as these from `mark`: https://vega.github.io/vega-lite/docs/selection.html#mark. 

For e.g. setting the `cursor` to a `pointer` when hovering over a select type.

ChristopherDavisUCI:
Hi @mattijn, yes, I think so, does this count? https://github.com/altair-viz/altair/blob/5607dfcff455f3897782ac6b54d9e9d927a78f27/altair/vegalite/v5/api.py#L503-L506
I tried to copy-paste from the Vega-Lite documentation on parameters, but I might have missed some (like the `expr` you caught a few days ago).

binste:
> Thanks @binste for letting me know about that downside of setting `__doc__` programmatically. Can you check this latest version and see if it's working in VS Code as you expect?

Docstrings defined like this are great, thanks! :)

mattijn:
Following the docstring I tried the following, from t[his doc-example](https://joelostblom.github.io/altair-docs/user_guide/interactions.html#interactions-rst-altair-source-17):
```python
point_mouseover = alt.selection_point(
    on='mouseover', 
    toggle=False, 
    empty=False
).mark(cursor='pointer')

make_example(point_mouseover)
```
But got an
```cmd
TypeError: 'GetAttrExpression' object is not callable
```
Wish: method chaining everwhere:).

But I think it needs to be done as such:
```python
point_mouseover = alt.selection_point(
    on='mouseover', 
    toggle=False, 
    empty=False, 
    mark=alt.MarkConfig(cursor='pointer')
)

make_example(point_mouseover)
```
But that gives me this error:
```txt
type        fields    resolve   
clear       nearest   toggle    
encodings   on                  

See the help for `PointSelectionConfig` to read the full description of these parameters
```
Nice error message though!

ChristopherDavisUCI:
Hi @mattijn, is the point that this option is only available for `selection_interval` and not `selection_point`?  But I might be missing the main issue.

mattijn:
Oops, my bad, `mark` is only valid for `selection_interval`:
```python
interval = alt.selection_interval(mark=alt.MarkConfig(cursor='pointer'))
make_example(interval)
```
All good then. I've another comment, but will add to another issue. Nice. Since this is also reviewed by @binste, will merge it in! Thanks @ChristopherDavisUCI

PR #2921, opened 2023-02-23
Move change introduced in #2824 to tools script

binste:
CC @joelostblom, no worries!  Moved a change introduced in https://github.com/altair-viz/altair/commit/389b87b7152381530d03e1ced319851da9e3a053#diff-ff5209c42c54727d16bcb580e10a0b8973304e32333df10d4362b7d4deaefcb6 to the schema generation so that `generate_schema_wrapper` again produces no changes if run against current files.


joelostblom (approved):
Thank you!

PR #2922, opened 2023-02-23
MAINT: Remove inheritance from object for classes

binste:
Small maintenance change. Only removes inheritance of `object` for all classes as this was only necessary to maintain compatibility with Python 2. In Python 3, it does not matter if `object` is specified or not and we can therefore remove it. See https://stackoverflow.com/a/45062077


joelostblom (approved):
Nice, thanks!

PR #2923, opened 2023-02-23
Deprecating selection

ChristopherDavisUCI:
As requested by @mattijn here: https://github.com/altair-viz/altair/pull/2908#issuecomment-1440698123

Add a deprecation decorator to `selection` and update the usage in a few spots.


altair/vegalite/v5/api.py
```diff
@@ -418,6 +418,10 @@ def _selection(type=Undefined, **kwds):
     return param(select=select, **param_kwds)
 
 
+@utils.deprecation.deprecated(
+    message="""'selection' is deprecated.
+    Although it is possible to create a point selection using 'selection(type="point")' and an interval selection using 'selection(type="interval")', for consistency, it is recommended to use 'selection_point()' or 'selection_interval()' instead."""
```
joelostblom, line 423:
```suggestion
   Use 'selection_point()' or 'selection_interval()' instead; these functions also include more helpful docstrings."""
```

joelostblom (changes requested):
Look good! One minor suggestion.

Do you think we should follow the same path for `alt.binding` for consistency? https://joelostblom.github.io/altair-docs/user_guide/API.html#api-functions

ChristopherDavisUCI:
Good catch @mattijn, I was searching for `type="interval"` instead of `type='interval'`.  Should we always use one or the other, at least for this little case?

mattijn:
Regarding:

> Do you think we should follow the same path for `alt.binding` for consistency? https://joelostblom.github.io/altair-docs/user_guide/API.html#api-functions

No. Not yet at least. See eg https://github.com/altair-viz/altair/issues/2760, which uses `alt.binding(input='color')`

Issue #2918, opened 2023-02-22
Reduce the number of entries available in the top level API


joelostblom:
Currently, the top level of the Altair API is huge, there are 573 entries in the tab completion menu of `alt.<TAB>`. This can make it rather hard to find what you're looking for unless you know the first few character of the name already, and exploring the API from scratch as a new user must be intimidating. Many of these seem unhelpful to have exposed at the top-level,  e.g. there are 70 different entries that relate to conditions, but I only use `alt.condition` myself. I suggest that the top level API only includes the most relevant classes and functions that we think are helpful on a day to day basis, and the rest go under sub-modules.

[The Altair API](https://binste.github.io/altair-docs/user_guide/API.html#api-reference) consists of four parts:

1. API functions
2. Top level objects
3. Encoding channels
4. Low level schema wrappers

I have listed my comments below on how we could reduce these, and I would love to hear what you think @mattijn, @ChristopherDavisUCI, @binste  (and anyone else that wants to chime in)

> **Note:** In all the changes I suggest below we should not break any old code, just remove the class/function from tab completion and link it to the new place in the API where this class/function now resides.

## API functions

Introduced specifically by Altair as conveniences and we should keep all of them (although maybe `check_fields_and_encodings` is meant to be a private function based on the name and since there is no docstring?).

![image](https://user-images.githubusercontent.com/4560057/220750231-3c0303c0-2230-42a3-b8d1-5eb14b5ff5d9.png)
(just showing the first few)

## Top level objects

The are only eight of these and we can probably keep all of them. Alternatively, since all except `Chart` have corresponding API functions and are rarely used directly we could move them to a charts sub domain: `alt.charts.HConcatChart`, etc, to make it more clear that the preferred usage is the API method `alt.hconcat`, etc.

![image](https://user-images.githubusercontent.com/4560057/220750304-c182e02f-b907-4eb1-83d4-1c9cbf839f5e.png)

## Encoding channels

There is wasted space here by having each channel repeated three times, e.g. `Angle`, `AngleDatum`, `AngleValue`. Especially as the most common use of datum and value is by using the top level API functions `alt.datum` and `alt.value`. I see to it possible ways of improving the situation:

1. Move the top level classes to methods on the corresponding encoding class: instead of `AngleDatum` we have `Angle.datum`.
2. Moved the top level classes to methods on the API function: instead of `AngleDatum`, we would have `alt.datum.Angle`.

[...]

PR #2927, opened 2023-02-25
Restrict top-level autocompletion for VS Code (PyLance)

binste:
Currently, this PR is only meant to show some code related to discussions happening in #2918, https://github.com/microsoft/pylance-release/discussions/3709#discussioncomment-5093960 and #2614.

As described in https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface, I expanded the wildcard import from vegalite and added `py.typed`. However, `selection` still shows up in autocompletion although it should be considered private by PyLance as it is not mentioned in `__all__`:

image

Opening `altair/__init__.py` in VS Code shows that PyLance is aware that it should be private so this is somehow ignored by the code completion?

image

---
Before this approach could be merged, we'd need to:

- [ ] Understand what implications are of adding `py.typed`. Altair is far off from being “type complete” as it is described [here](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#type-completeness). Maybe what we want to achieve is possible without adding `py.typed` for now?
- [ ] Automate expansion of `from .vegalite import *`


dangotbanned:
Closing as superseded by #3455, #3482

PR #2928, opened 2023-02-25
include section on expressions for interaction

### LINKED ISSUE #2904: Expand docs on expressions
Previously, expressions were mostly (only?) used with `transform_calculate` and it makes sense that they were described at this location in the docs. However, it is now possible to use `alt.expr`/`alt.ExprRef` with expression strings elsewhere so I think it makes sense to explain this clearly in the docs. A natural place could be under "Interactive charts" just above the heading "Future examples". That page currently already talks a lot about parameters in expressions, but we don't touch upon `expr` so it is unclear how these two relates.

A few things I think we should cover:

1. What is an expression?
    - This can be taken from the transform_calculate docs and expanded upon a bit.
2. What is the difference between using a string and `alt.expr`
    - This can be taken from the transform_calculate docs and expanded upon a bit.
    - If we think `alt.expr` should be preferred due to tab completion, maybe we should update some examples in the docs because most seem to use expression strings directly. If we don't think `alt.expr` is preferred, then is it even worth highlighting in the first place?
3. Where can an expression be used and where can it not be used?
4. In which use cases could you use either an expression or a param and when is an expression preferred over a param and vice versa?
5. How to reference a column inside an expression via `datum`
    - Agree on a syntax to use for importing datum. Currently the docs are a mix between `alt.datum` and `alt.expr.datum` (maybe we should always use the former and hide the latter from the public API?):
        - `alt.expr.datum`:
            - https://joelostblom.github.io/altair-docs/user_guide/marks/area.html
            - https://joelostblom.github.io/altair-docs/user_guide/marks/geoshape.html
            - https://joelostblom.github.io/altair-docs/user_guide/compound_charts.html
        - `alt.datum`
            - https://joelostblom.github.io/altair-docs/user_guide/transform/calculate.html
            - https://joelostblom.github.io/altair-docs/user_guide/transform/filter.html
            - ... and many more
6. What is `ExprRef` and `Expr` and should they ever be used (if not, should they be hidden from the public API?).
    - https://joelostblom.github.io/altair-docs/user_guide/generated/core/altair.Expr.html
7. This page use `alt.expr.if_()` which I haven't seen documented anywhere, how does that relate to `alt.condition`? https://joelostblom.github.io/altair-docs/gallery/us_population_over_time.html and https://joelostblom.github.io/altair-docs/gallery/us_population_pyramid_over_time.html
8. Make docs link to `expr` instead of `ExprRef` as per https://github.com/altair-viz/altair/pull/2886#issuecomment-1436113654

mattijn:
This is an attempt to resolve #2904. While not addressing all the details of that issue, I feel this PR captures expressions sufficient to understand the capabilities.

### DIFF
--- a/doc/user_guide/interactions.rst
+++ b/doc/user_guide/interactions.rst
@@ -687,6 +687,60 @@ since they can also be accessed in expression strings:
        selector
     )
 
+Expressions for Interaction
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Altair allows custom interactions by utilizing the expression language of Vega for writing basic formulas. A Vega expression string is a well-defined set of JavaScript-style operations.
+To simplify building these expressions in Python, Altair provides the ``altair.expr`` module, which offers constants and functions to construct expressions using Python syntax. Both JavaScript-syntax and Python-syntax are supported within Altair to define an expression.
+
+In the following example, we define a range connected to a parameter named `width`. We then assign two expressions within the ``alt.param()`` using both JavaScript and Python-syntax.
+Using these two expressions defined as a parameter, we can connect them to an encoding channel option, such as the title color of the axis. If the width is below 200, then the color is ``red``; otherwise, the color is ``blue``.
+
+.. altair-plot::
+
+    bind_range = alt.binding_range(min=100, max=300, name='my chart width ')
+    param_width = alt.param(name='width', bind=bind_range)
+
+    param_color_js_expr = alt.param(expr=f"{param_width.name} < 200 ? 'red' : 'blue'")
+    param_color_py_expr = alt.param(expr=alt.expr.if_(param_width < 200, 'red', 'blue'))
+
+    chart = alt.Chart(df).mark_point(color='black').encode(
+        x=alt.X('xval').axis(titleColor=param_color_js_expr),
+        y=alt.Y('yval').axis(titleColor=param_color_py_expr)
+    ).add_params(
+        param_width, 
+        param_color_js_expr, 
+        param_color_py_expr
+    )
+    chart
+
+In this example, we use a JavaScript-style ternary operator ``f"{param_width.name} < 200 ? 'red' : 'blue'"`` which is equivalent to the Python function ``alt.expr.if_(param_width < 200, 'red', 'blue')``.
+The expressions defined as parameters also need to be added to the chart within ``.add_params()`` to be usable within the chart. The ``alt.expr()`` utility function can be used directly inline.
+
+Expressions can be included within a chart specification using two approaches. One approach is to assign an expression within a parameter definition, as shown above. The second approach is to use an inline expression using the ``altair.expr()`` utility function.
+
+Here, we modify the chart above to change the color of the points based on an inline expression, using JavaScript-style syntax:
+
+.. altair-plot::
+
+    chart = chart.mark_point(color=alt.expr(f"{param_width.name} < 200 ? 'red' : 'blue'"))
+    chart
+
+We apply the expression here to style the mark properties ``color`` based on the width. 
+Inline expressions defined by ``alt.expr(...)`` are not a parameter and, therefore, do not need to be added within the `.add_params()`.
+
+Another option to include an expression within a chart specification is as a value definition within ``alt.value()``. Here, we expand the chart from above to change the size of the point based on the width of the chart. If the width is below ``200``, then the size has a value of ``40``; otherwise, it is ``600``:
+
+.. altair-plot::
+
+    chart.encode(size=alt.value(alt.expr(f"{param_width.name} < 200 ? 40 : 600")))
+
+To summarize expressions:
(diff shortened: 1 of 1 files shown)


doc/user_guide/interactions.rst
```diff
@@ -715,6 +715,62 @@ With interval selections, the ``bind`` property can be set to the value of ``"sc
...
+The second approach is to use an inline expression using the ``altair.expr()`` utility function.
+
+Here, we modify the chart above to change the color of the points based on an inline expression, using JavaScript-style syntax:
+
+.. altair-plot::
+
+    chart = chart.mark_point(color=alt.expr(f"{param_width.name} < 200 ? 'red' : 'blue'"))
+    chart
+
+We apply the expression here to style the mark properties ``color`` based on the width.
```
joelostblom, line 758:
```suggestion
Here, we modify the chart above to change the size of the points based on an inline expression. Instead of creating a conditional statement, we use the value of the expression as the size directly and therefore only need to specify the name of the parameter.

.. altair-plot::

    chart.mark_point(size=alt.expr(param_width.name))

```

doc/user_guide/interactions.rst
```diff
@@ -715,6 +715,62 @@ With interval selections, the ``bind`` property can be set to the value of ``"sc
...
+    chart
+
+We apply the expression here to style the mark properties ``color`` based on the width. 
+Inline expressions defined by ``alt.expr(...)`` are not a parameter and, therefore, do not need to be added within the ``.add_params()``.
+
+Another option to include an expression within a chart specification is as a value definition within ``alt.value()``. Here, we expand the chart from above to change the size of the point based on the width of the chart. If the width is below ``200``, then the size has a value of ``40``; otherwise, it is ``600``:
+
+.. altair-plot::
+
+    chart.encode(size=alt.value(alt.expr(f"{param_width.name} < 200 ? 40 : 600")))
```
joelostblom, line 765:
```suggestion
    chart.encode(size=alt.value(alt.expr(param_width.name)))
```

I made this modification exactly the same as the previous one so that it is easier to see how the two approaches are similar and different

doc/user_guide/interactions.rst
```diff
@@ -715,6 +715,62 @@ With interval selections, the ``bind`` property can be set to the value of ``"sc
...
+
+.. altair-plot::
+
+    bind_range = alt.binding_range(min=100, max=300, name='my chart width ')
+    param_width = alt.param(name='width', bind=bind_range)
+
+    param_color_js_expr = alt.param(expr=f"{param_width.name} < 200 ? 'red' : 'blue'")
+    param_color_py_expr = alt.param(expr=alt.expr.if_(param_width < 200, 'red', 'blue'))
+
+    chart = alt.Chart(df).mark_point(color='black').encode(
```
joelostblom, line 735:
```suggestion
    param_color_js_expr = alt.param(expr=f"{param_width.name} < 200 ? 'red' : 'black'")
    param_color_py_expr = alt.param(expr=alt.expr.if_(param_width < 200, 'red', 'black'))

    chart = alt.Chart(df).mark_point().encode(
```

This simplifies seeing how two of the later examples I added are similar

doc/user_guide/interactions.rst
```diff
@@ -715,6 +715,62 @@ With interval selections, the ``bind`` property can be set to the value of ``"sc
...
+
+We apply the expression here to style the mark properties ``color`` based on the width. 
+Inline expressions defined by ``alt.expr(...)`` are not a parameter and, therefore, do not need to be added within the ``.add_params()``.
+
+Another option to include an expression within a chart specification is as a value definition within ``alt.value()``. Here, we expand the chart from above to change the size of the point based on the width of the chart. If the width is below ``200``, then the size has a value of ``40``; otherwise, it is ``600``:
+
+.. altair-plot::
+
+    chart.encode(size=alt.value(alt.expr(f"{param_width.name} < 200 ? 40 : 600")))
+
```
joelostblom, line 766:
```suggestion
`Some parameter names have special meaning in Vega-Lite <https://vega.github.io/vega-lite/docs/parameter.html#built-in-variable-parameters>`_, for example, naming a parameter `width` will automatically link it to the width of the chart. In the example below, we also modify the chart title to show the value of the parameter:

.. altair-plot::
    bind_range = alt.binding_range(min=100, max=300, name='Chart width: ')
    param_width = alt.param('width', bind=bind_range)
    
    # In Javascript, a number is converted to a string when added to an existing string,
    # which is why we use this nested quotation.
    title=alt.Title(alt.expr(f'"This chart is " + {param_width.name} + " px wide"'))
    alt.Chart(df, title=title).mark_point().encode(
        x=alt.X('xval'),
        y=alt.Y('yval')
    ).add_params(
        param_width,
    )

```

joelostblom (changes requested):
This is great, thank you! I finally understand how expression works more properly. I have some suggestions to make it a bit simpler, and maybe we could add one sentence about when expressions are preferred over parameters and vice versa (if it is easy to summarize in a sentence)?

joelostblom (approved):

PR #2932, opened 2023-02-26
Fix deprecation warning for SelectableGroup in Python 3.7

binste:
Fixes deprecationwarning discussed in https://github.com/altair-viz/altair/discussions/2599#discussioncomment-5113306. I used the same approach as they use in SQLAlchemy https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/util/compat.py#L139 which should provide the broadest possible compatibility by relying on the new `select` interface if it is present and else fall back on the old `get`, independent of if the standard library or `importlib_metadata` is used. CC @mattijn. Don't see any deprecation warnings related to this anymore for any Python version we test against.

### DIFF
--- a/altair/utils/plugin_registry.py
+++ b/altair/utils/plugin_registry.py
@@ -122,10 +122,7 @@ class PluginRegistry(Generic[PluginType]):
     def names(self) -> List[str]:
         """List the names of the registered and entry points plugins."""
         exts = list(self._plugins.keys())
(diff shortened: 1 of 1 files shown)


mattijn:
Thanks for pushing this last bit @binste!

binste:
Thank you @mattijn! Really excited about the release candidate.

PR #2929, opened 2023-02-25
use selection_point or selection_interval syntax

mattijn:
More examples that were not (yet) included in https://github.com/altair-viz/altair/pull/2923

PR #2933, opened 2023-02-26
Fix a few remaining ocurrences of `type=`

joelostblom:
Follow up on #2929

### DIFF
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -410,7 +410,7 @@ def _selection(type=Undefined, **kwds):
         select = core.PointSelectionConfig(type="point", **kwds)
         warnings.warn(
             """The types 'single' and 'multi' are now
-        combined and should be specified using "type='point'".""",
+        combined and should be specified using "selection_point()".""",
             utils.AltairDeprecationWarning,
         )
     else:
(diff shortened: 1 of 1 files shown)


altair/vegalite/v5/api.py
```diff
@@ -461,7 +461,7 @@ def selection_interval(
     zoom=Undefined,
     **kwds,
 ):
-    """Create a selection parameter with `type='interval'`.  Selection parameters define data queries that are driven by direct manipulation from user input (e.g., mouse clicks or drags). Selection parameters with `type='interval'` are used to select a continuous range of data values on drag.  (The current alternative is `type='point'`, which is used to select multiple discrete data values, and which is created using 'selection_point'.)
+    """Create an interval selection parameter. Selection parameters define data queries that are driven by direct manipulation from user input (e.g., mouse clicks or drags). Interval selection parameters are used to select a continuous range of data values on drag, whereas point seleection parameters (`selection_point`) are used to select multiple discrete data values.)
```
mattijn, line 464:
```suggestion
    """Create an interval selection parameter. Selection parameters define data queries that are driven by direct manipulation from user input (e.g., mouse clicks or drags). Interval selection parameters are used to select a continuous range of data values on drag, whereas point selection parameters (`selection_point`) are used to select multiple discrete data values.)
```

mattijn (approved):

PR #2940, opened 2023-03-04
Fail build workflow if generate_schema_wrapper produces changes

### LINKED ISSUE #2906: PR tests should fail if generate_schema_wrapper produces changes
Altair has many files which are auto-generated based on the Vega-Lite schema. For development, it is often convenient to modify those directly but before merging into master, these changes need to be applied to the relevant scripts in the `tools` folder which is a step that is easy to forget.

It would be great if the tests for a PR fail if changes were applied to auto-generated files. One way could be to add a new step in [build.yml](https://github.com/altair-viz/altair/blob/ef8ff946bf610c19ccbecfbbd6b624004c25fb32/.github/workflows/build.yml) which runs `python tools/generate_schema_wrapper.py` and then checks if any files were changed with `git ls-files --modified`.

binste:
Closes #2906. Tested that it works on another branch in my forked repo, see the failed workflow [here](https://github.com/binste/altair/actions/runs/4330777174/jobs/7562209179):

image

Tests for this PR fail right now because test_schemapi.py is not "clean" as I added some modifications in the wrong place. Will fix this in another PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,6 +26,28 @@ jobs:
           pip install .[dev]
           # pip install "selenium<4.3.0"
           # pip install altair_saver
+      - name: Test that schema generation has no effect
+        run: |
+          python tools/generate_schema_wrapper.py
+          # This gets the paths of all files which were either deleted, modified
+          # or are not yet tracked by Git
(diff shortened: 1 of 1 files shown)


joelostblom (approved):
This is great to have!  Will merge after confirming with #2941

PR #2941, opened 2023-03-04
Apply changes of #2931 in tools folder

binste:
In #2931 I didn't apply the changes in the correct tools script. Created #2940 to prevent this from happening in the future.


joelostblom (approved):
Thank you!

PR #2946, opened 2023-03-06
Shorten page title

joelostblom:
Currently there is a window width that makes the navbar looks like this, which I think is rather ugly:

image

This PR changes the title so that the navbar always looks like the following until it is collapsed in the hamburger menu:

image


binste (approved):
Looks good to me

PR #2949, opened 2023-03-06
Type hints: Improve for encoding channel attributes

binste:
This improves on the existing and very nice implementation done by @ChristopherDavisUCI.

* Parse nested anyOf definitions which were so far represented as a single entry with non-specific `**kwds`.
* I also added types for the items in lists, so instead of a generic `list` type hint we get more specific ones such as `List[str]`. Also works for altair classes.
* Furthermore, it type hints enums as [literals](https://mypy.readthedocs.io/en/stable/literal_types.html). This tells a type checker that only the specified values are accepted. Also helpful in VS Code:

image

All of the above changes together turn this:

```python
   def sort(self, **kwds) -> 'Angle':
```

into this

```python
    def sort(self, _: List[float], **kwds) -> 'Angle':
        ...

    @overload
    def sort(self, _: List[str], **kwds) -> 'Angle':
        ...

    @overload
    def sort(self, _: List[bool], **kwds) -> 'Angle':
        ...

    @overload
    def sort(self, _: List[core.DateTime], **kwds) -> 'Angle':
        ...

    @overload
    def sort(self, _: Literal["ascending", "descending"], **kwds) -> 'Angle':
```

In many cases, this is just an improvement over the existing type hints. However, for some such as `X.title` this fixes wrong type hints which will lead to errors when checking code with mypy. The old implementation does not correctly recognize that title can be of type `str` or `List[str`:

```python
    @overload
    def title(self, **kwds) -> 'X':
        ...

    @overload
    def title(self, _: None, **kwds) -> 'X':
        ...
```

New:
```python
    def title(self, _: str, **kwds) -> 'X':
        ...

    @overload
    def title(self, _: List[str], **kwds) -> 'X':
        ...

    @overload
    def title(self, _: None, **kwds) -> 'X':
```

See #2951 for an overview of all efforts related to type hints in Altair.

### DIFF
--- a/altair/vegalite/v5/schema/channels.py
+++ b/altair/vegalite/v5/schema/channels.py
@@ -1,11 +1,17 @@
 # The contents of this file are automatically written by
 # tools/generate_schema_wrapper.py. Do not modify directly.
 
+import sys
 from . import core
 import pandas as pd
 from altair.utils.schemapi import Undefined, with_property_setters
 from altair.utils import parse_shorthand
-from typing import overload, Type
+from typing import overload, List
+
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
 
 
 class FieldChannelMixin:
@@ -322,7 +328,7 @@ class Angle(FieldChannelMixin, core.FieldOrDatumDefWithConditionMarkPropFieldDef
     _encoding_name = "angle"
 
     @overload
-    def aggregate(self, _: str, **kwds) -> 'Angle':
+    def aggregate(self, _: Literal["average", "count", "distinct", "max", "mean", "median", "min", "missing", "product", "q1", "q3", "ci0", "ci1", "stderr", "stdev", "stdevp", "sum", "valid", "values", "variance", "variancep"], **kwds) -> 'Angle':
         ...
 
     @overload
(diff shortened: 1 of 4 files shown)


mattijn:
Thanks @binste, took me a while to notice that there are top and down arrows in the type hint so you can iterate through them. Here I'm at `3/5`: 
image
Which gives me hints on the `LocalMultiTimeUnit` from within the VL-schema. This is good.

But if I make a mistake in this, eg `yearmonthdatehour` instead of `yearmonthdatehours` 
```cmd
SchemaValidationError: 'yearmonthdatehour' is an invalid value for `timeUnit`:

'yearmonthdatehour' is not one of ['year', 'quarter', 'month', 'week', 'day', 'dayofyear', 'date', 'hours', 'minutes', 'seconds', 'milliseconds']
'yearmonthdatehour' is not one of ['utcyear', 'utcquarter', 'utcmonth', 'utcweek', 'utcday', 'utcdayofyear', 'utcdate', 'utchours', 'utcminutes', 'utcseconds', 'utcmilliseconds']
```
It seems it only returns the `SingleTimeUnit` anyOf as suggestion (`LocalSingleTimeUnit` and `UtcSingleTimeUnit`). This observation is maybe more connected to https://github.com/altair-viz/altair/issues/2873 and not necessarily to this PR.

binste:
Nice catch! I propose a fix in #2957. Over time with a better understanding of the VL json schema, these error messages can hopefully be further improved.

PR #2950, opened 2023-03-07
Type hints: Add static type checker mypy

binste:
Adds mypy as a linter and fixes all current errors. This is a prerequisite for the subsequent tasks outlined in #2951.


mattijn:
Thanks for this PR @binste!
Can we make use of the pipe operator introduced in 3.10 through typing_extensions?

Eg
```python 
Dict[str, Any] | None
```
Instead of
```python
Optional[Dict[str, Any]]
```

binste:
Unfortunately no. The pipe operator throws a SyntaxError prior to 3.10, see https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-x-y-syntax-for-unions. This can be circumvented by doing the "Future Annotations Import"  `from __future__ import annotations` which seems to be a highly debated topic in the Python world and support by default for this behaviour has been postponed a few times, see https://github.com/python/cpython/issues/82786. I'd therefore suggest that we stick to `Union` and `Optional` to have the broadest possible compatibility.

mattijn:
That's unfortunate. It would improve readability of the type definitions.

PR #2952, opened 2023-03-07
Add missing descriptions to tables created with altair-object-table

binste:
#2892 added missing attribute descriptions to the docstrings and hence to the API reference sections in the docs. This PR applies the same fix to the tables created with `altair-object-table` throughout the user guide. The functionality used to fetch the description is the same as in the other PR (`SchemaInfo.deep_description`).

For example on the marks page it turns this:

image

Into:
image

You can browse the updated docs at https://binste.github.io/altair-docs/


mattijn:
Thanks @binste, nice!

PR #2954, opened 2023-03-10
Temporarily use default data transformer while saving a chart

### LINKED ISSUE #2953: save using vl-convert doesn't work with json data_transformer
In altair 5.0.0rc1, vl-convert produces images without any data points when the json transformer is enabled.

Example code:
```python
import altair as alt

alt.data_transformers.enable('json')

chart = alt.Chart(
    data.iris()
).mark_point().encode(
    x='sepalWidth', y='sepalLength'
)
chart.save('test.png', engine='vl-convert', scale_factor=2)
```
In my jupyter notebook, I see
visualization (8)

but the created `test.png` looks like:
test

binste:
Fixes #2953. See that issue for the related discussion. This PR moves the context manager added in #2724 into the `save` function so that it's already used when converting the chart to a dictionary. For all other lines, only the indentation changed.

### DIFF
--- a/altair/utils/mimebundle.py
+++ b/altair/utils/mimebundle.py
@@ -1,5 +1,4 @@
 from .html import spec_to_html
-from ..vegalite.v5.data import data_transformers
 
 
 def spec_to_mimebundle(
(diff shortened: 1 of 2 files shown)


mattijn:
What happens when a data URL references a remote JSON file? Will that also be in-lined? Do we have an example of this?

binste:
The default data transformer does not modify URLs which are explicitly passed and so the data is not in-lined in these cases. vl-convert can read remote JSON files and the saving works. Just double-checked this with the following example:

```python
import altair as alt

source = data.movies.url

    alt.X("IMDB_Rating:Q", bin=True),
    y='count()',
)

chart.save('test.png', engine='vl-convert')
```

PNG looks good. And using the debugger and stepping into the saving function, you can observe that the "data" attribute still contains the URL and not an inlined dataset: `{'url': 'https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/movies.json'}`.

mattijn:
great

PR #2957, opened 2023-03-12
Include all options in enum error messages

binste:
Fixes what @mattijn found in https://github.com/altair-viz/altair/issues/2949#issuecomment-1465028524. So far, for the error messages only the errors at the same leaf level of the "error tree" where considered. For errors which show that a value is not part of an enum, it seems to make sense to find and list all enums in leaves.

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -84,9 +84,14 @@ def _get_most_relevant_errors(
     if len(errors) == 0:
         return []
 
+    # Start from the first error on the top-level as we want to show
+    # an error message for one specific error only even if the chart
+    # specification might have multiple issues
+    top_level_error = errors[0]
+
     # Go to lowest level in schema where an error happened as these give
-    # the most relevant error messages
-    lowest_level = errors[0]
+    # the most relevant error messages.
+    lowest_level = top_level_error
     while lowest_level.context:
         lowest_level = lowest_level.context[0]
 
@@ -102,9 +107,25 @@ def _get_most_relevant_errors(
             # In this case we are still at the top level and can return all errors
             most_relevant_errors = errors
         else:
-            # Return all errors of the lowest level out of which
+            # Use all errors of the lowest level out of which
             # we can construct more informative error messages
             most_relevant_errors = lowest_level.parent.context
+            if lowest_level.validator == "enum":
+                # There might be other possible enums which are allowed, e.g. for
+                # the "timeUnit" property of the "Angle" encoding channel. These do not
+                # necessarily need to be in the same branch of this tree of errors that
+                # we traversed down to the lowest level. We therefore gather
+                # all enums in the leaves of the error tree.
+                enum_errors = _get_all_lowest_errors_with_validator(
+                    top_level_error, validator="enum"
+                )
+                # Remove errors which already exist in enum_errors
+                enum_errors = [
+                    err
+                    for err in enum_errors
+                    if err.message not in [e.message for e in most_relevant_errors]
+                ]
+                most_relevant_errors = most_relevant_errors + enum_errors
 
     # This should never happen but might still be good to test for it as else
     # the original error would just slip through without being raised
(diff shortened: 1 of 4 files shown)


mattijn:
Thanks @binste! Now the following returns the right error suggestions:
```python
import altair as alt
x_tu = alt.X().timeUnit('yearmonthdatehour')
alt.Chart().encode(x_tu)
```
```cmd
SchemaValidationError: 'yearmonthdatehour' is an invalid value for `timeUnit`:

'yearmonthdatehour' is not one of ['year', 'quarter', 'month', 'week', 'day', 'dayofyear', 'date', 'hours', 'minutes', 'seconds', 'milliseconds']
'yearmonthdatehour' is not one of ['utcyear', 'utcquarter', 'utcmonth', 'utcweek', 'utcday', 'utcdayofyear', 'utcdate', 'utchours', 'utcminutes', 'utcseconds', 'utcmilliseconds']
'yearmonthdatehour' is not one of ['yearquarter', 'yearquartermonth', 'yearmonth', 'yearmonthdate', 'yearmonthdatehours', 'yearmonthdatehoursminutes', 'yearmonthdatehoursminutesseconds', 'yearweek', 'yearweekday', 'yearweekdayhours', 'yearweekdayhoursminutes', 'yearweekdayhoursminutesseconds', 'yeardayofyear', 'quartermonth', 'monthdate', 'monthdatehours', 'monthdatehoursminutes', 'monthdatehoursminutesseconds', 'weekday', 'weeksdayhours', 'weekdayhoursminutes', 'weekdayhoursminutesseconds', 'dayhours', 'dayhoursminutes', 'dayhoursminutesseconds', 'hoursminutes', 'hoursminutesseconds', 'minutesseconds', 'secondsmilliseconds']
'yearmonthdatehour' is not one of ['utcyearquarter', 'utcyearquartermonth', 'utcyearmonth', 'utcyearmonthdate', 'utcyearmonthdatehours', 'utcyearmonthdatehoursminutes', 'utcyearmonthdatehoursminutesseconds', 'utcyearweek', 'utcyearweekday', 'utcyearweekdayhours', 'utcyearweekdayhoursminutes', 'utcyearweekdayhoursminutesseconds', 'utcyeardayofyear', 'utcquartermonth', 'utcmonthdate', 'utcmonthdatehours', 'utcmonthdatehoursminutes', 'utcmonthdatehoursminutesseconds', 'utcweekday', 'utcweeksdayhours', 'utcweekdayhoursminutes', 'utcweekdayhoursminutesseconds', 'utcdayhours', 'utcdayhoursminutes', 'utcdayhoursminutesseconds', 'utchoursminutes', 'utchoursminutesseconds', 'utcminutesseconds', 'utcsecondsmilliseconds']
```

One question, maybe you now this directly, why isn't the following directly returning an error:
```python
x_tu
```
```cmd
X({
  timeUnit: 'yearmonthdatehour'
})
```
Is it because the encoding channel properties are only inferred based on the encoding channel in which it is defined? 
So an encoding channel property can not be tested against the jsonschema without a corresponding encoding channel?

binste:
The main part where the schema is validated is when `.to_dict` is called, see https://github.com/altair-viz/altair/blob/master/altair/utils/schemapi.py#L551. When you simply print `x_tu` which still represents the `X` encoding channel, not the `timeUnit` property, this does not happen but it happens when you display a chart as the dictionary is used in the renderer.

Some objects are already being validated when there instantiated by calling `to_dict` in the `__init__` method, see https://github.com/altair-viz/altair/blob/master/altair/utils/schemapi.py#L367. For encoding channels [this is deactivated](https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/schema/channels.py#L13816) as it requires the context of the chart to infer some attributes. I think expanding shorthands is such a case.

`alt.TimeUnit` does not depend on its context and is therefore valid at instantiation and gets validated:
image

mattijn:
Thanks for digging deep! So we are seeing the behavior introduced by https://github.com/altair-viz/altair/pull/546. More clear now.

In line with your comment here, https://github.com/altair-viz/altair/issues/2850#issuecomment-1400649740, maybe better to alias `DEBUG_MODE` with `VALIDATE_ON_INIT` and deprecate the former. Will open another issue for this.

PR #2962, opened 2023-03-14
Remove $ from bash examples on installation page

palewire:
In my experience, many newbies are thrown by the $ and think they have to type it themselves.


mattijn:
Make sense!👍

PR #2968, opened 2023-03-15
Consolidate naming history into overview

palewire:
These edits correct the formal name of the project to the new Vega-Altair style and move the origin of the name to this space, where it seems more appropriate to me.


mattijn:
Thanks @palewire!

PR #2969, opened 2023-03-15
Reduce line-height in the gallery thumbnail headlines

palewire:
This is what we have now:

Screen Shot 2023-03-15 at 3 36 31 PM

Here's how this change should turn out:

Screen Shot 2023-03-15 at 3 36 20 PM


mattijn:
Took me a while to see the change. Its subtle.

PR #2970, opened 2023-03-15
Trim index page language

palewire:
We're saying the same thing more than once here. It might be nice to tighten things up so our homepage has a stronger punch.


mattijn:
There are currently some conflicts. Can you resolve these please?

palewire:
I believe that's done.

mattijn:
Thanks @palewire!

PR #2944, opened 2023-03-04
Simplify and clean up tool folder

binste:
Closes #2907

PR #2973, opened 2023-03-18
Maintenance: Remove test_schemapi.py in tools folder

binste:
Removes the `test_schemapi.py` file and the enclosing folder in the `tools` folder. I forgot to already do this in #2944 after which `test_schemapi.py` is directly maintained in the `tests` folder and no longer copied over.


mattijn:
Thanks @binste. Seems we were doing similar things, let me update my PR based on this:)

binste:
Thanks @mattijn! :)

Issue #2873, opened 2023-02-05
Combine multiple error messages


binste:
Based on feedback from @joelostblom [here](https://github.com/altair-viz/altair/pull/2842#issuecomment-1405578872):

> On brief comment now, do you think it is confusing to see three error messages like in your example above? I don't know what would be a feasible solution technically here, but I am thinking that instead of
> 
> ```
> 'asdf' is not one of ['zero', 'center', 'normalize']
> 'asdf' is not of type 'null'
> 'asdf' is not of type 'boolean'
> ```
> 
> the message would be clearer as:
> 
> ```
> 'asdf' needs to be one of ['zero', 'center', 'normalize'] or of type 'null' or boolean'
> ```

Also see subsequent comments in that thread. https://github.com/altair-viz/altair/pull/2842#discussion_r1089687484 is also relevant.

binste:
`alt.Chart().encode(alt.Angle().sort("invalid_value"))` gives:

```python
'invalid_value' is not of type 'array'
'invalid_value' is not of type 'array'
'invalid_value' is not of type 'array'
'invalid_value' is not of type 'array'
```

Error messages should be deduplicated based on `.message` attribute. Furthermore, this should ideally show all literal values which are acceptable such as ["x", "y", ...]. 

Maybe one for the long list but we could investigate if it make sense to show the error messages of all leaves in this error tree that we're traversing to get the relevant error messages. Or should it be dependent on what kind of object the user provided as a value, e.g. if a string such as `"invalid_value"` is provided then it should show acceptable string values but if an unsupported type is passed then it should show which types are accepted?

joelostblom:
I also saw some duplicated rounds of messages in https://github.com/altair-viz/altair/issues/2896

binste:
FYI, I'm working on this as part of #2915 and #2913. Feel free to assign to me

PR #2975, opened 2023-03-18
Deduplicate error messages

### LINKED ISSUE #2896: Multiple rounds of the same error message
I noticed that some specs shows the same error multiple time, for example:

```py
source = data.cars()
alt.Chart(source).mark_text(align="right").encode(
    alt.Text("Horsepower:N", bandPosition='4')
)
```

```
'4' is not of type 'number'
Additional properties are not allowed ('field' was unexpected)
'4' is not of type 'number'
Additional properties are not allowed ('bandPosition', 'field', 'type' were unexpected)
```

Sometimes these messages can become quite long:

```python
alt.Text("Horsepower:N", type=dict(text="Horsepower", align="right"))
```

```
{'text': 'Horsepower', 'align': 'right'} is not one of ['quantitative', 'ordinal', 'temporal', 'nominal']
{'text': 'Horsepower', 'align': 'right'} is not of type 'string'
Additional properties are not allowed ('field' was unexpected)
{'text': 'Horsepower', 'align': 'right'} is not one of ['quantitative', 'ordinal', 'temporal', 'nominal', 'geojson']
{'text': 'Horsepower', 'align': 'right'} is not of type 'string'
Additional properties are not allowed ('field', 'type' were unexpected)
```

Granted that these multiple errors are likely rare this is probably not high priority (and maybe even desirable for some reasons that escapes me right now), but I thought I would document them anyways.

binste:
Fixes #2896 and [this comment](https://github.com/altair-viz/altair/issues/2873#issuecomment-1465150688):

> `alt.Chart().encode(alt.Angle().sort("invalid_value"))` gives:
> 
> ```python
> 'invalid_value' is not of type 'array'
> 'invalid_value' is not of type 'array'
> 'invalid_value' is not of type 'array'
> 'invalid_value' is not of type 'array'
> ```
> 
> Error messages should be deduplicated based on `.message` attribute

Further improvements to messages generated out of multiple errors can be tackled separately as part of #2873


mattijn:
Thanks @binste!

After this PR, the following examples return as such:

```python
import altair as alt

alt.Chart().encode(alt.Angle().sort("invalid_value"))
```
```cmd
SchemaValidationError: 'invalid_value' is an invalid value for `sort`:

'invalid_value' is not of type 'array'

alt.Chart(...)
```

```python
source = data.cars()
alt.Chart(source).mark_text(align="right").encode(
    alt.Text("Horsepower:N", bandPosition='4')
)
```
```cmd
SchemaValidationError: '4' is an invalid value for `bandPosition`:

'4' is not of type 'number'
Additional properties are not allowed ('field' was unexpected)
Additional properties are not allowed ('bandPosition', 'field', 'type' were unexpected)

alt.Chart(...)
```

```python
alt.Text("Horsepower:N", type=dict(text="Horsepower", align="right")).to_dict()
```
```cmd
SchemaValidationError: '{'text': 'Horsepower', 'align': 'right'}' is an invalid value for `type`:

{'text': 'Horsepower', 'align': 'right'} is not one of ['quantitative', 'ordinal', 'temporal', 'nominal']
{'text': 'Horsepower', 'align': 'right'} is not of type 'string'
```

No duplication of error messages anymore. Will merge once tests pass.

One question, why did you had to add the `$` sign?

binste:
Great, thanks for double-checking! Regarding the `$`, the expected messages we use in the tests are treated as regex by pytest. `$` matches the end of the line, else this test:

```python
SchemaValidationError: 'invalid_value' is an invalid value for `sort`:

'invalid_value' is not of type 'array'
```

would also pass if the message looks like this:

```python
SchemaValidationError: 'invalid_value' is an invalid value for `sort`:

'invalid_value' is not of type 'array'
'invalid_value' is not of type 'array'
'invalid_value' is not of type 'array'
'invalid_value' is not of type 'array'
```

I then also added it to the existing examples as an additional restriction just to be safe.

Issue #2951, opened 2023-03-07
Type hints: Summary issue


binste:
Currently, Altair only has a few type hints to help IDEs provide better autocompletion suggestions. More type hints would further help IDEs provide a better development experience as well as static type checkers to catch potential errors before they appear at runtime. Especially applications (e.g. a company-internal web application) and frameworks (e.g. Streamlit, Great Expectations) which want to tightly integrate with Altair could greatly profit from this.

I'd suggest the following approach for Altair which is in line with [Using mypy with an existing codebase](https://mypy.readthedocs.io/en/stable/existing_code.html). Very open for inputs!

## Should happen before the v5 release
- [x] https://github.com/altair-viz/altair/pull/2949 Improvements in here could wait until after v5 but it also fixes some wrong type hints

## v5 or minor release
I think it's preferable to have the following items more or less bundled into one release from which on Altair is declared as a library supporting type checks. This could be a minor release (5.1?) as it should not affect any runtime behaviour. Goal is to have as much as possible of the "public API" of Altair typed and then incrementally do the rest later on if we see value in it.

- [x] https://github.com/altair-viz/altair/pull/2950 after #2949 is merged. This is a prerequisite for the following tasks so that mypy can tell us if the type hints are reasonable. We're in a fortunate position that we have many example charts in the `tests` folder which mypy can use for this.
- [ ] With mypy added in the previous step, we should aim to type as much as possible of the "public API" of Altair so users can get the most benefit out of this exercise. "internal" code can follow later. How about the following prioritization:
  - [x] `schemapi.py`: #3142
  - [x] `altair/vegalite/v5/api.py`: #3143
  - [x] Parts of folders "vegalite", "v5", and "utils": #2976
  - [ ] For the autogenerated files, we should infer types from the VL schema, see https://github.com/altair-viz/altair/issues/2854: #3208
  - [x] Remaining files in utils: mimebundle.py, save.py: #3248 
  - [ ] Finishing touches #3263 
    - [x]  Type hint `Undefined` as `UndefinedType`. Search for `# Remove "ignore" statement once Undefined is no longer typed as Any` in code which was introduced in #2950 to make mypy pass.
    - [x] Go through current code and search for `# type: ignore` and see if any of those mentions can now be removed or better type hinted
[...]

PR #2976, opened 2023-03-18
Type hints: Parts of folders "vegalite", "v5", and "utils"

binste:
I think it's easier to type hint and review if we can iteratively do it for a few files, review, merge and then continue so I'll stop here and will do the other files in these folders once this is merged. 

See #2951 for the overall progress of type hinting Altair.

- [x] Once #3081 is merged, use `DataFrameLike` protocol from there instead of [SupportsDataFrame](https://github.com/altair-viz/altair/pull/2976/files#diff-478a9046db15a358587cf77454cd2cf9cc4723435276cad0864c268d17e07650R34). Or use `SupportsDataFrame`?


altair/utils/data.py
```diff
@@ -2,63 +2,84 @@
...
+    import pyarrow.lib
+
+
+class _SupportsGeoInterface(Protocol):
+    __geo_interface__: MutableMapping
+
+
+class _SupportsDataframe(Protocol):
+    def __dataframe__(self, *args, **kwargs):
+        ...
```
jonmmease, line 36: We can replace this with `_DataFrameLike` now:

https://github.com/altair-viz/altair/blob/f3938bff6b32446623ad56455224c9b341a0f066/altair/utils/core.py#L40-L42

altair/utils/data.py
```diff
@@ -91,34 +114,71 @@ def limit_rows(data, max_rows=5000):
...
+    type: Literal["json"]
+
+
+class _CsvFormatDict(TypedDict):
+    type: Literal["csv"]
+
+
+class _DataJsonUrlDict(TypedDict):
+    url: str
+    format: _JsonFormatDict
```
jonmmease, line 166: Should format be optional?

altair/utils/data.py
```diff
@@ -2,63 +2,84 @@
...
+class _SupportsDataframe(Protocol):
+    def __dataframe__(self, *args, **kwargs):
+        ...
+
+
+_DataType = Union[dict, pd.DataFrame, _SupportsGeoInterface, _SupportsDataframe]
+_TDataType = TypeVar("_TDataType", bound=_DataType)
+
+_VegaLiteDataDict = Dict[str, Union[str, dict, List[dict]]]
+_ToValuesReturnType = Dict[str, Union[dict, List[dict]]]
```
jonmmease, line 43: I'm not sure the best way to do this, but should the type encode that this dictionary only has a top-level "values" key?

altair/utils/core.py
```diff
@@ -339,20 +347,20 @@ def to_list_if_array(val):
             # Work around bug in to_json for categorical types in older versions of pandas
             # https://github.com/pydata/pandas/issues/10778
             # https://github.com/altair-viz/altair/pull/2170
-            col = df[col_name].astype(object)
+            col = df[col_name].astype(object)  # type: ignore[call-overload]
```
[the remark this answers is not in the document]
binste: Hmm I removed it and mypy is still happy... Not sure what changed but I now configured mypy to fail if it detects redundant `type: ignore` statements and removed all of them.

altair/utils/data.py
binste: Yeah I agree it would be nice and we could use a `TypedDict`. Unfortunately, these lines in the function allow a dictionary to have more then just `values` as a top-level key in case `data` is a dictionary:

```python
    elif isinstance(data, dict):
        if "values" not in data:
            raise KeyError("values expected in data dict, but not present.")
        return data
```

I think there is currently no way to use a `TypedDict` with some fixed keys and then allow for arbitrary extra keys, see [this discussion over in the mypy repo](https://github.com/python/mypy/issues/4617).

altair/utils/data.py
binste: So far this is only used as a return type for `to_json` which always returns `format: _JsonFormatDict` so I think it's accurate for that usage. But maybe you mean it could be confused to mean that a data json dictionary always needs format specified? I renamed the classes to make it clearer that these are not general schemas but specific type hints for those functions.

Does that address it?

binste:
Thank you @jonmmease for the review! I addressed your comments and left two open for you to check. In any case, I'll leave it open for a while to see if Mattijn has any inputs. There is no rush to get this merged.

jonmmease (approved):
Thanks for the updates and clarifications. lgtm!

altair/utils/core.py
```diff
@@ -200,7 +201,12 @@ def infer_dtype(value):
 ]
 
 
-def infer_vegalite_type(data):
+InferredVegaLiteType = Literal["ordinal", "nominal", "quantitative", "temporal"]
```
mattijn, line 204: I think it can also be a `geojson` type here? See https://altair-viz.github.io/user_guide/generated/core/altair.Type.html for the options. For my own understanding, can the available type options also be inferred from the schema here? And if so, is there a reason not to use this? Eg. if type `array` is added to Vega-Lite (not sure if this happens) then type support for this requires a manual update here?

mattijn (commented):
Thanks for this PR @binste! One in-line comment + question.

altair/utils/core.py
binste: I think I see where you're coming from. However, this type is used to annotate what values the `infer_vegalite_type` function can return. It is not used in the sense of a schema for all possible vega lite types. It should therefore only contain the types that can be returned and this would only change if we manually update that function. Currently, the function cannot return `geojson`.

Keeping the concepts of types and VL schema separate might be challenging during this whole process. I also struggle with it and I agree that where possible we should infer one from the other.

As a side note: In the last commit I renamed `InferredVegaLiteType` to `_InferredVegaLiteType` so it's private which is in line with the discussion we had [here](https://github.com/altair-viz/altair/pull/3081#issuecomment-1587799122).

altair/utils/core.py
```diff
@@ -200,7 +201,12 @@ def infer_dtype(value):
 ]
 
 
-def infer_vegalite_type(data):
+_InferredVegaLiteType = Literal["ordinal", "nominal", "quantitative", "temporal"]
+
+
+def infer_vegalite_type(
+    data: Union[np.ndarray, pd.Series]
```
mattijn, line 208: Another question, here you mention: `data: Union[np.ndarray, pd.Series]`. When I look to where this is used: https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/api.py#L2411. I cannot understand why we need to be restrictive on the input types from numpy and pandas. I really like types like `DataFrameLike` and `SupportsGeoInterface` for input data features.

altair/utils/core.py
mattijn: Thanks for commenting. That clarifies it a bit more!
Not sure yet about the public names I'm not sure we should place words like `Return` and `Func` in type definitions.

altair/utils/core.py
mattijn: I tried to define a type hint named `ArrayLike1D` something that can serve for one-dimensional data of various forms, but not sure if this is possible yet (eg. https://github.com/python/mypy/issues/12280).

Otherwise, maybe something like they define as type hint for the `values` of a polars series? There they define an `ArrayLike` type as such (see [here](https://github.com/pola-rs/polars/blob/main/py-polars/polars/series/series.py#L139-L147):
```python
ArrayLike = Union[
    Sequence[Any],
    "Series",
    "pa.Array",
    "pa.ChunkedArray",
    "np.ndarray",
    "pd.Series",
    "pd.DatetimeIndex",
]
```
And in the docstring for values of the series they mention:
```
values : ArrayLike, default None
    One-dimensional data in various forms. Supported are: Sequence, Series,
    pyarrow Array, and numpy ndarray.
```
I don't think this type-hint is strictly 1D as the docstring suggest, but it can serve as a reference. 
Somehow wished it would be possible to include `range` as well (ref: https://github.com/altair-viz/altair/issues/2877).

binste: You're right, that's too restrictive. I used the type hints which were already present in the docstring but seems like this function supports anything that the pandas function `infer_type` can handle. According to [their own type hints](https://github.com/pandas-dev/pandas/blob/main/pandas/_libs/lib.pyx#L1403) this is all `object` so basically everything. I'd suggest that we type hint it the same to be consistent with pandas. Implemented in e974bc96d7ba07dc09f0a1d866c00c30c8b34c07

Btw, the function `infer_vegalite_type` is used only [here](https://github.com/altair-viz/altair/blob/master/altair/utils/core.py#L576). Your link points to the usage of `infer_encoding_types`.

mattijn: That is flexible enough, thanks for the change.

mattijn (approved):

binste:
Thank you for the reviews!

Just fyi, I won't be able to do any development in the next 2 weeks as I'm travelling. I will continue with the type hints in July/August.

PR #2979, opened 2023-03-20
Display more helpful error message when two fields strings are used in condition

### LINKED ISSUE #2935: Unhelpful error message when a field is incorrectly used as the outer else value in a `condition`
The underlying issue in this spec is that conditions can't have an outer value in the else condition that equals a field, [it has to be a value or a datum](https://stackoverflow.com/a/75598865/2166823), but the error message gives no hint that this is the case:

```py
import altair as alt

source = data.cars()

alt.Chart(source).mark_circle(size=60).encode(
    x='Horsepower',
    y='Miles_per_Gallon',
    color='Origin',
    tooltip=alt.condition(
        alt.datum.Origin == 'USA',
        'Name',
        'Miles_per_Gallon'
    )
)
```

```
Tool tip has no parameter "field"
...
```

The old error message was not very helpful here either, but at least not as directly misleading. On altair 4.2 I see this:

```
SchemaError: '#/definitions/FieldOrDatumDefWithCondition<MarkPropFieldDef,(Gradient|string|null)>' is not a 'uri-reference'

Failed validating 'format' in metaschema['properties']['$ref']:
    {'format': 'uri-reference', 'type': 'string'}

On schema['$ref']:
    '#/definitions/FieldOrDatumDefWithCondition<MarkPropFieldDef,(Gradient|string|null)>'
```

joelostblom:
This is an attempt to resolve https://github.com/altair-viz/altair/issues/2935.

If there are other string that can be passed to conditions that are not references for fields, then that would be an issue for this PR, but I can't think of any on the top of my head. It could also be argued that maybe this should be fixed in VegaLite, but it seems helpful and straightforward to fix it directly in Altair.

### DIFF
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -775,7 +775,7 @@ def condition(predicate, if_true, if_false, **kwargs):
     elif isinstance(if_true, str):
         if isinstance(if_false, str):
             raise ValueError(
-                "A field cannot be used in both the `if_true` and `if_false` branches of a condition. One of the them has to specify a `value` or `datum` definition."
+                "A field cannot be used as both the `if_true` and `if_false` values of a condition. One of them has to specify a `value` or `datum` definition."
             )
         else:
             if_true = utils.parse_shorthand(if_true)
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -774,7 +774,7 @@ def condition(predicate, if_true, if_false, **kwargs):
         if_true = if_true.to_dict()
     elif isinstance(if_true, str):
         if isinstance(if_false, str):
-            raise ValueError('A field cannot be used in both the `if_true` and `if_false` branches of a condition. One of the them has to specify a `value` or `datum` definition.')
+            raise ValueError("A field cannot be used in both the `if_true` and `if_false` branches of a condition. One of the them has to specify a `value` or `datum` definition.")
         else:
             if_true = utils.parse_shorthand(if_true)
             if_true.update(kwargs)
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -773,8 +773,11 @@ def condition(predicate, if_true, if_false, **kwargs):
         # dict in the appropriate schema
         if_true = if_true.to_dict()
     elif isinstance(if_true, str):
-        if_true = utils.parse_shorthand(if_true)
-        if_true.update(kwargs)
+        if isinstance(if_false, str):
+            raise ValueError('A field cannot be used in both the `if_true` and `if_false` branches of a condition. One of the them has to specify a `value` or `datum` definition.')
+        else:
+            if_true = utils.parse_shorthand(if_true)
+            if_true.update(kwargs)
     condition.update(if_true)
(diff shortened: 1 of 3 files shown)


altair/vegalite/v5/api.py
```diff
@@ -773,8 +773,13 @@ def condition(predicate, if_true, if_false, **kwargs):
         # dict in the appropriate schema
         if_true = if_true.to_dict()
     elif isinstance(if_true, str):
-        if_true = utils.parse_shorthand(if_true)
-        if_true.update(kwargs)
+        if isinstance(if_false, str):
+            raise ValueError(
+                "A field cannot be used in both the `if_true` and `if_false` branches of a condition. One of the them has to specify a `value` or `datum` definition."
```
mattijn, line 778: _`branches of a condition`_ is this the correct way to describe this? I've never heard of this before.

joelostblom: I was hesitant about writing it like that too, but went with it because [the VegaLite docs use something similar](https://vega.github.io/vega-lite/docs/condition.html#field)"

> When using a conditional field definition, only value or datum may be specified as the else (outer) branch. 

Do you think changing the error message to the following would be clearer?

> A field cannot be used as both the `if_true` and `if_false` values of a condition. One of the them has to specify a `value` or `datum` definition.

mattijn: Yes, I think that's better (single them).

joelostblom: Updated!

altair/vegalite/v5/api.py
```diff
@@ -773,8 +773,13 @@ def condition(predicate, if_true, if_false, **kwargs):
         # dict in the appropriate schema
         if_true = if_true.to_dict()
     elif isinstance(if_true, str):
-        if_true = utils.parse_shorthand(if_true)
-        if_true.update(kwargs)
+        if isinstance(if_false, str):
+            raise ValueError(
+                "A field cannot be used as both the `if_true` and `if_false` values of a condition. One of them has to specify a `value` or `datum` definition."
```
mattijn, line 778:
```suggestion
                "A field cannot be used for both the `if_true` and `if_false` values of a condition. One of them has to specify a `value` or `datum` definition."
```
Please ignore the suggested change if it not improves the wording.

tests/utils/test_schemapi.py
```diff
@@ -610,6 +610,20 @@ def test_chart_validation_errors(chart_func, expected_error_message):
         chart.to_dict()
 
 
+def test_multiple_field_strings_in_condition():
+    selection = alt.selection_point()
+    expected_error_message = "A field cannot be used as both the `if_true` and `if_false` values of a condition. One of them has to specify a `value` or `datum` definition."
```
mattijn, line 615:
```suggestion
    expected_error_message = "A field cannot be used for both the `if_true` and `if_false` values of a condition. One of them has to specify a `value` or `datum` definition."
```

altair/vegalite/v5/api.py
joelostblom: I think either is fine so I will accept yours!

mattijn (approved):

PR #2981, opened 2023-03-22
Restructure and clarify interactive docs

joelostblom:
As I was adding the new interactive examples from https://github.com/altair-viz/altair/issues/2760, I noticed several things I think we could express more clearly so I went through and overhauled this page. It is mostly clarification, and making titles a bit more succinct so that they are easier to get an overview from. I also restructured some sections mostly the widget one to try to give a more structured walk-through of the different options (I'm open to ideas if anyone thinks my suggestion could be improved).

Docs are live here so that it is easier to review https://joelostblom.github.io/altair-docs/user_guide/interactions.html#

Before:
image

After:
image

### DIFF
--- a/doc/user_guide/interactions.rst
+++ b/doc/user_guide/interactions.rst
@@ -560,6 +560,7 @@ after a selection has been made in a radio button or drop-down
 .. altair-plot::
 
     # Make radio button less cramped by adding a space after each label
+    # The spacing will only show up in your IDE, not on this doc page
     options = ['Europe', 'Japan', 'USA']
     labels = [option + ' ' for option in options]
 
--- a/doc/user_guide/interactions.rst
+++ b/doc/user_guide/interactions.rst
@@ -46,8 +46,8 @@ Here is a simple scatter-plot created from the ``cars`` dataset:
     cars = data.cars.url
 
     alt.Chart(cars).mark_circle().encode(
-        x='Miles_per_Gallon:Q',
-        y='Horsepower:Q',
+        x='Horsepower:Q',
+        y='Miles_per_Gallon:Q',
         color='Origin:N'
     )
 
@@ -67,8 +67,8 @@ In order to use this variable in the chart specification, we explicitly add it t
     op_var = alt.param(value=0.1)
 
     alt.Chart(cars).mark_circle(opacity=op_var).encode(
-        x='Miles_per_Gallon:Q',
-        y='Horsepower:Q',
+        x='Horsepower:Q',
+        y='Miles_per_Gallon:Q',
         color='Origin:N'
     ).add_params(
         op_var
@@ -81,8 +81,8 @@ It's reasonable to ask whether all this effort is necessary. Here is a more natu
     op_var2 = 0.1
 
     alt.Chart(cars).mark_circle(opacity=op_var2).encode(
-        x='Miles_per_Gallon:Q',
-        y='Horsepower:Q',
+        x='Horsepower:Q',
+        y='Miles_per_Gallon:Q',
         color='Origin:N'
     )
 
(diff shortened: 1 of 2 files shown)


joelostblom:
I think I finally found a sensible way to organize the widget section in my last commit!

doc/user_guide/interactions.rst
```diff
@@ -569,69 +540,107 @@ With point selections, an input element can be added to the chart to establish a
...
+
+So far we have seen the use of selections
+to lookup points with precisely matching values in our data.
+This is often useful,
+but sometimes we might want to make a more complex comparison
+than an exact match.
+For example,
+we might want to create a condition
+where we want to access the value of a slider
+to use directly in a comparison with values of the data.
```
mattijn, line 635:
```
For example,
we might want to create a condition
where we want to access the value of a slider
to use directly in a comparison with values of the data.
```
It is a bit hard for me to understand this sentence, especially the last part. After reading it two times I get what you mean. Maybe split in two sentences?

doc/user_guide/interactions.rst
```diff
@@ -695,12 +704,148 @@ often provides a more convenient syntax
...
+
+    alt.Chart(source).mark_circle().encode(
+        x=alt.X('value:Q', title=''),
+        y='Horsepower',
+        color='Origin',
+    ).add_params(
+        selection
+    ).transform_filter(
+        selection
+    )
```
mattijn, line 824:
```suggestion
```
I think we can delete this part. Just stick to the new routine of altair v5. This page is so extensive already.

doc/user_guide/interactions.rst
```diff
@@ -805,53 +1008,3 @@ Some possible use cases for the above interactivity are not currently supported
 1. If we are using a ``selection_point``, it would be natural to want to return information about the chosen data point, and then process that information using Python.  This is not currently possible (and as of December 2021 it does not seem likely to become possible any time soon), so any data processing will have to be handled using tools such as ``transform_calculate``, etc. You can follow the progress on this in the following issue: https://github.com/altair-viz/altair/issues/1153.
```
mattijn, line 1008:
```suggestion
1. If we are using a ``selection_point``, it would be natural to want to return information about the chosen data point, and then process that information using Python. This is not currently possible, so any data processing will have to be handled using tools such as ``transform_calculate``, etc. You can follow the progress on this in the following issue: https://github.com/altair-viz/altair/issues/1153.
```

mattijn (commented):
Great improvements @joelostblom! I left a few comments in-line.

doc/user_guide/interactions.rst
```diff
@@ -263,8 +240,49 @@ We can modify the brush definition, and leave the rest of the code unchanged:
 
     chart.encode(x='Acceleration:Q') | chart.encode(x='Miles_per_Gallon:Q')
```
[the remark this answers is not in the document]
joelostblom: I agree with this and updated the examples on the page

joelostblom (commented):
Thank you for the helpful comments! I have updated accordingly.

doc/user_guide/interactions.rst
```diff
@@ -261,10 +244,51 @@ We can modify the brush definition, and leave the rest of the code unchanged:
...
+For example, in ``transform_filter(brush)``,
+we are again using the selection parameter ``brush`` as a predicate.
+Data points which evaluate to ``True`` (i.e., data points which lie within the selection) are kept,
+and data points which evaluate to ``False`` are filtered out.
+
+It is not possible to both select and filter in the same chart,
+so typically this functionality will be used when at least two sub-charts are present.
+In the following example,
+we attach the selection parameter to the upper chart,
+and then filter data in the lower chart based selection in the upper chart.
```
mattijn, line 263:
```suggestion
and then filter data in the lower chart based on the selection in the upper chart.
```

doc/user_guide/interactions.rst
```diff
@@ -695,12 +719,125 @@ often provides a more convenient syntax
...
+but it is often more convenient to switch to a selection parameter
+and specifying a field/encoding.
+
+Logic-Driven Comparisons
+------------------------
+
+Data-driven lookups and comparisons are very helpful
+but sometimes we might want to make a logic comparison
+in the conditional statement instead
+that is independent of the data.
```
mattijn, line 734:
```suggestion
A logic comparison is a type of comparison
that is based on logical rules and conditions,
rather than on the actual data values themselves.
```

mattijn (approved):
Few textual improvements. Looks good

joelostblom:
Great! I accepted all your suggestions. Merging.

PR #2985, opened 2023-03-23
Trim that eliminates redundancy

palewire:
Pandas is listed first in the list. That's good enough, I think.


joelostblom (approved):
Thanks! Actually I think we should clarify the text a bit now that we have supported for multiple dataframe packages. I will do that in a follow up PR.

PR #2986, opened 2023-03-23
Clarify differences between pandas and other dataframe packages

joelostblom:


mattijn (approved):
This is better indeed. With another method to include data, the following issue becomes more relevant again: https://github.com/altair-viz/altair/issues/2761.
Maybe keep this page simple, with references to subpages?

joelostblom:
Yes, I think it makes sense to merge this now and then we can move the complexity into subpages with just a simple example on this main page as you said in a PR that addresses  #2761

PR #2991, opened 2023-03-24
Simplify syntax

joelostblom:
As per the discussion in https://github.com/altair-viz/altair/discussions/2988#discussioncomment-5422305

### DIFF
--- a/tests/examples_arguments_syntax/bar_chart_with_labels.py
+++ b/tests/examples_arguments_syntax/bar_chart_with_labels.py
@@ -7,19 +7,12 @@ This example shows a basic horizontal bar chart with labels created with Altair.
 import altair as alt
 from vega_datasets import data
 
+
 source = data.wheat()
 
-bars = alt.Chart(source).mark_bar().encode(
-    x='wheat:Q',
-    y="year:O"
+base = alt.Chart(source).encode(
+    x='wheat',
+    y="year:O",
+    text='wheat'
 )
-
-text = bars.mark_text(
-    align='left',
-    baseline='middle',
-    dx=3  # Nudges text to right so it doesn't appear on top of the bar
-).encode(
-    text='wheat:Q'
-)
-
-(bars + text).properties(height=900)
+base.mark_bar() + base.mark_text(align='left', dx=2)


mattijn (commented):
One question. The `text` encoding channel is not included within the `mark_bar()`.
Should the base not be just `x` and `y` and include the `text` as part of the `mark_text().encode(text='wheat')`?

mcp292:
Personally, I thought that was the cool part: that you define your encoding and then they fit the respective marks which either use them in full, or in part.

joelostblom:
To me, there is a conceptual difference between Altair and VegaLite here. In Altair, we can think of it as if it is the Chart object that takes the encoding, and then any number of marks can be added to it, as in the example in this PR:

```py
    x='wheat',
    y="year:O",
    text='wheat'
)
base.mark_bar() + base.mark_text(align='left', dx=1)
```

However, in VegaLite every mark has to be coupled to its own encoding (I think), so the reason the above spec works is that Altair expands it into [a VegaLite spec](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAZgAY2AacaYsiygAlMiRoVYcAvpO4AbAIYBPTACcUAbVABbeSoDWeEIUUAHTChAAjXSBkhMUSHRJJURzAA9xb+Jll0LbEZMeXFuYzMLAEdSeSI4QlC4NHM7D0Nff0Dg0JBw03NkEBi44kTiFNtuRQysLKLlG3zIoogVZyh5WVtpTm1dAzcIwvcvPJAuxCgLWUwYMJA6dOQAJjsHJxdDQk9vUEyAoqCQheHo2Pjy5NTuZf26w5Bj3OaRksukyrsanweLRrUrwsbQ6XR6kgAutw6KF5IZSCpukUxIQTJRkAB6DFgOhQAB0ACtKHQ-MkVHioJhCBioCYtBiUgh5ABaGGJShUygAATQAEY8SsAJx4tgYtnyDHPQiEyjQKogAAklDAwR0FhRaMxDMwTLxCASjFIljxcAgGOVqoljJZsgSmAZAFY8QA2PH8olyyRAA) that contains a redundant `text` encoding for the bars:

```json
  "layer": [
    {
      "mark": {"type": "bar"},
      "encoding": {
        "text": {"field": "wheat", "type": "quantitative"},
        "x": {"field": "wheat", "type": "quantitative"},
        "y": {"field": "year", "type": "ordinal"}
      }
    },
    {
      "mark": {"type": "text", "align": "left", "dx": 2},
      "encoding": {
        "text": {"field": "wheat", "type": "quantitative"},
        "x": {"field": "wheat", "type": "quantitative"},
        "y": {"field": "year", "type": "ordinal"}
      }
    }
  ],
```

I believe that being able to think of an `encoding` as a property of a `Chart` and that any `mark` added to that chart can make use of those encodings is a useful level of abstraction to think about the Altair syntax even if it leads to slight increase in the VegaLite verbosity (notably, it is also similar to how users can think of their charts in ggplot and seaborns new api)

PR #3005, opened 2023-03-30
Correct the method-based syntax for a few more gallery examples

joelostblom:


tests/examples_methods_syntax/isotype.py
```diff
@@ -69,7 +69,7 @@
 alt.Chart(source).mark_point(filled=True, opacity=1, size=100).encode(
     alt.X('x:O').axis(None),
     alt.Y('animal:O').axis(None),
-    alt.Row('country:N').header(title=''),
+    alt.Row('country:N').title(''),
```
mattijn, line 72:
```suggestion
    alt.Row('country:N').title(None),
```
I prefer `None` over `''`.

tests/examples_methods_syntax/layered_chart_with_dual_axis.py
```diff
@@ -11,7 +11,7 @@
 source = data.seattle_weather()
 
 base = alt.Chart(source).encode(
-    alt.X('month(date):T').axis(title=None)
+    alt.X('month(date):T').title('')
```
mattijn, line 14:
```suggestion
    alt.X('month(date):T').title(None)
```
Or there is another reason why you opt for `''`?

mattijn (commented):
Two inline comments/suggestions

tests/examples_methods_syntax/isotype.py
joelostblom: This is interesting. I think `None` is more semantically correct but `''`  is more convenient to type and I remember that when I started to use Altair a few years ago, there was some context where `None` didn't work, but `''` did. It is possible that I was just doing something wrong back then, but I will see if I can find it again.

mattijn: Might be, I can remember something similar. But I would consider that as a limitation of Vega-Lite at that moment.

joelostblom:
@mattijn I forgot about this for some while, but I noticed it now and went ahead with your suggestion of using `None` instead of `""` as I agree that it is more idomatic.

mattijn (approved):

PR #3008, opened 2023-04-01
use `ruff` as linter

mattijn:
This PR add ruff as linter over flake8. This in order to work towards a complete pyproject.toml-based build, see corresponding PR https://github.com/altair-viz/altair/pull/3007.

I've moved the flake8 configuration to the corresponding ruff section in the pyproject.toml file. I tried to synchronize the same tests for now. But still had to apply some fixes in the code to get in sync
- This PR include code fixes for B028, B904, B006 and B018. 
- I've excluded TID252 (relative imports) and B905 (valid from python>3.10). 
- E203, E266 and W503 checks are not yet available in ruff, see https://github.com/charliermarsh/ruff/issues/2402, so these cannot be excluded yet (as they were for flake8).

Ready for review.


mattijn:
I cannot assign you as reviewer @binste, but would you be willing to check this?

altair/utils/schemapi.py
```diff
@@ -564,11 +564,17 @@ def to_dict(self, validate=True, ignore=None, context=None):
             try:
                 self.validate(result)
             except jsonschema.ValidationError as err:
-                raise SchemaValidationError(self, err)
+                raise SchemaValidationError(self, err) from err
```
binste, line 567:
```suggestion
                # We do not raise `from err` and therfore ignore 
                # the B904 linting rule as else the resulting
                # traceback is very long as it contains part
                # of the Vega-Lite schema. It would also first
                # show the less helpful ValidationError instead of
                # the more user friendly SchemaValidationError
                raise SchemaValidationError(self, err)  # noqa: B904
```

tools/schemapi/schemapi.py
```diff
@@ -562,11 +562,17 @@ def to_dict(self, validate=True, ignore=None, context=None):
             try:
                 self.validate(result)
             except jsonschema.ValidationError as err:
-                raise SchemaValidationError(self, err)
+                raise SchemaValidationError(self, err) from err
```
binste, line 565:
```suggestion
                # We do not raise `from err` and therfore ignore 
                # the B904 linting rule as else the resulting
                # traceback is very long as it contains part
                # of the Vega-Lite schema. It would also first
                # show the less helpful ValidationError instead of
                # the more user friendly SchemaValidationError
                raise SchemaValidationError(self, err)  # noqa: B904
```

altair/utils/schemapi.py
```diff
@@ -564,11 +564,17 @@ def to_dict(self, validate=True, ignore=None, context=None):
...
         return result
 
     def to_json(
-        self, validate=True, ignore=[], context={}, indent=2, sort_keys=True, **kwargs
+        self,
+        validate=True,
+        ignore=None,
```
binste, line 573: Can't comment in the docstring below but could you add an `(optional)` for ignore there same as for context?

tools/schemapi/schemapi.py
```diff
@@ -562,11 +562,17 @@ def to_dict(self, validate=True, ignore=None, context=None):
...
         return result
 
     def to_json(
-        self, validate=True, ignore=[], context={}, indent=2, sort_keys=True, **kwargs
+        self,
+        validate=True,
+        ignore=None,
```
binste, line 571: Same remark here about docstring as in altair/utils/schemapi.py

setup.cfg
```diff
@@ -1,16 +0,0 @@
...
-exclude = 
-  .git,
-  build,
-  __pycache__,
-  tests/examples_arguments_syntax,
-  tests/examples_methods_syntax,  
-  altair/vegalite/v?/schema
-
-[metadata]
-description-file = README.md
```
binste, line 15: Can the metadata seciton here already be removed or should that happen as part of #3007?

binste (changes requested):
Sure! Always happy to review.

Thank you for tackling this and also for going through the rules to see what is available! I'm amazed by the speed of ruff and how many rules are already implemented... Maybe we can even use it as a replacement for black at one point: https://github.com/charliermarsh/ruff/issues/1904

Looks mostly good to me. Added some smaller code suggestions. Could you also replace the mention of flake8 in contributing.md?

.github/workflows/lint.yml
```diff
@@ -21,9 +21,10 @@ jobs:
       run: |
         black --diff --color .
         black --check .
-    - name: Lint with flake8
-      run: |
-        flake8 .  --statistics
+    - name: Lint with ruff
+      run: |        
+        ruff check .
```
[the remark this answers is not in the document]
mattijn: No, not intentional. First tried to get the same behaviour as what we had with flake8, but that did not gave me a trace of the errors. Let me remove the line on `ruff --statistics .`

Makefile
```diff
@@ -5,7 +5,8 @@ install:
 
 test :
 	black --diff --color --check .
-	flake8 . --statistics
+	ruff check .
+	ruff --statistics .
```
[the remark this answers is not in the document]
mattijn: as in lint.yml, deleted this line. `ruff check .` is sufficient.

tools/schemapi/schemapi.py
mattijn: Done

setup.cfg
mattijn: Valid point. Should be part of #3007. Undelete metadata section.

altair/utils/schemapi.py
mattijn: done

mattijn (commented):
Thanks for the review @binste. I've one inline question remaining.

mattijn:
Hm, it seems there are some sync issues with VSCode replies. Let me respond using the replies on github website.

.github/workflows/lint.yml
```diff
@@ -21,9 +21,10 @@ jobs:
...
+        ruff --statistics .
```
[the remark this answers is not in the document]
mattijn: as in lint.yml, deleted this line. ruff check . is sufficient.

altair/utils/schemapi.py
mattijn: I've modified your suggestion a little bit to `raise from None` instead of `raise # noqa: B209`. I'm not entirely sure if that leads to the same behavior. Do you have an example how you checked this?

tools/schemapi/schemapi.py
mattijn: Same change as above, `raise from None` instead of `rase # noqa: B904`

mattijn:
> Maybe we can even use it as a replacement for black at one point: https://github.com/charliermarsh/ruff/issues/1904

Hm, didn't realize this, but I think we can remove `black` for format suggestions already? The link you refer to is on ruff autoformat, but I don't think we have black autoformat enabled?

In that case `black --check .` should raise the same as `ruff check .`., agree? Or I miss something here?

binste:
I didn't find any options for ruff to enable checks on code formatting. Also tested it on some code where black correctly shows an error but ruff does not. The [FAQ](https://beta.ruff.rs/docs/faq/) state that *"As a project, Ruff is designed to be used alongside Black and, as such, will **defer implementing stylistic lint rules** that are obviated by autoformatting."* So I'd assume that the checking comes hand-in-hand with the autoformat capability and we need to wait for that.

altair/utils/schemapi.py
binste: Double-checked and that works as well. It gives the same traceback as in master which is good. I tested it with:

```python
import altair as alt

(
    alt.Chart(data.barley())
    .mark_bar()
    .encode(
        x=alt.X("variety", unknown=2),
        y=alt.Y("sum(yield)", stack="asdf"),
    )
)
```

charliermarsh:
Sorry for the drive-by comment but I just saw this is in Ruff's dependency graph and as a long-time Altair user I can't help but mention how cool it is to be able to support the project. Let me know if you ever run into issues with Ruff or have questions on usage etc.

joelostblom:
That's great, thank you for taking the time to comment and offering your support @charliermarsh !

Issue #2914, opened 2023-02-21
Show python types instead of javascript types in error messages


joelostblom:
Originally mentioned in https://github.com/altair-viz/altair/pull/2842#issuecomment-1400813951.

Currently data types in the error messages are javascript rather than python. For example an error message could look like this:

```py
alt.Chart(data.barley()).mark_bar().encode(
    x=alt.X('variety'),
    y=alt.Y('sum(yield)', stack='null'),  # should be eg. stack=None
)
```
```
'null' is not one of ['zero', 'center', 'normalize']
'null' is not of type 'null'
'null' is not of type 'boolean'
```

A would probably be more useful if the ever measured message look like this:

```
'null' is not one of ['zero', 'center', 'normalize']
'null' is not of type NoneType
'null' is not of type Bool
```

Or even simplifying the `NoneType` further:

```
'null' is not one of ['zero', 'center', 'normalize']
'null' is not of type Bool
'null' is not None
```

There is a [dictionary in the code]( and https://github.com/altair-viz/altair/pull/2842#issuecomment-1400938604.) for conversion between javascript and python types that could be useful here, but we might need to create something new as well

dangotbanned:
@joelostblom @binste 
Fully agree we should do this

Issue #2913, opened 2023-02-21
`alt.datum` and `alt.value` generates incorrect error messages


joelostblom:
Originally reported in https://github.com/altair-viz/altair/pull/2883#issuecomment-1427383912. When using the datum encoding then the message is incorrect:

```py
alt.Chart().mark_point().encode(x=alt.datum(1, wrong_name=1))
```

```
SchemaValidationError: Invalid specification

        altair.vegalite.v5.schema.channels.X, validating 'additionalProperties'

        Additional properties are not allowed ('datum', 'wrong_name' were unexpected)
```

`'datum'` should not be included above as an incorrect property to the X channel, instead `wrong_name` should be mentioned as an incorrect parameter to `datum`. The way to handle this is probably with special error checking inside the datum function (although maybe it needs to also consider which encoding channel it is used for?).

---

Also mentioned in https://github.com/altair-viz/altair/pull/2568#issuecomment-1439177226 for `alt.value`. This can look more confusing now that the error message gives parameter hints:

I found one spec that gives an incorrect hint:

```python
import altair as alt
from vega_datasets import data
cars = data.cars.url

alt.Chart(cars).mark_point().encode(
    x='Acceleration:Q',
    y='Horsepower:Q',
    color=alt.value(1)  # should be eg. alt.value('red')
)
```
```txt
SchemaValidationError: `Color` has no parameter named 'value'

Existing parameter names are:
shorthand      bin         legend   timeUnit   
aggregate      condition   scale    title      
bandPosition   field       sort     type       

See the help for `Color` to read the full description of these parameters
```
It is allowed to use `alt.value()` in this context, but in this case it should be a string and not a number.

joelostblom:
@binste This might be an issue to keep in mind when working on https://github.com/altair-viz/altair/issues/2915 since I believe it is also affected of `additionalProperties`

dangotbanned:
https://github.com/vega/altair/issues/2913#issuecomment-1729893515

# Minimal Repro

Both have identical behavior and meaning

```py
import altair as alt

>>> alt.Chart().mark_point().encode(x=alt.datum(1, bin=True))
>>> alt.Chart().mark_point().encode(x=alt.XDatum(1, bin=True))
```

You can get a different error by trying method chaining like

```py
import altair as alt

>>> alt.XDatum(1).bin(True)
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[6], line 1
[...]

PR #3009, opened 2023-04-02
Improve error prioritisation and messages

### LINKED ISSUE #2873: Combine multiple error messages

### LINKED ISSUE #2915: Some errors in layered or faceted specs raise the wrong error message
A spec that is not layered will return to correct error in this case:

```py
alt.Chart().mark_point().encode(tooltip=[{'wrong'}])
```

```
SchemaValidationError: '[{'field': {'wrong'}}]' is an invalid value for `tooltip`:

[{'field': {'wrong'}}] is not of type 'object'
[{'field': {'wrong'}}] is not of type 'object'
{'wrong'} is not valid under any of the given schemas
[{'field': {'wrong'}}] is not of type 'null'
```

However when the same error occurs within a layered spec, then the return error points in the wrong direction:

```py
alt.layer(
    alt.Chart().mark_point().encode(tooltip=[{'wrong'}])
)
```

```
SchemaValidationError: `LayerChart` has no parameter named 'mark'

Existing parameter names are:
layer        data          height    projection   usermeta   
autosize     datasets      name      resolve      view       
background   description   padding   title        width      
config       encoding      params    transform               

See the help for `LayerChart` to read the full description of these parameters
```

Same with a facet (but concatenation works fine)the:

```py
alt.Chart(pd.DataFrame({'a': [1]})).mark_point().encode(tooltip=[{'wrong'}]).facet()
```

```
SchemaValidationError: `Spec` has no parameter named 'mark'

Existing parameter names are:
args   

See the help for `Spec` to read the full description of these parameters
```

I don't have time to investigate right now but this might be due to the specific error handling of parameters rather than the new we have retreating errors without deep validation. In other words, we might need a more specific check for when to include the suggested parameters in #2565.

binste:
# Background
Error messages in Altair are mostly generated based on the validation of the chart specification against the Vega-Lite schema. This validation is performed by the `jsonschema` package. It's not straight forward to generate helpful error messages out of these specs and Altair 5 rc1 already contains various improvements over Altair 4 in this regard. One improvement was the replacement of the previous 'deep validation' approach with the error hierarchy returned by `jsonschema`, see #2842. The main benefits over 'deep validation'  were that it solved some error messages pointing to parts of the specifications which are actually correct, and the second one was that we can use the error hierarchy to craft more helpful error messages as we now have structured information on all the errors in the chart specification. This last benefit was already somewhat used in #2957.

This PR aims to further use the error hierarchy to improve error messages. It resolves #2915, resolves #2873 and partially addresses #2913.

# Suggested approach
See #2842 for a description of how the validation works in general. This PR keeps the separation between `validate_jsonschema` and `SchemaValidationError` the same so that:

* `validate_jsonschema`: Validates the chart spec against a JSON schema and prioritizes and prepares the relevant error messages
* `SchemaValidationError`: Creates a more user friendly error message out of the errors from `validate_jsonschema`

I tried to document the changes in the code and added more examples to `test_schemapi` so you can see how the errors look. However, let me know if I should write up a more detailed walkthrough. Below, I just show some examples:

```python
import altair as alt
```

## Aggregate multiple error messages which are about the same error in the chart specification (#2873)
```python
alt.Chart(data.barley()).mark_bar().encode(
    x=alt.X("variety"),
    y=alt.Y("sum(yield)", stack="asdf"),
)
```
```
'asdf' is an invalid value for `stack`. Valid values are:

- one of ['zero', 'center', 'normalize']
- of type 'null' or 'boolean'
```

If it would be just one `one of` line or only types then it is shown on the same line and not as a list item:
```python
alt.Chart(data.cars()).mark_text(align="right").encode(
    alt.Text("Horsepower:N", bandPosition="4")
)
```
```
'4' is an invalid value for `bandPosition`. Valid values are of type 'number'.
```

## Prefer errors from "deeper levels" of the schema (#2915)
Fixes both examples from #2915 for layered and faceted charts. See that issue for more details. This part happens in `schemapi.py:_subset_to_most_specific_json_paths`. For the `layer` example in #2915 it keeps the error for the tooltip as the other error has a json path `'$.layer[0]'` and the tooltip error is at `'$.layer[0].encoding.tooltip'`.

## Show all errors in the chart specification
In case a chart spec has multiple errors, so far we only showed one and as soon as that one was resolved the other showed up. As a user, I'd prefer to see all at once, else I sometimes think the change I did to resolve one error caused another error although the two are unrelated:

```python
alt.layer(
    alt.Chart().mark_point().encode(tooltip=[{"wrong"}]),
    alt.Chart().mark_line().encode(invalidChannel="unknown"),
)
```
```
SchemaValidationError: Error #1: '{'wrong'}' is an invalid value for `field`. Valid values are of type 'string' or 'object'.

Error #2: `Encoding` has no parameter named 'invalidChannel'

Existing parameter names are:
angle         key          order     strokeDash      tooltip   xOffset   
color         latitude     radius    strokeOpacity   url       y         
description   latitude2    radius2   strokeWidth     x         y2        
detail        longitude    shape     text            x2        yError    
fill          longitude2   size      theta           xError    yError2   
fillOpacity   opacity      stroke    theta2          xError2   yOffset   
href                                                                     

See the help for `Encoding` to read the full description of these parameters
```

# Open questions/todos

- [x] Messages for multiple errors are currently only separated by new lines and by the "Error #..." prefix. Any ideas on how we can improve the readability? I already tried separating the messages with a line ("------"...) but it's unclear what a good length for that line would be as it depends on the frontend where the message is displayed. If it's too long, it's wrapped onto a new line. An alternative might be to indent everything after the first line of a message which is similar to tracebacks and don't start on the first line, for example:
```
SchemaValidationError:
Error #1: `X` has no parameter named 'unknown'

    Existing parameter names are:
    shorthand      bin      scale   timeUnit   
    aggregate      field    sort    title      
    axis           impute   stack   type       
    bandPosition                               

    See the help for `X` to read the full description of these parameters

Error #2: 'asdf' is an invalid value for `stack`. Valid values are:

    - one of ['zero', 'center', 'normalize']
    - of type 'null' or 'boolean'
```
- [x] I'll do another code review myself but wanted to already put this up for review to get inputs, especially as we might still get it into version 5 which I think would be great!

What I think could be tackled in separate PRs:

* Replacing JavaScript types with Python types, e.g. `'null'` with `None`, can be tackled in a separate PR, see #2914.
* Datum example from #2913 is still not optimal as it now says `X has no parameter named 'wrong_argument'` although it should say something about `'wrong_argument' is not a valid parameter for 'datum' in this context`. The previous error messages was more confusing so I think this is an improvement and would suggest to keep #2913 open


binste:
PR is ready for review :) I'll do another review myself this weekend but wanted to already put this up to get your inputs in case you find the time, especially as we might still get it into version 5 which I think would be great!

binste:
Tests will fail until https://github.com/vega/schema/issues/8 is resolved.

joelostblom:
Wohoo! Exciting to see this ready for review :tada:  Thanks for all your hard work on the PR, it looks really awesome at first glance =)

> I'll do another review myself this weekend but wanted to already put this up to get your inputs in case you find the time, especially as we might still get it into version 5 which I think would be great!

I agree, and I think even getting this into 5 RC2 would be a good goal. It seems like we might want to wait for https://github.com/vega/schema/issues/8 before releasing RC 2 anyways?

I will try to review as soon as possible, but a deeper look will probably have to wait until end/middle of next week for me unfortunately. But I can say already that I like all the approaches taken here, and it is interesting that we know show more than just one Error. I agree with you that the formatting is important for this not to be confusing. To me the indented example you showed above is notably clearer than the other one. Maybe even tweaked to explicitly mention on the first line that there are multiple errors?

```
SchemaValidationError: Multiple errors were found.

Error 1: `X` has no parameter named 'unknown'

    Existing parameter names are:
    shorthand      bin      scale   timeUnit   
    aggregate      field    sort    title      
    axis           impute   stack   type       
    bandPosition                               

    See the help for `X` to read the full description of these parameters

Error 2: 'asdf' is an invalid value for `stack`. Valid values are:

    - one of ['zero', 'center', 'normalize']
    - of type 'null' or 'boolean'
```

Without indendation, maybe three `---` could work? Standardizing on three could be justified as this is what is enough in HTML for a horizontal rule and it adds some additional whitespace separation too:

```
SchemaValidationError: Multiple errors were found.

Error 1: `X` has no parameter named 'unknown'

Existing parameter names are:
shorthand      bin      scale   timeUnit   
aggregate      field    sort    title      
axis           impute   stack   type       
bandPosition                               

See the help for `X` to read the full description of these parameters

---

Error 2: 'asdf' is an invalid value for `stack`. Valid values are:

- one of ['zero', 'center', 'normalize']
- of type 'null' or 'boolean'
```

I am OK with either of those two approaches; I think the indentation is easier to quickly scan for the two errors, so I favor that a bit (unless there are any width issues width only indenting for multiple errors, but I can't think of why that would be the case). I also slightly favor having the `#` number prefixed removed, but that is very minor and I am really fine either way so you pick.

binste:
Happy to hear! :) And thank you for the first feedback. I fully agree with all your points and implemented them. I went for the indentation as I also think it's more readable then some solution with dividers. At most 3 errors are now shown.

binste:
Tests are failing because altair_viewer no longer supports VL 5.6 so we'll first need to merge #3022 once it's ready.

mattijn:
(did a quick check if this PR pass the bump to VL5.7.1)

altair/utils/schemapi.py
```diff
@@ -69,83 +126,181 @@ def validate_jsonschema(spec, schema, rootschema=None):
...
+
+
+def _subset_to_most_specific_json_paths(
+    errors_by_json_path: GroupedValidationErrors,
+) -> GroupedValidationErrors:
+    """Removes key (json path), value (errors) pairs where the json path is fully
+    contained in another json path. For example if `errors_by_json_path` has two
+    keys, `$.encoding.X` and `$.encoding.X.tooltip`, then the first one will be removed
+    and only the second one is returned. This is done under the assumption that
+    more specific json paths give more helpful error messages to the user.
```
joelostblom, line 173: This makes sense to me :+1:

altair/utils/schemapi.py
```diff
@@ -69,83 +126,181 @@ def validate_jsonschema(spec, schema, rootschema=None):
...
+                errors = deduplication_func(errors)
+            deduplicated_errors.extend(_deduplicate_by_message(errors))
+
+        # Removes any ValidationError "'value' is a required property" as these
+        # errors are unlikely to be the relevant ones for the user. They come from
+        # validation against a schema definition where the output of `alt.value`
+        # would be valid. However, if a user uses `alt.value`, the `value` keyword
+        # is included automatically from that function and so it's unlikely
+        # that this was what the user intended if the keyword is not present
+        # in the first place.
```
joelostblom, line 218: These comments are very helpful, thanks for including them!

altair/utils/schemapi.py
```diff
@@ -189,15 +344,108 @@ def _resolve_references(schema, root=None):
...
+        # Error messages for these errors look like this:
+        # "Additional properties are not allowed ('unknown' was unexpected)"
+        # Line below extracts "unknown" from this string
+        parameter_name = error.message.split("('")[-1].split("'")[0]
+        message = f"""\
+`{altair_cls.__name__}` has no parameter named '{parameter_name}'
+
+Existing parameter names are:
+{param_names_table}
+See the help for `{altair_cls.__name__}` to read the full description of these parameters"""
```
joelostblom, line 419: I think it is a bit nicer for the the indentation to use `inspect.cleandoc` here, is there an advantage of out-indenting like this instead?

altair/utils/schemapi.py
```diff
@@ -69,83 +126,181 @@ def validate_jsonschema(spec, schema, rootschema=None):
...
+        if not _contained_at_start_of_one_of_other_values(
+            json_path, list(errors_by_json_path.keys())
+        ):
+            errors_by_json_path_specific[json_path] = errors
+    return errors_by_json_path_specific
+
+
+def _contained_at_start_of_one_of_other_values(x: str, values: Sequence[str]) -> bool:
+    # Does not count as "contained at start of other value" if the values are
+    # the same. These cases should be handled separately
```
joelostblom, line 186: So if the values are exactly the same are they deduplicated elsewhere instead? Maybe in the function just below?

altair/utils/schemapi.py
```diff
@@ -247,82 +495,57 @@ def split_into_equal_parts(n, p):
...
+    ) -> str:
+        bullet_points: List[str] = []
+        errors_by_validator = _group_errors_by_validator(errors)
+        if "enum" in errors_by_validator:
+            for error in errors_by_validator["enum"]:
+                bullet_points.append(f"one of {error.validator_value}")
+
+        if "type" in errors_by_validator:
+            types = [f"'{err.validator_value}'" for err in errors_by_validator["type"]]
+            point = "of type "
```
joelostblom, line 512:
```suggestion
                bullet_points.append(f"One of {error.validator_value}")

        if "type" in errors_by_validator:
            types = [f"'{err.validator_value}'" for err in errors_by_validator["type"]]
            point = "Of type "
```

```suggestion
                bullet_points.append(f"One of {error.validator_value}")
```

What do you think of capitalizing these? So instead of the current version:

```
SchemaValidationError: 'asdf' is an invalid value for `stack`. Valid values are:

- one of ['zero', 'center', 'normalize']
- of type 'null' or 'boolean'
```

it would be:

```
SchemaValidationError: 'asdf' is an invalid value for `stack`. Valid values are:

- One of ['zero', 'center', 'normalize']
- Of type 'null' or 'boolean'
```

I don't feel strongly here, just a suggestion. I'm thinking this could help the bullet point sentences appear more separate from each other, but maybe it is just personal taste

joelostblom (approved):
Looks amazing and will be a huge help for all Altair users, thank you so much for tackling this @binste ! And sorry for taking some time to get to it. I have a few minor comments, but I also think this is good to go as is, so feel free to update/resolve as you see fir and we can merge it after. Although all the changes here seems to make sense to me and the tests are encompassing, I think it is important that we start running this ourselves and get it into the next RC, so that we get the chance to stumble upon error cases that are difficult to anticipate ahead of time.

altair/utils/schemapi.py
binste: Exactly, they are deduplicated by message on line 210

altair/utils/schemapi.py
binste: I agree that the code looks nicer but for me it was more difficult to make changes and refactor if it's indented as the indentation level had to match in all methods/functions which contribute parts of the error message such as `_format_params_as_table`. I therefore opted for the simpler maintenance and understandability over code style but happy to readjust, I don't feel strongly about it :)

altair/utils/schemapi.py
binste: I like it! Included in the latest commits

altair/utils/schemapi.py
joelostblom: Sounds like a good reason to me, let's keep what you have :+1:

joelostblom:
Great, thank you again! Merging this!

PR #3010, opened 2023-04-02
remove deprecated datasets.py

mattijn:
This PR removes the deprecated datasets.py. 

These utilities are deprecated since 2018 https://github.com/altair-viz/altair/commit/4e6eb6e5ef8270f924816ae59e7edd2bcfccaea3. I think it is save to delete this by now.


joelostblom (approved):
Agree, and 5.0 is a good release to remove something like this. Could you add a short note to the changelog and then merge?

mattijn:
Thanks @joelostblom for checking! Done and merging.

PR #3021, opened 2023-04-15
wip: doc for color scales

mattijn:
fix: https://github.com/altair-viz/altair/issues/2861

### DIFF
--- a/doc/user_guide/color_scales.rst
+++ b/doc/user_guide/color_scales.rst
@@ -1,0 +1,187 @@
+.. currentmodule:: altair
+
+.. _user-guide-color-scales:
+
+Color Scales
+============
+
+Effective visualization of scientific data requires careful attention
+to color selection. Choosing the right colors can be a complex task,
+involving considerations such as color perception, accessibility,
+and aesthetics. Fortunately, there are a number of resources available
+to help simplify this process, including pre-designed color schemes
+and tools for exploring and modifying color scales. In this guide,
+we'll explore some of the basics of color selection for scientific 
+data visualization, including different types of color scales and
+examples of their use.
+
+I. The Basics of Color Selection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Color selection is an essential aspect of creating clear and informative
+visualizations of scientific data. To make the most effective choices,
+it's important to understand the basics of color perception and theory.
+
+A. Color Perception and Theory
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+1. The importance of perceptually uniform color scales
+------------------------------------------------------
+
+Using colors that are perceptually uniform can help ensure that viewers
+can accurately perceive the relative differences between data points. By
+maintaining consistent differences in color perception across a scale, you
+can create visualizations that are not only informative but also easy to
+understand.
+
+.. code-block:: none
+
+    A comparison of perceptually uniform color scales and non-uniform
+    scales, showing how the choice of color space can affect the perceived
+    brightness and contrast of colors.
+
+2. Different types of color spaces
+----------------------------------
+
+Different color spaces, such as RGB or HSL, offer different advantages and
+can affect the perceived brightness and contrast of colors. By 
+understanding the nuances of color spaces, you can create visualizations
+that are not only informative but also visually appealing.
+
+B. Accessibility and Aesthetics
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+1. Considerations for color-blind viewers
+-----------------------------------------
+   
+Around 8% of men and 0.5% of women have some form of color vision deficiency.
+To make sure your visualizations are accessible to all viewers, it's
+important to choose colors that are distinguishable even for those with 
+color vision deficiencies. By using colors that have a high degree of
+contrast, you can create visualizations that are not only informative but
+also inclusive.
+
+.. code-block:: none
+
+    A comparison of color schemes that are accessible to color-blind
+    viewers versus those that are not.
+
+2. Guidelines for creating visually appealing presentations
+-----------------------------------------------------------
+
+While accuracy is the primary goal of visualizations, aesthetics also play
+an important role in engaging viewers and making data more memorable.
+By using a limited color palette, choosing harmonious colors, and balancing
+colors to create a sense of visual hierarchy, you can create visualizations
+that are not only informative but also visually stunning.
+
+.. code-block:: none
+    
+    A visualization that uses color to create a sense of visual hierarchy,
+    such as a heatmap or a scatterplot.
+
+
+II. Types of Color Scales
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This sections presents the three primary color scales used in data
+visualization: sequential, diverging, and categorical. The section
+includes examples of data that can be visualized using each scale and
+popular color schemes used in practice.
+
+A. Sequential Scales
+^^^^^^^^^^^^^^^^^^^^
+
+1. Examples of sequential data
+------------------------------
+
+Sequential scales are best suited for data that has an inherent ordering,
+such as data that varies from low to high, or over time. Examples of
+sequential data include temperature or population density, where the
+range of values is continuous and unbroken.
+
+1. Example color schemes for sequential data
+--------------------------------------------
+
+.. code-block:: none
+    
+    Sequential color schemes.
+
+B. Diverging Scales
+^^^^^^^^^^^^^^^^^^^
+
+1. Examples of diverging data
+-----------------------------
+
+Diverging scales are best suited for data that varies above and below a
+center point, such as positive and negative deviations from a mean.
+Examples of diverging data include measures of deviation, such as
+temperature anomalies or changes in sea level.
+
+1. Example color schemes for diverging data
+-------------------------------------------
+
+.. code-block:: none
+    
+    Diverging color schemes.
+
+
+C. Categorical Scales
+^^^^^^^^^^^^^^^^^^^^^
+
+1. Examples of categorical data
+-------------------------------
+
+Categorical scales are best suited for data that is unordered, such as
+different species or categories. Examples of categorical data include
+the types of flowers in a garden, or different political affiliations.
+
+1. Example color schemes for categorical data
+---------------------------------------------
+
+.. code-block:: none
+    
+    Categorical color schemes.
+
+III. Using Tools for Color Selection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A. Pre-Designed Color Palettes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+1. Example color palettes from the other packages
+-------------------------------------------------
+
+This subsection focuses on pre-designed color palettes that can be
+used in data visualization from other sources.
+
+B. Modifying Color Scales
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+1. Tools for adjusting brightness, saturation, and hue
+------------------------------------------------------
+
+This subsection provides an overview of the different ways in which
+brightness, saturation, and hue can be modified to create custom color
+scales. It explains how changing these attributes can help to emphasize
+or de-emphasize certain aspects of the data being visualized.
+
+2. Customizing color scales to match specific data requirements
+---------------------------------------------------------------
+
+This subsection explores the idea of creating custom color scales that
+are tailored to the specific needs of the data being visualized. It
+discusses how to use the tools introduced in III.B.1 to create color
+scales that are better suited to representing the nuances of the data,
+and provides examples of how this can be achieved in practice.
+
+Conclusion
+~~~~~~~~~~
+
+By utilizing the tools and resources available for color selection,
+data visualizers can create effective and visually appealing
+presentations that accurately convey scientific data. Whether using
+pre-designed color schemes or customizing color scales, it is
+essential to consider factors such as accessibility, aesthetics,
+and perceptual relationships in order to create the most effective
+visualizations possible.
--- a/doc/user_guide/data.rst
+++ b/doc/user_guide/data.rst
@@ -620,6 +620,7 @@ data before usage in Altair using GeoPandas for example as such:
    encodings/index
    marks/index
    transform/index
+   color_scales
    interactions
    compound_charts
    scale_resolve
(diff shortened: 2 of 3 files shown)


mattijn:
I'm struggling a bit with this one. @joelostblom, can you have a look if this is going in the right direction? Its maybe a bit too lengthy. I also doubt on the order of the sections. Currently the page starts with some color-theory including examples of good vs bad practices to continue with the predefined color schemes in Altair. I can imagine that one does not want to scroll through the theory each time to reach the available color schemes in Altair..

joelostblom:
Sorry for the delay in reviewing this @mattijn I had a go now but something is up with my environment, so I will try again tomorrow or during next week.

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+            chart_concat = chart_concat.configure_concat(spacing=1).configure_view(stroke=None)    
+        return chart_concat
+
+
+The basics of color selection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Color selection is an essential aspect of creating clear and informative
+visualizations of scientific data. To make the most effective choices,
+it's important to understand the basics of color perception and theory.
```
joelostblom, line 227:
```suggestion
Color selection is an essential aspect of creating clear and informative
data visualizations. To make the most effective color choices,
it's important to understand the basics of color perception and theory.
In this section we start by explaining colors as perceived with standard color vision
and proceed to discuss how common color vision deficiencies alter color perception.
```

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+
+We will explain the basics using two colors, ``blue`` and ``orange``: 
+ 
+.. altair-plot::
+    :remove-code:
+
+    percep_schemes = {
+        "two_color": ['blue','orange']
+    }    
+    plot_scheme("two_color", percep_schemes, cvd=False, continuous=False, method='hcl').configure_scale(rectBandPaddingInner=0.1)
```
joelostblom, line 240: In all the plots in this section, could we remove the "normal vision" label and only include it in the section that talks about different types of color vision? I think the blue and orange rectangles could look better if they were  made a bit smaller as well.

In the later plots, we might also want to consider calling it something like "Standard/reference/perfect color vision" instead of "normal". And likewise change "color blindness" to "color vision deficiencies" which is more descriptive of the partial loss in color vision that most people with this condition have (rather than a complete color blindness which is quite rare)

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+
+The basics of color selection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Color selection is an essential aspect of creating clear and informative
+visualizations of scientific data. To make the most effective choices,
+it's important to understand the basics of color perception and theory.
+
+Color perception and theory
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
joelostblom, line 230: Instead of starting with talking about how to define a path between two colors, what do you think about having a more data-centric explanation that explains the different scenarios for how color could be used with data? These would be varying the hue/color to represent categorical values and varying mainly luminance to represent change in a quantitative value (with two sub categories: sequential and diverging colormaps, both which should be perceptually uniform). Personally, I'm a fan of how the seaborn docs introduce this topic as it becomes really tangible how the concepts of color applies to data right across the bath. I think talking about categorical and numerical data also maps nicely to that these concepts are already used in channel encodings and maybe our categories would be `nominal` data on the one hand, and then `ordinal` and `quantitative` together (as discrete and continuous sequential (or diverging) color scales). What do you think?

The section you have is really interesting and I love that the color space plot is made directly in Altair, so I want to keep it as part of the "Sequential" subsection where we talk more about how these color palettes are defined.

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+    :remove-code:
+
+    percep_schemes = {
+        "two_color": ['blue','orange']
+    }    
+    plot_scheme("two_color", percep_schemes, cvd=False, continuous=False, method='hcl').configure_scale(rectBandPaddingInner=0.1)
+
+Now, lets say we want to define a gradient between ``blue`` and ``orange``.
+That means we have to go gradually from ``blue`` to ``orange``. 
+The interpolation between colors can follow several methods.
```
joelostblom, line 244:
```suggestion
Now, lets say we want to define a gradient between ``blue`` and ``orange``.
That means we have to gradually transition from ``blue`` to ``orange``,
which is also called an "interpolation" between the two colors.
```

Then after this I would have the paragraph that starts with "So what is the correct interpolation...", then show the color space plot, then show the ppaths and the corresponding sequential color schemes created by each path.

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+
+Color-blind viewers
+^^^^^^^^^^^^^^^^^^^
+   
+Around 8% of men and 0.5% of women have some form of color vision deficiency.
+To make sure your visualizations are accessible to all viewers, it's
+important to choose colors that are distinguishable even for those with 
+color vision deficiencies. By using colors that have a high degree of
+contrast, you can create visualizations that are not only informative but
+also inclusive.
```
joelostblom, line 355:
```suggestion
also inclusive.
The default color schemes that are used in Altair are designed
to be effective both for people with and without color vision deficiencies,
and the same goes for many of the available color scheme options.
However,
there are still some available color schemes that are not well-designed for people 
with color vision deficiency, as we can see in the example below.
```

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+       The red and green colors are used to represent different ends of a spectrum, 
+       but these two colors are not easy differentiated with some type of CVD.
+
+       .. altair-plot::
+         :remove-code:
+
+         plot_scheme("redyellowgreen", color_vision_deficiency, cvd=True, continuous=True)
+
+Guidelines for using colors
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
joelostblom, line 390: Love this section! One rule I would add is to start thinking about breaking our chart up into many smaller subcharts (e.g. faceting) when we get to around 5-8 different categorical colors:

image

doc/user_guide/color_usage.rst
joelostblom: I start my class on this with these three slides:

image

image

image

joelostblom: I also really like these from the mpl docs on what perceptually uniform looks like in terms of the lightness increase:

image

image

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+
+
+Color schemes
+~~~~~~~~~~~~~
+
+This sections presents the three primary color schemes used in data
+visualization: sequential, diverging, and categorical. 
+
+Sequential schemes
+^^^^^^^^^^^^^^^^^^
```
joelostblom, line 783: I think it is great that we are showing how each palette look for common color vision deficiencies here! I think it would also be great to include which ones are perceptually uniform, but maybe that is not possible? Or a small line chart / spark line maybe could work for the luminiosty? I also think that is nice that we show an example of the discrete version of the scheme, but maybe we could more clearly label that this is what it would look like for ordinal data (which I think is the main scenario where they would be used?)?

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+
+Sequential schemes are best suited for data that has an inherent ordering,
+such as data that varies from low to high, or over time. Examples of
+sequential data include temperature or population density, where the
+range of values is continuous and unbroken.
+
+.. altair-plot::
+    :hide-code:
```
joelostblom, line 791:
```suggestion
    :remove-code:    
```

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+Diverging schemes
+^^^^^^^^^^^^^^^^^
+
+Diverging schemes are best suited for data that varies above and below a
+center point, such as positive and negative deviations from a mean.
+Examples of diverging data include measures of deviation, such as
+temperature anomalies or changes in sea level.
+
+.. altair-plot::
+    :hide-code:
```
joelostblom, line 942:
```suggestion
    :remove-code:    
```

doc/user_guide/color_usage.rst
joelostblom: It also seems like the sequential multi-hue schemes (e.g. viridis) are missing?

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+         ).configure_concat(spacing=1).configure_view(stroke=None).resolve_scale(color='independent').configure_scale(rectBandPaddingInner=0.1)
+
+Cyclical schemes
+^^^^^^^^^^^^^^^^
+
+Cyclical color schemes may be used to highlight periodic patterns in continuous data.
+However, these schemes are not well suited to accurately convey value differences.
+
+.. altair-plot::
+    :hide-code:
```
joelostblom, line 1091:
```suggestion
    :remove-code:    
```

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+
+Categorical schemes
+^^^^^^^^^^^^^^^^^^^
+
+Categorical schemes are best suited for data that is unordered, such as
+different species or categories. Examples of categorical data include
+the types of flowers in a garden, or different political affiliations.
+
+.. altair-plot::
+    :hide-code:
```
joelostblom, line 1137:
```suggestion
    :remove-code:    
```

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+- distinct for all people, including colour-blind readers
+- distinct from black and white
+- distinct on screen and paper
+- matching well together
+
+.. altair-plot::   
+    :output: none
+
+    tol_schemes = {
+        "tol_bright" : ['#4477AA', '#EE6677', '#228833', '#CCBB44', '#66CCEE', '#AA3377', '#BBBBBB'],
```
joelostblom, line 1306: This might need some more explanation on how to use these custom palettes. Maybe the section could start with an intro of how to make your own palette (like what is in the color customization section right now) and then show some from this other source. Also, I am not familiar with Paul Tol, but maybe we can think about the presentation of his palettes so that it doesn't seem like the properties listed above are not true for the built-in altair color schemes, because in many cases they are.

doc/user_guide/color_usage.rst
```diff
@@ -0,0 +1,1525 @@
...
+
+    bars + text
+
+The lighter the bar, the higher the luminance. If the bar is light
+we like a text overlay that is black. The darker the bar, the lower
+the luminance. If the bar is dark, we like a text overlay that is white.
+
+In the expression above we have written this as a predicate. The text
+appear ``black`` if the luminance is above ``0.5`` and ``white`` when
+the luminance is below ``0.5``. The luminance is computed using the
```
joelostblom, line 1513:
```suggestion
        color=alt.expr("luminance(scale('color', datum.sum_yield)) > 0.408 ? 'black' : 'white'")
    )

    bars + text

The lighter the bar, the higher the luminance. If the bar is light
we like a text overlay that is black. The darker the bar, the lower
the luminance. If the bar is dark, we like a text overlay that is white.

In the expression above we have written this as a predicate. The text
appear ``black`` if the luminance is above ``0.408`` and ``white`` when
the luminance is below ``0.408``. The luminance is computed using the
```

This section is great! I love that expression can be used for this! The reason for the change I suggest here is to stay consistent with what is used in seaborn, which is also what I chose when adding this to the text of pandas heatmaps. We could add a note about this? (I was trying to look up the reason for picking this exact number in the first place, but based on [my comment here ](https://github.com/pandas-dev/pandas/pull/21263/files#r193573758)I don't think I ever knew =) )

joelostblom (changes requested):
This section is great! I would love to have more explanatory sections for good choices to make effective visualizations somehow linked to the Altair docs. I don't know exactly how this section will fit with the current color sections, but I think that is one of the things we will discuss on our call, so I just made some comments for now without knowing exactly where this will go in the end. Thanks for working on this!

dangotbanned:
@mattijn I haven't got anything to add on any of @joelostblom's comments.

I found all the examples with `altair` code really helpful!
The format of https://mattijn.github.io/altair-docs/user_guide/color_usage.html#guidelines-for-using-colors is a major 👍 👍 👍 

---

- https://github.com/Siddharth11/Colorful?tab=readme-ov-file#articles
- https://git.apcacontrast.com/documentation/thirdpartytools
- https://github.com/Myndex/SAPC-APCA?tab=readme-ov-file#design-guidelines

Some of these are more targeted at UX/App design, but might help bring a different perspective

mattijn:
For me, this kind of PR is quite energy demanding to do right. So, if others are excited to embrace this opportunity, please don't hesitate, chime in!

Last year I had access to a bit of funding from which I partly could do maintenance and doc-writing for Altair during working hours. In that period I also had asked my colleague to start writing a conceptual guide on best practices of effective data communication. That has reached concept level, see also this [.docx memo](https://github.com/user-attachments/files/16419336/Concept_DataVisualisation_MR.docx). Based on some discussions it was decided that it would be best if we can reduce this to something like 5 pages at maximum, otherwise it will become more book-like. It make sense to do some shifting of docs in this PR to the part of this, yet-to-be-written compact conceptual guide. This is now also defined on the [roadmap](https://github.com/orgs/vega/projects/9/views/3?pane=issue&itemId=59611113). My hope is still that the Vega project becomes a [NumFOCUS](https://numfocus.org/) project so it may be possible to get access to some funding to finish this _as work_. 

But like what I said, if others see this as an opportunity to contribute, please feel free!

dangotbanned:
Thanks for the context @mattijn 

> For me, this kind of PR is quite energy demanding to do right. So, if others are excited to embrace this opportunity, please don't hesitate, chime in!

I definitely can relate to this! But after skimming through [.docx memo](https://github.com/user-attachments/files/16419336/Concept_DataVisualisation_MR.docx) I feel quite out of my depth on the theory required for this task.

> My hope is still that the Vega project becomes a [NumFOCUS](https://numfocus.org/) project so it may be possible to get access to some funding to finish this _as work_.

Hopefully this will allow someone better equipt for this task to work on it.
I do have the time, for now, but I think I can utilise that more effectively in other areas of `altair`.

I see a lot of value in this kind of documentation and whenever it arrives - I'm sure it will be worth the wait

mattijn:
Include documenting accessibility, xref https://github.com/vega/altair/discussions/3314

Issue #3025, opened 2023-04-20
Documentation: Categories transposed in 'data' documentation


m-charlton:
The data documentation (`doc/user_guide/data.rst`) contains the following example:

```python
import altair as alt
import pandas as pd

data = pd.DataFrame({'x': ['A', 'B', 'C', 'D', 'E'],
                     'y': [5, 3, 6, 7, 2]})
alt.Chart(data).mark_bar().encode(
    x='x',
    y='y',
)
```

The description of the above code:

> For example, here we specify data via a pandas DataFrame and Altair automatically 
> detects that the x-column should be visualized on a quantitative scale and that the
> y-column should be visualized on a categorical (nominal) scale:

This strikes me as incorrect. The x-column is nominal. The y-column is
quantitative.

If we print the chart object then this would seem to back the above assertions.

```
"encoding": {
    "x": {
      "field": "x",
      "type": "nominal"
    },
    "y": {
      "field": "y",
      "type": "quantitative"
    }
  },
```

Fix in PR: #3026 Categories transposed in data documentation 

---

Please follow these steps to make it more efficient to solve your issue:

- [x] Since Altair is a Python wrapper around the Vega-Lite visualization grammar, [most bugs should be reported directly to Vega-Lite](https://github.com/vega/vega-lite/issues). You can click the Action Button of your Altair chart and "Open in Vega Editor" to create a reproducible Vega-Lite example and see if you get the same error in the Vega Editor.
- [x] Search for duplicate issues.
- [x] Use the latest version of Altair.
- [x] Describe how to reproduce the bug and include the full code and data to reproduce it, ideally using a sample data set from `vega_datasets`.

mattijn:
Thank you @m-charlton for raising the issue. You are right, that is incorrect. I've merged https://github.com/altair-viz/altair/pull/3026, so will close this issue!

PR #3026, opened 2023-04-20
Categories transposed in data documentation

m-charlton:
In an example column-x is specified as 'quantitative', when it should be 'nominal'. And column-y is 'nominal', when it should be 'quantitative'.


m-charlton:
Fix for #3025  "Documentation: Categories transposed in 'data' documentation"

mattijn:
Thank you @m-charlton for this PR. You are right, that is incorrect. All tests pass, merging. Thanks for the contribution!

PR #3029, opened 2023-04-21
Fix typo in docs causing code blocks to not render

joelostblom:
Addresses the documentation bug in https://github.com/altair-viz/altair/issues/3028


binste (approved):
Built the docs locally and can confirm that this fixes the rendering issue:

image

PR #3034, opened 2023-04-26
update vega expressions options

mattijn:
This PR updates the options for `alt.expr` according to https://vega.github.io/vega/docs/expressions/


binste (approved):
I parsed the vega page and compared it against the dictionary to check if any other entries need to be added or removed. Everything matches!

mattijn:
Yeah👍

PR #3039, opened 2023-05-02
Increase minimum required jsonschema (`>=4.0.01`) 

### LINKED ISSUE #3038: "'ValidationError' object has no attribute 'json_path'" on interactive legend example in RC
I have trouble passing the condition passing a point selection into the encode method. The [interactive legend example](https://altair-viz.github.io/gallery/interactive_legend.html) does not work on my end. I tried both RC1 and RC2, and in both cases, I get this:

`'ValidationError' object has no attribute 'json_path'`

… which seems to have been connected with issue #2715.

Is this a known issue? I do not get the error on version 4 with the old 

Thank you,
Marian

mattijn:
Fix #3038.

This PR increases the minimum required version of jsonschema to be `4.0.1` or higher.


mattijn:
Tests pass, merging!

PR #3042, opened 2023-05-06
Add info that altair_saver does not yet support Altair 5

binste:
Maybe it's good to highlight that `altair_saver` does not support Altair 5? Useful for new users who read the "Saving Charts" page in the docs but also for existing users which might rely on `altair_saver` and who read the changelog to see if anything could break when they upgrade to Altair 5.

Changelog (compared to screenshot I changed "consider switching" to "switch":

image

"Saving charts" doc:

image


mattijn:
Thanks @binste! Looks good, add a note how to install `vl-convert-python` using conda and removed the leading `$` so the prompts are easier to copy.

PR #3052, opened 2023-05-11
Be clearer about how vegafusion works

joelostblom:
@jonmmease What do you think of this update based on our discussion in https://github.com/hex-inc/vegafusion/issues/309?


jonmmease:
This looks great, thanks for making the update!

Issue #3051, opened 2023-05-11
Add altair sphinx extension back to package?


jtilly:
With version 5.0.0 (thank you for the release! 🥳) the altair sphinx extension was removed from the package. It used to be importable from `altair.sphinxext`. Would it be possible to bring that back?

There is a small number of projects on GitHub that use the extension in their own code base: https://github.com/search?q=%22altair.sphinxext%22&type=code including one of [ours](https://github.com/Quantco/glum/blob/4b03683a4c60e82733b721149ee7140d99d928bd/docs/conf.py#L26).

Thank you!

jedbrown:
We also use this, love altair, and would rather not copy `sphinxext` into our project. If this can't be re-added, what is the migration plan for projects that have been using `altair.sphinxext`?

binste:
Move of `sphinxext` to top-level happened in #2792. Honestly, I was simply not aware when reviewing it that other packages use this as well, although I can see now that it makes sense...

I think the main goal of that PR was to get the tests out of the altair module into a separate folder and `altairgallery` needs to import some functionality from the tests to fetch [the example charts](https://github.com/altair-viz/altair/blob/master/sphinxext/altairgallery.py#L26) and so it made sense to lift it out as well. Also, `sphinxext` relies on packages, such as sphinx and docutils, which are not part of the Altair requirements.

How about we only move the `altairplot` directive back into `altair.sphinxext` as it does not depend on the `tests` folder? We can add `ImportError`s with some context if `docutils` or `sphinx` are not installed. `altairgallery` only [shows up in Altair forks](https://github.com/search?q=%22altair.sphinxext.altairgallery%22&type=code) so it could be left in a separate folder. I could work on this.

jedbrown:
Thanks, that would be great. And if you really don't want it included with the rest of altair, it could be a separate package `altair-sphinxext` on PyPI.

jtilly:
> How about we only move the altairplot directive back into altair.sphinxext as it does not depend on the tests folder? 

Yes, that would be great!

binste:
I'll wait for @mattijn to chime in before working on this.

mattijn:
I was unaware that there were other packages depending on the `sphinxext` module when moving it out of Altair package. 

I've seen repositories struggling how to make these Altair plots appear in documentation. If it is a separate package I think it will make it easier to discover, install and use in other projects. 

[...]

PR #3056, opened 2023-05-13
Use altairplot Sphinx directive of sphinxext_altair package

binste:
See #3051 for context. Built docs and charts rendered fine.

- [x] Before merge, change requirement to package instead of git repo


binste:
Built the documentation locally and the charts rendered. All tests pass, merging.

PR #3057, opened 2023-05-13
Remove extra files in site-packages from wheel

jtilly:
Currently, when I install altair, I get a couple of extra files in my site-packages directory:

```
python -m env .env
pip install altair
cat .env/lib/python3.11/site-packages/README.md
cat .env/lib/python3.11/site-packages/LICENSE
cat .env/lib/python3.11/site-packages/pyproject.toml
```
These are all altair files. I think these files shouldn't be there 😉 

This PR removes them.

Note: I'm no hatch expert, so I don't fully understand what the benefit of adding these files in `tool.hatch.build` is.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,6 +100,9 @@ include = [
     "pyproject.toml"
 ]
 
+[tool.hatch.build.targets.wheel]
+only-include = ["/altair"]
+
 [tool.hatch.envs.default]
 features = ["dev"]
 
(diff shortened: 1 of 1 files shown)


binste:
Thanks for bringing it up! Indeed, that shouldn't be the case :) You now excluded it from wheels which definitely makes sense, but should we also exclude those files from source distributions, i.e. remove the paths from `INCLUDE`? The license is stored in the `.dist-info` package anyway, see [PEP 639](https://peps.python.org/pep-0639/#license-files-in-project-formats):

image

and README and pyproject.toml might not need to be distributed at all?

mattijn (approved):

mattijn:
Thanks @jtilly!

PR #3061, opened 2023-05-18
Add validation of Vega-Lite schema itself

### LINKED ISSUE #2801: Validate incoming jsonschemas from Vega-lite within tests
As was pointed out here https://github.com/altair-viz/altair/pull/2771#issuecomment-1369219098 it is recommended to validate the incoming static jsonschemas:

> > _The schema is included within Altair so it make sense to validate the vega-lite schema during development. Then we can catch these issues before it is shipped. Maybe we can set up a test for this within Github Action._
> 
> Yes! This is the thing I usually recommend -- if you have a static schema which you think is supposed to be valid, validate it in CI, then you'll know if something's wrong, but otherwise the library proceeds as normal.
> 

`check-jsonschema` could be used here, as is described in this comment: https://github.com/altair-viz/altair/pull/2771#issuecomment-1369245681

binste:
Closes #2801. I tested that the validation fails if I remove the `--disable-formats uri-reference` flag.

Thanks again @sirosen!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,3 +65,12 @@ jobs:
         run: |
           # pip install vl-convert-python
           pytest -m save_engine --doctest-modules tests
+      - name: Validate Vega-Lite schema
+        run: |
+          # We install all 'format' dependencies of jsonschema as check-jsonschema
+          # only does the 'format' checks which are installed.
+          # We can always use the latest jsonschema version here.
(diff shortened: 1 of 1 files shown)


sirosen:
Awesome, I'm glad I could help out!

Always feel free to ask for features or changes in the `check-jsonschema` repo if you run into issues or new use-cases.

PR #3066, opened 2023-05-23
TitleParam to Title in example gallery and sync scatterplot table

joelostblom:
Change TitleParam at a few instances and making sure the two scatter plot table examples are in sync.


binste (approved):
Looks good to me!

PR #3068, opened 2023-05-25
Fix bug in reconstructing layered charts with from_json/from_dict

### LINKED ISSUE #3067: JSON to Chart failure in v5.0.0
As part of a larger application, we're moving chart content around in JSON format via `Chart.to_json` and `Chart.from_json`. However, I'm encountering an error after updating to Altair v5.0.0. An example that used to work previously - based on the layered chart in https://altair-viz.github.io/altair-tutorial/notebooks/09-Geographic-plots.html

```python
import altair as alt

states = alt.topo_feature(data.us_10m.url, feature='states')
airports = data.airports()

background = alt.Chart(states).mark_geoshape(
    fill='lightgray',
    stroke='white'
).project('albersUsa').properties(
    width=500,
    height=300
)

points = alt.Chart(airports).mark_circle().encode(
    longitude='longitude:Q',
    latitude='latitude:Q',
    size=alt.value(10),
    tooltip='name'
)

chart = background + points

json = chart.to_json()
chart = alt.Chart.from_json(json)
```

In v4.2.2 this runs without any errors. In v5.0.0 this results in

```
Traceback (most recent call last):
  File "altair_scratchpad.py", line 25, in <module>
    chart = alt.Chart.from_json(json)
  File "C:\Users\romesh\miniconda3\envs\atomica38\lib\site-packages\altair\utils\schemapi.py", line 914, in from_json
    return cls.from_dict(dct, validate=validate)
  File "C:\Users\romesh\miniconda3\envs\atomica38\lib\site-packages\altair\vegalite\v5\api.py", line 2497, in from_dict
    return class_.from_dict(dct, validate=validate)  # type: ignore[attr-defined]
  File "C:\Users\romesh\miniconda3\envs\atomica38\lib\site-packages\altair\utils\schemapi.py", line 893, in from_dict
    return converter.from_dict(dct, cls)
  File "C:\Users\romesh\miniconda3\envs\atomica38\lib\site-packages\altair\utils\schemapi.py", line 1058, in from_dict
    return cls(**kwds)
  File "C:\Users\romesh\miniconda3\envs\atomica38\lib\site-packages\altair\vegalite\v5\api.py", line 2982, in __init__
    combined_dict, self.layer = _remove_layer_props(self, self.layer, layer_props)
  File "C:\Users\romesh\miniconda3\envs\atomica38\lib\site-packages\altair\vegalite\v5\api.py", line 3393, in _remove_layer_props
    if all(c[prop] is Undefined or c[prop] == chart[prop] for c in subcharts):
  File "C:\Users\romesh\miniconda3\envs\atomica38\lib\site-packages\altair\vegalite\v5\api.py", line 3393, in <genexpr>
    if all(c[prop] is Undefined or c[prop] == chart[prop] for c in subcharts):
  File "C:\Users\romesh\miniconda3\envs\atomica38\lib\site-packages\altair\utils\schemapi.py", line 711, in __getitem__
    return self._kwds[item]
KeyError: 'height'
```

binste:
Fixes #3067. When reconstructing a layered chart which has `height`or `width` defined on the top-level with `from_dict`, the code so far checked that the layers have that property set to `Undefined` or to the same value as the layered chart itself. However, it's often the case that the layers don't have the property set at all -> This fix handles those cases.

I also added a test to run a to_json and from_json roundtrip for all examples in the gallery so that we can hopefully catch future regressions to this functionality. I checked that these tests would have caught the bug reported in #3067.

@mattijn Would be great if this fix can still make it into 5.0.1 in case you or someone else has the time to review it.

### DIFF
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -3395,9 +3395,12 @@ def _remove_layer_props(chart, subcharts, layer_props):
             else:
                 raise ValueError(f"There are inconsistent values {values} for {prop}")
         else:
-            # Top level has this prop; subchart props must be either
-            # Undefined or identical to proceed.
-            if all(c[prop] is Undefined or c[prop] == chart[prop] for c in subcharts):
+            # Top level has this prop; subchart must either not have the prop
+            # or it must be Undefined or identical to proceed.
+            if all(
+                getattr(c, prop, Undefined) is Undefined or c[prop] == chart[prop]
+                for c in subcharts
+            ):
                 output_dict[prop] = chart[prop]
             else:
                 raise ValueError(f"There are inconsistent values {values} for {prop}")
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -3,6 +3,7 @@ import pkgutil
 
 import pytest
 
+import altair as alt
 from altair.utils.execeval import eval_block
 from tests import examples_arguments_syntax
 from tests import examples_methods_syntax
@@ -45,6 +46,39 @@ def test_render_examples_to_chart(syntax_module):
             raise AssertionError(
                 f"Example file {filename} raised an exception when "
                 f"converting to a dict: {err}"
+            ) from err
+
+
+@pytest.mark.parametrize(
+    "syntax_module", [examples_arguments_syntax, examples_methods_syntax]
+)
+def test_from_and_to_json_roundtrip(syntax_module):
+    """Tests if the to_json and from_json (and by extension to_dict and from_dict)
+    work for all examples in the Example Gallery.
+    """
+    for filename in iter_examples_filenames(syntax_module):
+        source = pkgutil.get_data(syntax_module.__name__, filename)
+        chart = eval_block(source)
+
+        if chart is None:
+            raise ValueError(
+                f"Example file {filename} should define chart in its final "
+                "statement."
+            )
+
+        try:
+            first_json = chart.to_json()
+            reconstructed_chart = alt.Chart.from_json(first_json)
+            # As the chart objects are not
+            # necessarily the same - they could use different objects to encode the same
+            # information - we do not test for equality of the chart objects, but rather
+            # for equality of the json strings.
+            second_json = reconstructed_chart.to_json()
+            assert first_json == second_json
+        except Exception as err:
+            raise AssertionError(
+                f"Example file {filename} raised an exception when "
+                f"doing a json conversion roundtrip: {err}"
             ) from err
 
 


mattijn:
Thanks for this PR @binste! 

On the branch of this PR 23 examples (arguments syntax only) are failing your introduced test without the patch you suggested. I applied the same test (arguments syntax only) to the 4.2.x branch and there all tests pass (0 fails) without requiring your suggested patch. 

This makes we wonder if we can discover why it is currently often the case that the layers don't have the property set at all. Another thing that makes me wonder, I always thought that -all- properties by default are set to `Undefined`, so in my mind it is or defined or `Undefined`, but not nothing. See any API reference (eg [alt.LayerChart](https://altair-viz.github.io/user_guide/generated/toplevel/altair.LayerChart.html)).

The following are just guesses, buy do we maybe miss something in the codegen.py/generate_schema_wrapper.py? There we append `=Undefined` at [L209 ](https://github.com/altair-viz/altair/blob/master/tools/schemapi/codegen.py#L209), [L252](https://github.com/altair-viz/altair/blob/master/tools/schemapi/codegen.py#L252) and [L520](https://github.com/altair-viz/altair/blob/master/tools/generate_schema_wrapper.py#L520), but is that all required? Or maybe something related to the `UndefinedType` (around [here](https://github.com/altair-viz/altair/blob/master/tools/schemapi/schemapi.py#L586))?

binste:
The tests pass in Altair 4 as the behavior of lifting `height` and `width` to the top-level of a layered chart were introduced in Altair 5 in https://github.com/altair-viz/altair/commit/c2a58b2f5ef211fd09d68e19c96582237387fddc and with it also the line of code which I patched.

For a normal `Chart`, they would be set to `Undefined`:
```python
import altair as alt
c = alt.Chart().mark_bar()
c._kwds
```
```python
{'data': Undefined,
 'mark': 'bar',
 'align': Undefined,
 'autosize': Undefined,
 'background': Undefined,
 'bounds': Undefined,
 'center': Undefined,
 'config': Undefined,
 'datasets': Undefined,
 'description': Undefined,
 'encoding': Undefined,
 'height': Undefined,
 'name': Undefined,
 'padding': Undefined,
 'params': Undefined,
 'projection': Undefined,
 'resolve': Undefined,
 'spacing': Undefined,
 'title': Undefined,
 'transform': Undefined,
 'usermeta': Undefined,
 'view': Undefined,
 'width': Undefined}
```

But I think it makes sense that `height` and `width` are not defined at all in the layers when reconstructing it as the layers are following the specifications of a `UnitSpec` which does not have those attributes. 

image

`layered_reconstructed.layer[0]._kwds`:

```python
{'mark': MarkDef({
   type: Mark('bar')
 }),
 'data': NamedData({
   name: 'empty'
 }),
 'description': Undefined,
 'encoding': Undefined,
 'name': Undefined,
 'params': Undefined,
 'projection': Undefined,
 'title': Undefined,
 'transform': Undefined}
```

This is also visible in the [Atlassian JSON Schema Viewer for a LayerSpec](https://json-schema.app/view/%23%2Fdefinitions%2FLayerSpec?url=https%3A%2F%2Fvega.github.io%2Fschema%2Fvega-lite%2Fv5.json) -> `layers` shows that it accepts an array of `LayerSpec` or `GenericUnitSpec<Encoding,AnyMark>`. `GenericUnitSpec` does not accept `width` or `height`, see [here](https://json-schema.app/view/%23%2Fdefinitions%2FLayerSpec/%23%2Fdefinitions%2FGenericUnitSpec%3CEncoding%2CAnyMark%3E?url=https%3A%2F%2Fvega.github.io%2Fschema%2Fvega-lite%2Fv5.json).

In my understanding, when converting to a Vega-Lite spec it doesn't matter if an attribute is not defined at all in the Altair objects or set to `Undefined` as we exclude all key, value pairs with value `Undefined` when converting to a dictionary, see [here](https://github.com/altair-viz/altair/blob/master/altair/utils/schemapi.py#L340).

Would you agree or do you think there is something we should still dig deeper into?

mattijn:
Thanks for the clarification @binste! It is more clear now. I thought with 'not set' it meant that it was a valid property for the specific component, but it was not given as defined or `Undefined`, but as you explained, this `GenericUnitSpec` does not accept these properties. Merging this. Thanks again!

PR #3071, opened 2023-05-27
Add Vega-Lite to Vega compiler registry and format arg to to_dict() and to_json()

jonmmease:
This PR is a first step towards the proposal discussed in https://github.com/altair-viz/altair/discussions/3054 to add a `chart.transformed_data()` function that relies on VegaFusion.

Because VegaFusion works on Vega specs, it's necessary to compile the Vega-Lite spec produced by `chart.to_dict()` into a Vega spec.  This PR adds a new `chart.to_vega()` method to perform this compilation.  Vega-Lite to Vega compilation support is provided by vl-convert.

To maximize future flexibility, this PR includes a new `PluginRegistry` for functions that compile Vega-Lite to Vega and registers a default implementation using vl-convert. My thinking is that this approach will make it possible to support the `to_vega()` method in contexts where vl-convert isn't available. For example, when running in pyodide, a custom plugin could be registered that uses the Vega-Lite JavaScript library directly. Or if running on an architecture that `vl-convert-python` isn't compiled for, a custom implementation could be registered that makes an API call to a remote server running vl-convert.

A couple of tests are added in tests/utils/test_compiler.py

### DIFF
--- a/altair/utils/compiler.py
+++ b/altair/utils/compiler.py
@@ -4,7 +4,7 @@ from altair.utils import PluginRegistry
 # ==============================================================================
 # Vega-Lite to Vega compiler registry
 # ==============================================================================
-VegaLiteCompilerType = Callable
+VegaLiteCompilerType = Callable[[dict], dict]
 
 
 class VegaLiteCompilerRegistry(PluginRegistry[VegaLiteCompilerType]):
--- a/altair/utils/compiler.py
+++ b/altair/utils/compiler.py
@@ -1,0 +1,11 @@
+from typing import Callable
+from altair.utils import PluginRegistry
+
+# ==============================================================================
+# Vega-Lite to Vega compiler registry
+# ==============================================================================
+VegaLiteCompilerType = Callable
+
+
+class VegaLiteCompilerRegistry(PluginRegistry[VegaLiteCompilerType]):
+    pass
--- a/altair/vegalite/compiler.py
+++ b/altair/vegalite/compiler.py
@@ -1,4 +1,0 @@
-from ..utils.compiler import VegaLiteCompilerRegistry
-
-
-__all__ = ("VegaLiteCompilerRegistry",)
--- a/altair/vegalite/compiler.py
+++ b/altair/vegalite/compiler.py
@@ -1,0 +1,4 @@
+from ..utils.compiler import VegaLiteCompilerRegistry
+
+
+__all__ = ("VegaLiteCompilerRegistry",)
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -8,7 +8,7 @@ import pandas as pd
 from toolz.curried import pipe as _pipe
 import itertools
 import sys
-from typing import cast
+from typing import cast, Optional, List, Dict, Any
 
 # Have to rename it here as else it overlaps with schema.core.Type
 from typing import Type as TypingType
@@ -816,8 +816,14 @@ class TopLevelMixin(mixins.ConfigMethodMixin):
 
     _class_is_valid_at_instantiation = False
 
-    def to_dict(self, *args, **kwargs) -> dict:
+    def to_dict(self, *args, format: str = "vega-lite", **kwargs) -> dict:
         """Convert the chart to a dictionary suitable for JSON export"""
+        # Validate format
+        if format not in ("vega-lite", "vega"):
+            raise ValueError(
+                f'The format argument must be either "vega-lite" or "vega". Received {repr(format)}'
+            )
+
         # We make use of three context markers:
         # - 'data' points to the data that should be referenced for column type
         #   inference.
@@ -828,7 +834,7 @@ class TopLevelMixin(mixins.ConfigMethodMixin):
 
         # note: not a deep copy because we want datasets and data arguments to
         # be passed by reference
-        context = kwargs.get("context", {}).copy()
+        context = (kwargs.get("context", {}) or {}).copy()
         context.setdefault("datasets", {})
         is_top_level = context.get("top_level", True)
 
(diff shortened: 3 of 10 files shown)


altair/vegalite/compiler.py
binste, line 1: I think we can skip this file and directly import from utils.compiler in `v5.compiler` as done in `v5.theme`. 

I think these a bit complicated structures are a leftover of when Altair supported multiple Vega-Lite versions. I think we can simplify this in a future major release but that's just a side remark and not important for this PR.

altair/vegalite/v5/compiler.py
```diff
@@ -0,0 +1,26 @@
+from ..compiler import VegaLiteCompilerRegistry
+
+vegalite_compilers = VegaLiteCompilerRegistry()
```
binste, line 3: We could already add an entry point group so that it's easier to register new compilers.

```suggestion
ENTRY_POINT_GROUP: str = "altair.vegalite.v5.vegalite_compiler"
vegalite_compilers = VegaLiteCompilerRegistry(entry_point_group=ENTRY_POINT_GROUP)
```

altair/vegalite/v5/compiler.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def vl_convert_compiler(vegalite_spec) -> dict:
```
binste, line 6:
```suggestion
def vl_convert_compiler(vegalite_spec: dict) -> dict:
```

altair/utils/compiler.py
```diff
@@ -0,0 +1,11 @@
+from typing import Callable
+from altair.utils import PluginRegistry
+
+# ==============================================================================
+# Vega-Lite to Vega compiler registry
+# ==============================================================================
+VegaLiteCompilerType = Callable
```
binste, line 7:
```suggestion
VegaLiteCompilerType = Callable[[dict], dict]
```

Let's just add some simple type hints. We'll probably expand these as part of the overall typing effort (#2951).

binste (changes requested):
This already looks great! Nice to already prepare for potential other compilers, e.g. for Pyodide as you mentioned.

altair/vegalite/v5/api.py
```diff
@@ -891,6 +892,23 @@ def to_html(
             requirejs=requirejs,
         )
 
+    def to_vega(self):
```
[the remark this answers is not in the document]
jonmmease: Yeah, adding a `format` arg to `to_dict()` makes good sense. The current signature is:

```python
    def to_dict(self, *args, **kwargs) -> dict:
        """Convert the chart to a dictionary suitable for JSON export"""
```

I could make this the new signature:
```python
    def to_dict(self, *args, format="vega-lite", **kwargs) -> dict:
        """Convert the chart to a dictionary suitable for JSON export"""
```

But the `*args` and `**kwargs` makes this a bit awkward, as I don't think they are intended to be part of the public API.  Do you have any thoughts here?

altair/vegalite/compiler.py
jonmmease: Yeah, sounds good

altair/vegalite/v5/api.py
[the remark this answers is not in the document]
binste: Good point. We should be able to specify these arguments explicitly, I created #3073.

altair/vegalite/compiler.py
jonmmease: Done in f5c52551792d01c0e9a3d9d8cf27b292e01efba6

altair/vegalite/v5/compiler.py
jonmmease: Done in c7e860774eea5468c612e486767a38a8a64acc77

altair/vegalite/v5/compiler.py
jonmmease: done in f043b485107ba02643755df873554978596b3482

altair/utils/compiler.py
jonmmease: Done in f043b485107ba02643755df873554978596b3482

jonmmease:
Converting to draft until #3073 is merged

jonmmease:
@binste, I think approach will work (adding `format` to `to_dict` and `to_json`), but I just wanted to mention something that's a bit awkward.

`to_dict` is defined in several places, (at the [TopLevelMixin-level](https://github.com/altair-viz/altair/blob/601ae167a66d77cd6b5f154a53c63a364afc7c16/altair/vegalite/v5/api.py#L818) and the [SchemaBase-level](https://github.com/altair-viz/altair/blob/601ae167a66d77cd6b5f154a53c63a364afc7c16/tools/schemapi/schemapi.py#L735)). In order to satisfy mypy, all of these signatures need to match, so it's necessary to add the format arg to the `SchemaBase.to_dict` method even though it can't be used at that level.  There's a similar issue for `to_json` as it's defined at the Schema-base level.

It would be kind of nice if we could use different names for the `TopLevelMixin` and `SchemaBase` implementations of to_dict. This way we wouldn't need to expose the `context` and `ignore` args in `TopLevelMixen.to_dict`, and we wouldn't need to include the `format` argument in `SchemaBase.to_dict`.  On the other hand, that may not be worth the effort.

binste:
Hmm I can't fully replicate this. When I add `new_argument` to `TopLevelMixin.to_dict` mypy complains that:

```python
altair/vegalite/v5/api.py:2547: error: Signature of "to_dict" incompatible with supertype "TopLevelMixin"  [override]
...
```
This error refers to `Chart.to_dict`. Once I also add `new_argument` to `Chart.to_dict` mypy is happy. Do you get any other errors?

jonmmease:
@binste, I've merged in master. Here is the mypy error I'm getting now:

```
cmd [3] | mypy altair tests
altair/vegalite/v5/api.py:2488: error: Definition of "to_json" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
altair/vegalite/v5/api.py:2606: error: Signature of "to_dict" incompatible with supertype "SchemaBase"  [override]
altair/vegalite/v5/api.py:2606: note:      Superclass:
altair/vegalite/v5/api.py:2606: note:          def to_dict(self, validate: bool = ..., ignore: Optional[List[str]] = ..., context: Optional[Dict[str, Any]] = ...) -> Dict[Any, Any]
altair/vegalite/v5/api.py:2606: note:      Subclass:
altair/vegalite/v5/api.py:2606: note:          def to_dict(self, validate: bool = ..., format: str = ..., ignore: Optional[List[str]] = ..., context: Optional[Dict[str, Any]] = ...) -> Dict[Any, Any]
altair/vegalite/v5/api.py:2606: note:          def to_dict(self, validate: bool = ..., ignore: Optional[List[str]] = ..., context: Optional[Dict[str, Any]] = ...) -> Dict[Any, Any]
altair/vegalite/v5/api.py:2606: note:          def to_dict(self, validate: bool = ..., format: str = ..., ignore: Optional[List[str]] = ..., context: Optional[Dict[str, Any]] = ...) -> Dict[Any, Any]
altair/vegalite/v5/api.py:2771: error: Definition of "to_dict" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
altair/vegalite/v5/api.py:2771: error: Definition of "to_json" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
altair/vegalite/v5/api.py:2891: error: Definition of "to_dict" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
altair/vegalite/v5/api.py:2891: error: Definition of "to_json" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
altair/vegalite/v5/api.py:2964: error: Definition of "to_dict" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
altair/vegalite/v5/api.py:2964: error: Definition of "to_json" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
altair/vegalite/v5/api.py:3035: error: Definition of "to_dict" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
altair/vegalite/v5/api.py:3035: error: Definition of "to_json" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
altair/vegalite/v5/api.py:3106: error: Definition of "to_dict" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
altair/vegalite/v5/api.py:3106: error: Definition of "to_json" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
altair/vegalite/v5/api.py:3198: error: Definition of "to_dict" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
altair/vegalite/v5/api.py:3198: error: Definition of "to_json" in base class "TopLevelMixin" is incompatible with definition in base class "SchemaBase"  [misc]
Found 14 errors in 1 file (checked 331 source files)
```

jonmmease:
In 6509aaad8dcde96e43d82f107d9879eff8b770d1 I add the format argument to the `SchemaBase` `to_dict` and `to_json` methods and mypy is happy.

I don't think there's any harm in this, but wanted to point it out. How do you feel about this aproach?

binste:
With this approach `format` shows up in the docstring of `to_dict` for objects such as `alt.Title`. I agree that there is not much harm in this as users use `to_dict` mostly used on top-level objects anyway.

Seems like mypy is also happy if instead of modifying the SchemaBase method we mark `format` and any subsequent arguments as keyword-only. Got the idea from [this SO answer](https://stackoverflow.com/a/60137300). This way, it's only present in the methods were it actually has an effect. What do you think? I pushed the relevant changes in a new commit directly to your branch. Hope that's ok but I thought it's easier instead of pasting code in here.

binste (approved):

jonmmease:
Thanks for looking into this. The keyword only arg approach is great! Merging

PR #3073, opened 2023-05-29
Explicitly specify arguments for to_dict and to_json methods for top-level chart objects

binste:
This PR explicitly specifies and documents the accepted arguments for `to_dict` and `to_json` on top-level chart objects (e.g. `Chart`, `LayerChart`, ...). Previously, these methods used `args` and `kwargs`. This improves the documentation for users, allows us to type hint the arguments, and it unblocks https://github.com/altair-viz/altair/pull/3071#discussion_r1208617735 where it is currently unclear how to best add an additional documented keyword argument `format`.

This might break existing code in case a user passes an unused argument, e.g. `to_dict(this_argument_is_not_used=1`). However, that seems acceptable to me as I don't think it was the intended behavior to allow for arbitrary arguments.

cc @jonmmease. 
Maybe someone else could review it as well. Just in case I'm missing an actual use case for the `args`/`kwargs` notation

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -827,10 +827,10 @@ class SchemaBase:
     def to_json(
         self,
         validate: bool = True,
-        ignore: Optional[List[str]] = None,
-        context: Optional[Dict[str, Any]] = None,
         indent: int = 2,
         sort_keys: bool = True,
+        ignore: Optional[List[str]] = None,
+        context: Optional[Dict[str, Any]] = None,
         **kwargs,
     ) -> str:
         """Emit the JSON representation for this object as a string.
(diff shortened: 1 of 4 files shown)


altair/utils/schemapi.py
```diff
@@ -816,36 +826,41 @@ def to_dict(self, validate=True, ignore=None, context=None):
...
-        validate=True,
-        ignore=None,
-        context=None,
-        indent=2,
-        sort_keys=True,
+        validate: bool = True,
+        ignore: Optional[List[str]] = None,
+        context: Optional[Dict[str, Any]] = None,
+        indent: int = 2,
+        sort_keys: bool = True,
```
jonmmease, line 833: While we're changing signatures, how would you feel about moving `indent` and `sort_keys` above `ignore` and `context`. Like this:

```python
    def to_json(
        validate: bool = True,
        indent: int = 2,
        sort_keys: bool = True,
        ignore: Optional[List[str]] = None,
        context: Optional[Dict[str, Any]] = None,
```

I doubt this would be breaking in practice as `ignore` and `context` aren't intended for use by end users, and even if they are being used it would be even more unusual to specify `validate` `ignore` and `context` as positional args.

This is top of mind because I'd rather add the format argument in #3071 before `ignore` and `context`.

jonmmease (commented):
Thanks for doing this! One suggestion inline for the argument order of `to_json`.

binste:
@mattijn @joelostblom Could someone of you review this PR as well just to be sure not to break anything? No pressure at all, just a gentle ping to make sure you see this PR :)

mattijn:
Thanks for the reminder, end of week or early next week I've time.

mattijn (approved):

mattijn:
Thanks @binste! Looks good to me too.

PR #3074, opened 2023-05-31
Improve troubleshooting docs regarding Vega-Lite 5

### LINKED ISSUE #3070: Include a suggestion to update frontend (Jupyterlab, ...) in mimetype error
Follow-up that comes out of #2585. Raised by @joelostblom:

> do you think we need to communicate the minimum version of JuptyerLab that support Altair 5 somewhere? I am thinking ideally directly in the error message if possible, but otherwise at least in the docs and release notes, what do you all think?

binste:
Resolves #3070. I updated the documentation which is linked from the error message. This should help users figure out what to do in case they use a Jupyterlab or notebook version which does not support the VL 5 mimetype renderer.


jonmmease (approved):
This looks like a helpful update. Thanks @binste

joelostblom:
Looks great, thank you!

joelostblom (approved):

PR #3076, opened 2023-06-02
Sanitize timestamps in arrow tables

### LINKED ISSUE #3050: pyarrow datetime is not supported using altair 5.0
I created an altair chart from a duckdb dataframe exported as an arrow table,  but I get this error 

TypeError: Object of type datetime is not JSON serializable

when I cast the field to string everything works fine.

jonmmease:
Closes #3050 by converting timestamps columns in Arrow tables to ISO-8601 strings before JSON serialization.

I added a test, though it doesn't look like we had any existing test coverage for the DataFrame interchange protocol support.

cc @djouallah

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -397,6 +397,24 @@ def sanitize_dataframe(df):  # noqa: C901
     return df
 
 
+def sanitize_arrow_table(pa_table):
+    """Sanitize arrow table for JSON serialization"""
+    import pyarrow as pa
+    import pyarrow.compute as pc
+
+    arrays = []
+    schema = pa_table.schema
+    for name in schema.names:
+        array = pa_table[name]
+        dtype = schema.field(name).type
+        if str(dtype).startswith("timestamp"):
+            arrays.append(pc.strftime(array))
+        else:
+            arrays.append(array)
+
+    return pa.Table.from_arrays(arrays, names=schema.names)
+
+
 def parse_shorthand(
     shorthand,
     data=None,
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -409,6 +409,13 @@ def sanitize_arrow_table(pa_table):
         dtype = schema.field(name).type
         if str(dtype).startswith("timestamp"):
             arrays.append(pc.strftime(array))
+        elif str(dtype).startswith("duration"):
+            raise ValueError(
+                'Field "{col_name}" has type "{dtype}" which is '
+                "not supported by Altair. Please convert to "
+                "either a timestamp or a numerical value."
+                "".format(col_name=name, dtype=dtype)
+            )
         else:
             arrays.append(array)
 
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -8,7 +8,7 @@ import pandas as pd
 from toolz import curried
 from typing import Callable
 
-from .core import sanitize_dataframe
+from .core import sanitize_dataframe, sanitize_arrow_table
 from .core import sanitize_geo_interface
 from .deprecation import AltairDeprecationWarning
 from .plugin_registry import PluginRegistry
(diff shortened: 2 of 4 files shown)


mattijn:
Thanks for pushing this @jonmmease! Would be interesting to see if we slowly can phase-in the dataframe interchange protocol for all dataframe-a-likes (and simultaneously phase-out the dependency on pandas).

Can you move the `_sanitize_arrow_table` to `core.py` below this function https://github.com/altair-viz/altair/blob/master/altair/utils/core.py#L287?

Would it also be possible to include a test including a `NaT` equivalent (`None` I think).

By the way, on windows the test is not passing. It gives the following error: 
```python
ArrowInvalid: Cannot locate timezone 'UTC': Timezone database not found at D:\Users\Hoek\Downloads\tzdata
```
See also https://github.com/apache/arrow/issues/31472 and https://github.com/apache/arrow/pull/35637#issuecomment-1560531201. That probably require some thought how we approach this.

jonmmease:
> Can you move the _sanitize_arrow_table to core.py below this function https://github.com/altair-viz/altair/blob/master/altair/utils/core.py#L287?

Done in 58f11ba7718a969809f1dc65e6f4ac504977a582

> Would it also be possible to include a test including a NaT equivalent (None I think).

Done in 9c6ea83e5b908c13401160514f97d45209c751e7

> Would be interesting to see if we slowly can phase-in the dataframe interchange protocol for all dataframe-a-likes (and simultaneously phase-out the dependency on pandas).

Yeah, that's an interesting possibility. I'm a bit leery of adding a required dependency on pyarrow since it's not available everywhere pandas is (e.g. pyodide).  But the pandas project itself is currently debating whether to take on pyarrow as a required dependency anyway, so that might be the future direction the ecosystem goes: https://github.com/pandas-dev/pandas/issues/52509, https://github.com/pandas-dev/pandas/pull/52711

mattijn:
I agree. I think it is unfortunate that we, for now, have to rely on pyarrow to get support for the current implementation of the dataframe interchange protocol.

If there is a WASM friendly option to support the dataframe interchange protocol, I would favor that one.

jonmmease:
> By the way, on windows the test is not passing

Given that this is a known upstream pyarrow issue, are you comfortable with this being merged?

mattijn:
Yes not a blocker from my side. Also noticed that wasm support for pyarrow is something that is actively worked on, see: https://github.com/pyodide/pyodide/issues/2933, that is positive.

Haven't had the change to test this PR, which I would like to do, please give me a few more days.

jonmmease:
> Haven't had the change to test this PR, which I would like to do, please give me a few more days.

Sure thing. Thanks for taking a look. I was thinking that if the test doesn't work on windows, maybe we should skip it when the current platform is windows.

mattijn:
Getting this `tz_data` folder on the right place was a bit strange for my machine, see https://github.com/apache/arrow/issues/35600#issuecomment-1580624761. 
Skipping this test on windows seems fine to me, see https://github.com/apache/arrow/pull/35735 how they've done this recently as well within pyarrow.

Can you also have a look to this [comment](https://github.com/altair-viz/altair/issues/3050#issuecomment-1571863504), which states:

> Note: in case of duration, the error should be raised.

Is this something we have to consider for here? Docs for duration: https://arrow.apache.org/docs/python/generated/pyarrow.duration.html. 

Also related, `Timedelta`:  
```python
import altair as alt
import pandas as pd
import pyarrow as pa
from altair.utils.data import to_values

td = pd.timedelta_range(0,periods=3,freq='h')
df = pd.DataFrame(td).reset_index()
df.columns = ['id', 'timedelta']

pa_table = pa.table(df)
values = to_values(pa_table)
values
```

> ```cmd
> {'values': [{'id': 0, 'timedelta': Timedelta('0 days 00:00:00')},
>   {'id': 1, 'timedelta': Timedelta('0 days 01:00:00')},
>   {'id': 2, 'timedelta': Timedelta('0 days 02:00:00')}]}
> ```

```python
alt.Chart(pa_table).mark_bar().encode(
    x = 'timedelta:O',
    y = 'id:O'
)
```

> ```cmd
> TypeError: Object of type Timedelta is not JSON serializable
> ```
See also https://github.com/altair-viz/altair/pull/974

jonmmease:
Thanks for the review @mattijn, test are massing. Merging!

PR #3094, opened 2023-06-30
Add VegaFusion data transformer with mime renderer, save, and to_dict/to_json integration

jonmmease:
## Overview
Following on from the discussion in https://github.com/altair-viz/altair/discussions/3054, this PR introduces a new data transformer named "vegafusion".  When activated, we use [VegaFusion](https://vegafusion.io/) to pre-evaluate data transformations before displaying charts with mime renderers, and before saving charts with `Chart.save`.

## The Data Transformer
The "vegafusion" data transformer replaces DataFrames in Vega-Lite specs with URLs of the form "table://{uuid}". These DataFrames are stashed in a local `WeakValueDictionary`, with the UUID as the key. The `table://` URL is a VegaFusion convention used to mean that the associated data will be provided using the `inline_datasets` argument to the `pre_transform_spec` function.

## Mimerenderer and Save
A function is defined in `_vegafusion_data.py` called `compile_with_vegafusion` that inputs a Vega-Lite spec containing these `table://` URLs, compiles it to a Vega spec, then invokes VegaFusion's `pre_transform_spec` function. The stashed DataFrames that correspond to the `table://` URLs are extracted from the `WeakValueDictionary` and passed to `pre_transform_spec` as the `inline_datasets` argument.  `compile_with_vegafusion` then returns the pre-transformed Vega spec as a dictionary.

When the "vegafusion" data transformer is enabled, the `default_renderer_base()` function calls `compile_with_vegafusion()` before returning a Vega mime bundle.  Similarly, the `spec_to_mimebundle` function has been updated to call `compile_with_vegafusion()` when the "vegafusion" data transformer is enabled. This way, saved charts have their transforms pre-evaluated before saving.

## Transformed Data
This PR also updates the `chart.transformed_data` functionality added in #3081 and #3084 to rely on the new "vegafusion" data transformere rather than the "vegafusion-inline" transformer that is provided by the `vegafusion` Python package. ("vegafusion-inline" will be deprecated in `vegafusion` once this is released).

## MaxRowsError
VegaFusion still has a notion of a maximum number of rows. But unlike the regular row limit, VegaFusion's row limit is applied after all supported data transformations have been applied. So you may hit it in the case of a large scatter chart.  I decided to reuse the existing infrastructure for setting and disabling max_rows (e.g. `alt.data_transformers.disable_max_rows()`). When this post-transformed limit is exceeded, we raise the same `MaxRowsError` exception, but with a message that explains that the limit is applied after data transformations.

This is open for discussion, but I set the "vegafusion" data transformer's row limit to 100k.  This is partly justified by the fact that VegaFusion will remove unused data columns. But also, practically speaking I haven't run into issues at this scale. One case where a high default limit is helpful is for creating fine-grained heatmaps (e.g. 300x300 get's you to 90k).

## Testing
I added a test for the `spec_to_mimebundle` updates, but still need to find a place to test the renderer path.

## Follow-up
After this PR, I want to work on how to document this. We can talk more about it, but I'm thinking of adding a section to the top of the "Large Datasets" section that recommends enabling this data transformer as the first step.  And perhaps also updating the default MaxRowsError with instructions on how to enable it.

### DIFF
--- a/altair/utils/_vegafusion_data.py
+++ b/altair/utils/_vegafusion_data.py
@@ -143,7 +143,7 @@ def compile_with_vegafusion(vegalite_spec: dict) -> dict:
     from altair import vegalite_compilers, data_transformers
 
     try:
-        import vegafusion as vf
+        import vegafusion as vf  # type: ignore
     except ImportError as e:
         raise ImportError(
             'The "vegafusion" data transformer requires the vegafusion-python-embed\n'
@@ -154,7 +154,11 @@ def compile_with_vegafusion(vegalite_spec: dict) -> dict:
         ) from e
 
     # Compile Vega-Lite spec to Vega
-    vega_spec = vegalite_compilers.get()(vegalite_spec)
+    compiler = vegalite_compilers.get()
+    if compiler is None:
+        raise ValueError("No active vega-lite compiler plugin found")
+
+    vega_spec = compiler(vegalite_spec)
 
     # Retrieve dict of inline tables referenced by the spec
     inline_tables = get_inline_tables(vega_spec)
--- a/altair/utils/_vegafusion_data.py
+++ b/altair/utils/_vegafusion_data.py
@@ -1,0 +1,189 @@
+from typing import TypedDict, Union, Dict, Set, MutableMapping, Final
+
+from toolz import curried
+import uuid
+from weakref import WeakValueDictionary
+
+from altair.utils.core import _DataFrameLike
+from altair.utils.data import _DataType, _ToValuesReturnType, MaxRowsError
+from altair.vegalite.data import default_data_transformer
+
+# Temporary storage for dataframes that have been extracted
+# from charts by the vegafusion data transformer. Use a WeakValueDictionary
+# rather than a dict so that the Python interpreter is free to garbage
+# collect the stored DataFrames.
+extracted_inline_tables: MutableMapping[str, _DataFrameLike] = WeakValueDictionary()
+
+# Special URL prefix that VegaFusion uses to denote that a
+# dataset in a Vega spec corresponds to an entry in the `inline_datasets`
+# kwarg of vf.runtime.pre_transform_spec().
+VEGAFUSION_PREFIX: Final = "table://"
+
+
+class _ToVegaFusionReturnUrlDict(TypedDict):
+    url: str
+
+
+@curried.curry
+def vegafusion_data_transformer(
+    data: _DataType, max_rows: int = 100000
+) -> Union[_ToVegaFusionReturnUrlDict, _ToValuesReturnType]:
+    """VegaFusion Data Transformer"""
+    if hasattr(data, "__geo_interface__"):
+        # Use default transformer for geo interface objects
+        # # (e.g. a geopandas GeoDataFrame)
+        return default_data_transformer(data)
+    elif hasattr(data, "__dataframe__"):
+        table_name = f"table_{uuid.uuid4()}".replace("-", "_")
+        extracted_inline_tables[table_name] = data
+        return {"url": VEGAFUSION_PREFIX + table_name}
+    else:
+        # Use default transformer if we don't recognize data type
+        return default_data_transformer(data)
+
+
+def get_inline_table_names(vega_spec: dict) -> Set[str]:
+    """Get a set of the inline datasets names in the provided Vega spec
+
+    Inline datasets are encoded as URLs that start with the table://
+    prefix.
+
+    Parameters
+    ----------
+    vega_spec: dict
+        A Vega specification dict
+
+    Returns
+    -------
+    set of str
+        Set of the names of the inline datasets that are referenced
+        in the specification.
+
+    Examples
+    --------
+    >>> spec = {
+    ...     "data": [
+    ...         {
+    ...             "name": "foo",
+    ...             "url": "https://path/to/file.csv"
+    ...         },
+    ...         {
+    ...             "name": "bar",
+    ...             "url": "table://inline_dataset_123"
+    ...         }
+    ...     ]
+    ... }
+    >>> get_inline_table_names(spec)
+    {'inline_dataset_123'}
+    """
+    table_names = set()
+
+    # Process datasets
+    for data in vega_spec.get("data", []):
+        url = data.get("url", "")
+        if url.startswith(VEGAFUSION_PREFIX):
+            name = url[len(VEGAFUSION_PREFIX) :]
+            table_names.add(name)
+
+    # Recursively process child marks, which may have their own datasets
+    for mark in vega_spec.get("marks", []):
+        table_names.update(get_inline_table_names(mark))
+
+    return table_names
+
+
+def get_inline_tables(vega_spec: dict) -> Dict[str, _DataFrameLike]:
+    """Get the inline tables referenced by a Vega specification
+
+    Note: This function should only be called on a Vega spec that corresponds
+    to a chart that was processed by the vegafusion_data_transformer.
+    Furthermore, this function may only be called once per spec because
+    the returned dataframes are deleted from internal storage.
+
+    Parameters
+    ----------
+    vega_spec: dict
+        A Vega specification dict
+
+    Returns
+    -------
+    dict from str to dataframe
+        dict from inline dataset name to dataframe object
+    """
+    table_names = get_inline_table_names(vega_spec)
+    tables = {}
+    for table_name in table_names:
+        try:
+            tables[table_name] = extracted_inline_tables.pop(table_name)
+        except KeyError:
+            # named dataset that was provided by the user
+            pass
+    return tables
+
+
+def compile_with_vegafusion(vegalite_spec: dict) -> dict:
+    """Compile a Vega-Lite spec to Vega and pre-transform with VegaFusion
+
+    Note: This function should only be called on a Vega-Lite spec
+    that was generated with the "vegafusion" data transformer enabled.
+    In particular, this spec may contain references to extract datasets
+    using table:// prefixed URLs.
+
+    Parameters
+    ----------
+    vegalite_spec: dict
+        A Vega-Lite spec that was generated from an Altair chart with
+        the "vegafusion" data transformer enabled
+
+    Returns
+    -------
+    dict
+        A Vega spec that has been pre-transformed by VegaFusion
+    """
+    from altair import vegalite_compilers, data_transformers
+
+    try:
+        import vegafusion as vf
+    except ImportError as e:
+        raise ImportError(
+            'The "vegafusion" data transformer requires the vegafusion-python-embed\n'
+            "and vegafusion packages. These can be installed with pip using:\n"
+            '    pip install "vegafusion[embed]"\n'
+            "Or with conda using:\n"
+            "    conda install -c conda-forge vegafusion-python-embed vegafusion"
+        ) from e
+
+    # Compile Vega-Lite spec to Vega
+    vega_spec = vegalite_compilers.get()(vegalite_spec)
+
+    # Retrieve dict of inline tables referenced by the spec
+    inline_tables = get_inline_tables(vega_spec)
+
+    # Pre-evaluate transforms in vega spec with vegafusion
+    row_limit = data_transformers.options.get("max_rows", None)
+    transformed_vega_spec, warnings = vf.runtime.pre_transform_spec(
+        vega_spec,
+        vf.get_local_tz(),
+        inline_datasets=inline_tables,
+        row_limit=row_limit,
+    )
+
+    # Check from row limit warning and convert to MaxRowsError
+    for warning in warnings:
+        if warning.get("type") == "RowLimitExceeded":
+            raise MaxRowsError(
+                "The number of dataset rows after filtering and aggregation exceeds\n"
+                f"the current limit of {row_limit}. Try adding an aggregation to reduce\n"
+                "the size of the dataset that must be loaded into the browser. Or, disable\n"
+                "the limit by calling alt.data_transformers.disable_max_rows(). Note that\n"
+                "disabling this limit may cause the browser to freeze or crash."
+            )
+
+    return transformed_vega_spec
+
+
+def using_vegafusion() -> bool:
+    """Check whether the vegafusion data transfomer is enabled"""
+    from altair.vegalite.v5.data import data_transformers
+
+    return data_transformers.active == "vegafusion"
--- a/altair/utils/_vegafusion_data.py
+++ b/altair/utils/_vegafusion_data.py
@@ -3,12 +3,12 @@ from toolz import curried
 import uuid
 from weakref import WeakValueDictionary
 
-from typing import Union, Dict, Set, MutableMapping, Final
+from typing import Union, Dict, Set, MutableMapping
 
 if sys.version_info >= (3, 8):
-    from typing import TypedDict
+    from typing import TypedDict, Final
 else:
-    from typing_extensions import TypedDict
+    from typing_extensions import TypedDict, Final
 
 from altair.utils.core import _DataFrameLike
 from altair.utils.data import _DataType, _ToValuesReturnType, MaxRowsError
--- a/altair/utils/_vegafusion_data.py
+++ b/altair/utils/_vegafusion_data.py
@@ -146,6 +146,7 @@ def compile_with_vegafusion(vegalite_spec: dict) -> dict:
     dict
         A Vega spec that has been pre-transformed by VegaFusion
     """
+    # Local import to avoid circular ImportError
     from altair import vegalite_compilers, data_transformers
 
     try:
@@ -194,6 +195,7 @@ def compile_with_vegafusion(vegalite_spec: dict) -> dict:
 
 def using_vegafusion() -> bool:
     """Check whether the vegafusion data transfomer is enabled"""
-    from altair.vegalite.v5.data import data_transformers
+    # Local import to avoid circular ImportError
+    from altair import data_transformers
 
     return data_transformers.active == "vegafusion"
--- a/altair/utils/_vegafusion_data.py
+++ b/altair/utils/_vegafusion_data.py
@@ -23,7 +23,7 @@ extracted_inline_tables: MutableMapping[str, _DataFrameLike] = WeakValueDictiona
 # Special URL prefix that VegaFusion uses to denote that a
 # dataset in a Vega spec corresponds to an entry in the `inline_datasets`
 # kwarg of vf.runtime.pre_transform_spec().
-VEGAFUSION_PREFIX: Final = "table://"
+VEGAFUSION_PREFIX: Final = "vegafusion+dataset://"
 
 
 class _ToVegaFusionReturnUrlDict(TypedDict):
--- a/altair/utils/_vegafusion_data.py
+++ b/altair/utils/_vegafusion_data.py
@@ -1,8 +1,14 @@
-from typing import TypedDict, Union, Dict, Set, MutableMapping, Final
-
+import sys
 from toolz import curried
 import uuid
 from weakref import WeakValueDictionary
+
+from typing import Union, Dict, Set, MutableMapping, Final
+
+if sys.version_info >= (3, 8):
+    from typing import TypedDict
+else:
+    from typing_extensions import TypedDict
 
 from altair.utils.core import _DataFrameLike
 from altair.utils.data import _DataType, _ToValuesReturnType, MaxRowsError
--- a/altair/utils/display.py
+++ b/altair/utils/display.py
@@ -163,6 +163,7 @@ def default_renderer_base(
     how to render the custom VegaLite MIME type listed above.
     """
     from altair.vegalite.v5.display import VEGA_MIME_TYPE, VEGALITE_MIME_TYPE
+
     assert isinstance(spec, dict)
     bundle: Dict[str, Union[str, dict]] = {}
     metadata: Dict[str, Dict[str, Any]] = {}
--- a/altair/utils/display.py
+++ b/altair/utils/display.py
@@ -4,6 +4,7 @@ import textwrap
 from typing import Callable, Dict, Optional, Tuple, Any, Union
 import uuid
 
+from ._vegafusion_data import compile_with_vegafusion, using_vegafusion
 from .plugin_registry import PluginRegistry, PluginEnabler
 from .mimebundle import spec_to_mimebundle
 from .schemapi import validate_jsonschema
@@ -161,9 +162,18 @@ def default_renderer_base(
     This renderer works with modern frontends (JupyterLab, nteract) that know
     how to render the custom VegaLite MIME type listed above.
     """
+    from altair.vegalite.v5.display import VEGA_MIME_TYPE, VEGALITE_MIME_TYPE
     assert isinstance(spec, dict)
     bundle: Dict[str, Union[str, dict]] = {}
     metadata: Dict[str, Dict[str, Any]] = {}
+
+    if using_vegafusion():
+        spec = compile_with_vegafusion(spec)
+
+        # Swap mimetype from Vega-Lite to Vega.
+        # If mimetype was JSON, leave it alone
+        if mime_type == VEGALITE_MIME_TYPE:
+            mime_type = VEGA_MIME_TYPE
 
     bundle[mime_type] = spec
     bundle["text/plain"] = str_repr
--- a/altair/utils/display.py
+++ b/altair/utils/display.py
@@ -162,6 +162,7 @@ def default_renderer_base(
     This renderer works with modern frontends (JupyterLab, nteract) that know
     how to render the custom VegaLite MIME type listed above.
     """
+    # Local import to avoid circular ImportError
     from altair.vegalite.v5.display import VEGA_MIME_TYPE, VEGALITE_MIME_TYPE
 
     assert isinstance(spec, dict)
--- a/altair/utils/mimebundle.py
+++ b/altair/utils/mimebundle.py
@@ -44,6 +44,7 @@ def spec_to_mimebundle(
     ----
     The png, svg, pdf, and vega outputs require the altair_saver package
     """
+    # Local import to avoid circular ImportError
     from altair.utils.display import compile_with_vegafusion, using_vegafusion
 
     if mode != "vega-lite":
--- a/altair/utils/mimebundle.py
+++ b/altair/utils/mimebundle.py
@@ -45,6 +45,7 @@ def spec_to_mimebundle(
     The png, svg, pdf, and vega outputs require the altair_saver package
     """
     from altair.utils.display import compile_with_vegafusion, using_vegafusion
+
     if mode != "vega-lite":
         raise ValueError("mode must be 'vega-lite'")
 
--- a/altair/utils/mimebundle.py
+++ b/altair/utils/mimebundle.py
@@ -124,13 +124,13 @@ def _spec_to_mimebundle_with_engine(spec, format, mode, **kwargs):
             if mode == "vega":
                 png = vlc.vega_to_png(
                     spec,
-                    scale=kwargs.get("scale_factor", 1.0),
+                    scale=kwargs.get("scale_factor", 1),
                 )
             else:
                 png = vlc.vegalite_to_png(
                     spec,
                     vl_version=vl_version,
-                    scale=kwargs.get("scale_factor", 1.0),
+                    scale=kwargs.get("scale_factor", 1),
                 )
             return {"image/png": png}
         else:
--- a/altair/utils/mimebundle.py
+++ b/altair/utils/mimebundle.py
@@ -44,8 +44,13 @@ def spec_to_mimebundle(
     ----
     The png, svg, pdf, and vega outputs require the altair_saver package
     """
+    from altair.utils.display import compile_with_vegafusion, using_vegafusion
     if mode != "vega-lite":
         raise ValueError("mode must be 'vega-lite'")
+
+    if using_vegafusion():
+        spec = compile_with_vegafusion(spec)
+        mode = "vega"
 
     if format in ["png", "svg", "pdf", "vega"]:
         return _spec_to_mimebundle_with_engine(
@@ -82,7 +87,7 @@ def _spec_to_mimebundle_with_engine(spec, format, mode, **kwargs):
         a dictionary representing a vega-lite plot spec
     format : string {'png', 'svg', 'pdf', 'vega'}
         the format of the mimebundle to be returned
-    mode : string {'vega-lite'}
+    mode : string {'vega-lite', 'vega'}
         The rendering mode.
     engine: string {'vl-convert', 'altair_saver'}
         the conversion engine to use
@@ -102,17 +107,29 @@ def _spec_to_mimebundle_with_engine(spec, format, mode, **kwargs):
         # from SCHEMA_VERSION (of the form 'v5.2.0')
         vl_version = "_".join(SCHEMA_VERSION.split(".")[:2])
         if format == "vega":
-            vg = vlc.vegalite_to_vega(spec, vl_version=vl_version)
+            if mode == "vega":
+                vg = spec
+            else:
+                vg = vlc.vegalite_to_vega(spec, vl_version=vl_version)
             return {"application/vnd.vega.v5+json": vg}
         elif format == "svg":
-            svg = vlc.vegalite_to_svg(spec, vl_version=vl_version)
+            if mode == "vega":
+                svg = vlc.vega_to_svg(spec)
+            else:
+                svg = vlc.vegalite_to_svg(spec, vl_version=vl_version)
             return {"image/svg+xml": svg}
         elif format == "png":
-            png = vlc.vegalite_to_png(
-                spec,
-                vl_version=vl_version,
-                scale=kwargs.get("scale_factor", 1.0),
-            )
+            if mode == "vega":
+                png = vlc.vega_to_png(
+                    spec,
+                    scale=kwargs.get("scale_factor", 1.0),
+                )
+            else:
+                png = vlc.vegalite_to_png(
+                    spec,
+                    vl_version=vl_version,
+                    scale=kwargs.get("scale_factor", 1.0),
+                )
             return {"image/png": png}
         else:
             # This should be validated above
--- a/altair/utils/save.py
+++ b/altair/utils/save.py
@@ -153,7 +153,7 @@ def save(
             mimebundle = spec_to_mimebundle(
                 spec=spec,
                 format=format,
-                mode=mode,
+                mode=inner_mode,
                 vega_version=vega_version,
                 vegalite_version=vegalite_version,
                 vegaembed_version=vegaembed_version,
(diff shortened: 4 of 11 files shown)


altair/utils/_vegafusion_data.py
```diff
@@ -0,0 +1,189 @@
...
+# Temporary storage for dataframes that have been extracted
+# from charts by the vegafusion data transformer. Use a WeakValueDictionary
+# rather than a dict so that the Python interpreter is free to garbage
+# collect the stored DataFrames.
+extracted_inline_tables: MutableMapping[str, _DataFrameLike] = WeakValueDictionary()
+
+# Special URL prefix that VegaFusion uses to denote that a
+# dataset in a Vega spec corresponds to an entry in the `inline_datasets`
+# kwarg of vf.runtime.pre_transform_spec().
+VEGAFUSION_PREFIX: Final = "table://"
```
mattijn, line 26: maybe rename this to `vf_source://`? Since you also prefix a DataFrame as `table_<x>`, [here](https://github.com/altair-viz/altair/pull/3094/commits/07f8081ee70ec756beeed02cd1d423f62bc9dc58#diff-4789a42d9187ced48479b5409346321a0d110b18d5163aea482d74557681b52aR37). This will keep space for the support for arrays/tensors/matrices once they will be introduced.

mattijn (commented):
Thanks @jonmmease! Without testing, one quick inline comment.

altair/utils/_vegafusion_data.py
jonmmease: This prefix is build into VegaFusion. But there is an alternative prefix supported that I can switch to `vegafusion+dataset://`.

jonmmease: renamed in [3a5517b](https://github.com/altair-viz/altair/pull/3094/commits/3a5517b55da95a7749e15973031b3e2ec38bcbd0)

mattijn:
Thanks @jonmmease! I've read the code-diff a few times and it looks solid to me. 

I've tested a bit and only have a few observations. maybe interesting for documenting purposes. I don't think these are bugs.
1. I can update `vegafusion-python-embed` independent from `vegafusion` (during installation). Is this intended? I've the feeling these versions should be in sync with each other, but not sure.
2. A roundtrip with a pandas dataframe as input including a column with datetime objects (without timezones connected) will have timezones connected upon doing `chart.transformed_data()`. 
3. As far I can see, this does not happen if the input is a pyarrow table. Not sure if (my installed version of) pyarrow support timezones.
4. `chart.transformed_data()` also works without `alt.data_transformers.enable("vegafusion")` activated.
5. If vegafusion is activated and I click 'Open in Vega Editor' I see the data inlined. But if I do `chart.to_dict()` I see it as such `"url": "vegafusion+dataset://table_72f60cdf_3ede_47cf_9eb6_53f65059a9f5"`.
6. The generated JSON code in the vega-editor is Vega JSON and not Vega-Lite. The order of attributes is not fixed upon re-rendering.
7. Can I force inline the vegafusion-generated data in python? Return `"values": []` instead of `"url": "vegafusion+dataset://"`?

Tests I did were similar to:
```python
from datetime import datetime
import pandas as pd
import pyarrow as pa
import altair as alt

alt.data_transformers.enable("vegafusion")
#alt.data_transformers.enable("default")

data = {    
    'date': [datetime(2004, 8, 30, 23, 15), datetime(2004, 9, 1, 12, 10)],
    'value': [102, 129]
 }
pd_df = pd.DataFrame(data)

pa_table = pa.table(data)
```
```python
# pandas dataframe
c = alt.Chart(pd_df).mark_point().encode(x='date:T', y='value:Q')
c.transformed_data()
```

> <html>
> <body>
> 
> 
>  date | value
>  -- | --
> 2004-08-30 23:15:00+02:00 | 102
> 2004-09-01 12:10:00+02:00 | 129
> 
> 
> </body>
> </html>

```python
c.to_dict()
```

> ```python
> {'config': {'view': {'continuousWidth': 300, 'continuousHeight': 300}},
>  'data': {'url': 'vegafusion+dataset://table_27a58b61_c93c_44ae_8476_31212b82f513'},
>  'mark': {'type': 'point'},
>  'encoding': {'x': {'field': 'date', 'type': 'temporal'},
>   'y': {'field': 'value', 'type': 'quantitative'}},
>  '$schema': 'https://vega.github.io/schema/vega-lite/v5.8.0.json'}
> ```

```python
c.to_dict(format='vega')
```

> ```python
> {'$schema': 'https://vega.github.io/schema/vega/v5.json',
>  'background': 'white',
>  'padding': 5,
>  'width': 300,
>  'height': 300,
>  'style': 'cell',
>  'data': [{'name': 'source_0',
>    'url': 'vegafusion+dataset://table_9d518c4b_c318_4e1f_b3fa_82d57b90275c',
>    'format': {'type': 'json', 'parse': {'date': 'date'}},
>    'transform': [{'type': 'filter',
>      'expr': '(isDate(datum["date"]) || (isValid(datum["date"]) && isFinite(+datum["date"]))) && isValid(datum["value"]) && isFinite(+datum["value"])'}]}],
>  'marks': [{'name': 'marks',
>    'type': 'symbol',
>    'style': ['point'],
>    'from': {'data': 'source_0'},
>    'encode': {'update': {'opacity': {'value': 0.7},
>      'fill': {'value': 'transparent'},
>      'stroke': {'value': '#4c78a8'},
>      'ariaRoleDescription': {'value': 'point'},
>      'description': {'signal': '"date: " + (timeFormat(datum["date"], \'%b %d, %Y\')) + "; value: " + (format(datum["value"], ""))'},
>      'x': {'scale': 'x', 'field': 'date'},
>      'y': {'scale': 'y', 'field': 'value'}}}}],
>  'scales': [{'name': 'x',
>    'type': 'time',
>    'domain': {'data': 'source_0', 'field': 'date'},
>    'range': [0, {'signal': 'width'}]},
>   {'name': 'y',
>    'type': 'linear',
>    'domain': {'data': 'source_0', 'field': 'value'},
>    'range': [{'signal': 'height'}, 0],
>    'nice': True,
>    'zero': True}],
>  'axes': [{'scale': 'x',
>    'orient': 'bottom',
>    'gridScale': 'y',
>    'grid': True,
>    'tickCount': {'signal': 'ceil(width/40)'},
>    'domain': False,
>    'labels': False,
>    'aria': False,
>    'maxExtent': 0,
>    'minExtent': 0,
>    'ticks': False,
>    'zindex': 0},
>   {'scale': 'y',
>    'orient': 'left',
>    'gridScale': 'x',
>    'grid': True,
>    'tickCount': {'signal': 'ceil(height/40)'},
>    'domain': False,
>    'labels': False,
>    'aria': False,
>    'maxExtent': 0,
>    'minExtent': 0,
>    'ticks': False,
>    'zindex': 0},
>   {'scale': 'x',
>    'orient': 'bottom',
>    'grid': False,
>    'title': 'date',
>    'labelFlush': True,
>    'labelOverlap': True,
>    'tickCount': {'signal': 'ceil(width/40)'},
>    'zindex': 0},
>   {'scale': 'y',
>    'orient': 'left',
>    'grid': False,
>    'title': 'value',
>    'labelOverlap': True,
>    'tickCount': {'signal': 'ceil(height/40)'},
>    'zindex': 0}]}
> ```

```python
c.save('c.png')  # success
```
```python
print(c.to_json())
```
> ```json
> {
>   "$schema": "https://vega.github.io/schema/vega-lite/v5.8.0.json",
>   "config": {
>     "view": {
>       "continuousHeight": 300,
>       "continuousWidth": 300
>     }
>   },
>   "data": {
>     "url": "vegafusion+dataset://table_6ca2f4b0_90aa_4f17_b792_d7943bbae44b"
>   },
>   "encoding": {
>     "x": {
>       "field": "date",
>       "type": "temporal"
>     },
>     "y": {
>       "field": "value",
>       "type": "quantitative"
>     }
>   },
>   "mark": {
>     "type": "point"
>   }
> }
> ```

```python
# pyarrow table
d = alt.Chart(pa_table).mark_point(tooltip=True).encode(x='date:T', y='value:Q')
d.transformed_data()
```

> ```python
> pyarrow.Table
> date: timestamp[ms]
> value: int64
> ----
> date: [[2004-08-30 21:15:00.000,2004-09-01 10:10:00.000]]
> value: [[102,129]]
> ```

```python
d.to_dict()
```

> ```python
> {'config': {'view': {'continuousWidth': 300, 'continuousHeight': 300}},
>  'data': {'url': 'vegafusion+dataset://table_5d35f0b9_8f6c_4c68_a5dd_8a5d0cf5213d'},
>  'mark': {'type': 'point', 'tooltip': True},
>  'encoding': {'x': {'field': 'date', 'type': 'temporal'},
>   'y': {'field': 'value', 'type': 'quantitative'}},
>  '$schema': 'https://vega.github.io/schema/vega-lite/v5.8.0.json'}
> ```python

```python
d.save('d.png')  # success
```
```python
print(d.to_json())  # print(d.to_json(format='vega'))
```

> ```json
> {
>   "$schema": "https://vega.github.io/schema/vega-lite/v5.8.0.json",
>   "config": {
>     "view": {
>       "continuousHeight": 300,
>       "continuousWidth": 300
>     }
>   },
>   "data": {
>     "url": "vegafusion+dataset://table_2412a125_3419_42ff_a469_9644fcb3a537"
>   },
>   "encoding": {
>     "x": {
>       "field": "date",
>       "type": "temporal"
>     },
>     "y": {
>       "field": "value",
>       "type": "quantitative"
>     }
>   },
>   "mark": {
>     "tooltip": true,
>     "type": "point"
>   }
> }
> ```

```python
# large table
alt.data_transformers.enable("vegafusion")
#alt.data_transformers.enable("default")

flights = pd.read_parquet(
    "https://vegafusion-datasets.s3.amazonaws.com/vega/flights_1m.parquet"
)

f = alt.Chart(flights).mark_bar().encode(
    alt.X("delay", bin=alt.Bin(maxbins=30)),
    alt.Y("count()")
)
```
```python
f.save('f.png')  # success
# print(f.to_json())
# f
f.transformed_data().head()
```
<html>
<body>


 bin_maxbins_30_delay | bin_maxbins_30_delay_end | __count
 -- | -- | --
0.0 | 20.0 | 392700
-20.0 | 0.0 | 419400
20.0 | 40.0 | 92700
140.0 | 160.0 | 2000
40.0 | 60.0 | 38400


</body>
</html>

mattijn (approved):

jonmmease:
Thanks for taking the time to go through this @mattijn! I'll respond in more detail tomorrow, but wanted to mention one thing. 

I debated about whether to change the behavior of `to_dict` and `to_json`. One possibility I considered was the apply the VegaFusion pre-transform logic when `format="vega"`, but not when `format=vega-lite" (default). But this seemed like it might end up being confusing though. Let me know if you have any thoughts.

Also, I want to wait until @binste is back and has a chance to look at this before merging.

jonmmease:
> 1. I can update `vegafusion-python-embed` independent from `vegafusion` (during installation). Is this intended? I've the feeling these versions should be in sync with each other, but not sure.

Yeah, I'm not sure of the best way to handle this. `vegafusion-python-embed` is technically an optional dependency because it's possible to use a subset of the functionality of the `vegafusion` package by connecting it to an instance of VegaFusion server over gRPC. This mode isn't full featured, or documented well, but the idea is to make it possible to use VegaFusion in architectures that aren't supported by connecting to an instance of VegaFusion server running somewhere else.

When you install with `vegafusion[embed]`, the versions are pinned exactly to match.  But I'm not sure if there's a way to get pip to constrain an optional dependencies version. Let me know if you have any ideas!

> 2. A roundtrip with a pandas dataframe as input including a column with datetime objects (without timezones connected) will have timezones connected upon doing `chart.transformed_data()`. 

This is expected (and needs to be documented). Vega-Lite and VegaFusion treat everything as UTC internally, so on output we pretty much need to assign some timezone.  I chose to have the output timezone match the local timezone that's used for timeunit calculations (this defaults to the kernel's local timezone but can be overridden with `vf.set_local_tz()`.

> 3. As far I can see, this does not happen if the input is a pyarrow table. Not sure if (my installed version of) pyarrow support timezones.

This conversion to the local timezone works with pandas as polars, but I did see a way to perform the conversion with pyarrow. There might be a way that I didn't find.

4. `chart.transformed_data()` also works without `alt.data_transformers.enable("vegafusion")` activated.

This is expected

6. If vegafusion is activated and I click 'Open in Vega Editor' I see the data inlined. But if I do `chart.to_dict()` I see it as such `"url": "vegafusion+dataset://table_72f60cdf_3ede_47cf_9eb6_53f65059a9f5"`.

See note above. Because VegaFusion works at the Vega level it's not possible to inline data into a Vega-Lite spec. So there are 3 options when it comes to `to_dict` and `to_json`.
 1. Do nothing, which passes through the "vegafusion+dataset://" urls
 2. Do nothing when `format="vega-lite"` (default), but inline then data when `format="vega"`
 3. Always inline the data and return a Vega spec, ignoring the `format` argument.

I wasn't really happy with (2) or (3), but I agree that (1) is also not great. Happy to consider other alternatives as well.

7. The generated JSON code in the vega-editor is Vega JSON and not Vega-Lite. The order of attributes is not fixed upon re-rendering.

Expected to be Vega. I haven't looked into the ordering of attributes. I'm not certain where this comes from, but there's probably somewhere we could sort the keys.

9. Can I force inline the vegafusion-generated data in python? Return `"values": []` instead of `"url": "vegafusion+dataset://"`?

Right now you can accomplish this with:

```
from altair.utils._vegafusion_data import compile_with_vegafusion
compile_with_vegafusion(chart.to_dict())
```

But it would probably be nice to have a simpler way to do this.

altair/utils/save.py
```diff
@@ -114,6 +114,8 @@ def save(
     **kwargs :
         additional kwargs passed to spec_to_mimebundle.
     """
+    from altair.utils._vegafusion_data import using_vegafusion
```
binste, line 117: Could you add a comment above this line why the import is in the function so that this does not get modified unintentionally later on? Due to circular imports? If the import could also be at the top of the file, I'd prefer that.

altair/utils/mimebundle.py
```diff
@@ -44,9 +44,15 @@ def spec_to_mimebundle(
     ----
     The png, svg, pdf, and vega outputs require the altair_saver package
     """
+    from altair.utils.display import compile_with_vegafusion, using_vegafusion
```
binste, line 47: See my comment on the import in `save.py`

altair/utils/mimebundle.py
```diff
@@ -102,17 +108,29 @@ def _spec_to_mimebundle_with_engine(spec, format, mode, **kwargs):
...
-            vg = vlc.vegalite_to_vega(spec, vl_version=vl_version)
+            if mode == "vega":
+                vg = spec
+            else:
+                vg = vlc.vegalite_to_vega(spec, vl_version=vl_version)
             return {"application/vnd.vega.v5+json": vg}
         elif format == "svg":
-            svg = vlc.vegalite_to_svg(spec, vl_version=vl_version)
+            if mode == "vega":
+                svg = vlc.vega_to_svg(spec)
```
binste, line 118: Is it intentional that the `vega_to_*` functions do not need to take the Vega version as an argument? I'd assume that once Vega 6 would come out, Altair and VLC need to be kept in sync anyway by a user and so they would both be upgraded at the same time and hence `vlc` would pick the correct Vega version again by default. Just want to check if that matches your understanding.

altair/utils/mimebundle.py
```diff
@@ -102,17 +108,29 @@ def _spec_to_mimebundle_with_engine(spec, format, mode, **kwargs):
...
         elif format == "png":
-            png = vlc.vegalite_to_png(
-                spec,
-                vl_version=vl_version,
-                scale=kwargs.get("scale_factor", 1.0),
-            )
+            if mode == "vega":
+                png = vlc.vega_to_png(
+                    spec,
+                    scale=kwargs.get("scale_factor", 1.0),
```
binste, line 126:
```suggestion
                    scale=kwargs.get("scale_factor", 1),
```
Integers should be fine here

altair/utils/mimebundle.py
```diff
@@ -102,17 +108,29 @@ def _spec_to_mimebundle_with_engine(spec, format, mode, **kwargs):
...
+                )
+            else:
+                png = vlc.vegalite_to_png(
+                    spec,
+                    vl_version=vl_version,
+                    scale=kwargs.get("scale_factor", 1.0),
```
binste, line 132:
```suggestion
                    scale=kwargs.get("scale_factor", 1),
```

altair/utils/display.py
```diff
@@ -161,10 +162,20 @@ def default_renderer_base(
     This renderer works with modern frontends (JupyterLab, nteract) that know
     how to render the custom VegaLite MIME type listed above.
     """
+    from altair.vegalite.v5.display import VEGA_MIME_TYPE, VEGALITE_MIME_TYPE
```
binste, line 165: See my comment on imports in `save.py`

altair/utils/_vegafusion_data.py
```diff
@@ -0,0 +1,199 @@
...
+
+from altair.utils.core import _DataFrameLike
+from altair.utils.data import _DataType, _ToValuesReturnType, MaxRowsError
+from altair.vegalite.data import default_data_transformer
+
+# Temporary storage for dataframes that have been extracted
+# from charts by the vegafusion data transformer. Use a WeakValueDictionary
+# rather than a dict so that the Python interpreter is free to garbage
+# collect the stored DataFrames.
+extracted_inline_tables: MutableMapping[str, _DataFrameLike] = WeakValueDictionary()
```
binste, line 21: I didn't know about `WeakValueDictionary`, very cool!

In the [docs](https://docs.python.org/3/library/weakref.html) it says:

> Not all objects can be weakly referenced. Objects which support weak references include class instances, functions written in Python (but not in C), instance methods, sets, frozensets, some [file objects](https://docs.python.org/3/glossary.html#term-file-object), [generators](https://docs.python.org/3/glossary.html#term-generator), type objects, sockets, arrays, deques, regular expression pattern objects, and code objects.

Do you know if Pandas/Polars/PyArrow dataframes all support weak references?

altair/utils/_vegafusion_data.py
```diff
@@ -0,0 +1,199 @@
...
+                "the limit by calling alt.data_transformers.disable_max_rows(). Note that\n"
+                "disabling this limit may cause the browser to freeze or crash."
+            )
+
+    return transformed_vega_spec
+
+
+def using_vegafusion() -> bool:
+    """Check whether the vegafusion data transfomer is enabled"""
+    from altair.vegalite.v5.data import data_transformers
```
binste, line 197: See my comment on imports in `save.py`. Also maybe better to use the top-level `altair.data_transformers` so it's not dependent on the VL version?

altair/utils/_vegafusion_data.py
```diff
@@ -0,0 +1,199 @@
...
+class _ToVegaFusionReturnUrlDict(TypedDict):
+    url: str
+
+
+@curried.curry
+def vegafusion_data_transformer(
+    data: _DataType, max_rows: int = 100000
+) -> Union[_ToVegaFusionReturnUrlDict, _ToValuesReturnType]:
+    """VegaFusion Data Transformer"""
+    if hasattr(data, "__geo_interface__"):
```
binste, line 38: I think this escape hatch for geo data is not yet implemented in the other parts of this PR such as the rendering and saving of charts. The following example from the Altair docs raises a `PanicError`:

```python
import altair as alt
import geopandas as gpd

alt.data_transformers.enable("vegafusion")

url = "https://naciscdn.org/naturalearth/110m/cultural/ne_110m_admin_0_countries.zip"
gdf_ne = gpd.read_file(url)  # zipped shapefile
gdf_ne = gdf_ne[["NAME", "CONTINENT", "POP_EST", 'geometry']]

alt.Chart(gdf_ne).mark_geoshape()
```

Somehow `using_vegafusion` needs to be aware of this distinction as well?

altair/utils/_vegafusion_data.py
```diff
@@ -0,0 +1,199 @@
...
+        # Use default transformer for geo interface objects
+        # # (e.g. a geopandas GeoDataFrame)
+        return default_data_transformer(data)
+    elif hasattr(data, "__dataframe__"):
+        table_name = f"table_{uuid.uuid4()}".replace("-", "_")
+        extracted_inline_tables[table_name] = data
+        return {"url": VEGAFUSION_PREFIX + table_name}
+    else:
+        # Use default transformer if we don't recognize data type
```
binste, line 47: Same comment about escape hatch as for geo data.

altair/utils/_vegafusion_data.py
```diff
@@ -0,0 +1,199 @@
...
+    vegalite_spec: dict
+        A Vega-Lite spec that was generated from an Altair chart with
+        the "vegafusion" data transformer enabled
+
+    Returns
+    -------
+    dict
+        A Vega spec that has been pre-transformed by VegaFusion
+    """
+    from altair import vegalite_compilers, data_transformers
```
binste, line 149: See my comment on imports in `save.py`

altair/utils/_vegafusion_data.py
```diff
@@ -0,0 +1,199 @@
...
+        vega_spec,
+        vf.get_local_tz(),
+        inline_datasets=inline_tables,
+        row_limit=row_limit,
+    )
+
+    # Check from row limit warning and convert to MaxRowsError
+    for warning in warnings:
+        if warning.get("type") == "RowLimitExceeded":
+            raise MaxRowsError(
```
binste, line 184: I like the increased row limit for this transformer! I'll open a separate issue where we can discuss if we in general want to raise the limit somewhat as devices/frontends have become more powerful.

binste (changes requested):
I really like the general approach of integrating and enabling VegaFusion via a data transformer and the implementation in this PR! 🥳  I added some comments inline.

Regarding the to_dict/to_json discussion, I also think this should produce something usable but I see that it's not straightforward. What do you think about, if the vegafusion transformer is called, to:

* raising an exception if `to_dict/to_json` are called without `format` specified (default `vega-lite` or with `format="vega-lite"`
* Mention in error that only `format="vega"` is supported for this transformer

This would make it clear to a user that with the `vegafusion` transformer a `vega-lite` spec is not possible and they can explicitly opt in to receive a `vega` spec in case they want to send the spec to e.g. a frontend in a web application.

I'd be ok if this is implemented in a follow-up PR and we could also continue the discussion there or in the existing discussion thread on the VegaFusion integration.

Your ideas on the documentation sound good to me.

pyproject.toml
```diff
@@ -66,7 +66,7 @@ dev = [
     "vega_datasets",
     # following is rejected by PyPI
     "altair_viewer @ git+https://github.com/altair-viz/altair_viewer.git",
-    "vl-convert-python",
+    "vl-convert-python==0.10.3",
```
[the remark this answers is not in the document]
jonmmease: This is pinned to prevent update to 0.11.1 due to https://github.com/vega/vl-convert/issues/78.  I'll remove it one I get the fix out in 0.11.2

altair/utils/_vegafusion_data.py
jonmmease: I believe the "(but not in C)" qualifier is related to native functions, not instances of Python classes that include native implementation.

Here's an example of this working with pandas and polars:

```python
>>> import pandas as pd
>>> import polars as pl
>>> from weakref import WeakValueDictionary
>>> dfs = dict(df1=pd.DataFrame({"a": [1, 2, 3]}), df2=pl.DataFrame({"a": [1, 2, 3]}))
>>> weak_dfs = WeakValueDictionary(df1=dfs["df1"])
>>> list(weak_dfs.keys())
['df1', 'df2']
>>> del dfs["df1"]
>>> list(weak_dfs.keys())
['df2']
>>> del dfs["df2"]
>>> list(weak_dfs.keys())
[]
```

tests/utils/test_mimebundle.py
```diff
@@ -221,3 +221,26 @@ def test_spec_to_json_mimebundle():
...
+
+def test_vegafusion_spec_to_vega_mime_bundle(vegalite_spec):
+    with alt.data_transformers.enable("vegafusion"):
+        bundle = spec_to_mimebundle(
+            spec=vegalite_spec,
+            mode="vega-lite",
+            format="vega",
+        )
+        # Returned bundle will be vega
+        vega_spec = bundle["application/vnd.vega.v5+json"]
```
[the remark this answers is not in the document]
jonmmease: Done in 85fa7b49e611ef07f39203a727319b9c7a69bb3f

altair/utils/save.py
jonmmease: Done in 6e4c8c4fc3d93e8cbd9e296ad79c11b957729eeb

altair/utils/mimebundle.py
jonmmease: Done in 6e4c8c4fc3d93e8cbd9e296ad79c11b957729eeb

altair/utils/display.py
jonmmease: Done in 6e4c8c4fc3d93e8cbd9e296ad79c11b957729eeb

altair/utils/_vegafusion_data.py
jonmmease: Done in 6e4c8c4fc3d93e8cbd9e296ad79c11b957729eeb

altair/utils/_vegafusion_data.py
jonmmease: Done in 6e4c8c4fc3d93e8cbd9e296ad79c11b957729eeb

altair/utils/mimebundle.py
jonmmease: Done in da1ac3f3eaded57a675f4cd3a2bc55f62296c638

altair/utils/mimebundle.py
jonmmease: Done in da1ac3f3eaded57a675f4cd3a2bc55f62296c638

altair/utils/_vegafusion_data.py
jonmmease: Thanks for catching and flagging this. The issue here is that the JSON-to-Arrow logic VegaFusion uses is choking on the inline geojson data. I think VegaFusion needs to be better about gracefully passing through complex JSON that Arrow fails to parse.

jonmmease: Tracking in https://github.com/hex-inc/vegafusion/issues/352 and will be fixed in VegaFusion 1.4.0-rc2

altair/utils/mimebundle.py
binste: Makes sense, thanks!

altair/utils/_vegafusion_data.py
binste: Thanks for checking! Looks good to me as well.

altair/utils/_vegafusion_data.py
binste: Great to hear, thanks!

altair/utils/_vegafusion_data.py
jonmmease: I think this should work fine now with `vegafusion>=1.4.0rc2`. If you have a chance, could you try that out from your end?

jonmmease:
@binste, I updated `to_dict`/`to_json` to apply pre-transformation when "vegafusion" is enabled and `format="vega"` and to raise when `format="vega-lite"`. I also used the `context` argument to `to_dict` to disable for internal usage. See what you think!

altair/utils/_vegafusion_data.py
binste: Can confirm that the example I posted now works with 1.4.0rc2.

binste (changes requested):
I like the implementation with the `ValueError` and using `context`! I think it needs to be extended on line 903 in `api.py` (can't comment there) as it now raises this error for a layered chart:

```python
import altair as alt

alt.data_transformers.enable("vegafusion")

source = data.wheat()

    x='year:O',
    y='wheat:Q'
)

rule = alt.Chart(source).mark_rule(color='red').encode(
    y='mean(wheat):Q'
)

layer_chart = (bar + rule).properties(width=600)
layer_chart.to_json(format="vega")
```

```python
Cell In[3], line 1
----> 1 layer_chart.to_dict(format="vega")

File [~/Library/CloudStorage/Dropbox/Programming/altair/altair/vegalite/v5/api.py:903](https://file+.vscode-resource.vscode-cdn.net/Users/stefanbinder/Library/CloudStorage/Dropbox/Programming/altair/~/Library/CloudStorage/Dropbox/Programming/altair/altair/vegalite/v5/api.py:903), in TopLevelMixin.to_dict(self, validate, format, ignore, context)
    898 context["top_level"] = False
    900 # TopLevelMixin instance does not necessarily have to_dict defined
    901 # but due to how Altair is set up this should hold.
    902 # Too complex to type hint right now
--> 903 vegalite_spec = super(TopLevelMixin, copy).to_dict(  # type: ignore[misc]
    904     validate=validate, ignore=ignore, context=context
    905 )
    907 # TODO: following entries are added after validation. Should they be validated?
    908 if is_top_level:
    909     # since this is top-level we add $schema if it's missing

File [~/Library/CloudStorage/Dropbox/Programming/altair/altair/utils/schemapi.py:807](https://file+.vscode-resource.vscode-cdn.net/Users/stefanbinder/Library/CloudStorage/Dropbox/Programming/altair/~/Library/CloudStorage/Dropbox/Programming/altair/altair/utils/schemapi.py:807), in SchemaBase.to_dict(self, validate, ignore, context)
    805     if "mark" in kwds and isinstance(kwds["mark"], str):
    806         kwds["mark"] = {"type": kwds["mark"]}
--> 807     result = _todict(
    808         kwds,
    809         context=context,
    810     )
    811 else:
    812     raise ValueError(
    813         "{} instance has both a value and properties : "
    814         "cannot serialize to dict".format(self.__class__)
    815     )

File [~/Library/CloudStorage/Dropbox/Programming/altair/altair/utils/schemapi.py:340](https://file+.vscode-resource.vscode-cdn.net/Users/stefanbinder/Library/CloudStorage/Dropbox/Programming/altair/~/Library/CloudStorage/Dropbox/Programming/altair/altair/utils/schemapi.py:340), in _todict(obj, context)
    338     return [_todict(v, context) for v in obj]
    339 elif isinstance(obj, dict):
--> 340     return {k: _todict(v, context) for k, v in obj.items() if v is not Undefined}
    341 elif hasattr(obj, "to_dict"):
    342     return obj.to_dict()

File [~/Library/CloudStorage/Dropbox/Programming/altair/altair/utils/schemapi.py:340](https://file+.vscode-resource.vscode-cdn.net/Users/stefanbinder/Library/CloudStorage/Dropbox/Programming/altair/~/Library/CloudStorage/Dropbox/Programming/altair/altair/utils/schemapi.py:340), in (.0)
    338     return [_todict(v, context) for v in obj]
    339 elif isinstance(obj, dict):
--> 340     return {k: _todict(v, context) for k, v in obj.items() if v is not Undefined}
    341 elif hasattr(obj, "to_dict"):
    342     return obj.to_dict()

File [~/Library/CloudStorage/Dropbox/Programming/altair/altair/utils/schemapi.py:338](https://file+.vscode-resource.vscode-cdn.net/Users/stefanbinder/Library/CloudStorage/Dropbox/Programming/altair/~/Library/CloudStorage/Dropbox/Programming/altair/altair/utils/schemapi.py:338), in _todict(obj, context)
    336     return obj.to_dict(validate=False, context=context)
    337 elif isinstance(obj, (list, tuple, np.ndarray)):
--> 338     return [_todict(v, context) for v in obj]
    339 elif isinstance(obj, dict):
    340     return {k: _todict(v, context) for k, v in obj.items() if v is not Undefined}

File [~/Library/CloudStorage/Dropbox/Programming/altair/altair/utils/schemapi.py:338](https://file+.vscode-resource.vscode-cdn.net/Users/stefanbinder/Library/CloudStorage/Dropbox/Programming/altair/~/Library/CloudStorage/Dropbox/Programming/altair/altair/utils/schemapi.py:338), in (.0)
    336     return obj.to_dict(validate=False, context=context)
    337 elif isinstance(obj, (list, tuple, np.ndarray)):
--> 338     return [_todict(v, context) for v in obj]
    339 elif isinstance(obj, dict):
    340     return {k: _todict(v, context) for k, v in obj.items() if v is not Undefined}

File [~/Library/CloudStorage/Dropbox/Programming/altair/altair/utils/schemapi.py:336](https://file+.vscode-resource.vscode-cdn.net/Users/stefanbinder/Library/CloudStorage/Dropbox/Programming/altair/~/Library/CloudStorage/Dropbox/Programming/altair/altair/utils/schemapi.py:336), in _todict(obj, context)
    334 """Convert an object to a dict representation."""
    335 if isinstance(obj, SchemaBase):
--> 336     return obj.to_dict(validate=False, context=context)
    337 elif isinstance(obj, (list, tuple, np.ndarray)):
    338     return [_todict(v, context) for v in obj]

File [~/Library/CloudStorage/Dropbox/Programming/altair/altair/vegalite/v5/api.py:2677](https://file+.vscode-resource.vscode-cdn.net/Users/stefanbinder/Library/CloudStorage/Dropbox/Programming/altair/~/Library/CloudStorage/Dropbox/Programming/altair/altair/vegalite/v5/api.py:2677), in Chart.to_dict(self, validate, format, ignore, context)
   2673     copy.data = core.InlineData(values=[{}])
   2674     return super(Chart, copy).to_dict(
   2675         validate=validate, format=format, ignore=ignore, context=context
   2676     )
-> 2677 return super().to_dict(
   2678     validate=validate, format=format, ignore=ignore, context=context
   2679 )

File [~/Library/CloudStorage/Dropbox/Programming/altair/altair/vegalite/v5/api.py:926](https://file+.vscode-resource.vscode-cdn.net/Users/stefanbinder/Library/CloudStorage/Dropbox/Programming/altair/~/Library/CloudStorage/Dropbox/Programming/altair/altair/vegalite/v5/api.py:926), in TopLevelMixin.to_dict(self, validate, format, ignore, context)
    924 if context.get("pre_transform", True) and _using_vegafusion():
    925     if format == "vega-lite":
--> 926         raise ValueError(
    927             'When the "vegafusion" data transformer is enabled, the \n'
    928             "to_dict() and to_json() chart methods must be called with "
    929             'format="vega". \n'
    930             "For example: \n"
    931             '    >>> chart.to_dict(format="vega")\n'
    932             '    >>> chart.to_json(format="vega")'
    933         )
    934     else:
    935         return _compile_with_vegafusion(vegalite_spec)

ValueError: When the "vegafusion" data transformer is enabled, the 
to_dict() and to_json() chart methods must be called with format="vega". 
For example: 
    >>> chart.to_dict(format="vega")
    >>> chart.to_json(format="vega")
```

binste (approved):
Works great! Looking forward to the next Altair release where we can include this. 🥳  Whereas originally I wasn't sure how to integrate VegaFusion without making it confusing for users, I think this solution now is a big usability improvement and ties in well with the existing concepts of Altair (data_transformers, to_dict, ...).

jonmmease:
Thanks for the reviews @mattijn and @binste! Merging this now, and I'll work on updating the documentation next week.

PR #3100, opened 2023-07-09
Drop support for Python 3.7 which is end-of-life

binste:
Python 3.7 is [end-of-life as of 27.06.2023](https://devguide.python.org/versions/). Many other packages (e.g. [Streamlit](https://docs.streamlit.io/library/changelog), [python-jsonschema](https://github.com/python-jsonschema/jsonschema/releases/tag/v4.18.0), ...) have already dropped support for it in minor releases and I'd recommend to do the same to simplify maintenance. Users with 3.7 can still install Altair 5.0.1 which has been a rather stable release so far.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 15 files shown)


mattijn (approved):
Looks good @binste!

PR #3116, opened 2023-07-22
 copied  percentage calculation in example

thoemi123:
I added @jonmmease 's implementation of the percentage calculation (https://github.com/altair-viz/altair/issues/3101) to the [diverging stacked bar chart example](https://altair-viz.github.io/gallery/diverging_stacked_bar_chart.html).

It was mentioned in  (https://github.com/altair-viz/altair/discussions/3103) that this would be helpful.

The example now slightly differs from the [Vega-Lite example](https://vega.github.io/vega-lite-v2/examples/bar_diverging_stacked.html) since there is a small error in the hard-coded data provided in the Vega-Lite example (percentage should not be 0 when there are 11 answers): 
```
{
        "question": "Question 2",
        "type": "Strongly agree",
        "value": 11,
        "percentage": 0,
        "percentage_start": 63.6,
        "percentage_end": 63.6
      }
```
@binste would be great if you can again test run the documentation on your machine since it did not work properly on the github code-space.

### DIFF
--- a/tests/examples_arguments_syntax/diverging_stacked_bar_chart.py
+++ b/tests/examples_arguments_syntax/diverging_stacked_bar_chart.py
@@ -5,336 +5,248 @@ This example shows a diverging stacked bar chart for sentiments towards a set of
 """
 # category: bar charts
 import altair as alt
+import pandas as pd
 
-source = alt.pd.DataFrame([
-      {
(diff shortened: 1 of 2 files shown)


jonmmease:
Thanks a lot @thoemi123!

PR #3121, opened 2023-07-23
Fix contributing descriptions

thomend:
This PR will resolve https://github.com/altair-viz/altair/issues/3014


joelostblom:
Thanks for the update, I would be OK merging this as is without adding guidelines for two dataframes

binste:
Thanks @thomend! How about adding something like "If the example requires multiple dataframes then this does not apply." at the end of the bullet point which mentions 'sources'? On mobile right now so I cant link to the exact location.

thomend:
> Thanks @thomend! How about adding something like "If the example requires multiple dataframes then this does not apply." at the end of the bullet point which mentions 'sources'? On mobile right now so I cant link to the exact location.

Thanks for the input, @binste, I added this in the contributing guidelines and marked the PR as ready for review.

binste (approved):
Looks good, thanks @thomend!

PR #3124, opened 2023-07-25
Add a code of conduct

joelostblom:
As discussed during our call today, this PR adds a Code of Conduct. I suggest we use the [Contributor Covenant's](https://www.contributor-covenant.org/) code of conduct which is a popular CoC used in many projects including the Vega and [Vega-Lite repos](https://github.com/vega/.github/blob/master/CODE_OF_CONDUCT.md). The only difference is that I took the latest version (2.1) instead of the version used in Vega-Lite (1.4). [All the changes between the two versions can be seen here](https://github.com/EthicalSource/contributor_covenant/blob/release/changelog), the biggest one being the detailed section of how to enforce violations.

I put down the same email address as in the Vega and Vega-Lite CoC's for reporting violations which I think makes sense since these projects are so tightly linked, but @domoritz let me know if you prefer that we use a different email.

Maybe this would also be a good time to discuss changing the default branch from `master` to `main` to avoid any negative associations with the word "master" and to bring Altair in line with Vega and Vega-Lite in this regard as well?

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,132 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual
+identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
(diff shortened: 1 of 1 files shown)


jonmmease:
Thanks for opening the PR, @joelostblom. I want to read through in more detail before hitting "approve", but this makes a lot of sense. I listed all of the current maintainers as reviewers for the PR, as it would be good to get everyone's affirmative approval.

I'm also totally on board with swapping our dev branch to `main`

domoritz:
Thanks for pushing this forward. I think it would be good to update the Vega one to 2.1 as well to be in sync. I'll make a pull request. I think @jheer or @arvind own the email for reporting violations, actually.

arvind:
I'm happy to add other members of the core Vega/Vega-Lite/Altair developer teams to the conduct email address. We have it setup just to forward incoming emails out to our personal/institutionally-affiliated email addresses, so others in a similar fashion. Perhaps DM me on the Vega Slack, or send me an email if you'd like to be added?

jonmmease (approved):
Looking forward to having this in place, thanks again @joelostblom!

ChristopherDavisUCI (approved):
Thanks for putting this together so quickly @joelostblom 

It seems like a very natural choice and I didn't notice anything concerning in it.

mattijn (approved):

binste (approved):
Looks good to me too, thanks @joelostblom! I'd also like to be on the mailing list, I'll reach out separately via Slack.

Also on board with changing master to main!

jonmmease:
Great! merging

PR #3130, opened 2023-07-31
Fix accidental requirement of Pandas 1.5. Bump minimum Pandas version to 0.25. Run tests with it

binste:
[This import](https://github.com/altair-viz/altair/pull/3114/files#diff-d110e8643c6feff37c22c1a5c7d213b8384177718b8e26edd8e1b8f2ed027a0aR18) introduced in #3114 is not available in Pandas < 1.5 and Altair currently supports Pandas >= 0.18. This PR adds a test run with the lowest supported Pandas version.

@jonmmease Do you know of a good way to make that type check independent of the pandas import? Else, we could fall back on not type-hinting it if necessary. Feel free to push directly to this branch if you have an idea and want to try it out in the GitHub actions workflow.

Pandas 0.18 is from 2016 and we could for sure consider bumping it but I don't think we can go to Pandas 1.5 just yet. Especially not in a minor release of Altair.

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -324,7 +324,13 @@ def sanitize_dataframe(df: pd.DataFrame) -> pd.DataFrame:  # noqa: C901
 
     for col_name, dtype in df.dtypes.items():
         dtype_name = str(dtype)
-        if dtype_name == "string":
+        if dtype_name == "category":
+            # Work around bug in to_json for categorical types in older versions of pandas
+            # https://github.com/pydata/pandas/issues/10778
+            # https://github.com/altair-viz/altair/pull/2170
+            col = df[col_name].astype(object)
+            df[col_name] = col.where(col.notnull(), None)
+        elif dtype_name == "string":
             # dedicated string datatype (since 1.0)
             # https://pandas.pydata.org/pandas-docs/version/1.0.0/whatsnew/v1.0.0.html#dedicated-string-data-type
             col = df[col_name].astype(object)
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -15,7 +15,6 @@ from types import ModuleType
 import jsonschema
 import pandas as pd
 import numpy as np
-from pandas.core.interchange.dataframe_protocol import Column as PandasColumn
 
 from altair.utils.schemapi import SchemaBase
 from altair.utils._dfi_types import Column, DtypeKind, DataFrame as DfiDataFrame
@@ -25,13 +24,16 @@ if sys.version_info >= (3, 10):
 else:
     from typing_extensions import ParamSpec
 
-from typing import Literal, Protocol
+from typing import Literal, Protocol, TYPE_CHECKING
 
 try:
     from pandas.api.types import infer_dtype as _infer_dtype
 except ImportError:
     # Import for pandas < 0.20.0
     from pandas.lib import infer_dtype as _infer_dtype  # type: ignore[no-redef]
+
+if TYPE_CHECKING:
+    from pandas.core.interchange.dataframe_protocol import Column as PandasColumn
 
 _V = TypeVar("_V")
 _P = ParamSpec("_P")
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -632,7 +632,7 @@ def parse_shorthand(
 
 
 def infer_vegalite_type_for_dfi_column(
-    column: Union[Column, PandasColumn],
+    column: Union[Column, "PandasColumn"],
 ) -> Union[_InferredVegaLiteType, Tuple[_InferredVegaLiteType, list]]:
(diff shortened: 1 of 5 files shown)


jonmmease:
Ahh, thanks for flagging this @binste, and for working on testing with older pandas.

I'll look into alternatives for the type signature, it might be enough to wrap the import in [`typing.TYPE_CHECKING`](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#typing-type-checking), although maybe the mypy tests would fail with the older version of pandas in this case.

In terms of pandas versions, I personally think we could bump the pandas requirement up to 1.0 (released 3.5 years ago) in an Altair minor version, but I agree that 1.5 (released less than 1 year ago) would be too aggressive.

jonmmease:
huh, looks like the `if` condition in the GitHub actions isn't working

jonmmease:
I'm really stumped as to why GitHub actions is always evaluating the "Maybe uninstall optional dependencies" and "Maybe install lowest supported Pandas version" jobs, ignoring the `if:` conditions (or always evaluating them to true).

binste:
I think I had the same issue recently over in #3118. I'll give it a try.

binste:
Seems like you're trying it already. In #3118 I got it working for jsonschema with this line https://github.com/altair-viz/altair/pull/3118/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R32

jonmmease:
I'll try that again (I think again 😄 )

jonmmease:
Oh, looks like I needed single quotes. The issue now is with installing pandas 0.18.0 in Python 3.8:

```
Run pip install pandas==0.18
Collecting pandas==0.18
  Downloading pandas-0.18.0.tar.gz (7.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 7.1/7.1 MB 24.2 MB/s eta 0:00:00
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Installing backend dependencies: started
  Installing backend dependencies: finished with status 'done'
  Preparing metadata (pyproject.toml): started
  Preparing metadata (pyproject.toml): finished with status 'error'
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [4 lines of output]
      <string>:39: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
      error in pandas setup command: 'install_requires' must be a string or list of strings containing valid project/version requirement specifiers; Expected end or semicolon (after version specifier)
          pytz >= 20[11](https://github.com/altair-viz/altair/actions/runs/5716874306/job/15489393829?pr=3130#step:6:12)k
               ~~~~~~~^
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.
```

Looks like pandas 0.18.1 didn't ship wheels for Python 3.8. I'll check what the oldest version of pandas is that includes wheels for 3.8 is.

binste:
Thanks for figuring out to use single quotes, weird...

Bumping to minimum version 0.25.3 sounds very reasonable to me too. We can rename this PR and do it all in here. I think   there is some code logic somewhere which is for Pandas 0.18. I'll try to find and remove that one too while we're at it.

I can offer to do the Pandas bump and cleanup  later this week if you prefer to focus on the issue with the type hint.

jonmmease:
Sounds good. I'll take this PR and update the min pandas and numpy versions and fix the type hint issue (I'll need to update these versions to get CI working).  I won't look for any pandas/numpy compatibility code that could be removed, so if you could do a pass on that at some point that would be great!

binste:
Great, I removed some part of the code which is now redundant and clarified another one which we need to keep but for other reasons compared to why it was originally implemented.

@jonmmease Could you do a review of my last commits? Feel free to merge afterwards. Thanks!

altair/utils/core.py
```diff
@@ -42,26 +39,6 @@ def __dataframe__(self, *args, **kwargs) -> DfiDataFrame:
...
-            # pandas < 0.21.0 don't support skipna keyword
-            infer_dtype._supports_skipna = False  # type: ignore[attr-defined]
-        else:
-            infer_dtype._supports_skipna = True  # type: ignore[attr-defined]
-    if infer_dtype._supports_skipna:  # type: ignore[attr-defined]
-        return _infer_dtype(value, skipna=False)
-    else:
-        return _infer_dtype(value)
-
-
```
jonmmease, line 64: 🎉

jonmmease (approved):
Looks good. Thanks!

PR #3133, opened 2023-08-01
MAINT: Use importlib.metadata and packaging instead of deprecated pkg_resources

### LINKED ISSUE #3132: pkg_resources is deprecated
It's used in `import_pyarrow_interchange` in `data.py`. Currently generates this deprecation warning in our CI

```python
../../../../Application Support/hatch/env/virtual/altair/Ac6cCVwn/altair/lib/python3.10/site-packages/pkg_resources/__init__.py:121
  /Users/stefanbinder/Library/Application Support/hatch/env/virtual/altair/Ac6cCVwn/altair/lib/python3.10/site-packages/pkg_resources/__init__.py:121: DeprecationWarning: pkg_resources is deprecated as an API
    warnings.warn("pkg_resources is deprecated as an API", DeprecationWarning)
```
We can replace `pkg_resources` with the `packaging` package and which I introduce in #3118 anyway and `importlib.metadata`. This is also the recommended way to do it in https://setuptools.pypa.io/en/latest/pkg_resources.html. I can look at this.

binste:
Closes #3132

### DIFF
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -3,8 +3,7 @@ import os
 import random
 import hashlib
 import warnings
-from importlib.metadata import version as importlib_version
-from importlib.metadata import PackageNotFoundError
+from importlib.metadata import version as importlib_version, PackageNotFoundError
 from typing import Union, MutableMapping, Optional, Dict, Sequence, TYPE_CHECKING, List
 from types import ModuleType
 
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -3,12 +3,15 @@ import os
 import random
 import hashlib
 import warnings
+from importlib.metadata import version as importlib_version
+from importlib.metadata import PackageNotFoundError
 from typing import Union, MutableMapping, Optional, Dict, Sequence, TYPE_CHECKING, List
 from types import ModuleType
 
 import pandas as pd
 from toolz import curried
 from typing import TypeVar
+from packaging.version import Version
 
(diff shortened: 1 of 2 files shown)


altair/utils/data.py
```diff
@@ -3,12 +3,15 @@
 import random
 import hashlib
 import warnings
+from importlib.metadata import version as importlib_version
+from importlib.metadata import PackageNotFoundError
```
jonmmease, line 7: nit: These could be consolidated to a single import line

```python
from importlib.metadata import version as importlib_version, PackageNotFoundError
```

pyproject.toml
```diff
@@ -21,7 +21,8 @@ dependencies = [
     "jsonschema>=3.0",
     "numpy",
     "pandas>=0.18",
-    "toolz"
+    "toolz",
+    "packaging"
```
jonmmease, line 25: The `Version` class will be really handy going forward. Is `packaging` a pure Python package that should work everywhere?

jonmmease (approved):
One non-blocking nit and one non-blocking question inline. Thanks for taking care of this!

altair/utils/data.py
binste: Was not aware that this is a valid syntax if an import is renamed. Thanks! Done in 81f1146fb36f

pyproject.toml
binste: Exactly. Pure Python package, lightweight with a wheel file being ~48KB, maintained by the Python Packaging Authority.

binste:
Thanks for the review!

PR #3142, opened 2023-08-06
Type hint schemapi.py

binste:
See #2951 for an overview of the type hinting effort. For some methods which construct classes based on their input parameters I took the shortcut of typing them as `Any` for now as I think to make it more specific we would need some more complicated type hints. These are not super relevant for users anyway but happy to take suggestions on how to improve it!

We need to keep the type of `Undefined` as `Any` as long as not all parts of the code are type hinted where `Undefined` is used. Afterwards we can change it to the proper `UndefinedType`.

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -927,7 +927,7 @@ class SchemaBase:
         _wrapper_classes: Optional[Iterable[Type["SchemaBase"]]] = None,
         # Type hints for this method would get rather complicated
         # if we want to provide a more specific return type
-    ) -> Any:
+    ) -> "SchemaBase":
         """Construct class from a dictionary representation
 
         Parameters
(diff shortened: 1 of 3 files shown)


altair/utils/schemapi.py
```diff
@@ -873,12 +915,19 @@ def to_json(
...
     @classmethod
-    def from_dict(cls, dct, validate=True, _wrapper_classes=None):
+    def from_dict(
+        cls,
+        dct: dict,
+        validate: bool = True,
+        _wrapper_classes: Optional[Iterable[Type["SchemaBase"]]] = None,
+        # Type hints for this method would get rather complicated
+        # if we want to provide a more specific return type
+    ) -> Any:
```
jonmmease, line 930: Would `_TSchemaBase` work here? I seems like the returned object should be a `SchemaBase` subclass.

jonmmease (approved):
Looks good. One non-blocking question

altair/utils/schemapi.py
binste: When I used `SchemaBase` as a return type, I got an error for `Chart.from_dict` in `api.py` so I left it at that. But after your comment I revisited this and indeed the `Chart.from_dict` type was wrong :) -> Done in 2732c8f402ea594191168ee89

binste:
Thanks for the review @jonmmease!

PR #3148, opened 2023-08-12
Update Vega-Lite to 5.14.1. Add transform_extent

binste:
So far there were no changes to Altair necessary apart from rerunning the automatic code generation. For the tests to pass, we need a new version of `vl-convert` with support for VL 5.14.1. @jonmmease Would that work for you? I think even if a new VL version comes out in the next 1-2 weeks, although I'm not aware of one, we could still stick to VL 5.14.1 for Altair 5.1.

See https://github.com/altair-viz/altair/discussions/3145#discussioncomment-6693697 for the related release discussion.


jonmmease:
Thanks for working on this @binste! Yes, I'll get a vl-convert update out shortly

jonmmease:
Vega-Lite 5.14.1 added to `vl-convert-python` in version 0.12.0 (just released)

altair/vegalite/v5/schema/mixins.py
```diff
@@ -1275,6 +1277,14 @@ def configure_title(self, *args, **kwargs) -> Self:
         copy.config["title"] = core.TitleConfig(*args, **kwargs)
         return copy
 
+    @use_signature(core.FormatConfig)
+    def configure_tooltipFormat(self, *args, **kwargs) -> Self:
```
jonmmease, line 1281: I'm not sure where it is in the codegen logic, but it would be nice to convert tooltipFormat to snake_case here (so it would be `configure_tooltip_format`).

doc/releases/changes.rst
```diff
@@ -6,11 +6,16 @@ Release Notes
...
 
+- Update Vega-Lite from version 5.8.0 to version 5.14.1;
+  see `Vega-Lite Release Notes <https://github.com/vega/vega-lite/releases>`_.
+
 Enhancements
 ~~~~~~~~~~~~
 - Add ``chart.transformed_data()`` method to extract transformed chart data using VegaFusion (#3081, #3084, #3102)
 - Add ``"vegafusion"`` data transformer with mime renderer, save, and ``to_dict``/``to_json`` integration (#3094, #3107)
 - Support field encoding inference for objects that support the DataFrame Interchange Protocol (#3114)
+- Support grouped bars inside time axis with time bins (see `Vega-Lite Release Notes <https://github.com/vega/vega-lite/releases/tag/v5.9.0>`_)
```
jonmmease, line 17: Maybe call out the new extent transform as well

jonmmease:
I don't know if it actually introduces any challenges, but the new `extent` transform is a bit unique in that it's a transform that outputs a param.

From https://vega.github.io/vega-lite/docs/extent.html, it looks like the param is provided as just a string, so I guess that should work fine in Altair as well.  It doesn't need to be this PR, but we should make a note that we need to add a new extent transform section to the docs, and it would be good to port over the example from https://vega.github.io/vega-lite/docs/extent.html.

altair/vegalite/v5/schema/mixins.py
binste: I'd also prefer snake_case but we would then also need to change some of the other `configure_` methods such as `configure_axisXTemporal`, `configure_headerColumn`, ...

binste:
Thanks for making a new release of vl-convert and the feedback! Good point about `extent`, I missed that one when going through the release logs.

I'll add it in this PR in the coming days

- [ ] transform_extent method in `api.py`
- [ ] Doc page
- [ ] Test in test_transforms
- [ ] Add to changes.rst

binste:
Implemented `transform_extent` method and added a new page to the documentation.

doc/user_guide/transform/index.rst
```diff
@@ -96,6 +97,7 @@ Transform                                  Supported
 :ref:`user-guide-bin-transform`              ✔
 :ref:`user-guide-calculate-transform`        ✔
 :ref:`user-guide-density-transform`
+:ref:`user-guide-extent-transform`
```
binste, line 100: @jonmmease Does VegaFusion support the extent transform? The `.transformed_data` method worked on the chart example I built in the doc page but as I use the parameter value as single values it does not return it anyway so not sure if that's a good test.

altair/vegalite/v5/schema/mixins.py
jonmmease: Ahh, ok, I didn't recall that we had others already in mixed case. Maybe we address this in Altair 6 some day 😃

doc/user_guide/transform/index.rst
jonmmease: Yes, VegaFusion supports extent, so I think we're good there

jonmmease (approved):
Thanks for pulling all of this together, and adding the new docs for extent. lgtm!

doc/user_guide/transform/index.rst
binste: Cool, I'll add the tick in the docs

binste:
Thanks for the review! :)

PR #3151, opened 2023-08-14
JupyterChart: Add support for params defined in the extent transform

jonmmease:
Addresses https://github.com/altair-viz/altair/issues/3149.

This PR adds support for exposing the params defined in the new `extent` transform in JupyterChart

## Example
```python
import altair as alt
import pandas as pd

df = pd.DataFrame(
    [
        {"a": "A", "b": 28},
        {"a": "B", "b": 55},
        {"a": "C", "b": 43},
        {"a": "D", "b": 91},
        {"a": "E", "b": 81},
        {"a": "F", "b": 53},
        {"a": "G", "b": 19},
        {"a": "H", "b": 87},
        {"a": "I", "b": 52},
    ]
)

base = alt.Chart(df, title="A Simple Bar Chart with Lines at Extents").transform_extent(
    extent="b", param="b_extent"
)
bars = base.mark_bar().encode(x="b", y="a")
lower_extent_rule = base.mark_rule(stroke="firebrick").encode(
    x=alt.value(alt.expr("scale('x', b_extent[0])"))
)
upper_extent_rule = base.mark_rule(stroke="firebrick").encode(
    x=alt.value(alt.expr("scale('x', b_extent[1])"))
)

chart = bars + lower_extent_rule + upper_extent_rule
jchart = alt.JupyterChart(chart)

jchart.params
```
```
Params({'b_extent': None})
```
```python
jchart
```
visualization

```python
jchart.params.b_extent
```
```
[19, 91]
```

## Limitations
As demonstrated above, params defined in `extent` don't have initial values defined in the spec, so their value is `None` until the chart is displayed. 

We could potentially fill in this initial value using VegaFusion, but that would be more complicated and it would introduce a change in behavior based on whether VegaFusion is installed.

Also, params defined in transforms can't be set to custom values from the outside. I haven't done anything in this PR to prevent setting the value, it just doesn't do anything.

cc @mattijn


altair/jupyter/jupyter_chart.py
```diff
@@ -185,6 +185,10 @@ def _on_change_chart(self, change):
                     clean_value = param.value if param.value != alt.Undefined else None
                     initial_params[clean_name] = clean_value
 
+        # Handle the param generated by the extent transform
```
mattijn, line 188:
```suggestion
        # Handle the params generated by transforms
```

mattijn (commented):
Thanks @jonmmease! Regarding:

> Also, params defined in transforms can't be set to custom values from the outside.

Do you know why is this? A feature of Vega-lite or Altair?

In fact what I tried to do yesterday, was to see if I could use a Python `list` to set a parameter (as opposed to a value). 

This worked great when manually defining this parameter (without the `transform_extent`). Not a blocker for me, just wondering why there is this differentiation.

jonmmease:
> Do you know why is this? A feature of Vega-lite or Altair?

I think this is a Vega characteristic (Vega has the same notion of transforms being able to output signals). I think what's happening is that these transform output signals are reactive functions of the data that's input by the transform, so any changes from the outside are immediately overwritten when the dataflow is re-evaluated.

altair/jupyter/jupyter_chart.py
```diff
@@ -241,3 +245,32 @@ def _on_change_selections(self, change):
...
+    # Handle recursive case
+    for prop in ("layer", "concat", "hconcat", "vconcat"):
+        for child in getattr(chart, prop, []):
+            transform_params.update(collect_transform_params(child))
+
+    # Handle chart's own transforms
+    transforms = getattr(chart, "transform", [])
+    transforms = transforms if transforms != alt.Undefined else []
+    for tx in transforms:
+        if isinstance(tx, alt.ExtentTransform):
```
mattijn, line 273: Instead of checking for instance `alt.ExtentTransform` could we check for `hasattr` `param`? In order to capture potential new transforms that behave similar?

mattijn (approved):

altair/jupyter/jupyter_chart.py
jonmmease: Good idea. Done in ba8b1fad1898f30c290649b01584385345fcf9e9

jonmmease:
mypy is failing with:
```
Run mypy altair tests
altair/utils/core.py:331: error: Incompatible types in assignment (expression has type "Series[Any]", variable has type "str")  [assignment]
altair/utils/core.py:332: error: "str" has no attribute "where"  [attr-defined]
altair/utils/core.py:332: error: "str" has no attribute "notnull"  [attr-defined]
altair/utils/core.py:336: error: Incompatible types in assignment (expression has type "Series[Any]", variable has type "str")  [assignment]
altair/utils/core.py:337: error: "str" has no attribute "where"  [attr-defined]
altair/utils/core.py:337: error: "str" has no attribute "notnull"  [attr-defined]
altair/utils/core.py:344: error: Incompatible types in assignment (expression has type "Series[Any]", variable has type "str")  [assignment]
altair/utils/core.py:345: error: "str" has no attribute "where"  [attr-defined]
altair/utils/core.py:345: error: "str" has no attribute "notnull"  [attr-defined]
altair/utils/core.py:380: error: Incompatible types in assignment (expression has type "Series[Any]", variable has type "str")  [assignment]
altair/utils/core.py:381: error: "str" has no attribute "where"  [attr-defined]
altair/utils/core.py:381: error: "str" has no attribute "notnull"  [attr-defined]
altair/utils/core.py:388: error: Incompatible types in assignment (expression has type "Series[Any]", variable has type "str")  [assignment]
altair/utils/core.py:389: error: "str" has no attribute "isnull"  [attr-defined]
altair/utils/core.py:390: error: "str" has no attribute "astype"  [attr-defined]
altair/utils/core.py:395: error: "str" has no attribute "where"  [attr-defined]
altair/utils/core.py:395: error: "str" has no attribute "notnull"  [attr-defined]
Found [17](https://github.com/altair-viz/altair/actions/runs/5892711500/job/15982613681?pr=3151#step:7:18) errors in 1 file (checked 342 source files)
```

Looks like mypy 1.5 was released recently, and may take a bit of effort to update. Going to try pinning to 1.4 and see if that fixes CI.

binste:
This is caused by the new version 2.0.3 of `pandas-stubs` which seems to provide better type hints for `pd.DataFrame.columns`. I think previously `col` in that function just had the type `Any` but now mypy knows that it is `str` when it's first called but afterwards we reuse `col` as a `pd.Series. I'll create a separate PR to fix it.

PR #3154, opened 2023-08-17
MAINT: Fix type hint errors which came up with new pandas-stubs release

binste:
@jonmmease This should fix the errors you found in #3151. Feel free to merge afterwards so it doesn't block you!


jonmmease (approved):
Thanks @binste!

PR #3155, opened 2023-08-19
doc: Add tooltip to Line example with custom order

NickCrews:
It wasn't obvious how the year was used to create the order. Now you can see that one endpoint is 1956, the other is 2010.

I considered adding mark_text(), but that seemed heavier weight and potentially distracting.

Now, the chart has this tooltip when I hover over the last point:
image


mattijn:
Thanks for the improving this example @NickCrews!

PR #3156, opened 2023-08-19
docs: examples: add line plot with custom order

NickCrews:
Followup to https://github.com/altair-viz/altair/pull/3155.

I think this use case should be more easily discoverable, so it should be in the example gallery. Let me know if I need to do something else, or if just creating this file will trigger discovery and build.

I tweaked the prose from https://altair-viz.github.io/user_guide/marks/line.html#connected-scatter-plot-line-chart-with-custom-path just a little bit

### DIFF
--- a/tests/examples_arguments_syntax/line_custom_order.py
+++ b/tests/examples_arguments_syntax/line_custom_order.py
@@ -1,0 +1,25 @@
+"""
+Line Chart with Custom Order
+----------------------------
+By default, the line’s path (order of points in the line) is
+determined by data values on the temporal/ordinal field.
+However, a field can be mapped to the order channel for a custom path.
+
+For example, to show a pattern of data change over time between
+gasoline price and average miles driven per capita we use
+order channel to sort the points in the line by time field (year).
+The earliest year (1956) is one endpoint and the latest year (2010)
+is the other endpoint.
+"""
+# category: line charts
+import altair as alt
+from vega_datasets import data
+
+source = data.driving()
+
+alt.Chart(source).mark_line(point=True).encode(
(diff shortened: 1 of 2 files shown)


mattijn:
Thanks for this PR @NickCrews! Make sense to surface this a bit more in the example gallery.

But Altair specifications for the example gallery require a tiny bit more effort. Namely, every example should have an `arguments_syntax` and `methods_syntax` implementation. Each implementation must be saved as a stand alone script in the `tests/examples_arguments_syntax` and `tests/examples_methods_syntax` directories.

Currently you have added the Altair specification using a `methods_syntax` to the `examples_arguments_syntax` folder. 

Hopefully this is clear enough to make the additional changes. Thanks again for the PR!

NickCrews:
Thanks for the explanation! Is that correct now?

NickCrews:
PS, in a followup commit should I reverse the x and y on all of these charts? Seems like gas price is the more logical independent variable, and miles driven is the dependent.

mattijn:
Looks good now, regarding syntax. Thanks for the follow-up!

Regarding changing `x`/`y` order. It  would be good to do this in a separate PR. Also, for that change I would like to have an approval of another maintainer. 

Thanks again for this PR!

mattijn (approved):

NickCrews:
Oh wait 🤦 this is already present as a scatter plot example at https://altair-viz.github.io/gallery/connected_scatterplot.html.

I would vote to remove that one, and keep this one. This is more of a line plot, since we use mark_line(), and the connectivity of the points is significant.

mattijn:
No problem. I agree with you that it make more sense to place it in the line chart section. Can you make the change (in a new PR as this PR is merged already).

PR #3158, opened 2023-08-19
docs: line: Improve prose on custom ordering

NickCrews:
I didn't like
- How verbose the title was
- How it called this a "scatter plot". It is not.

also:
- was explicit in how the one endpoint is earliest, other is latest
- replaced weird "smart" apostrophe with basic vanilla one
- Normed capitalization of following example

### DIFF
--- a/doc/user_guide/marks/line.rst
+++ b/doc/user_guide/marks/line.rst
@@ -160,7 +160,7 @@ We can also use line grouping to create a line chart that has multiple parts wit
     import altair as alt
     import pandas as pd
 
-    source = pd.DataFrame({ 
+    source = pd.DataFrame({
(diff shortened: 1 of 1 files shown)


doc/user_guide/marks/line.rst
```diff
@@ -352,6 +354,5 @@ By mapping geographic coordinate data to ``longitude`` and ``latitude`` channels
         lookup="airport",
         from_=lookup_data
     )
-    
-    background + line
 
+    background + line
```
NickCrews, line 358: I don't think this is actually a delta?? Is the trailing newline now gone?

mattijn (approved):
Thanks for improving this @NickCrews!

PR #3159, opened 2023-08-19
docs: examples: remove connected_scatterplot

NickCrews:
Caused by
https://github.com/altair-viz/altair/pull/3156#issuecomment-1684684458

I move over the reference to the NYTimes
article because I think that is valuable.


mattijn:
Tests are failing because of this line:

https://github.com/altair-viz/altair/blob/main/tests/test_transformed_data.py#L25

Can you update this test to reflect the new file?

mattijn (approved):
Thanks for the changes @NickCrews

PR #3160, opened 2023-08-21
Refactor optional import logic and verify minimum versions

jonmmease:
This PR moves the optional import logic for pyarrow, vegafusion, and vl-convert to `util._importers` and checks for minimum supported versions of vegafusion (1.4.0) and vl-convert (0.12.0).

### DIFF
--- a/altair/utils/_importers.py
+++ b/altair/utils/_importers.py
@@ -1,0 +1,82 @@
+from types import ModuleType
+from packaging.version import Version
+
+
+def import_vegafusion() -> ModuleType:
+    min_vf_version = "1.4.0"
+    try:
+        import vegafusion as vf  # type: ignore
+
+        if Version(vf.__version__) < Version(min_vf_version):
+            raise ImportError(
+                f"The vegafusion package must be version {min_vf_version} or greater. "
+                f"Found version {vf.__version__}"
+            )
+        return vf
+    except ImportError as err:
+        raise ImportError(
+            'The "vegafusion" data transformer and chart.transformed_data feature requires\n'
+            f"version {min_vf_version} or greater of the 'vegafusion-python-embed' and 'vegafusion' packages.\n"
+            "These can be installed with pip using:\n"
+            f'    pip install "vegafusion[embed]>={min_vf_version}"\n'
+            "Or with conda using:\n"
+            f'    conda install -c conda-forge "vegafusion-python-embed>={min_vf_version}" '
+            f'"vegafusion>={min_vf_version}"\n\n'
+            f"ImportError: {err.args[0]}"
+        ) from err
+
+
+def import_vl_convert() -> ModuleType:
+    min_vlc_version = "0.12.0"
+    try:
+        import vl_convert as vlc
+
+        if Version(vlc.__version__) < Version(min_vlc_version):
+            raise ImportError(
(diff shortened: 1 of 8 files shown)


altair/utils/_importers.py
```diff
@@ -0,0 +1,82 @@
...
+            f"version {min_vlc_version} or greater of the 'vl-convert-python' package. \n"
+            f"This can be installed with pip using:\n"
+            f'   pip install "vl-convert-python>={min_vlc_version}"\n'
+            "or conda:\n"
+            f'   conda install -c conda-forge "vl-convert-python>={min_vlc_version}"\n\n'
+            f"ImportError: {err.args[0]}"
+        ) from err
+
+
+def import_pyarrow_interchange() -> ModuleType:
```
jonmmease, line 52: For consistency with the others, I updated this to use `pa.__version__` directly rather than `importlib.metadata.version`. Does anyone recall whether there was a particular reason to use importlib?

altair/utils/_transformed_data.py
```diff
@@ -133,10 +124,9 @@ def transformed_data(chart, row_limit=None, exclude=None):
             raise ValueError("Failed to locate all datasets")
 
     # Extract transformed datasets with VegaFusion
-    datasets, warnings = runtime.pre_transform_datasets(
+    datasets, warnings = vf.runtime.pre_transform_datasets(
         vega_spec,
         dataset_names,
-        get_local_tz(),
```
jonmmease, line 139: This is now the default value of the `local_tz` argument in VegaFusion 1.4.0.

altair/utils/_importers.py
```diff
@@ -0,0 +1,82 @@
...
+            f'    pip install "vegafusion[embed]>={min_vf_version}"\n'
+            "Or with conda using:\n"
+            f'    conda install -c conda-forge "vegafusion-python-embed>={min_vf_version}" '
+            f'"vegafusion>={min_vf_version}"\n\n'
+            f"ImportError: {err.args[0]}"
+        ) from err
+
+
+def import_vl_convert() -> ModuleType:
+    min_vlc_version = "0.12.0"
```
binste, line 30: Is this version number determined by what Vega-Lite version is bundled in VLC? If yes, could you add a bullet point to the `Updating the Vega-Lite version` section in `NOTES_FOR_MAINTAINERS.md` to update this version number every time we update Vega-Lite?

altair/utils/_importers.py
binste: `__version__` is not a required property of a Python package, I think it could for example be defined only in pyproject.toml without exposing `__version__`. I've also recently seen a deprecation warning of a package when accessing `__version__`. `importlib.metadata.version` is a more general and robust way to retrieve the version information based on the metadata of a package.

Although `__version__` works for all these packages, I'd slightly prefer `importlib`. What do you think?

binste (approved):
LGTM! I only added some minor comments.

altair/utils/_importers.py
jonmmease: These are good points, I'll look at switching over to `importlib`.

jonmmease:
could you glance over this once more @binste? I switched to `importlib` and added a section to `NOTES_FOR_MAINTAINERS`, so I think it's good to go.

binste:
Looks good, thanks @jonmmease! :)

PR #3173, opened 2023-08-28
Add missing arguments to Parameter and add type hints

binste:
I noticed in #3143 that `alt.param` sets some attributes on `alt.Parameter` which do not exist on that class when its initialised. That's a problem for mypy as all attributes should exist from the beginning. This PR adds `empty`, `param`, and `param_type` and type hints for the whole class.

I tested with the [Interval examples](https://altair-viz.github.io/user_guide/interactions.html#interval-selections) from the docs to make sure that indeed `empty=Undefined` leads to `empty` not being present in the final chart dict which is the old behaviour when `empty` was not present on the parameter instance at all if it was not defined.

### DIFF
--- a/altair/__init__.py
+++ b/altair/__init__.py
@@ -489,6 +489,7 @@ __all__ = [
     "TypedFieldDef",
     "URI",
     "Undefined",
+    "UndefinedType",
     "UnitSpec",
     "UnitSpecWithFrame",
     "Url",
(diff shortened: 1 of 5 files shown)


joelostblom (approved):
Thank you @binste ! I am not very familiar with type but this looks harmless, so approving to avoid it going stale. Feel free to merge.

binste:
Thanks @joelostblom :)

PR #3175, opened 2023-08-28
Fix doctest and run doctests in altair module

### LINKED ISSUE #3174: Example in docstring of get_inline_table_names does not work. Blocks conda-forge release.
@jonmmease The example in the docstring of `get_inline_table_names` does not retrieve the table name. I don't know if there is a bug in the function or it's just the docstring that needs to be adapted.

```python
spec = {
    "data": [
        {
            "name": "foo",
            "url": "https://path/to/file.csv"
        },
        {
            "name": "bar",
            "url": "table://inline_dataset_123"
        }
    ]
}
from altair.utils._vegafusion_data import get_inline_table_names
get_inline_table_names(spec)  # Returns an empty set
```
This came up in https://github.com/conda-forge/altair-feedstock/pull/49

Btw, this did not come up in the tests as docstring tests are only run in the `tests` folder. We should probably change the pytest command to include the altair package -> `pytest --doctest-modules altair tests`.

jonmmease:
Closes https://github.com/altair-viz/altair/issues/3174 by running doctests within the altair module and fixing the failing doctest.

### DIFF
--- a/altair/utils/_vegafusion_data.py
+++ b/altair/utils/_vegafusion_data.py
@@ -72,7 +72,7 @@ def get_inline_table_names(vega_spec: dict) -> Set[str]:
     ...         },
     ...         {
     ...             "name": "bar",
-    ...             "url": "table://inline_dataset_123"
+    ...             "url": "vegafusion+dataset://inline_dataset_123"
     ...         }
     ...     ]
     ... }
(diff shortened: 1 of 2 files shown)


binste (approved):
Thanks! Sorry for not creating the PR myself, I was not sure if it's just the docstring or something bigger that is wrong.

PR #3202, opened 2023-09-22
Add tutorial on numpy tooltip images

### LINKED ISSUE #3198: Add example showing how to render numpy image arrays as tooltip images
I think it could be helpful to show how images that are represented as numpy arrays can be rendered in tooltips in altair. I can add a doc example of this. Maybe in [the tutorials/case studies](https://altair-viz.github.io/case_studies/exploring-weather.html) section? We only have one example there currently. Another option would be to combine it with https://altair-viz.github.io/gallery/image_tooltip.html and create a new page in the user guide on images, but that 's more involved.

Here is the code and a video of the output. Note that this would add scipy as a documentation dependency (I could probably get around that, but I have another example I want to add that requires scipy so thought I might as well ask now if it is ok to add it). The images are not too large, the size of the chart saved as an html file is around 200kb.

1. Create some example image arrays with blobs in them and measure the area of the blobs.
    ```python
    import numpy as np
    import pandas as pd
    from scipy import ndimage as ndi
    
    rng = np.random.default_rng([ord(c) for c in 'altair'])
    n_rows = 200
    
    def create_blobs(img_width=96, n_dim=2, thresh=0.0001, sigmas=[0.1, 0.2, 0.3]):
        """Helper function to create blobs in the images"""
        shape = tuple([img_width] * n_dim)
        mask = np.zeros(shape)
        points = (img_width * rng.random(n_dim)).astype(int)
        mask[tuple(indices for indices in points)] = 1
        return ndi.gaussian_filter(mask, sigma=rng.choice(sigmas) * img_width) > thresh
    
    df = pd.DataFrame({
        'img1': [create_blobs() for _ in range(n_rows)],
        'img2': [create_blobs(sigmas=[0.15, 0.25, 0.35]) for _ in range(n_rows)],
        'group': rng.choice(['a', 'b', 'c'], size=n_rows)
    })
    df[['img1_area', 'img2_area']] = df[['img1', 'img2']].applymap(np.mean)
    df
    ```

2. Convert the numpy arrays to base64 encoded strings that will show in the tooltip

    ```python
    from io import BytesIO
    from PIL import Image, ImageDraw
    import base64
    
    
    def create_tooltip_image(df_row):
        # Concatenate images to show together in the tooltip
        img_gap = np.ones([df_row['img1'].shape[0], 10])  # 10 px white gap between imgs
        img = Image.fromarray(
            np.concatenate(
                [
                    df_row['img1'] * 128,  # grey
                    img_gap * 255,  # white
                    df_row['img2'] * 128
                ],
                axis=1
            ).astype('uint8')
        )
        
        # Optional: Burn in labels as pixels in the images
        ImageDraw.Draw(img).text((3, 0), 'img1', fill=255)
        ImageDraw.Draw(img).text((3 + df_row['img1'].shape[1] + img_gap.shape[1], 0), 'img2', fill=255)
        
        # Convert to base64 encoded image string that can be displayed in the tooltip
        buffered = BytesIO()
        img.save(buffered, format="PNG")
        img_str = base64.b64encode(buffered.getvalue()).decode()
        return f"data:image/png;base64,{img_str}"
    
    # The column with the image must be called "image" in order for it to trigger the image rendering in the tooltip
    df['image'] = df[['img1', 'img2']].apply(create_tooltip_image, axis=1)
    
    # Dropping the images since they are large an no longer needed
    df = df.drop(columns=['img1', 'img2'])
    df
    ```
3. Create a chart to show the images

    ```python
    import altair as alt
    
    alt.Chart(df, title='Area of grey blobs').mark_circle().encode(
        x='group',
        y=alt.Y(alt.repeat(), type='quantitative'),
        tooltip=['image'],
        color='group'
    ).repeat(
        ['img1_area', 'img2_area']
    )
    ```

https://github.com/altair-viz/altair/assets/4560057/45ccc43f-c8a4-4b3b-bb42-ed0b18cd9703

joelostblom:
close #3198 

I thought about trying to fit this in the user guide, but it seems to be more of a tangent to me. Maybe we will make a user guide page on images in the future when there is a raster mark and we could also move this here, but for now my thinking is that it fits better in a tutorial as a more specific and advanced use case. I think having some more complex tutorials like this could be interesting, so feel free to contribute some if there are ideas (I can add another one on laying points out in the shape of density distributions from previous work).

### DIFF
--- a/doc/case_studies/numpy-tooltip-images.rst
+++ b/doc/case_studies/numpy-tooltip-images.rst
@@ -116,6 +116,15 @@ as expected.
         title='Comparison of blob areas'
     )
 
+Note that when including images as part of the chart data,
+the chart size often increases several-fold.
+The size of the chart above would have been 19 Kb without the images,
+but with the images added it is 760 Kb.
+While this is a 20x size increase,
+the base64 encoding is still quite storage efficient;
+if we would have included the images in their original Numpy array format
+the chart size would have been 35Mb!
+
 If we want to have even more fun and get a bit more sophisticated,
 we could show one chart at a time
 and update what is shown on the y-axis
--- a/doc/case_studies/numpy-tooltip-images.rst
+++ b/doc/case_studies/numpy-tooltip-images.rst
@@ -1,0 +1,195 @@
+.. _numpy-tooltip-imgs:
+
+Displaying Numpy Images in Tooltips
+-----------------------------------
+
+In this tutorial,
+you’ll learn how to display images stored as Numpy arrays
+in tooltips with any Altair chart.
+
+First,
+we create some example image arrays with blobs of different sizes.
+We measure the area of the blobs
+in order to have a quantitative measurement to plot.
+
+.. altair-plot::
+    :output: repr
+
+    import numpy as np
+    import pandas as pd
+    from scipy import ndimage as ndi
+
+    rng = np.random.default_rng([ord(c) for c in 'altair'])
+    n_rows = 200
+
+    def create_blobs(blob_shape, img_width=96, n_dim=2, sizes=[0.05, 0.1, 0.15]):
+        """Helper function to create blobs in the images"""
+        shape = tuple([img_width] * n_dim)
+        mask = np.zeros(shape)
+        points = (img_width * rng.random(n_dim)).astype(int)
+        mask[tuple(indices for indices in points)] = 1
+        if blob_shape == 'circle':
+            im = ndi.gaussian_filter(mask, sigma=rng.choice(sizes) * img_width)
+        elif blob_shape == 'square':
+            im = ndi.uniform_filter(mask, size=rng.choice(sizes) * img_width, mode='constant') * rng.normal(4, size=(img_width, img_width))
+        return im / im.max()
+
+    df = pd.DataFrame({
+        'image1': [create_blobs('circle') for _ in range(n_rows)],
+        'image2': [create_blobs('square', sizes=[0.3, 0.4, 0.5]) for _ in range(n_rows)],
+        'group': rng.choice(['a', 'b', 'c'], size=n_rows)
+    })
+    # Compute the area as the proportion of pixels above a threshold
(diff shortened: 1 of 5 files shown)


doc/case_studies/numpy-tooltip-images.rst
```diff
@@ -0,0 +1,195 @@
...
+    )
+    metric_param = alt.param(
+        value='image1_area',
+        bind=metric_dropdown
+    )
+    alt.hconcat(
+        # This first chart is the axis title and is only needed because
+        # Vega-Lite does not yet support passing an expression directly to the axis title
+        alt.Chart().mark_text(angle=270, dx=-150, fontWeight='bold').encode(
+            alt.TextValue(alt.expr(f'{metric_param.name}'))
```
joelostblom, line 170: Interestingly, sphinx does not allow us to reference the actual name of the `metric_param` as a string, which works in jupyterlab and the vega editor. The only way around it I found was to use an f-string with `metric_param.name` which is what we use elsewhere in the docs and it's probably more convenient anyways since we don't need to manually set a name for the param.

binste: Hmm weird, can't explain why that happens but I agree that using `.name` is more convenient.

binste (approved):
I like the idea of expanding the tutorial section with more complex use cases and this is a great one! Thanks @joelostblom. LGTM :)

jonmmease:
I haven't reproduced locally, but I really like the example!

Thinking it might be worth calling out the fact that this may result in a large chart specification as all of the images are included in the spec (not just the one at a time that shows up in the tooltip).

joelostblom:
Thanks both! And good note about image sizes, I included this section:

> Note that when including images as part of the chart data,
the chart size often increases several-fold.
The size of the chart above would have been 19 Kb without the images,
but with the images added it is 760 Kb.
While this is a 20x size increase,
the base64 encoding is still quite storage efficient;
if we would have included the images in their original Numpy array format
the chart size would have been 35Mb!

(Which reflects the sizes when the chart is saved as an html file)

Issue #2877, opened 2023-02-07
Altair should accept iterables where currently list/tuple/ndarray is required


davidgilbertson:
My scenario: I want the x-axis of my chart to show values in increments of 7 (autocorrelation plot). I do this by providing the tick values explicitly:
```py
alt.X(
    "lag:Q",
    axis=alt.Axis(
        values=range(0, len(chart_df), 7),
    ),
)
```
This errors, because `range` is not a list, and [this line](https://github.com/altair-viz/altair/blob/97ff1ebc3a485be1c00a34e5d19801eca5287e6f/altair/utils/schemapi.py#L89) in Altair checks for quite specific types
```py
    elif isinstance(obj, (list, tuple, np.ndarray)):
        return [_todict(v, validate, context) for v in obj]
```

I think this is overly restrictive, this should be able to handle any sort of iterable (generators, sets, iterators, etc)

Some options:
* Check for `isinstance(obj, collections.abc.Iterable)` - this check should come last because dicts are iterable.
* Check for `isinstance(obj, collections.abc.Sequence)` - but beware this doesn't include Numpy arrays.

IMO Iterable ([docs](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable)) is the way to go.

Also, if you wanted to handle lots of other data types without much effort (e.g. PyTorch and TensorFlow tensors) you could do what Matplotlib do and cast everything to NumPy arrays with `nd.array(iterable)`. That way NumPy does all the work (checking for `__array__` methods, etc) and I would expect it to be pretty robust.

You could even type these as [`ArrayLike`](https://numpy.org/doc/stable/reference/typing.html#arraylike), borrowing from NumPy, similar to what Pandas does.

joelostblom:
Thanks for raising this, I agree that it would be useful to be more flexible in what types of iterables that are accepted at various places in Altair. I like the idea of using the same approach as matplotlib as this would already have been battle tested and work well for them and we just need to make sure it also works here. A PR here would be welcome and we can continue discussing there once we see what a solution to this could look like. Related https://github.com/altair-viz/altair/issues/2808

joelostblom:
fixed in #3501

PR #3208, opened 2023-09-29
Type hints: Infer types from vegalite schema for autogenerated code

### LINKED ISSUE #2854: infer argument types from vegalite schema
Based on these two comments in this PR https://github.com/altair-viz/altair/pull/2846:

> _Originally posted by @mattijn in https://github.com/altair-viz/altair/issues/2846#issuecomment-1401606735_
> 
> I understand that adding another dependency just to reduce 5 lines of code is probably not worth it. I also agree that there may be other recent type-related changes in the core of Python that makes it something to consider if we look a bit broader than this PR alone. 
> 
> We previously had two types of discussions about types in Altair. 
> 
> - One is about improving the existing Python code in the /tools folder in the Altair repository, such as this issue (https://github.com/altair-viz/altair/issues/2493) and related PR (https://github.com/altair-viz/altair/pull/2614, btw I now notice that this PR also uses the `TypeVar` bound approach).
> 
> - The other is about including types in the generated Python code of Altair, originally (https://github.com/altair-viz/altair/pull/2670) adapted to (https://github.com/altair-viz/altair/pull/2681) and then reverted back to the original in (https://github.com/altair-viz/altair/issues/2714). 
> 
> Especially the result of the second bullet has set the type `Any` to anything that is `Undefined` finishing this discussion on, 'what is the type of `Undefined`?'. 
> 
> However, with your impressive PRs regarding jsonschema and types, I was wondering, since types are included in the vega-lite jsonschema, if these types can also be inferred for each argument? Maybe recent type-related changes in the core of Python can make this easier? It is merely an open question from which I don't know enough if it is worthwhile to consider.

and

> _Originally posted by @binste in https://github.com/altair-viz/altair/issues/2846#issuecomment-1404066238_
> 
> Thanks for writing this down! This is very helpful. It should be possible to create Python type hints based on the vl schema, especially given that this parsing is already done for the docstrings, see e.g. https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/schema/channels.py#L12509 where the `stack` argument to `alt.X` is documented as:
> ```
> stack : anyOf(:class:`StackOffset`, None, boolean)
> ```
> 
> There are probably also synergies with replacing the vega-lite types with Python types in all error messages (suggested by you in https://github.com/altair-viz/altair/pull/2842#issuecomment-1400813951) and docstrings. 
> 
> As more type hints shouldn't break anyones code, I don't think we would need to necessarily include it in the release candidate but I'm definitely intrigued in giving this a try. I really like the development experience you get for a library with well-defined type hints such as better autocompletion as well as early warnings if you pass a wrong value to an argument before even executing the code, for example if you'd pass an integer to `stack`. The same linters could then of course be used in CI/CD pipelines.
> 
> I'll try to get a draft PR going in the coming weeks so we can further explore how well it works.

binste:
This will close #2854. See #2951 for an overview of the type hint effort


binste:
This is ready to be reviewed. All autogenerated files now contain type hints which are inferred based on the Vega-Lite schema. I'm aware that this is a complex one to review. However, all Altair examples in the `tests` folder pass mypy (which was not always the case when working on this PR...) which I think can give us a lot of confidence.

In the documentation, I configured Sphinx to not show the type hints in the class signature as else it gets too cluttered. The type hints are then shown in the docstrings:

image

Apart from the usual benefits of static type checking, I really like the improved autocompletion I'm getting in VS Code:

image
image

joelostblom:
Wow, that autocompletion is really helpful! I wish jupyterlab supported it out of the box; but I don't think it does without https://github.com/jupyter-lsp/jupyterlab-lsp if I recall correctly.

Is there any chance that this also enables autocompletion of column names when doing `.encode(x=<TAB>` and/or `alt.X(<TAB>`? I think that would be super helpful to, but I can open a separate issue for that if it is not covered here. I'm thinking it might be possible to propagate all fields from the charts `data` to have the same autocompletion status as the strings in your screenshots.

binste:
It's a great idea but it's not covered by this PR. Happy to have a look at it in a separate issue! Not sure how Pandas does thisso that column names show up in e.g. `df["col...`. Maybe we can use something similar.

mattijn:
Thanks @binste! 
Based on your comments above I can really see the value of this PR.

But in other cases, I'm wondering if we can make this more useful than what I currently get. See for example this screenshot:
image
Is it right if I don't see support for the `str` type hint (`shorthand` argument is not coming from the schema)

While I think, _as a user_, I only can do something as such for the `color` encoding channel:
```python
[
    type('foo:Q'), 
    type(alt.Color()), 
    type(alt.value('')), 
    type(alt.datum), 
    type(alt.ColorDatum(''))
]
```
```python
[str,
 altair.vegalite.v5.schema.channels.Color,
 dict,
 altair.expr.core.DatumType,
 altair.vegalite.v5.schema.channels.ColorDatum]
```

binste:
Nice catch! `str` for shorthands is indeed missing in `core.FacetedEncoding` and hence in the `.encode` method. Another cool thing with the increased type hint coverage is that this error now also shows up after merging #3143  :) The more type hints we have, the more stable this gets.

I'll try to fix it soon and look into your other point as well.

binste:
* Added shorthand to encoding channels
image

* Fixed mypy issues which now shoed up after merging #3143
* Switched to ruff as formatter for the autogenerated files


@mattijn Regarding the types for the color encoding channel, I agree that the types you listed are the ones that users are usually familiar with. However, the other types work as well, for example `alt.ColorDef`:

image

The type hints serve sort of a dual purpose here, one is for actual type checking, where I think this exhaustive list of types is what is needed, and the other one is to improve autocompletion. For autocompletion, probably a shorter list of "the most common" types would be better but that would mean that we need to compromise on the completeness of the type hints for type checking. Does this explanation make sense and address your concern?

binste:
The more I think about this the more I'm comming around to the idea of limiting the type hints on at least the encode method to improve autocompletion...

mattijn:
I understand it is a bit complicated to decide what is right. From a machine perspective these are correct type hints and I'm sure they all actually work, but the power of Altair to human users is that you do not need to know and use these low-level classes to write a specification that compiles correctly. I'm in favour of having readable autocompletion, but also one that is helpful to users. I really hope that no one starts to use or try `alt.FieldOrDatumDefWithConditionDatumDefGradientstringnull()`, because it appears in the autocompletion as type hint.

joelostblom:
I'm in favor of prioritizing the user experience with regards to the autocompletion from type hints (unless there are some substantial drawbacks)

binste:
Yep, this is much better:

image

All mypy tests still pass so I think it captures the usual input types. Logic is:
* Always `str` and the channel class, e.g. `str | alt.Color` as well as datum `alt.ColorDatum` and value `alt.ColorValue` definition if accepted
* If datum or value definitions are accepted, also include a type hint for a dictionary which would come out of `alt.datum` or `alt.value`. 

Some things to note regarding the type hints for `datum` (`_Datum`) and `value` (`_Value`):
* they are used to make the type hints more readable when they show up in IDEs so that users know they can use
`alt.datum` or `alt.value`.
* I can't type them without underscore (`datum`) as this would then be imported into other modules due to how imports are set up in Altair and then it conflicts with the actual `datum` or `value`. 
* We can't type these more accurately as `TypedDict` do not accept arbitrary extra keys where as alt.value and alt.datum do. Also see https://github.com/python/mypy/issues/4617#issuecomment-367647383
* If we want to use both value and datum it needs to be a combined type alias (`_DatumOrValue`) as IDEs such as VS Code simply show the first type alias in the signature  if they refer to the same type.

What do you think? :)

joelostblom:
I agree this looks better! What do you think about simplifying even further and removing `ColorDatum` and `ColorValue` in favor of having only `_DatumOrValue`? We almost never use `ColorDatum` and `ColorValue` in the documentation and only explain `datum` and `value` in the narrative part of the docs, so if I saw the mention of these in the autocompletion and then went to read up about what they do in the docs, I wouldn't really get any wiser.

Are there any advantages (except channel-specific tab completion) to these channel specific datum and value classes? If there are advantages we should write about it in the docs, but if there are not I wonder if we treat these classes more as an internal detail and the user-facing interface is always the `datum` and `value` functions that return a dictionary?

mattijn:
hm, I treat `alt.value('')` and `alt.datum('')` as a shorthand for `alt.<encodingchannel>Value('')` and `alt.<encodingchannel>Datum('')`. Just like `'foo:Q'` is a shorthand for `alt.<encodingchannel('')>`.

Where the shorthand `foo:Q` is of type `str`, `alt.value('')` and `alt.datum('')` are just returning type `dict`. One could also write `{'value': ''}` or `{'datum': ''}`.

I would say, from a type perspective, I'm OK with this:
```
color: str | Color | dict | ColorValue | ColorDatum
```

See also this recording showing what I mean:

https://github.com/altair-viz/altair/assets/5186265/0385e5ca-e72c-4800-ab87-628c1371c968

When using a `dict` I get no hints, when using `alt.datum()` I get minimal hints and when using alt.ColorDatum()` I get complete list of what is possible.
So while we not using it as much in the docs, I leaning towards still seeing a bit of value in using `ColorValue | ColorDatum` as type hints...

joelostblom:
Yeah, the more elaborate help docstring pop up is the main benefit I can think of. Although I guess also the method-based syntax is only available for alt.ColorDatum etc. Actually I see now that we do have some info on this in the docs at the end of this section https://altair-viz.github.io/user_guide/encodings/index.html#datum-and-value.

I agree it's helpful to think of `alt.datum` as a shorthand, although I do think it is a bit different from the field shorthand, since there are three levels here:

```
dict -> alt.datum -> alt.ColorDatum
```

whereas for fields there are only two levels:

```
str -> alt.Color
```

Thinking about this more, maybe it is the middle level with `alt.datum` that is the least useful, since as you said, it only offers marginal docstring help/tab completion over typing out the dictionary directly, is not that much more convenient to type than a dict, and does not have an analogy in the fields shorthands. 

From that angle, I would agree with the docstring types you proposed `color: str | Color | dict | ColorValue | ColorDatum` since it de-emphasizes the use of `alt.datum` (I don't think seeing the `dict` type will be an intuitive indication for using `alt.datum`). If we think that `alt.datum` is an important shorthand, I think it would make sense to have it more explicitly indicated in the docstring.

Partly related, using `alt.datum('value')` less could also clear potential confusion with that there is also `alt.datum.Column` which we use to construct comparative JS expressions (I remember this was confusing for me initially). Maybe we should use it less in the docs too in favor of either the dict or `alt.ColorDatum` etc...

mattijn:
Valid points. Didn't know we had this docs section on it, thanks again @ChristopherDavisUCI!
Regarding:

> If we think that `alt.datum` is an important shorthand, I think it would make sense to have it more explicitly indicated in the docstring.

Better docstrings for this would be great!

For me the following three are a kind of _group_, which have value being there:
- alt.datum()
- alt.value()
- alt.expr()

But the returning types of these are not equal (2x `dict`, 1x `ExprRef`). We could investigate if we can change the returning types of `alt.datum()` and `alt.value()`, but currently a type hint of `dict` seems most appropriate. Not forget too mention, a `dict` is also valid for an encoding channel (eg.`x={'field':'a', 'type':'nominal'}`)

binste:
Thank you both for the insightful comments! Seems like there is more in this discussion that we could take out of then just improved type hints :) I also found it confusing in the beginning (and sometimes still) to be able to use `alt.datum.Column`, `alt.datum("something")`, or `alt.datum.label/value` in places such as `labelExpr`. I prefer to use the more explicit `alt.ColorDatum`.

I see that `_ValueOrDatum` doesn't really indicate that a user could use `alt.value` or `alt.datum` -> I changed it to just `dict` so we have:

```
color: str | Color | dict | ColorValue | ColorDatum
```

image


This looks clean and intuitive to me. It also captures e.g. `x={'field':'a', 'type':'nominal'}`, thanks for catching this!

joelostblom:
> `color: str | Color | dict | ColorValue | ColorDatum`

I think this looks good, should we put `dict` before `Color` for more aesthetic capitalization? `color: str | dict | Color | ColorValue | ColorDatum`. I'm fine either way.

> Not forget too mention, a dict is also valid for an encoding channel (eg.x={'field':'a', 'type':'nominal'})

Ah yes, thanks for reminding me about that. So the more correct mental model would probably look like this:

datum/value:

```
alt.datum -> dict -> alt.ColorDatum
```

fields:

```
str -> dict -> alt.Color
```

So `alt.datum` is the shortcut for the dict, similar to `str` for a column/field name, and not the other way around.

However, while the convenience of using a string such as `'col_name'` is notable in that is saves us from typing out the entire `{'field': 'col_name'}`, the convenience of using `alt.datum('col_name')` vs `{'datum': 'col_name'}` is not as drastic as it is as many characters. I also can't see how we could improve the docstring to the same level at alt.ColorDatum, so `alt.datum` does seem less useful, and maybe we should start favoring `alt.ColorDatum` etc in the docs.

Another interesting observation I just made, is that it is also possible to type `alt.Color(datum=...)`, which I wasn't aware of and it is not documented in the docstring. It currently breaks if we try to append a method such as `.title()`, but that could be fixed. I like the mental model for this ("use `alt.Color` for anything that is not just a column name") and it does make the autocompletion docstring less crowded as well as reduces the number of objects at top-level which makes for less overwhelming autocompletion. Maybe it should eventually become `alt.Color.datum()` as suggested in https://github.com/altair-viz/altair/issues/2918, I'm not sure).

An example:

```py
import altair as alt
import pandas as pd

source = pd.DataFrame({
  'x': range(3),
  'y': range(3)
})

    x='x',
    y='y'
)

# Option 1
points + alt.Chart().mark_text().encode(
    x=alt.XDatum(1),
    y=alt.YValue(120),
    text=alt.TextDatum('Text here')
)

# Option 2
points + alt.Chart().mark_text().encode(
    x=alt.X(datum=1),
    y=alt.Y(value=120),
    text=alt.Text(datum='Text here')
)

```

mattijn:
> it is also possible to type `alt.Color(datum=...)`, which I wasn't aware of and it is not documented in the docstring

Interesting! Didn't know this either!
Another point I just realized, some encodings also accept `list` as type, for example, `tooltip` and `detail`. I don't think this is currently covered.

binste:
Didn't know this as well! :)

Oh yeah, mypy somehow does not complain about it... I added a `list` type to the channels which support it (tooltip, detail, and order). We could be more specific what can be passed in that list but then the type hints again get rather large: `Sequence[Union[str, alt.Tooltip, dict, alt.TooltipValue]]`(would need to use Sequence due to invariance of a list)

mattijn:
Thanks @binste. Can you elaborate a bit on the `_ParameterProtocol` type? You had to define it for [circular reasons](https://github.com/altair-viz/altair/blob/0292f14050aef97835d0e3f74918a1a827b88953/tools/generate_schema_wrapper.py#L54C8-L59), but how can I make use of this type? 
See for example here, on hovering `domain` within `.scale(domain=[-15, 45] ..`:
image
It suggests that I can use something as such? 
```python
my_param = alt.param(value=[-15, 45])
.scale(domain=my_param ..)  # actually works :)
```
Another question based on the type hints, and maybe it is something I've to get used to, but for me it was a bit confusing that `Sequence[]` is a type of `list`? You write a very valuable section on it around [here](https://github.com/altair-viz/altair/blob/0292f14050aef97835d0e3f74918a1a827b88953/tools/schemapi/utils.py#L261-L276), but I was wondering if we just could use `list` here? I understand that we then don't define what is possible within the list, but currently it feels a bit duplicated, that the mentioned duplicate types in the type hint are nested variants within the `Sequence` type.

binste:
> Side note, please ignore if not related, is this related to this issue: #2877? Would it be possible to define a type for `ArrayLike`? Eg using `range(-15, 46, 60)` for scale-domain will currently raise:
> 
> ```python
> SchemaValidationError: 'range(-15, 46, 60)' is an invalid value for `domain`. Valid values are of type 'array', 'string', or 'object'.
> 
> 'unaggregated' was expected
> ```
> 
> Or is this merely a matter of converting an `ArrayLike` as the `range` in this case into a `list` within Altair, as is aimed for in #3237?

We'll need to first implement the conversion of anything that looks like a [Sequence](https://docs.python.org/3/library/stdtypes.html#sequence-types-list-tuple-range) (or just any iterable?) to a list as in the Vega-Lite specification we need them to be lists (i.e. an `array` in JSON terminology). Afterwards, we could then update the type hints to accept any iterable/sequence. We might be able to stick with `Sequence` as a type hint.

binste:
GitHub shows our comments in this PR out of order for me :) Just something to be aware of:

image

mattijn:
Side note, please ignore if not related, is this related to this issue: https://github.com/altair-viz/altair/issues/2877? Would it be possible to define a type for `ArrayLike`? Eg using `range(-15, 46, 60)` for scale-domain will currently raise:
```python
SchemaValidationError: 'range(-15, 46, 60)' is an invalid value for `domain`. Valid values are of type 'array', 'string', or 'object'.

'unaggregated' was expected
```
Or is this merely a matter of converting an `ArrayLike` as the `range` in this case into a `list` within Altair, as is aimed for in https://github.com/altair-viz/altair/pull/3237?

binste:
> Thanks @binste. Can you elaborate a bit on the `_ParameterProtocol` type? You had to define it for [circular reasons](https://github.com/altair-viz/altair/blob/0292f14050aef97835d0e3f74918a1a827b88953/tools/generate_schema_wrapper.py#L54C8-L59), but how can I make use of this type? See for example here, on hovering `domain` within `.scale(domain=[-15, 45] ..`: image It suggests that I can use something as such?
> 
> ```python
> my_param = alt.param(value=[-15, 45])
> .scale(domain=my_param ..)  # actually works :)
> ```

It means that wherever `_ParameterProtocol` is given, you can use an `alt.Parameter` which is usually created from `alt.param` or `alt.selection_*`. You would not use `_ParameterProtocol` directly, it's purely there for the benefit of a type checker. It's usually used for structural subtyping ("duck-typing"). For example, if you have a function that accepts any type which has a `get_number` method which returns an `int` and does not care about anything else about this type, you can do:
```
class Proto(Protocol):
    def get_number(self) -> int:
        ...
```
and use this as a type hint:

```python
class A:
    def get_number(self) -> int:
         return 2

```



As you mentioned, here I had to use it to prevent circular imports as ideally, we could use `alt.Parameter` directly. Does that make sense? Happy to elaborate 




> Another question based on the type hints, and maybe it is something I've to get used to, but for me it was a bit confusing that `Sequence[]` is a type of `list`? You write a very valuable section on it around [here](https://github.com/altair-viz/altair/blob/0292f14050aef97835d0e3f74918a1a827b88953/tools/schemapi/utils.py#L261-L276), but I was wondering if we just could use `list` here? I understand that we then don't define what is possible within the list, but currently it feels a bit duplicated, that the mentioned duplicate types in the type hint are nested variants within the `Sequence` type.

mattijn:
Thanks for the clear explanation @binste. Much appreciated! 

I feel there is a balance between how we suggest/hint how users should use the library and what is function-wise the correct type hint. It is great that we can make our type checker happy, but I'm hoping we also can use these type hints to be useful in communicating how users can use the altair API. I feel the battle that you've had with mypy (much appreciated!), but I'm trying to review as being an average user..

For my understanding, would it be possible to make a small example how this currently leads to circular imports? 

I'm somehow hoping we can circumvent this definition of `_ParameterProtocol` by restructuring pieces of code or by introducing some more advanced OOP methods.

binste:
> I feel the battle that you've had with mypy (much appreciated!), but I'm trying to review as being an average user..

That's much appreciated and exactly what this PR needs! It's easy to get too focused on just adding type hints everywhere and making mypy happy so I think the feedback from Joel and you have already improved this a lot :)
 
> For my understanding, would it be possible to make a small example how this currently leads to circular imports?
> 
> I'm somehow hoping we can circumvent this definition of `_ParameterProtocol` by restructuring pieces of code or by introducing some more advanced OOP methods.

Sure! I hope as well that there is a feasible solution which I'm missing. `Parameter` is defined in [`api.py#L189`](https://github.com/binste/altair/blob/autoinfer_types/altair/vegalite/v5/api.py#L189). However, the autogenerated files `core.py`, `channels.py`, and `mixin.py` are imported at the beginning of `api.py`: https://github.com/binste/altair/blob/autoinfer_types/altair/vegalite/v5/api.py#L17. Taking `core.py` as an example, if we would import `Parameter` from `api.py`, then we have the circular import. Parameter also requires an import of the `expr` module which again imports `core.py` to complicate things.

To keep it maintainable, I think code should be imported from `autogenerated code in schema folder` -> `'hand-written' code in api.py, data.py, ...` -> `top-level imports making it all accessible under the altair namespace`.

mattijn:
Thanks again for the extra explanations. I start to understand the issue. Would it be possible to define an abstract base class of the `Parameter` within `core.py` instead of the `_ParameterProtocol`? So the `Parameter` class within `api.py` can then become a subclass of the `Parameter` within `core.py`.  Something as such:

##### `api.py`:

```python
import core
import expr
from typing import Union, Optional
from typing import Dict as TypingDict


class Parameter(core.Parameter, expr.MySomethingExpr, object):
    _counter: int = 0

    @classmethod
    def _get_name(cls) -> str:
        cls._counter += 1
        return f"param_{cls._counter}"

    def to_dict(self) -> TypingDict[str, Union[str, dict]]:
        return {"param": self.name}

    def _to_expr(self) -> str:
        return f"{self.name}"

    def __init__(self, name: Optional[str] = None) -> None:
        if name is None:
            name = self._get_name()
        self.name = name


def selection(name) -> Parameter:
    parameter = Parameter(name)
    return parameter
```

##### `core.py`:

```python
from abc import ABC, abstractmethod
from typing import Union
from typing import Dict as TypingDict


class Parameter(ABC):
    _counter: int

    @classmethod
    @abstractmethod
    def _get_name(cls) -> str:
        pass

    @abstractmethod
    def to_dict(self) -> TypingDict[str, Union[str, dict]]:
        pass

    @abstractmethod
    def _to_expr(self) -> str:
        pass


class MySomethingCore:
    def __init__(self, param: Parameter):
        self.param = param
```

##### `expr.py`:

```python
class MySomethingExpr:
    ...
```

This code passed mypy and when in use, result in something as such:
```python
import api
from core import MySomethingCore

my_sth_core = MySomethingCore(api.selection('my_param'))
type(my_sth_core.param), my_sth_core.param.name
```

> ```cmd
> (api.Parameter, 'my_param')
> ```

I've to admit, I tried to implement this POC within this PR, but I could not understand the errors that mypy gave. I've no idea if they were related to the change I tried to implement.

binste:
Thanks for the example and playing around with it! I can replicate the mypy errors when implementing it in this PR and I can't figure out what the issue is with this approach... It's not due to the naming of the class, I can rename the protocol to `Parameter` and everything works but as soon as it is a subclass of `ABC` it breaks down. Weird. If no one else has an idea here, I think we'll need to stick to it being a `Protocol` but we can of course change the naming.

I assume the improvement would be that `Parameter` is more intuitive for users then `_ParameterProtocol` right? I do see the value in the simpler name but I find classes with the same name rather tricky in Altair as everything is imported upwards with `from ... import *` statements. It then depends on the order of the imports here https://github.com/altair-viz/altair/blob/main/altair/vegalite/v5/__init__.py#L3 to make sure that the correct `Parameter` makes it to the top level.

How about renaming it to `_Parameter`? No user should instantiate a `alt.Parameter` directly anyway but instead use `param` so I think the underscore would not really matter. However, it would still clearly signal that this is not something to be used directly. I implemented it in the latest commit. What do you think @mattijn?

mattijn:
I agree with your point that it is not a good idea use the same name, make sense to do it like you've implemented it.

Another point. With this PR the tooltip looks like this:
image

The type hints are very detailed, but there is no extra documentation on the arguments. But in the current main branch it looks as follows:
image

All the arguments are explained in detail after the `Parameter` heading.
Can we get these tooltips back again?

binste:
Oh, sure! I added it back in the latest commit:

image


 The docstring got lost because I had to create a custom signature for `encode` and I forgot to add a docstring there so this should not be the case anywhere else.

mattijn:
Thanks @binste! Can you do the same for the `mark_circle()` etc, it seems still missing as well.

image


I was playing around and checking the type hints here and there (it's a lot, but informative!), but when I check the type hints of `transform_filter` I notice it accept also `Any` among others, that is a bit strange I think, and here I see it accept `Parameter` instead of `_Parameter` elsewhere.
image

binste:
The `mark_` methods never had a more extensive docstring than this, it's the same on main. But I agree that we should have this! I opened https://github.com/altair-viz/altair/issues/3262 to track it as I'd prefer to do it in a separate PR as it is not directly related to type hints.

Regarding `transform_filter`, this is because Pylance (and probably mypy as well) can't resolve `expr.core.Expression` due to how `expr` module is both a module but also a callable:

image

But good that you bring it up, I realised that if I change the import statement to `from ...expr import core as _expr_core` and then use that instead of `expr.core`, it works:

image

I only used `_Parameter` in the autogenerated files where otherwise we would have circular imports. In all other files, I went with `Parameter` directly. No preference if we want to consistently use the `_Parameter` protocol or use `Parameter` where possible (current implementation).

mattijn (approved):
Thanks for all the hard work here @binste! I approve:).

Since this involves the Altair API's user-facing side, having an additional review would be much appreciated. No rush, but would you be willing to have a look to his @jonmmease?

binste:
Thanks @mattijn for the very helpful review! :) Sounds good, I'll see if someone else finds a moment to have a look at the changes and in the meantime, I'll already try to prepare the final changes in a separate PR so that we can mark Altair as typed.

jonmmease (approved):
Wow, thanks for the epic work on this @binste! This is a great step forward for Altair as a foundational library in the PyData ecosystem.

After playing with the branch for a bit, I only came up with one non-blocking suggestion:

Would it be possible to replace `BinParams`, `ImputeParams`, `TitleParams` with the shorter `Bin`, `Impute`, and `Title` aliases?

Screenshot 2023-11-22 at 6 28 41 AM

binste:
Thanks @jonmmease for the kind words and the review! :)

I agree that this would be nicer. Unfortunately, these aliases are defined in `api.py` and we'd need them in the autogenerated files -> Leads to circular imports. It's the same reason why we have to use the `_Parameter` protocol instead of `Parameter` directly.

PR #3218, opened 2023-10-06
[Doc] Add integers to four digit year format example

### LINKED ISSUE #1365: Better deal with year integer?
Hi Jake,

https://stackoverflow.com/questions/55082460/strange-year-values-on-x-axis/55082695#55082695 seems like a common problem that users doesn't know that integer could be a year.  

We should at least document about it in https://altair-viz.github.io/user_guide/times_and_dates.html for now. 

I'm also filing an issue in Vega to see if we can fix it there.  https://github.com/vega/vega/issues/1681

### LINKED ISSUE #3162: Add an example of reading in years in a convenient manner
As per @mattijn suggestion here https://github.com/altair-viz/altair/discussions/3140#discussioncomment-6714090. Since it is hard to discover it would be great to include it in the docs.

ChiaLingWeng:
This will close #3162 .
Add an example as below comparing using quantitative, ordinal and temporal data type.
image

With explanation:
To plot the year data as four digit format; i.e. without thousand separator, we recommend converting integer to string data type first, then storing the data as temporal, since directly convert integer to temporal data type would result in an error.

And also close #1365
image

### DIFF
--- a/doc/user_guide/encodings/index.rst
+++ b/doc/user_guide/encodings/index.rst
@@ -170,7 +170,7 @@ Effect of Data Type on Color Scales
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 As an example of this, here we will represent the same data three different ways,
 with the color encoded as a *quantitative*, *ordinal*, and *nominal* type,
-using three vertically-concatenated charts (see :ref:`vconcat-chart`):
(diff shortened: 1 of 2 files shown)


joelostblom:
Thanks for this @ChiaLingWeng ! I made some modifications directly to the PR and will go ahead and merge.

PR #3220, opened 2023-10-09
Do not display search keyboard shortcut on home page

joelostblom:
There are a few style changes in the latest pydata sphinx theme. The Altair docs home page looks like this:

image

After the changes in the latest pydata version it will look like this:

image

I like the centered navigation titles, but think the different color banner goes less well with the altair logo (although I don't feel strongly about this), and I definitely don't like that a keyboard shortcut is being displayed in the search box by default. After this PR, the home page looks like this:

image

Once you click the search box, the keyboard shortcut still shows up:

image

I tried changing back the header color also, but couldn't quite figure out how to do it and want to avoid putting too much time into changing CSS rules. I tried something like this

```css
/* with and without ":after" */
.bd-header-announcement:after {
    background-color: #daebf1 !important;
}
```

### DIFF
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -70,9 +70,16 @@ properly displayed on mobile devices and not restricted to 20% */
   vertical-align: text-bottom;
   margin-right: 3px;
 }
-  
+
 .full-width-plot {
   width: 100%;
+}
+
+/* This hides the Ctrl + K from the search box on the start page
+ * to make it less distracting on the home page.
+ * The shortcut still shows up when clicking the search box */
+.search-button-field > .search-button__kbd-shortcut {
+    display: none;
 }
 
 /* Configurations for the start page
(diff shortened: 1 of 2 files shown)


doc/_static/custom.css
binste, line 1: Can you use the latest version as a minimum requirement in pyproject.toml to make sure that we use it from now on? `"pydata-sphinx-theme>=0.14.1",`

binste (approved):
Thanks @joelostblom! I agree that it looks better without the shortcut and that the blue color is better for our docs.

The CSS you provided works well for me if I include it in `custom.css`:

```css
.bd-header-announcement:after {
    background-color: #daebf1 !important;
}
```

Could you give it another try? Maybe your browser cached the css? Otherwise, looks good to me.

joelostblom:
You're right, it must have been cached in my browser. I didn't even need the `after` part. Merging when tests are passing.

PR #3231, opened 2023-10-20
docs: pandas -> Pandas

domoritz:


mattijn:
> When using the project name pandas, please use it in lower case, even at the beginning of a sentence.
See https://pandas.pydata.org/about/citing.html

domoritz:
Today I learned. I always spelled it wrong then. 

Note that this page has other mentions of pandas and spells it as "Pandas". So we will want to update those, right?

mattijn:
Trying to give them the right credits.. I also think they deserve a capital p;). 
I did a quick search and had 51 hits in the Altair docs where we use pandas with a capital p, so at least some of these require changing indeed..

domoritz:
Thanks for making the changes. It looks odd to me as well but I prefer consistency over either option.

PR #3233, opened 2023-10-20
doc: update numpy-tooltip-images.rst

mattijn:
I see a future warning in Github Action for this page: https://altair-viz.github.io/case_studies/numpy-tooltip-images.html

> ```
> FutureWarning: DataFrame.applymap has been deprecated. Use DataFrame.map instead.
> ```

My personal preference is to avoid usage of map/applymap/apply lambda functions, but I changed both `applymap` and `apply` to `map`. Not sure if that is right.

Btw, where is this tutorial about? _Area of blobs_ what does that mean?

### DIFF
--- a/doc/case_studies/numpy-tooltip-images.rst
+++ b/doc/case_studies/numpy-tooltip-images.rst
@@ -8,9 +8,11 @@ you’ll learn how to display images stored as Numpy arrays
 in tooltips with any Altair chart.
 
 First,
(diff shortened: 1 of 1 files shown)


joelostblom (approved):
Thanks and sorry for missing this PR originally.  I added a clarification of what "blob" is referring to (the amorphous objects of white pixels in the images) and will merge when tests are passing

PR #3235, opened 2023-10-20
maint: GitHub action include `python-version` 3.12

mattijn:
Python 3.12 was released recently, this PR updates our GitHub action to also test against python 3.12. This PR is currently failing because pyarrow does not yet publish builds for python 3.12, see https://github.com/apache/arrow/issues/37880. 

By including `python-version` 3.12 I dropped 3.9 to avoid the number of tests that are run in parallel. So it becomes `python-version: ["3.8", "3.10", "3.11", "3.12"]`, not sure if that is a good idea.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.8", "3.10", "3.11", "3.12"]
         jsonschema-version: ["3.0", "latest"]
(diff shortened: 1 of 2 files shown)


jonmmease:
> By including python-version 3.12 I dropped 3.9 to avoid the number of tests that are run in parallel. So it becomes python-version: ["3.8", "3.10", "3.11", "3.12"], not sure if that is a good idea.

👍  Yeah, I think that's reasonable

binste:
Thanks! Sounds good to me too.

binste:
Could you add 3.12 also under `classifiers` in `pyproject.toml`?

mattijn:
Tests passing, merging

PR #3243, opened 2023-10-29
maint: change to ruff for formatting

### LINKED ISSUE #3241: investigate if we can switch from `black` to `ruff` for formatting
With the recent introduction of ruff formatter ([docs](https://docs.astral.sh/ruff/formatter/)), we can investigate if we can reduce the number of development dependencies by 1.

All of the `black` related config within `pyproject.toml` can be rewritten to `ruff` style config, but we also use `black` to format code within `tools/update_init_file.py` ([around here](https://github.com/altair-viz/altair/blob/main/tools/update_init_file.py#L60-L61)), as such:
```python
# Format file content with black
new_file_content = black.format_str("\n".join(new_lines), mode=black.Mode())
```
I don't think this usage of `black` is possible with `ruff` yet, but please correct me if I'm wrong.

([initial attempt in this branch](https://github.com/altair-viz/altair/compare/main...ruff-format))

mattijn:
Close https://github.com/altair-viz/altair/issues/3241.

With this PR we switch from black to ruff for formatting and so we can reduce our dependencies by 1 (thank you `black`, you were awesome). There are a few minimal formatting changes with associated with this PR. 
I excluded `tests/test_transformed_data.py` for now as I didn't like the introduced reformatting there.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
-    name: black-ruff-mypy
+    name: ruff-mypy
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python 3.10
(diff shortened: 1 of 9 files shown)


binste (approved):
LGTM, thanks @mattijn!

PR #3244, opened 2023-10-30
Support saving to PDF with VlConvert 1.0

jonmmease:
VlConvert 1.0 was just released and it includes support for saving charts to PDF. This PR bumps the minimum version of vl-convert-python to 1.0.0 and adds support for saving to PDF with the `vl-convert` engine.

This brings vl-convert to feature parity with `altair_saver`, so I added a deprecation warning when the altair_saver engine is used.  I'm thinking we could drop altair_saver support in Altair 5.3.0.


mattijn:
Thanks for this PR @jonmmease, quite awesome to see all of this progress around vl-convert! 
I run into a few issues with this PR. When I run:
```python
import altair as alt
import pandas as pd

source = pd.DataFrame({
    'b': [28, 55, 43, 91, 81, 53, 19, 87, 52],
})

chart = alt.Chart(source).mark_bar().encode(x='a', y='b')
chart.save('chart.pdf')
```
I get the following error:

> ```cmd
> ValueError: Saving charts in 'pdf' format requires the altair_saver package: see http://github.com/altair-viz/altair_saver/
> ```
Which surprises me. I was expecting an error coming from the `import_vl_convert`:
```python
alt.utils._importers.import_vl_convert()
```
> ```cmd
> ImportError: The vl-convert-python package must be version 1.0.0 or greater. Found version 0.14.0
> ```
But if I do
```python
import vl_convert as vlc
print(vlc.__version__)
vlc.vegalite_to_pdf(chart.to_json())
```
I get a successful pdf:

> ```cmd
> 0.14.0
> b'%PDF-1.7...'
> ```

We should probably be raising an `importlib.metadata.PackageNotFoundError` around [here](https://github.com/altair-viz/altair/blob/main/altair/utils/_importers.py#L43-L44) and [there](https://github.com/altair-viz/altair/blob/1ceaa0f1ecb65b882d65ce56d60cbad7aa3e52d1/altair/utils/mimebundle.py#L197), so this `ImportError` of a too low package version is properly raised and not caught in the try..except clause.

And based on the vl-convert v0.14.0 [release](https://github.com/vega/vl-convert/releases/tag/v0.14.0) page, I'm wondering if the bump to vl-convert 1.0 is already required part of this PR?

mattijn:
I add a commit: https://github.com/altair-viz/altair/pull/3244/commits/07ae53a7dbd9d786e400cf194eee8539aa7e39f0. think a `RuntimeError` is probably better over a `ImportError` when a version of a soft dependency is detected which is too low. This will presents the correct error with above example as well.

jonmmease:
Thanks for the review and for trying it out @mattijn!  I'm happy with change you made. Are you ready for this to be merged?

> And based on the vl-convert v0.14.0 [release](https://github.com/vega/vl-convert/releases/tag/v0.14.0) page, I'm wondering if the bump to vl-convert 1.0 is already required part of this PR?

Yeah, for PDF vl-convert-python 0.14.0 would be ok (though there was at least one PDF bug fix since 0.14.0), but I want to open a PR with offline HTML support after this and that will require 1.0.0.

mattijn:
Yes, fine with merging. In a follow up PR will also try to improve the current error message if a too low version is detected. I like the part that you get some info how to upgrade using pip or conda, but currently you only get to see this if you don't have the soft dependency installed. Also mentioning that it is a soft dependency is helpful I think.
Thanks @jonmmease! Looking forward to offline support🥳

PR #3248, opened 2023-11-04
Type hint utils/save.py and utils/mimebundle.py

binste:
Adds type hints to utils/save.py and utils/mimebundle.py. After this PR and #3208, there are only a few minor steps left in #2951 🥳


mattijn:
Thanks @binste! Looks good to be restrictive on output types. Almost there indeed!

PR #3249, opened 2023-11-04
doc: barchart highlighting values beyond a single threshold

### LINKED ISSUE #2996: Add VL example 'Bar Chart Highlighting Values beyond a Threshold' to gallery
I find https://vega.github.io/vega-lite/examples/layer_bar_annotations.html an interesting example which could be added to the Altair Example Gallery:

image

I already started [in this comment](https://github.com/altair-viz/altair/issues/2915#issuecomment-1484121673) but the code needs to be cleaned up a bit.

mattijn:
This PR fix https://github.com/altair-viz/altair/issues/2996.

There is still a bit of discussion how to introduce a barchart with multiple thresholds, see https://github.com/altair-viz/altair/pull/3227, but I think we now have an example-worthy method for a barchart highlighting values beyond a single threshold.

### DIFF
--- a/tests/examples_arguments_syntax/bar_chart_with_single_threshold.py
+++ b/tests/examples_arguments_syntax/bar_chart_with_single_threshold.py
@@ -8,17 +8,17 @@ import pandas as pd
 import altair as alt
 
 source = pd.DataFrame({
-    'Day': [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15], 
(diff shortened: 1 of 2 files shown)


binste (approved):
Thank you @mattijn ! Could you align all quotation marks to the same style (I prefer double quotations but we're not consistent with it in the docs)? Apart from that, LGTM.

mattijn:
Thanks for the review @binste! Modified as requested. Merging.

PR #3251, opened 2023-11-05
Use vl-convert for offline html export

jonmmease:
This PR takes advantage of HTML bundling support in vl-convert 1.0 to replace the dependency on altair_viewer when calling `chart.save('chart.html', inline=True)` to create an HTML file with inline JavaScript dependencies.

The vl-convert `javascript_bundle()` function returns a bundle with Vega, Vega-Lite, and Vega-Embed which can be included in a `<script>` tag. This function actually performs bundling and minification and takes 1-2s the first time it's called. The result is cached and subsequent calls are much faster.

I had to update the vl-convert minimum dependency to 1.0.1 because I discovered a bundling bug during integration.


mattijn:
Thanks @jonmmease! 
Would it be possible to update the allowed arguments and docstrings [here](https://github.com/altair-viz/altair/blob/main/altair/vegalite/v5/api.py#L1077-L1112), for `chart.save()`, using the complete arguments and docstrings from [here](https://github.com/altair-viz/altair/blob/main/altair/utils/save.py#L71-L130) (`save` function within `altair.utils.save`)?

Maybe [this](https://github.com/altair-viz/altair/blob/main/altair/vegalite/v5/api.py#L1035-L1043), for `chart.to_html()`, is incomplete too?

I noticed this when there was no insight on the new argument in the tool tip when trying to save a chart:
image

jonmmease:
Thanks for the review @mattijn, this is ready for another look

mattijn (approved):
Looks good now! This is great @jonmmease!

Issue #2875, opened 2023-02-06
"Open in Vega Editor" does not work in VS Code Notebooks


JannisKoeksel:
The actions View Source, View Compiled Vega and Open in Vega Editor provided next to the plot do not work in VS Code. 
The links redirect to the top of the notebook, so I'm guessing the event handlers do not work as intended.
I am using Version 4.2.2.

mattijn:
Thanks for raising @JannisKoeksel! I noticed the same. Are you able to figure out why this is not working? Is there documentation by MS VSCode that can support this issue how we could improve this?

ChristopherDavisUCI:
Does this look related? https://github.com/microsoft/vscode-jupyter/issues/11294

JannisKoeksel:
So there are two problems: 
1) Opening links from VS Code via window.open does not work. I have not found any documentation about this, but all my attempts have failed.
2) Vega Editor gets its definition via window.postMessage 

Problem one could be solved by by e.g. using ipywidgets.widgets.Button and displaying them instead of the vega-embed actions. Then using python webbrowser to open a new tab.
Although I personally find the widgets aesthetically unpleasing they should work for "View Source" and "View Compiled Vega".

Regarding the second issue, the only viable solution I can think of is to dynamically write a html file or use data: base64 encoding with the definition and embedding vega editor as an iframe. Then the postMessage mechanism should work again. 
There seams to be an issue with opening data-links and vscode-remote-links thought.  And I have not found a solution that does not include running a server on the remote host or syncing files. 
Sadly I am not very familiar with the internals of VS Code so best I can do is guessing.

@ChristopherDavisUCI yes is seams to be the same issue.  I would really like to use this feature so disabling it is my least favorite solution.

ChristopherDavisUCI:
This definitely seems worth looking into.

As a not-very-satisfying workaround, do you know about `c.to_json()` or `print(c.to_json())` or `pprint(c.to_json())`, where `c` is the Altair Chart?  That should produce something that can be pasted into the Vega Editor.

JannisKoeksel:
No I wasn't aware, but  `print(c.to_json())`  works for me. For my use case the extra steps required aren't to bad either.

joelostblom:
I opened https://github.com/microsoft/vscode-jupyter/issues/13261 where we can continue this discussion. Since it is an issue with VS Code and not with Altair, I will close this issue.

PR #3252, opened 2023-11-05
Support converting a Chart to a Vega editor URL

### LINKED ISSUE #3188: Feature: Shareable link to eg https://vega.github.io/editor
I filed the main issue at https://github.com/vega/editor/issues/1365 because I thought that was a better place for it, but just posting this here for visibility. Please chime in there with your thoughts!

jonmmease:
Closes https://github.com/altair-viz/altair/issues/3188

This PR adds a `Chart.to_url` method that uses VlConvert to generate a Vega editor URL that opens the chart's specification in the online Vega editor.

```python
import altair as alt

chart = alt.Chart(data.cars.url).mark_point().encode(
    x='Horsepower:Q',
    y='Miles_per_Gallon:Q',
    color='Origin:N'
)

print(chart.to_url())
```
https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAZgAY2AacaYsiygAlMiRoVYcAvpO50AhoTl4QpAE4AbFCDGEADpWQB6Q2DpQAdACtKdTOrhpV5qJkKGougLaG0mBHIBaeUVKV0oAATQARnMAJgBOczZDYLkTOVVKK0poEBkQTwyAa2VCAE9dTC1dCBJxfMwoSDoSJFQedQhVZXg7Oi0AeVVEEhBucsqtKAhPEjlNfIAPHqx1fuQQQS7QmuxMbvGKqo2AR1I5IjhFYl887jKVvq0AWTh1TEoAfUrVT4BxeadKBjEATY4gM4XYjXBxVaTcAAklDAjEwhS0On0Rh8fjk5gQV0YpAARuY4BBDMjUYUcf4AvZCJgfABWcxRVkxay5SRAA

cc @NickCrews

### DIFF
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -21,6 +21,7 @@ from ... import utils, expr
 from .display import renderers, VEGALITE_VERSION, VEGAEMBED_VERSION, VEGA_VERSION
 from .theme import themes
 from .compiler import vegalite_compilers
+from ...utils._importers import import_vl_convert
 from ...utils._vegafusion_data import (
     using_vegafusion as _using_vegafusion,
     compile_with_vegafusion as _compile_with_vegafusion,
@@ -1054,6 +1055,25 @@ class TopLevelMixin(mixins.ConfigMethodMixin):
             fullhtml=fullhtml,
             requirejs=requirejs,
         )
+
+    def to_url(self, fullscreen: bool = False):
+        """Convert a chart to a URL that opens the chart specification in the Vega chart editor
+
+        Convert a chart to a sharable URL that opens the chart in the online Vega chart editor.
+        The chart specification (including any inline data) is encoded in the URL.
+
+        This method requires the option vl-convert-python dependency
+
+        Parameters
+        ----------
+        fullscreen : bool
+            If True, editor will open chart in fullscreen mode. Default False
+        """
+        vlc = import_vl_convert()
+        if _using_vegafusion():
+            return vlc.vega_to_url(self.to_dict(format="vega"), fullscreen=fullscreen)
+        else:
+            return vlc.vegalite_to_url(self.to_dict(), fullscreen=fullscreen)
 
     def save(
         self,
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -1057,11 +1057,9 @@ class TopLevelMixin(mixins.ConfigMethodMixin):
 
     def to_url(self, fullscreen: bool = False) -> str:
         """Convert a chart to a URL that opens the chart specification in the Vega chart editor
-
-        Convert a chart to a sharable URL that opens the chart in the online Vega chart editor.
         The chart specification (including any inline data) is encoded in the URL.
 
-        This method requires the option vl-convert-python dependency
+        This method requires that the vl-convert-python package is installed.
 
         Parameters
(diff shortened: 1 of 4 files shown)


altair/vegalite/v5/api.py
```diff
@@ -1055,6 +1055,27 @@ def to_html(
             requirejs=requirejs,
         )
 
+    def to_url(self, fullscreen: bool = False):
```
binste, line 1058:
```suggestion
    def to_url(self, fullscreen: bool = False) -> str:
```

altair/vegalite/v5/api.py
```diff
@@ -1055,6 +1055,27 @@ def to_html(
...
+        """Convert a chart to a URL that opens the chart specification in the Vega chart editor
+
+        Convert a chart to a sharable URL that opens the chart in the online Vega chart editor.
+        The chart specification (including any inline data) is encoded in the URL.
+
+        This method requires the option vl-convert-python dependency
```
binste, line 1064: Removes duplication and improves language
```suggestion
        """Convert a chart to a URL that opens the chart specification in the Vega chart editor
        The chart specification (including any inline data) is encoded in the URL.

        This method requires that the vl-convert-python package is installed.
```

binste (approved):
Looks great @jonmmease. Not sure if that was the intention but it helps with #2875 which is nice.

Only added two minor comments.

jonmmease:
Thanks for the review and suggestions @binste, I've accepted them.

> Not sure if that was the intention but it helps with https://github.com/altair-viz/altair/issues/2875 which is nice.

I hadn't come across that issue, but yeah this will provide a reasonable workaround for Visual Studio Code users.

altair/vegalite/v5/api.py
```diff
@@ -1055,6 +1055,25 @@ def to_html(
             requirejs=requirejs,
         )
 
+    def to_url(self, fullscreen: bool = False) -> str:
```
NickCrews, line 1058: Consider: forcing all args to be keyword-only so that in the future, if we add more params to this function, they are not order dependent and we can re-order them as needed without breaking anyone. Not sure if this is overkill. 

to_url(False) is also not very obvious what False does. Requiring a kwarg improves that.

On the other hand, requiring a kwarg is more typing.

I think I fall 60% in favor of requiring kwarg but am fine with keeping it as is

NickCrews (commented):
Looks great! Thank you!

Could consider adding auto copy to clipboard support, but I lean towards that being out of scope

altair/vegalite/v5/api.py
jonmmease: Good idea! Done in [5d440f9](https://github.com/altair-viz/altair/pull/3252/commits/5d440f98a9cceadd88cb51fa4d1f78556a067b0a)

jonmmease:
> Could consider adding auto copy to clipboard support, but I lean towards that being out of scope

Yeah, this is a possible future extension. Another idea is to optionally use the Python [webbrowser](https://docs.python.org/3/library/webbrowser.html) module to open the URL in the editor.

mattijn:
Thanks for the PR @jonmmease and reviews @NickCrews, @binste! This is great and valuable (especially in VSCode, like Stefan said.)

PR #3256, opened 2023-11-08
[Doc] Add scatter plot with shaded area example

### LINKED ISSUE #2031: How to build a classification boundary with Altair?
Omitting the dot points of the data, how to build a classification boundary like this image?
classifier decision boundary

ChiaLingWeng:
This will close #2031.
 Also add how to use area mark to create more flexible area (e.g. area between 2 lines),
but may need to find more proper way to show these 2 shaded area in one plot.
image


joelostblom (approved):
Looks great, thanks again @ChiaLingWeng ! I deleted the example in the method based folder since we only need that if there are differences in the code compared to the example in the argument-based folder (in this case they were the same)

PR #3269, opened 2023-11-17
Fix broken JupyterWidget chart by pinning Vega

### LINKED ISSUE #3268: Vega 5.26.1 breaks `JupyterChart`
First posted in anywidget:

- https://github.com/manzt/anywidget/issues/381

It seems there is a bug in latest patch release of Vega (https://github.com/vega/vega/pull/3829), that breaks the altair "hello, world":

```python
import altair as alt

source = data.cars()

    x="Horsepower",
    y="Miles_per_Gallon",
    color=alt.condition(brush, "Origin", alt.value("lightgray"))
).add_params(
    brush
)

bars = alt.Chart(source).mark_bar().encode(
    y="Origin",
    color="Origin",
    x="count(Origin)"
).transform_filter(
    brush
)
```

```python
alt.jupyter.JupyterChart(points & bars)
```

image

This occurs because the  JS `embed` dependency is coming from a CDN and `vega-embed`'s package.json says the patch fix is within semver. The CDN [esm.sh](https://esm.sh/) let's you pin an exact dependency version in your import URL, so you can fix this bug by changing the import in 

```diff
-- import embed from "https://cdn.jsdelivr.net/npm/vega-embed@6/+esm";
++ import embed from "https://esm.sh/vega-embed@6?deps=vega@5.26.0";
```

jonmmease:
Closes #3268

### DIFF
--- a/NOTES_FOR_MAINTAINERS.md
+++ b/NOTES_FOR_MAINTAINERS.md
@@ -49,6 +49,15 @@ of some docstrings.
 Major version updates (e.g. Vega-Lite 1.X->2.X) have required substantial
 rewrites, because the internal structure of the schema changed appreciably.
 
+## Updating the Vega-Lite in JupyterChart
+To update the Vega-Lite version used in JupyterChart, update the version in the esm.sh URL in `altair/jupyter/js/index.js`.
+
+For example, to update to Vega-Lite 5.15.1, Vega 5 and Vega-Embed 6, the URL should be:
+
+```javascript
+import embed from "https://esm.sh/vega-embed@6?deps=vega@5&deps=vega-lite@5.15.1";
+```
+
 ### Updating vl-convert version bound
 When updating the version of Vega-Lite, it's important to ensure that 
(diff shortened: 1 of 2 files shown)


binste (approved):
Thanks @jonmmease! 

Does this mean that we continuously need to bump Vega versions? Do you know of a good way to test when this is resolved so we can stop pinning it again? Could you add a note to `NOTES_FOR_MAINTAINERS.md` that this version should be kept up-to-date?

mattijn:
Was wondering the same. How does this interfere with:
```python
alt.VEGAEMBED_VERSION, alt.VEGA_VERSION  # alt.VEGALITE_VERSION, 
```

> ```cmd
> ('6', '5')
> ```

jonmmease:
I'll take another look at this when I update Vega-Lite. It might work to unpin the version and switch to the latest from https://esm.sh/.

jonmmease:
I updated the esm.sh URL to use the same vega embed, vega, and vega-lite versions as the rest of Altair.  I added a note to `NOTES_FOR_MAINTAINERS` with instructions for updating this URL when we update vega-lite.

Issue #1232, opened 2018-11-27
Interactive charts - from a map!


AnneEstoppey:
Hello there,

I was wondering if it would be possible to have interactive chart also working on a map.

Meaning that by selecting a group of points on my map, I would be able to see the corresponding data in my bar chart, for instance. Similarly as below, just that the scatter point would be, in fact, a map.

I don't think I have seen that in any other package (but I should perhaps investigate). I think it would be a really great additional interactive feature.

If you need sample datasets (in my case in the exploration well public data of Norway), please let me know.

Thanks in advance,
Cheers
Anne

![interactive_chart](https://user-images.githubusercontent.com/35219455/49073610-23a3ee80-f233-11e8-870a-cc659cd83490.jpg)

jakevdp:
Currently, no, that's not possible. This would require adding interval selection support to geographic projections in Vega-Lite. This is an open issue in Vega-Lite: https://github.com/vega/vega-lite/issues/3306

jakevdp:
though note that some interactions (based on single/multi selections rather than interval selections) do work currently, e.g. https://altair-viz.github.io/gallery/us_state_capitals.html

AnneEstoppey:
Oh ok I see. Hmm. I hope 'they' find a way. That would be really cool. 

Yes the utility from your link, I can do that in mapbox. 
I am intensively using mapbox for more elaborate maps. 
Maybe I should also ask them about interactivity between a map and a graph...

Cheers
Anne

jakevdp:
No, we don't really have control over any of that in Altair. Altair just produces JSON specifications that are interpreted and rendered by the vega stack... so all aspects of the rendering are in the domain of vega & vega-lite.

AnneEstoppey:
Ok, I understand, thanks a lot for information, that makes sense.

I'll check the vega github :-)

Cheers
Anne

joelostblom:
@AnneEstoppey is this what your are looking for https://stackoverflow.com/a/65948144/2166823?

binste:
This was added in Vega-Lite 5.7 and hence is available in Altair 5. See https://github.com/vega/vega-lite/pull/6953 for a Vega-Lite example.

PR #3275, opened 2023-11-25
doc: add example, interval selection on a map

mattijn:
This PR adds a new example showing how one can use the selection interval on a map:

image

ref: #1232

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,8 @@ dev = [
     "types-setuptools",
     "pyarrow>=11",
     "vegafusion[embed]>=1.4.0",
-    "anywidget"
+    "anywidget",
+    "geopandas",
 ]
 doc = [
     "sphinx",
@@ -82,7 +83,6 @@ doc = [
     "numpydoc",
     "pillow>=9,<10",
     "pydata-sphinx-theme>=0.14.1",
-    "geopandas",
     "myst-parser",
     "sphinx_copybutton",
     "sphinx-design",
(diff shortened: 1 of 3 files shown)


mattijn:
@binste, do you know how I can avoid this mypy error?
```
error: Cannot find implementation or library stub 
for module named "geopandas"  [import-not-found]
```

binste:
In the `lint` workflow, only normal and `dev` dependencies are being installed. `geopandas` is a doc dependency so far and hence, mypy can't find it. As we now use it in this example which is also a test, geopandas should be moved from `doc` dependencies to `dev` in pyproject.toml. As geopandas is not a typed package as far as I know, you'll first get another error:
```
tests/examples_methods_syntax/interval_selection_map_quakes.py:11: error: Skipping analyzing "geopandas": module is installed, but missing library stubs or py.typed marker  [import-untyped]
```
which you can solve by adding `geopandas.*` to the modules under `[[tool.mypy.overrides]]` in `pyproject.toml`.

These two steps should make mypy happy :)

binste (approved):
Great example! Could you harmonize the usage of single and double quotes to one of them (my preference is double quotes). Apart from that, looks good to me.

mattijn:
Thanks for the review @binste. Applied your suggestions. Tests are happy, so merging.

binste:
I like the idea of "happy" tests. Petition for GitHub to replace the red X and green check mark with grumpy and happy smilies ;)

PR #3278, opened 2023-12-02
Update save.py to use utf-8 instead of None per default

franzhaas:
I experience issues with the Null on windows. It appears as if when saving HTML, the template is already utf-8 and when writing the files with the system default code page, there are encoding issues.

Alternatively I see another options to solve my problem.: 

set it only in the

`elif format == "html":`

branch...

### DIFF
--- a/altair/utils/save.py
+++ b/altair/utils/save.py
@@ -142,7 +142,12 @@ def save(
 
         if format == "json":
             json_spec = json.dumps(spec, **json_kwds)
-            write_file_or_filename(fp, json_spec, mode="w", encoding=kwargs.get("encoding", "utf-8"))
+            write_file_or_filename(
+                fp,
+                json_spec,
+                mode="w",
+                encoding=kwargs.get("encoding", "utf-8")
+            )
         elif format == "html":
             if inline:
                 kwargs["template"] = "inline"
--- a/altair/utils/save.py
+++ b/altair/utils/save.py
@@ -142,7 +142,7 @@ def save(
 
         if format == "json":
             json_spec = json.dumps(spec, **json_kwds)
-            write_file_or_filename(fp, json_spec, mode="w", encoding="utf-8")
+            write_file_or_filename(fp, json_spec, mode="w", kwargs.get("encoding", "utf-8"))
         elif format == "html":
             if inline:
                 kwargs["template"] = "inline"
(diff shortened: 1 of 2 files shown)


jonmmease:
Thanks for the contribution @franzhaas. Setting the default encoding to `utf-8` when saving our text-based formats (svg and html) makes sense. The tests are failing because it's not valid to set an encoding when saving to binary formats (png and pdf).

So I think you'll need to move the default encoding up a level and set it for only the `svg` and `html` branches.

franzhaas:
I think it's now ok.

...I need to get into the habit of using my account to run CI...

altair/utils/save.py
```diff
@@ -157,7 +157,9 @@ def perform_save():
                 json_kwds=json_kwds,
                 **kwargs,
             )
-            write_file_or_filename(fp, mimebundle["text/html"], mode="w")
+            write_file_or_filename(
+                fp, mimebundle["text/html"], mode="w", encoding="utf-8"
```
jonmmease, line 161:
```suggestion
                fp, mimebundle["text/html"], mode="w", encoding=kwargs.get("encoding", "utf-8")
```

altair/utils/save.py
```diff
@@ -142,7 +142,7 @@ def perform_save():
 
         if format == "json":
             json_spec = json.dumps(spec, **json_kwds)
-            write_file_or_filename(fp, json_spec, mode="w")
+            write_file_or_filename(fp, json_spec, mode="w", encoding="utf-8")
```
jonmmease, line 145:
```suggestion
            write_file_or_filename(fp, json_spec, mode="w", kwargs.get("encoding", "utf-8"))
```

franzhaas:
I think it is ok now...

I personally would prefer to not have the encoding parameter explicit, working by default and the ability to pass an IO object when specific encodings are necessary appear to be a  good and compact solution too me.

jonmmease:
I think this example is so slow because it ends up needing to render 50k tick labels since the x-encoding is nominal. Is this representative of your data?  I'm wondering if you can reproduce this behavior for a chart that doesn't have so much text, that might help narrow down where the memory is still being held.

franzhaas:
I am a bit confused about that one... I cant remember to have touched this file... and i see nothing in the logs... also it didn't trigger on my own CI run...

but sure i fixed it...

PR #3282, opened 2023-12-13
doc: histogram with gradient color

mattijn:
This PR adds an example of a histogram with a gradient color. Sometimes it is desirable to emphasise that higher/lower values are more/less favourable (here green/red).

image

### DIFF
--- a/tests/examples_arguments_syntax/histogram_gradient_color.py
+++ b/tests/examples_arguments_syntax/histogram_gradient_color.py
@@ -2,7 +2,7 @@
 Histogram with Gradient Color
 -----------------------------
 This example shows how to make a histogram with gradient color. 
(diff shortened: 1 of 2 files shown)


joelostblom:
Pretty =) Could we switch the color scheme to something that is easier to differentiate for people with red-green color vision deficiencies? Maybe `redyellowblue` or `pinkyellowgreen` (or really any of the other diverging ones)?

mattijn:
Thanks for the review @joelostblom! I've changed the example to use the scheme `pinkyellowgreen`, which is better for red-green color vision deficiencies: 
image

joelostblom (approved):
Great, thank you!

binste:
I've just used this in a chart, thanks! :)

PR #3297, opened 2023-12-23
Fix type checker errors for libraries such as Polars where Altair uses dataframe interchange protocol

### LINKED ISSUE #3294: Passing basic polars dataframes to alt.Chart results in Pylance errors
I took the default boilerplate chart code from an altair example and simply converted the `cars` Pandas DataFrame to a Polars one (for testing purposes as I am using Polars), and while the code appeared to still execute, Pylance is catching some type issues.  Restarted VSCode and still see the same problem:

image

binste:
Closes #3294 . As discussed in that issue, I switched to Protocols over ABC classes as other libraries such as Polars of course don't use our own classes but instead have their own implementations -> ABC classes won't type check but Protocols work. 

After that change Polars dataframes still did not satisfy mypy so I digged a bit into the interchange protocol. There are minor differences in how Pandas, Polars, etc. have implemented type hints for the protocol and sometimes even the protocol itself. For example, in Polars `CategoricalDescription.is_dictionary` has type `Literal[True]` as it is indeed always True in the case of Polars. However, the interchange protocol only requires `bool`. As [`TypedDict`s are invariant](https://peps.python.org/pep-0589/#type-consistency), this will not satisfy a type checker as the types would need to be exactly the same. As an example:

```python
from typing import TypedDict


class PolarsDict(TypedDict):
    y: int


class AltairDict(TypedDict):
    y: int


class AltairDict2(TypedDict):
    y: int | None


z: AltairDict = PolarsDict(x=1, y=2)  # this is fine
t: AltairDict2 = PolarsDict(x=1, y=2)  # this is not as the TypedDict is invariant
```

To not continuously have to deal with such minor differences, I removed all type hints including methods and attributes from the interchange protocols which are not relevant for Altair, e.g. methods which we do not use anywhere in our code base.

With the changes here, all examples in #3294 now pass mypy. If you see errors while testing, try deleting the `.mypy_cache` folder before running mypy.

### DIFF
--- a/altair/utils/_dfi_types.py
+++ b/altair/utils/_dfi_types.py
@@ -5,7 +5,7 @@
 #
 # These classes are only for use in type signatures
 import enum
-from typing import Any, Iterable, Optional, Tuple, Protocol
+from typing import Any, Iterable, Optional, Tuple, Protocol, Literal
 
 
 class DtypeKind(enum.IntEnum):
@@ -120,8 +120,6 @@ class DataFrame(Protocol):
     ``__dataframe__`` method of a public data frame class in a library adhering
     to the dataframe interchange protocol specification.
     """
-
-    version: int = 0  # version of the protocol
 
     def __dataframe__(
         self, nan_as_null: bool = False, allow_copy: bool = True
--- a/altair/utils/_dfi_types.py
+++ b/altair/utils/_dfi_types.py
@@ -7,19 +7,6 @@ import enum
 from typing import Any, Dict, Iterable, Optional, Sequence, Tuple, TypedDict, Protocol
 
 
-class DlpackDeviceType(enum.IntEnum):
-    """Integer enum for device type codes matching DLPack."""
-
-    CPU = 1
-    CUDA = 2
-    CPU_PINNED = 3
-    OPENCL = 4
-    VULKAN = 7
-    METAL = 8
-    VPI = 9
-    ROCM = 10
-
-
 class DtypeKind(enum.IntEnum):
     """
     Integer enum for data types.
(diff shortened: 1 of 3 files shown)


binste:
Hmm, pyright is not yet happy. I'll see if I can figure out why.

binste:
Now, pyright is happy as well :) The last issue I fixed comes actually more from the Polars codebase where they define `DataFrame.version` as a property in an underlying Protocol but then in `PolarsDataFrame` define it as an attribute. Pyright still sees it as a property which does not go along with the definition of it as an attribute in our own codebase, see https://github.com/microsoft/pyright/issues/2204. In short, I removed `version` from the `DataFrame` protocol as we don't need it anyway.

I also had to add a mroe explicit signature for `__dataframe__`, see https://github.com/microsoft/pyright/issues/2204 for more context if needed.

jonmmease (approved):
Works for me. Thanks for digging through this!

mattijn:
Thanks @binste, can we add a note in the release notes? Not sure if it counts as an enhancement or bug fix.

binste:
---
For reference, here the code I used to test the type hints:

test_types.py:
```python
import polars as pl
import numpy as np
import altair as alt
import pyarrow as pa
import pandas as pd

content = {
    "a": range(5),
    "b": np.random.rand(5),
}
df_pl = pl.DataFrame(content)
df_pd = pd.DataFrame(content)
df_pa = pa.Table.from_pydict(content)

alt.Chart(df_pl).mark_point().encode(
    x="a",
    y="b",
)
alt.Chart(df_pd).mark_point().encode(
    x="a",
    y="b",
)
alt.Chart(df_pa).mark_point().encode(
    x="a",
    y="b",
)
```

```bash
mypy test_types.py
pyright test_types.py
```

PR #3299, opened 2023-12-23
Docs: Add section on displaying Altair charts in dashboards

binste:
Adds a very brief section on dashboards to give users a quick overview which frameworks have support for Altair. These are the ones I'm aware off but please let me know if I missed anything!

This was previously discussed in https://github.com/altair-viz/altair/pull/2762 and https://github.com/altair-viz/altair/pull/2415.

image

### DIFF
--- a/doc/user_guide/display_frontends.rst
+++ b/doc/user_guide/display_frontends.rst
@@ -153,6 +153,9 @@ Package                                                                         
 `Streamlit <https://docs.streamlit.io/library/api-reference/charts/st.altair_chart>`_                                                ✔                                              
 ===================================================================================================================================  ===================================  =============================
 
+The above mentioned frameworks all require you to run a web application on a server if you want to share your work with others. A web application gives you a lot of flexibility, you can for example fetch data from a database based on the value of a dropdown menu in the dashboard. However, it comes with some complexity as well. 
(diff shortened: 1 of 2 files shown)


mattijn:
Is quarto useful to mention here?

binste:
Sure! I added a note on Quarto and Jupyter Book. I hope the distinction is clear enough but I think this section is also mainly useful for users who are already somewhat familiar with these topics and just want to look up which frameworks provide support. However, if you see an easy way how we could clarify the distinction, please let me know!

doc/user_guide/display_frontends.rst
```diff
@@ -135,10 +135,33 @@ Optionally, for offline rendering, you can use the mimetype renderer::
     # Optional in VS Code
     alt.renderers.enable('mimetype')
 
+.. _display_dashboards:
+
+Dashboards
+----------
+Altair is compatible with common Python dashboarding packages. Some of them even provide support for reading out :ref:`parameters <user-guide-interactions>` from the chart.
```
joelostblom, line 142: I would say "Many" instead of "Some" after all the fantastic recent progress on this front via JupyterChart and dash_vega_components =)

joelostblom (approved):
LGTM, thanks @binste ! One minor comment, but good with you merging as is too.

doc/user_guide/display_frontends.rst
binste: Fully agree! :)

binste:
Thanks for the inputs!

PR #3304, opened 2024-01-02
Add embed_options to JupyterChart and pass them through in "jupyter" renderer

jonmmease:
* Adds an `embed_options` traitlet to `JupyterChart` for configuring the Vega Embed options (e.g. whether to display actions). 

 * Passes through the active embed_options to `JupyterChart` when constructed in the `"jupyter"` renderer.


binste (approved):
Tested it and works like a charm, thanks Jon!

Issue #3290, opened 2023-12-20
Make signatures and docstrings easy to read again


joelostblom:
I think it is important that our docstrings are easy to read, so that it is easy to quickly look up help without going to the documentation site. Up until altair 5.1.2, the signatures were easy to parse for a human:

![image](https://github.com/altair-viz/altair/assets/4560057/4395bc00-fb5c-4cb8-9246-87da0eab4938)

Although it is unclear what all the `Undefined` means above, I get a clear, succinct overview of which arguments exist. Since v 5.2, the signatures are now notably harder to read and directly intimidating at first glance:

![image](https://github.com/altair-viz/altair/assets/4560057/c5087d68-664e-44b5-9f7d-a72787946ac0)
 
The story is similar for the docstring (although less severe), e.g. it used to be relatively easy to understand what the `sort` parameter does:

![image](https://github.com/altair-viz/altair/assets/4560057/de33591a-9b6f-4434-879e-5414bd91f7ab)

But now that section is it is non-really understandable:

![image](https://github.com/altair-viz/altair/assets/4560057/7245a4ee-6bb2-4da0-80fc-92f4f4f8a76c)

I think this stems mainly from https://github.com/altair-viz/altair/pull/3208 so tagging @binste @mattijn @jonmmease since we were all part of the discussion there. Does anyone have ideas of how we can keep the docstrings human readable while keeping the benefits from all the work @binste has put in on making altair properly typed? There are at least two angles here: the formatting of the docstring (which is mostly up to the IDE), and the content (which is entirely up to us). I would be in favor of keeping the old signatures in the help popup if possible; I'm unsure about what is best in the docstring itself.

I only included screenshots from jupyterlab above, but the situation seems to be similar in vscode:

![image](https://github.com/altair-viz/altair/assets/4560057/90096a3a-fe5e-4c30-9f02-5d98602654a7)

Our [online documentation](https://altair-viz.github.io/user_guide/generated/channels/altair.Color.html) retains the old easy to read signatures (I'm not sure if we added an exception or just haven't regenerated it since 5.2):

![image](https://github.com/altair-viz/altair/assets/4560057/14d41004-fbb3-4635-bbfb-0e71af43b74f)

Although the docstring for the parameters has updated to the new more verbose syntax:

![image](https://github.com/altair-viz/altair/assets/4560057/a0449c9a-ec52-4441-8fd7-8442cf233055)

binste:
[...]

PR #3307, opened 2024-01-04
Type hints: Simplify to improve user experiences

binste:
With this PR I'm trying to address the feedback raised in #3290. It relaxes some of the type hints to improve the readability of the type hints in an IDE such as VS Code and Jupyterlab as well as in the docstrings:

* Instead of showing all specific Altair classes, use SchemaBase from which all autogenerated classes inherit from anyway. As an example, in the definition of Color.aggregate, we could replace Aggregate, ArgmaxDef, ArgminDef, NonArgAggregateOp, and the different Dict[required[argmax]]... etc. with just dict and SchemaBase. Same in docstrings.

Old:
image

New:
image

* Fix deduplication of type hints and flatten out nested type hints to one level, i.e. we no longer have `Union`s within other `Union`s.
* Remove unnecessary class name + dict from beginning of docstrings:

Old:

image

New:

image


binste:
The latest commit adds back the full list of class names (instead of `SchemaBase`) in the docstrings based on the input from @joelostblom [here](https://github.com/altair-viz/altair/issues/3290#issuecomment-1877650610)

joelostblom (commented):
Thank you for working on this @binste ! I think your changes here makes the docstrings easier to read overall. A few comments:

> Fix deduplication of type hints and flatten out nested type hints to one level, i.e. we no longer have Unions within other Unions.

Nice! I noticed that `Literals` are now also combined which I think is helpful as well:

image

However, it seems like they are not combined in the docstring (if this is linked to that we expand `SchemaBase` in the docstring, then let's just leave it as is):

image

For some reason `Literal` don't seem to be combined in JupyterLab, but I'm guessing that is a jlab issue and nothing we can improve here:

image

One question regarding the first VSCode screenshot above, I noticed that we have the type `Sequence` repeated several times. Is that something we can change on our end so that it says `Sequence[str | bool | float | dict | SchemaBase]` instead?

---

Somewhat related, I noticed that the docstring says that we accept `Sequence[str]` as the `shorthand` to e.g. `alt.Color`, is this correct? When I try with a list of strings

```
alt.Chart().mark_point().encode(alt.Color(['test:Q']))
```

I get this error:

```
SchemaValidationError: '[{'field': 'test', 'type': 'quantitative'}]' is an invalid value for `color`. Valid values are of type 'object'.
```

joelostblom:
Another thing I noticed is that the signature to the `mark_` methods does not include the `Literal` for the color options:

image

whereas the docstring for `MarkDef` does:

image

I honestly don't think that's something we need to change, just something I noticed when reading through the help popups.

binste:
Thanks for the review Joel! Here some first feedback:

### Literals
In the code, the literals are not combined so seems like VS Code is doing this by itself. I'm somewhat indifferent if its easier to read if they are combined or not. If they are not combined, it gives a certain grouping but it's also more verbose. In any case, combining Literals in the type hints itself would require some larger rewriting of the code generation. If that's ok, I'd prefer to not tackle this in the current PR but happy to work on a follow up issue if you have a strong preference! :) Basically, when I'm flattening the type hints, a literal is already represented as a string, ready for code generation, e.g. `'Literal["ascending", "descending"]`. I'd need to change that to keep the meaning of `Literal` and it's content `"ascending", "descending"` separat.

### Sequence
Same story as with literals.

binste:
### Sequence[str] for shorthand
Struggling to remember the story here. In [`channels.py`](https://github.com/altair-viz/altair/blob/main/altair/vegalite/v5/schema/channels.py) around line 40, there is some code which processes lists and tuples. This is done for all channels which is probably why I added the type hints.However, seems like this wouldn't produce a valid spec for all channels as you've shown for color. It works for tooltip where you can do: `alt.Tooltip(["column_1", "column_2"])`. I'll open up a follow-up issue on this. Maybe we can infer from the VL schema when it is valid and when it isn't.

binste:
> Another thing I noticed is that the signature to the `mark_` methods does not include the `Literal` for the color options (but it includes `Literal` for other options such as the cursor):
> 
> ...
> 
> I honestly don't think that's something we need to change, just something I noticed when reading through the help popups.

Nice catch! Weird... It's the same for me but the type hints are there in the code:

image

Seems like an issue with vs code, maybe it struggles with the length of these type hints? ;)

joelostblom (approved):
Thanks for the explanations! No need to do any large re-write of the code for combining the `Literal`s; let's keep it the way it is and leave it up to the IDEs to take care of that.

This looks good from my side of things, but maybe let's leave open a couple of days to see if anyone wants to chime in on the main idea of abbreviating the signatures to use `SchemBase`  and then the full expansion only in the docstring.

mattijn (approved):
Its difficult to get this right. There is a balance between making human-users happy and providing meaningful type-related context. Its great that we try to improve this step by step.

binste:
Thank you both! Merging :)

PR #3340, opened 2024-02-25
Docs: Add another version of the 'Multiline Tooltip' exmaple which uses the standard tooltip

binste:
We already have https://altair-viz.github.io/gallery/multiline_tooltip.html which shows how to add multiline tooltips based on text marks. However, I often prefer to use the proper Vega tooltip also in these multiline cases as it's easier to make them readable then the text marks. We already have an example for this in the [Pivot](https://altair-viz.github.io/user_guide/transform/pivot.html) documentation but I think it would be nice to add one also to the example gallery.

This PR adds an adjusted version of https://altair-viz.github.io/gallery/multiline_tooltip.html which uses the standard Vega tooltip:

image


tests/examples_arguments_syntax/multiline_tooltip_standard.py
```diff
@@ -0,0 +1,52 @@
...
+    index=pd.RangeIndex(100, name="x"),
+)
+source = source.reset_index().melt("x", var_name="category", value_name="y")
+
+# The basic line
+line = (
+    alt.Chart(source)
+    .mark_line(interpolate="basis")
+    .encode(x="x:Q", y="y:Q", color="category:N")
+)
```
joelostblom, line 27: I'm curious, do you find this more convenient than the approach we use elsewhere in the docs?

```
alt.Chart().mark_line(),encode(
    ...
)
```

I'm not suggesting you change it here, just curious about what you use on a day to day basis.

joelostblom (changes requested):
Nice! I agree that this has multiple advantages over the other tooltip and I think it would be a good addition. I'm tempted to say that we could put them in the same example since they share about half the code in the example, but maybe that would be too busy. Instead, could you standardize the syntax between this and the old example so that the code for generating the data, `source`, `nearest`, `line`, and `points` look the same and come in the same order? I think that will make them easier to compare. I'm OK with either updating the old file to the syntax you use here or the other way around; whichever you prefer.

tests/examples_arguments_syntax/multiline_tooltip_standard.py
binste: Good point, they should be the same! I harmonised the code style, mainly adapting the new example to the old one so it's consistent with the approach we use in other places. I also slightly cleaned up the old one.

Personally, I just always use a code formatter (black or ruff). I love that I don't have to think about it and can get nicely formatted code with a hotkey or by saving a file :) The style I used before comes from ruff.

joelostblom (approved):
This looks good to me, thanks @binste ! Let's merge after the other PR is merged and tests are passing.

PR #3342, opened 2024-02-25
Relax type hint for 'indent' in to_json method

### LINKED ISSUE #3341: type hint for to_json(indent=) should be Optional
In https://github.com/altair-viz/altair/pull/3071 and https://github.com/altair-viz/altair/pull/3073 types were added to some `to_json()` methods (well, #3071 added the method entirely to `TopLevelMixin`) that specify `indent` as an `int`. In these, `indent` is only passed along to `json.dumps`, where it should be possible to pass `None` or a string as well: https://docs.python.org/3/library/json.html#json.dump

With `indent` typed as `int` and defaulting to 2 in the altair API, there is no way to get the compact `indent=None` serialization in a way that typechecks.

I think the types should be relaxed to `str | int | None` or something similar.

binste:
Closes #3341

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -982,7 +982,7 @@ class SchemaBase:
     def to_json(
         self,
         validate: bool = True,
-        indent: int = 2,
+        indent: Optional[Union[int, str]] = 2,
         sort_keys: bool = True,
         *,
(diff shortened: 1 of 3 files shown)


jonmmease (approved):
lgtm. Thanks!

PR #3350, opened 2024-03-03
Docs: Modifying scale of "multiple interactions" example and adding legend adjustments

d-trigo:
This is an attempt to modify the original "multiple interaction" example from the Altair-Vega documentation website in regard to the scale created with the checkbox condition for big budget films within the movies dataset. I had experimented with the example on my own to see how the Altair-Vega syntax worked when I realized that the quantitative scale for the alt.Size argument regarding a film's big budget designation wasn't fully accurate considering how the variable is entered. From what I read, the initial transformation in the chart creation creates a nominal scale between two categories of films, which can be labelled as "above $100 mil (big budget)" and "below $100 mil (not big budget)." However, the default scale being set to ':Q' for this variable means that the scale incorrectly assumes that the values of "100" and "10" (for big budget and non big-budget films respectively) are within a quantitative scale which then suggests to the viewer that it is displaying a continuous scale of films (via dot size) relative to production budget between $20 mil to $100 mil when the "Big Budget Films" box is checked.

I'm not sure if this is intentional to showcase how ':Q' automatically creates legend intervals when given any potential range of data, but if it is not I think it would be best to modify the scale and legend of this to be nominal in nature to clarify the purpose of the Big Budget Film checkbox. This modified version assigns '1' to films above $100 mil in budget and '0' for films vice versa so that, with a 0 -> 1 scale, alt.Scale will give a smaller mark size for lower-budget films before scaling up for big-budget films. I also include an example of how legends can be manually adjusted by adding a title and a defined range for mark size increase (from 25 to 150) between films below and above $100 mil.

Edit: As a small note, I forgot to re-add the intro text and comment for it being an interactive chart for the multiple interactions webpage so I re-added it via a 2nd and 3rd commit to this request. Hopefully that should be enough to correct that, this is my first pull request here so apologies for any errors I might have made doing so.

Screen Shot 2024-03-02 at 10 58 45 PM

### DIFF
--- a/tests/examples_methods_syntax/multiple_interactions.py
+++ b/tests/examples_methods_syntax/multiple_interactions.py
@@ -1,3 +1,16 @@
+"""
+Multiple Interactions
+=====================
+This example shows how multiple user inputs can be layered onto a chart. The four inputs have functionality as follows:
+
+* Dropdown: Filters the movies by genre
+* Radio Buttons: Highlights certain films by Worldwide Gross
+* Mouse Drag and Scroll: Zooms the x and y scales to allow for panning.
+
+
+
+"""
+
 import altair as alt
 from vega_datasets import data
 
--- a/tests/examples_methods_syntax/multiple_interactions.py
+++ b/tests/examples_methods_syntax/multiple_interactions.py
@@ -10,7 +10,7 @@ This example shows how multiple user inputs can be layered onto a chart. The fou
 
 
 """
-
+# category: interactive charts
 import altair as alt
 from vega_datasets import data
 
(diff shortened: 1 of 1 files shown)


joelostblom:
Thanks for taking the time to contribute @d-trigo ! I agree with you that it is more appropriate to use a nominal scale here. I think we can simplify even further and not require a legend formatting step. I noticed a few other things when I reviewed the PR also, such as putting the checkbox chart closer to the legend and including a note in the intro text about the checkbox. I made a suggestion in my latest commit, let me know if you think that aligns with your original intent of the PR and I will go ahead and merge it. Thanks again!

d-trigo:
Your revisions looks great, thank you!

joelostblom:
Great, merged! Thanks again!

PR #3357, opened 2024-03-12
Add timeseries filtering example to illustrate subsetting of columns from selector values

joelostblom:
I answered [this SO question](https://stackoverflow.com/questions/78146533/calculate-new-series-based-on-data-in-filtered-altair-chart/78148205#78148205) and thought it would be interesting with an example like that there too. The main thing it illustrates is how to select column based on selection values, but also puts it all together in a mini dashboards.

The code is somewhat long, but not long enough to be in a tutorial I think. I chose the "interactive" gallery category, but I think "advanced calc" would also work. There are better ways to compute time series correlation, but that's a parenthesis here.

https://github.com/altair-viz/altair/assets/4560057/1c2f9b53-b229-4391-bd62-80b1d1c3f5a9

### DIFF
--- a/tests/examples_arguments_syntax/interactive_column_selection.py
+++ b/tests/examples_arguments_syntax/interactive_column_selection.py
@@ -29,7 +29,7 @@ ex_ts = pd.DataFrame(
 select_x = alt.selection_point(fields=['level_0'], name='select_x', value='b')
 select_y = alt.selection_point(fields=['level_1'], name='select_y', value='d')
 heatmap = alt.Chart(
-    ex_ts.drop(columns='date').corr().stack().reset_index(name='correlation'),
+    ex_ts.drop(columns='date').corr().stack().reset_index().rename(columns={0: 'correlation'}),
     title='Click a tile to compare timeseries',
     height=250,
     width=250,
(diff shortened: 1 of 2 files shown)


joelostblom:
The reason for the failing test is that pandas added `numeric_only` in 1.5 (at least to [corr](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.corr.html)) and [in 2.0 the default was changed to `False`](https://github.com/pandas-dev/pandas/issues/46072). That means that we need set `numeric_only=True` as in this example, but in our CI we test the oldest version altair supports which is 0.25. Since the docs build, I believe this should be fine?

jonmmease:
What are you proposing? I think it's a good idea for the docs to use a more modern version of pandas.

jonmmease (approved):
This looks good!  In the future, I wonder if we should split out the ci jobs that run the examples so that we can use more up to date dependency versions. But not something to worry about today.

joelostblom:
Thank you for reviewing; merging!

mattijn:
Nice example @joelostblom!

binste:
Agree, nice one :) And also thanks for the other PRs on improving the docs!

PR #3374, opened 2024-03-20
Update README badges to reflect updated tool chain

mattijn:
This PR removes the obsolete code style black badge and replaces it with a badge indicating that this is a type checked library using mypy

Moreover under the development section I've added three more badges indicating we use hatch, ruff and pytest for development.

Before:
image

After:
image

---

Before:
image

After:
image


joelostblom (approved):
LGTM

PR #3376, opened 2024-03-22
fix: Don't error when embed_options are None

jonmmease:
Fix for https://github.com/vega/vl-convert/issues/159


joelostblom (approved):
LGTM, can you add a changelog entry and then merge?

PR #3377, opened 2024-03-22
fix: Handle Date32 columns in Arrow tables and Polars DataFrames

### LINKED ISSUE #3280: Polars Dataframe with "date" dtype would not be rendered by Altair. Two workarounds (with and without pandas) are proposed.
## Issue when altair (version 5.2.0) chart would not render if **any of the column** is in ["**date**" temporal data type](https://pola-rs.github.io/polars/user-guide/transformations/time-series/parsing/). 

I found that whenever there is a `date` data type column in the polars dataframe, the altair would not render the chart properly. Only when there is no `date` column, either by removing it completely, or by casting it to pandas or different data type as below.

### Code to reproduce issue

```python
import polars as pl
import altair as alt

# Importing data from vegadataset via url

dfx = pl.read_csv( "https://raw.githubusercontent.com/vega/vega-datasets/main/data/seattle-weather.csv",
try_parse_dates=True)

dfx

```
CleanShot 2023-12-10 at 18 43 55@2x

As shown above, `try_parse_dates=True` the date column will be "`date`" format. 

Using altair 5.2.0 version, if plotted using `rect` encoding the output will not be rendered.

```python
import altair as alt

alt.Chart(dfx).mark_rect().encode(
    alt.X("date(date):O").axis(labelAngle=0, format="%e").title("Day"),
    alt.Y("month(date):O").title("Month"),
    alt.Color("max(temp_max):Q", scale=alt.Scale(scheme="redblue")).title("Max Temp"),
)
```

CleanShot 2023-12-10 at 14 17 58@2x

There are two workarounds to avoid this issue:

### 1. Convert to pandas dataframe using `to_pandas()` before using altair.
This means the altair will use the familiar format (pandas dataframe) instead of polars, however that means we wouldn't get the benefit of using polars dataframe later on (e.g. speed, etc.). 

Perhaps works best if we are not planning to do any data wrangling with polars later on.

```python
dfx = (
    pl.read_csv(
        "https://raw.githubusercontent.com/vega/vega-datasets/main/data/seattle-weather.csv",
        try_parse_dates=True
    )
    .to_pandas()
)

dfx
```
CleanShot 2023-12-10 at 18 44 23@2x

Using the same code to make a chart from altair would work:

CleanShot 2023-12-10 at 18 44 57@2x

However, if we plan to use polars as the main dataframe library, then the second method is better.

### 2.  Parse the date column to `datetime` type instead of `date` data type. 
Casting the str type to `datetime[μs]` before we make the plot in altair.

> This works when the `try_parse_dates=False` (which is the default setting in `read_csv`)

```python
dfx = (
    pl.read_csv(
        "https://raw.githubusercontent.com/vega/vega-datasets/main/data/seattle-weather.csv")
    .with_columns(pl.col("date").str.to_datetime())
)

dfx.head()
```
CleanShot 2023-12-10 at 18 50 40@2x

Then just do the same chart plotting using altair would then work. 

CleanShot 2023-12-10 at 18 51 22@2x

jonmmease:
Closes #3280

This includes:
 1. Support sanitizing Date32 columns in pyarrow Tables
 2. For objects that follow the DataFrame interchange protocol, check for direct arrow conversion methods by name ("arrow", "to_arrow", and "to_arrow_table"), and if found use this instead of pyarrow `from_dataframe` function (which doesn't support Date32 yet).
 3. To make mypy happy, I updated the `DataFrameLike` protocol to be `@runtime_checkable` (which adds support for `isinstance` checks), and switched our `if hasattr(data, '__dataframe__'):` calls to `if isinstance(data, DataFrameLike):` calls. This way mypy knows that `data` is a `DataFrameLike` inside the branch and is happy for it to be passed to the `arrow_table_from_dfi_dataframe` function.
 4. Update VegaFusion tests for VegaFusion 1.6.6 and update the minimum VegaFusion version to 1.6.6.


altair/vegalite/v5/api.py
```diff
@@ -56,7 +56,7 @@ def _dataset_name(values: Union[dict, list, core.InlineDataset]) -> str:
         values = values.to_dict()
     if values == [{}]:
         return "empty"
-    values_json = json.dumps(values, sort_keys=True)
+    values_json = json.dumps(values, sort_keys=True, default=str)
```
jonmmease, line 59: For computing the values hash, I think it's fine to fallback to string representation for types not supported by `json.dumps` (`datetime.date` in this case).

tests/utils/test_mimebundle.py
```diff
@@ -241,7 +241,7 @@ def check_pre_transformed_vega_spec(vega_spec):
 
     # Check that the bin transform has been applied
     row0 = data_0["values"][0]
-    assert row0 == {"a": "A", "b": 28, "b_end": 28.0, "b_start": 0.0}
+    assert row0 == {"a": "A", "b_end": 28.0, "b_start": 0.0}
```
jonmmease, line 244: VegaFusion 1.6.6 strips out the "description" encoding field (which isn't used in the canvas renderer), so it's able to do a better job dropping unused columns.

altair/utils/data.py
```diff
@@ -346,3 +338,23 @@ def curry(*args, **kwargs):
...
+    # Give this preference over the pyarrow from_dataframe function since the object
+    # has more control over the conversion, and may have broader compatibility.
+    # This is the case for Polars, which supports Date32 columns in direct conversion
+    # while pyarrow does not yet support this type in from_dataframe
+    for convert_method_name in ("arrow", "to_arrow", "to_arrow_table"):
+        convert_method = getattr(dfi_df, convert_method_name, None)
+        if callable(convert_method):
+            result = convert_method()
+            if isinstance(result, pa.Table):
+                return result
```
mattijn, line 357: It is a joy to read this type of code diff. Really nice approach!

mattijn (approved):
Thanks @jonmmease, really nice PR! Literature level code👍.
I added one more commit with a few more `isinstance(data, DataFrameLike)` over `hasattr(data, __dataframe__)`

I'm surprised that pandas has no a `to_arrow()` method or something a like.

mattijn:
All tests pass. Merging! Thanks again @jonmease!

PR #3379, opened 2024-03-23
feature: Add browser renderer to open charts in external browser and update chart.show() to display chart

jonmmease:
This WIP PR tries out the idea I mentioned in https://github.com/altair-viz/altair/issues/2866#issuecomment-1919453217. 
It adds a new `"browser"` renderer that opens the chart in a new browser tab using a one-shot web server and the Python webbrowser module. I like this approach because it avoids the need to write temporary files just to open them in the browser (Although we do need to document that this means that you can't refresh the page or open the url in another browser tab).

The browser renderer's `using` argument can be used to configure what browser should be used, the `port` argument to configure what port the one-shot server should (defaults to random), and the `inline` argument to configure whether to embed all JavaScript dependencies in the html bundle so that the renderer works offline.

This PR also updates `chart.show()` to invoke the active renderer as a side-effect. This uses `IPython.display.display` for all renderer except the browser renderer. There's a special case for the browser renderer that invokes `_repr_mimebundle_` directly so that it works from a standard python repl, without IPython installed.

If we want to go forward with this approach, I'll update the docs to replace the [Altiar Viewer](https://altair-viz.github.io/user_guide/display_frontends.html#altair-viewer) section.

cc @fhg-isi if you have a chance to give this branch a try.


jonmmease:
Fix for faiing VegaFusion tests in https://github.com/altair-viz/altair/pull/3377

altair/vegalite/v5/api.py
```diff
@@ -2678,29 +2679,29 @@ def serve(
...
         ----------
-        embed_opt : dict (optional)
+        embed_options : dict (optional)
             The Vega embed options that control the display of the chart.
-        open_browser : bool (optional)
-            Specify whether a browser window should be opened. If not specified,
-            a browser window will be opened only if the server is not already
-            connected to a browser.
+        using: str or iterable of str
+            Name of the web browser to open (e.g. "chrome", "firefox", etc.).
```
joelostblom, line 2696: Should we mention that it can also be a path?

mattijn (approved):
I tried several things from command line and notebook through a vscode notebook and was not yet able to break it. So I would say it works as expected.
I hadn't seen it before, but apparently this is a feature missed by people (https://github.com/altair-viz/altair_viewer/issues/59#issuecomment-1694506433).

I don't know the previous behavior, but maybe we should at the meaning of a oneshot request in the docstring of `.show()` as I was wondering why the chart was not working anymore after a page refresh or opening the same address:port in an additional tab.

altair/utils/_show.py
```diff
@@ -0,0 +1,75 @@
...
+    using: Union[str, Iterable[str], None] = None,
+    port: Optional[int] = None,
+):
+    """
+    Display an html document in a web browser without creating a temp file.
+
+    Instantiates a simple http server and uses the webbrowser module to
+    open the server's URL
+
+    Based on
```
joelostblom, line 17: Incomplete sentence

altair/utils/_show.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Parameters
+    ----------
+    html: str
+        HTML string to display
+    using: str or iterable of str
+        Name of the web browser to open (e.g. "chrome", "firefox", etc.).
```
joelostblom, line 24: Same comment here with the path

joelostblom (commented):
Thanks @jonmmease! This looks good to me an work as expected when I try the PR locally. I only have a few minor inline comments.

I did notice that when trying the first example from https://vegafusion.io/, it works even when I haven't enabled the vegafusion data_transformer (ie doesn't raise the max rows error). I think that's unrelated to this PR and related to how `chart.save()` works. I think that makes sense for saving charts since we don't need to worry about large notebook sizes from saving multiple figures. Do you think there are any extra concerns with `show` in the sense that it sends the entire (possibly huge) spec to the browser instead of to a file?

jonmmease:
Thanks @joelostblom, that's a good call out that this approach disables the max rows error. I think it makes sense to use the same rules here as for the inline html renderer. I'll make this change!

altair/vegalite/v5/api.py
jonmmease: I didn't realize it could be, yeah that's useful information

jonmmease:
@mattijn Thanks for the call out that we should document the one-shot implications.

jonmmease:
@mattijn and @joelostblom, I just pushed a change to how this works that I think is a lot nicer.

First, I added a `"browser"` renderer (name up for discussion, this is what we called it in plotly). This renderer exports the chart to HTML and opens it in a browser tab as a side effect.  The `using`, `port`, and `inline` arguments can be passed during renderer initialization to configure things, like a regular renderer:

```python
alt.renderers.enable("browser", using="chrome", port=1234, inline=True)
...
chart
```
Then, in an IPython setting, the chart will open in a browser tab automatically when it's the last value in a cell/command.

Then I updated the `show` method to trigger the active renderer as a side effect.  I think this is useful in general, because for renderers that display charts inline, it makes it easy to display multiple charts in a cell (without concatenation). For example, with the `default` renderer enabled:

Screenshot 2024-03-24 at 3 12 46 PM

When the browser renderer is enabled, calling `chart.show()` will open the chart in the browser, even if it's not the final value in the cell/command.  Additionally, calling `chart.show()` with the `browser` renderer works in a standard (non-IPython) python script or REPL. So you can have a Python scrip with several calls  to `chart.show()` in it, run it from the command line with `python my_script.py`, and each chart will open in the browser.

jonmmease:
I updated the docs and marked this as ready for review. @mattijn, the change since you approved was to add the `"browser"` renderer and have `chart.show()` invoke the active renderer.

altair/vegalite/v5/api.py
jonmmease: The docs don't seem to mention this: https://docs.python.org/3/library/webbrowser.html#webbrowser.register. Does this work for you?

altair/utils/_show.py
jonmmease: Done in 478944342eff67dedc49fe699252d97d680e07bc

mattijn:
I like the approach! I noticed that in practice there are still several people liking to explicitly call the `.show()` command to visualize the chart.

doc/releases/changes.rst
binste, line 1: Now that https://github.com/altair-viz/altair/pull/3380 is merged, you can just put these changes at the end of `RELEASING.md`. I'll then use it when making the release

doc/user_guide/display_frontends.rst
```diff
@@ -179,57 +184,83 @@ If you are using a dashboarding package that is not listed here, please `open an
...
+The ``"browser"`` renderer supports the following keyword argument configuration options:
+
+- The ``using`` argument may be used to specify which system web browser to use. This
+  may be set to a string to indicate the single browser that must be used (e.g. ``"safari"``),
+  or it may be set to a list of browser names where the first available browser is used. See the
+  documentation for the `webbrowser module`_ for the list of supported browser names. If not
+  specified, the system default browser is used.
+- The ``offline`` argument may be used to specify whether JavaScript dependencies should
+  be loaded from an online CDN or embedded alongside the chart specification. When ``offline``
+  is ``False`` (The default), JavaScript dependencies are loaded from an online CDN, and so
```
binste, line 243:
```suggestion
  is ``False`` (the default), JavaScript dependencies are loaded from an online CDN, and so
```

doc/user_guide/display_frontends.rst
```diff
@@ -179,57 +184,83 @@ If you are using a dashboarding package that is not listed here, please `open an
...
+  then opens the designated browser pointing to this server's URL. This approach does not require
+  the creation of temporary HTML files on disk, and it's memory efficient as there are no long-lived
+  web server processes required. A limitation of this approach is that the chart will be lost if the
+  browser is refreshed, and it's not possible to copy the chart URL and paste it in another browser
+  tab.
+- When used in IPython-based environments, the ``"browser"`` renderer will automatically open the
+  chart in the browser when the chart is the final value of the cell or command. This behavior is not
+  available in the standard ``python`` REPL. In this case, the ``chart.show()`` method may be used to
+  manually invoke the active renderer and open the chart in the browser.
+- This renderer is not compatible with remote environments like Binder or Colab.
```
binste, line 262: Could we use links also for "Binder" and "Colab" so that it's consistent with Ipython and Spyder above? 

```suggestion
- This renderer is not compatible with remote environments like `Binder`_ or `Colab`_.
```

doc/user_guide/display_frontends.rst
```diff
@@ -678,3 +709,7 @@ see :ref:`display-general`.
 .. _Vega: https://vega.github.io/vega/
 .. _VSCode-Python: https://code.visualstudio.com/docs/python/python-tutorial
 .. _Zeppelin: https://zeppelin.apache.org/
+.. _IPython: https://ipython.org/
```
binste, line 712:
```suggestion
.. _Binder: https://mybinder.org/
.. _Colab: https://colab.google/
.. _IPython: https://ipython.org/
```

binste (approved):
Only some minor nits. Nice feature and implementation @jonmmease! Good to see that we can remove the mentions of altair_viewer from the docs and provide an alternative to users.

joelostblom (approved):
Thanks for the updates @jonmmease ! When I tested the latest changes locally after, I was initially surprised that it is no longer enough to call `.show()` on an individual chart to open it in the browser; but then I read your updates to the docs in more details and realized that I have to enable the new `'browser'` renderer:

image

The difference between using `show` and not become clear when I also test in the in the regular python console:

image

I would expect that this makes things more convenient overall, since the IPython console is for working interactively and it's helpful that all charts open in the browser without the need to call `show` when the new renderer is active. The only thing that becomes less convenient is showing a single chart or switching between showing and not showing charts, but that is probably a rare workflow.

Since this differs from how the old altair_viewer approach worked where it was enough to call `show`, but you had to do it for every chart, I think it's great that you have clearly mentioned it as two separate items in the release notes. Maybe we should document the behavior of showing multiple charts from the same cell using `show`/`display` somewhere in the docs as well (doesn't have to be in this PR though)?

altair/vegalite/v5/api.py
joelostblom: It does, but since it is not in the official docs, then I don't think we need to mention it either

jonmmease:
Thanks for reviews everyone. merging for 5.3.0!

PR #3380, opened 2024-03-23
docs: Remove release notes and fully capture them in GitHub Releases

### LINKED ISSUE #3339: Only track release notes in one place
We currently track release notes in 2 places, the [documentation](https://altair-viz.github.io/releases/changes.html) and in [GitHub Releases](https://github.com/altair-viz/altair/releases). The docs use restructured text whereas GitHub Releases uses Markdown so we can't just copy-paste content from one into the other but it requires a rewrite. In addition, as it's easier to add images and do formatting in Markdown, the GH Releases pages of the last few releases look much better and are more extensive than what we have in the docs. Thanks @mattijn for taking the time every release to write this!

How about we only maintain the release notes in GitHub Releases? We can simply link to it from the documentation. No information gets lost and it's easier for us to maintain and we can write better release notes which is beneficial for users.

I'm happy to create a PR for the docs and to make sure that everything is captured in GitHub Releases. Thoughts?

binste:
Closes #3339. I added the missing releases to GitHub Releases and made sure that all content from the docs is captured in there.

"Release Notes" in the navigation bar now links directly to GitHub:
image

And the old /releases/changes.html URL redirects to GitHub after showing a message that the content got moved:

image

I moved the draft release notes for 5.3.0 into `RELEASING.md` so we don't loose them. I'll remove them when I'll do the 5.3 release. To capture release notes in the future, see Joel's suggestion [here](https://github.com/altair-viz/altair/issues/3339#issuecomment-2012679908)


binste:
@jonmmease https://github.com/altair-viz/altair/actions/runs/8402900588/job/23012797705?pr=3380 fails due to some different outputs from VegaFusion. Maybe something to do with the latest VF release? Would you mind taking a look? Otherwise, I can look into it later this week.

jonmmease:
Yes, it's from a change in the latest VegaFusion release. I fixed the tests in https://github.com/altair-viz/altair/pull/3377, which I think is ready to merge as soon as someone has time to look over it.

joelostblom (approved):
These changes look good to me. Thanks for taking the time to move all the old releases to GitHub!

doc/releases/changes.rst
```diff
@@ -1,711 +1,9 @@
...
+:orphan:
 
 Release Notes
 =============
 
-Version 5.3.0 (unreleased month day, year)
---------------------------
-- Update Vega-Lite from version 5.16.3 to version 5.17.0;
-  see `Vega-Lite Release Notes <https://github.com/vega/vega-lite/releases>`_.
+We have moved the release notes to GitHub. You'll be automatically redirected there in 3 seconds, or you can click `here <https://github.com/altair-viz/altair/releases>`_.
```
mattijn, line 6: Is there an urgent need to redirect automatically?

binste: No urgent need. I thought redirecting in general is good to not lose users. An additional click is always an opportunity for someone to drop out. And 3 seconds felt like enough time for me to read the message but no strong preference on how we handle this. What do you think?

binste: Just stumbled on pandas doing the same with around 1 second of wait time: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.to_json.html

mattijn: Double negative alert: I thought not to auto redirect in order to not lose users..

At least we have the same intention😀!

mattijn: I don't notice a redirect on the pandas website, it stays within the pandas documentation page isn't it?

binste: Ah I think I see what you mean! Same intentions but I think not redirecting might indeed be a better way to achieve this. I'll remove the automated redirect so users can decide themselves if they want to go to GitHub or stay in the Altair docs.

Re the Pandas docs: Maybe they changed that... Yesterday it showed briefly a message before redirecting. Now it directly goes to `stable/reference` instead of `stable/generated`. Anyway :)

PR #3381, opened 2024-03-23
chore: Add templates for PRs and automated release notes

joelostblom:
Continuation of https://github.com/altair-viz/altair/issues/3339. This is an example of a PR template we could use to encourage contributors to use conventional commit style for the title. In addition to the conventional commit homepage, angular has [a short description](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type) of what each title prefix indicates.

To be sorted into specific sections in the automated GitHub release notes, we would also need to label each PR. I could not find a way to automate this. The closest is this action, but it is outdated https://github.com/marketplace/actions/conventional-release-labels. I don't think it is a big deal to add labels manually and if we forget a label the PR will still be included in the automated release notes draft, just not under the correct section, but it's easy to move it manually.

I think the relevant title prefixes and labels would be the following:

| Title prefix                       | Label         |
|------------------------------------|---------------|
| fix                                | bug           |
| feat                               | enhancement   |
| docs                               | documentation |
| chore,build,refactor,style,test,ci | maintenance   |

As @binste pointed out, we could also add an automated check for the PR title, e.g. via https://github.com/marketplace/actions/semantic-pull-request which VL uses.


.github/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+Thanks for contributing to Altair!
```
binste, line 1: I like the text! Could we wrap it in `<!-- Here the text -->` so that we don't even have to delete the message before submitting the PR? See how Ibis handles this: https://raw.githubusercontent.com/ibis-project/ibis/main/.github/PULL_REQUEST_TEMPLATE.md

I think I've also seen this in other repos as well.

.github/ISSUE_TEMPLATE/feature-request.md
```diff
@@ -4,12 +4,10 @@ about: 'Suggest a new idea or feature'
 labels: 'enhancement'
 ---
 
-Type here...
+Thanks for taking the time to suggest a new feature!
 
----
-
-Please follow these steps to make it more efficient to respond to your feature request.
+Please read the guidelines below to make it more efficient to respond to your feature request, and then delete this message before submitting the issue.
```
binste, line 9: See my comment on the pull request template on using the comments syntax to hide the message by default when submitting the issue.

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -4,13 +4,11 @@ about: 'Create a bug report to help us improve'
 labels: 'bug'
 ---
 
-Type here...
+Thanks for taking the time to report a bug!
 
----
-
-Please follow these steps to make it more efficient to solve your issue:
+Please read the guidelines below to make it more efficient to solve your bug, and then delete this message before submitting the issue.
```
binste, line 9: See other comments regarding "comment" syntax

.github/release.yml
```diff
@@ -0,0 +1,17 @@
...
+    - title: Enhancements
+      labels:
+        - enhancement
+    - title: Bug Fixes
+      labels:
+        - bug
+    - title: Maintainance
+      labels:
+        - maintainance
+    - title: documentation
```
binste, line 12:
```suggestion
    - title: Documentation
```

binste (changes requested):
Thanks for pulling this together @joelostblom! I really like that we can add some structure to PR titles and make the release process easier.

I agree with you that it's totally fine if the automatically generated release notes still require us to sort the PR's which were not labelled. That's a quick thing to do but it's also great to already get the automatic categorization with `release.yml`, I wasn't aware of this feature :)

I added some minor comments. Regarding `perf`, mild preference to put it into `Enhancements`. It's something which makes the library better, I think `Maintenance` should be things which the users ideally don't notice.

.github/pull_request_template.md
joelostblom: Yeah, I have seen this in other repos too, however I'm a bit split on whether it is preferable. The reason I kept it as uncommented text here is so that it can be read in the rich preview; I find that the links makes the comment cluttered and hard to parse.

For the issues, the answer is to use "Issue forms" instead of the regular templates. This renders links and formatting nicely while not including it in the actual issue. In my latest commit I switched both our issue templates over, [you can see a demo in this other repo](https://github.com/joelostblom/test-pr-templates/issues/new/choose) I made.

image

image

Unfortunately, [forms don't exist for PR ](https://github.com/orgs/community/discussions/4881).

joelostblom:
Thanks for the review @binste. I agree `perf` is better suited for `enhancement` and updated the table above. I replied to your inline comment regarding formatting for the templates.

binste (approved):
The issue templates are really nice! And I see your point regarding the readability of the comments. Thanks!

PR #3383, opened 2024-03-24
perf: Improve performance of `Chart.from_dict`

### LINKED ISSUE #3382: `alt.Chart.from_dict` slow due to large number of calls to `utils.schemaapi._use_referencing_library`
I'm the author of a [FOSS library](http://github.com/moj-analytical-services/splink) that makes extensive use of `alt.Chart.from_dict`.

I noticed this is quite slow, and I think I have identified the root cause. Specifically, `alt.Chart.from_dict` makes a very large number of calls to [`utils.schemaapi._use_referencing_library`](https://github.com/altair-viz/altair/blob/b1b9dba8a1cd11f9a5e18ed38d4fd5acf3961ca7/altair/utils/schemapi.py#L203). In turn, this makes repeated calls to `importlib`, resulting in more than half of its runtime in `importlib.metadata`

Since the result of `utils.schemaapi._use_referencing_library` is a bool and will not change within a Python session (it only changes if the user installs a different version of the `jsonschema` package), it can be computed a single time, resulting in significantly improved performace.

## Reprex

<details>
<summary>
Reprex including timings
</summary>

```python
import json
import time
import urllib.request

import altair as alt

spec_json_url = "https://gist.githubusercontent.com/RobinL/de91f5d95aa29ee0b5464bca4bfd3dcf/raw/c391db54295a8f43e3258aa6c42a8a39f239b4bc/match_weights.json"

with urllib.request.urlopen(spec_json_url) as url:
    data = json.loads(url.read().decode())

start_time = time.time()
chart = alt.Chart.from_dict(data)
end_time = time.time()
execution_time = end_time - start_time
print(f"Execution time: {execution_time} seconds")

spec_json_url = "https://gist.githubusercontent.com/RobinL/de91f5d95aa29ee0b5464bca4bfd3dcf/raw/c391db54295a8f43e3258aa6c42a8a39f239b4bc/waterfall.json"

with urllib.request.urlopen(spec_json_url) as url:
    data = json.loads(url.read().decode())

start_time = time.time()
chart = alt.Chart.from_dict(data)
end_time = time.time()
execution_time = end_time - start_time
print(f"Execution time: {execution_time} seconds")
```

</details>

The timings for the above on `main` are:

```
Execution time: 0.9597208499908447 seconds
Execution time: 1.2827000617980957 seconds
```

on a 2019 macbook pro.

After removing the repeated `importlib` calls, the timings are:

```
Execution time: 0.2081460952758789 seconds
Execution time: 0.3407268524169922 seconds
```

## Solution

I've opened a [PR](https://github.com/altair-viz/altair/pull/3383) that I think should address this

RobinL:
Closes #3382

This PR substantially improves performance of `alt.Chart.from_dict`.  See #3382 for a detailed description and reprex.

Grateful for any feedback on this PR, e.g. whether the team would prefer I leave the current function as is, or use `functools.lru_cache`, instead of the top-level `jsonschema_version_str = importlib_version("jsonschema")` statement.  Or some other solution.

WIP whilst I make sure the builds/tests pass

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -71,6 +71,8 @@ _DEFAULT_JSON_SCHEMA_DRAFT_URL: Final = "http://json-schema.org/draft-07/schema#
 # class-level _class_is_valid_at_instantiation attribute to False
 DEBUG_MODE: bool = True
 
+jsonschema_version_str = importlib_version("jsonschema")
+
 
 def enable_debug_mode() -> None:
     global DEBUG_MODE
@@ -203,7 +205,6 @@ def _get_json_schema_draft_url(schema: dict) -> str:
 def _use_referencing_library() -> bool:
     """In version 4.18.0, the jsonschema package deprecated RefResolver in
     favor of the referencing library."""
-    jsonschema_version_str = importlib_version("jsonschema")
     return Version(jsonschema_version_str) >= Version("4.18")
 
 
--- a/tools/schemapi/schemapi.py
+++ b/tools/schemapi/schemapi.py
@@ -71,6 +71,7 @@ DEBUG_MODE: bool = True
 
 jsonschema_version_str = importlib_version("jsonschema")
 
+
 def enable_debug_mode() -> None:
     global DEBUG_MODE
     DEBUG_MODE = True
--- a/tools/schemapi/schemapi.py
+++ b/tools/schemapi/schemapi.py
@@ -69,6 +69,7 @@ _DEFAULT_JSON_SCHEMA_DRAFT_URL: Final = "http://json-schema.org/draft-07/schema#
 # class-level _class_is_valid_at_instantiation attribute to False
 DEBUG_MODE: bool = True
 
+jsonschema_version_str = importlib_version("jsonschema")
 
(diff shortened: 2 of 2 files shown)


jonmmease:
Thanks for the investigation and the fix @RobinL! I think this is a reasonable way to go, but would like to hear from @binste as well

RobinL:
Thanks very much for the quick response, will await feedback from binste.  

Also to note: I've tried to run `python tools/generate_schema_wrapper.py` locally but these results in a significant diff to `altair/utils/schemapi.py` even if I run it before having made any changes to `tools/schemapi/schemapi.py`.  

Very happy to look into this further, but if we go ahead with the proposed simple fix, it might be trivial to sort by the maintainers, and I'm a bit unsure how to proceed.

jonmmease:
This is straightforward enough that with the passing tests I'm fine merging it. We can make any adjustments before the 5.3.0 release if anyone want to weight in afterward.

binste:
Nice catch @RobinL! Thanks for finding and fixing it :)

Thanks Jon for the review! Fully agree, changes look good so we can release it in 5.3.0 as-is.

PR #3387, opened 2024-03-26
fix: Support falling back to pandas when pyarrow is installed but too old

### LINKED ISSUE #3385: minimum pyarrow version enforced even if pandas is installed
The error we are facing in an environment says:
```python
RuntimeError: The pyarrow package must be version 11.0.0 or greater. Found version 6.0.1
```

And is caused by these lines:
https://github.com/altair-viz/altair/blob/main/altair/utils/core.py#L591-L592

```python
    # if data is specified and type is not, infer type from data
    if "type" not in attrs:
        if pyarrow_available() and data is not None and isinstance(data, DataFrameLike):
            ...

        elif isinstance(data, pd.DataFrame):
            # Fallback if pyarrow is not installed or if pandas is older than 1.5
```
In that particular environment pandas is installed by default and we are not able to upgrade pyarrow. 

Now the altair specifications errors as the code never tries the pandas approach as it has found a pyarrow version that is too old.

jonmmease:
Closes #3385 by catching `RuntimeError` along with `ImportError` in `pyarrow_available`.

This was a regression introduced in https://github.com/altair-viz/altair/pull/3244 when the `ImportError` was swapped out for a `RuntimeError` without updating `pyarrow_available`.

@mattijn, could you give this a try to see if it takes care of the issue of an old pyarrow with pandas?

### DIFF
--- a/altair/utils/_importers.py
+++ b/altair/utils/_importers.py
@@ -93,5 +93,5 @@ def pyarrow_available() -> bool:
     try:
         import_pyarrow_interchange()
         return True
-    except ImportError:
+    except (ImportError, RuntimeError):
         return False


mattijn:
How can we report to a user if the version of pyarrow is too low when pandas is explicitly not installed?

jonmmease:
> How can we report to a user if the version of pyarrow is too low when pandas is explicitly not installed?

I guess this is something we would want to work into your PR at https://github.com/altair-viz/altair/pull/3384

mattijn:
Valid point!

jonmmease:
Ready for this to merge @mattijn?

mattijn:
I haven't had time to test this, but the fix looks legitimate and appropriate👍

PR #3394, opened 2024-04-04
feat: Extend `.interactive()` with tooltip and legend 

### LINKED ISSUE #3393: Add more features to `.interactive()`?
### What is your suggestion?

I'm wondering if it would be helpful if `.interactive()` did more by default. I'm primarily thinking about adding a tooltip (of the plotted encodings), and an interactive legend (based on opacity). There should of course be some way to turn these on/off, so that they don't interfere with manually set up bindings (maybe `chart.interactive(legend=False)`.

_No response_

joelostblom:
These are some initial ideas for how we could extend the `.interactive()` method to become more useful. I think having a clickable legend and a tooltip by default would cover many of the basic interactive needs for situations when one does not want more fine-grained control.

https://github.com/altair-viz/altair/assets/4560057/39e65490-d0da-486d-a7de-15efeb9993f3

We would still need to add this to the concat charts etc, but I thought I would add to #3393 by creating an example of what this could look like in the code.

I think the addition of the tooltip is pretty straightforward and harmless. The addition of the legend raises the question of what would happen when there is an existing opacity encoding. My suggestion would be to detect this and print a message that the interactive legend has been automatically disabled (and setting `legend=False`) would suppress that message.

I chose to link the legend click to opacity as I thought this was less intrusive than color or a transform filter, but happy to hear other thoughts (I know that e.g. plotly filters points based on legend clicks)

close https://github.com/altair-viz/altair/issues/3393


altair/vegalite/v5/api.py
```diff
@@ -3047,7 +3061,57 @@ def interactive(
...
+                #     in [
+                #         "angle",
+                #         "radius",
+                #         "color",
+                #         "fill",
+                #         "shape",
+                #         "size",
+                #         "stroke",
+                #     ]
+                # ]
```
joelostblom, line 3103: I am thinking it's more elegant if we set the legend encodings only to encodings that actually exist in the spec. However, `to_dict(validate=False)` currently does not work for channels as per https://github.com/altair-viz/altair/issues/3075.

It is possible to do something like this instead, but it seems clunky:

```py
[
    enc for enc in vars(chart.encoding)['_kwds']
    if not isinstance(chart.encoding[enc], alt.utils.schemapi.UndefinedType)
]
```

binste:
I'll need to find some more time to review and be helpful but I really like the idea, thanks for getting started on this! I received the feedback from Plotly users that Plotly does more "interactive" things out of the box so I think there could be demand for this.

dwootton:
Really excited to see the progress on .interactive!  Stefan makes a great point –  ease of use around interactions is key for a positive first impression. This aligns with the user feedback we gathered during Altair Express testing.

Since this PR is still a draft, consider the following _discussion points_ rather than direct implementation feedback:

Multi-View Filtering:  How should filtering behave in multi-view charts? In Altair Express, confusion arose when interactions were chart-specific.  Propagating interactions across the entire spec proved more intuitive (ie `alx.highlight_brush() + my_multi_view_chart` added a brush to each chart and cross filtered the entire visualization. Would a similar approach make sense here? 



Interaction Effect Generalization: I love the opacity fade effect for scatterplots. Feels intuitive and looks nice when plots are relatively balanced like the cars datasets.  However, for broader chart types, I've found filtering generally provides better usability. For example with bar charts, fades don’t rescale the y axis and as such they're less helpful for investigating lower-count bars. 


I think plotly's approach works pretty well for bar charts. 


and it generally helpful for rescaling skewed charts.  


Great work bringing up changes for .interactive; Let me know your thoughts on these points. I'm happy to brainstorm solutions and explore potential trade-offs with you. Longer term might make sense to roll up some of altair express's code to .interactive.

joelostblom:
Thanks both! I agree that this first interactive impression is important and it's great to hear that this is backed up by your research findings @dylan. Thank you so much for sharing some of the insights from your research!

> Interaction Effect: I love the opacity fade effect for scatterplots. Feels intuitive and looks nice when plots are relatively balanced like the cars datasets. However, for broader chart types, I've found filtering generally provides better usability. ...

I 100% agree with this. I think the fade looks nicer and it doesn't have the effect of the axes domain rescaling upon selection with creates a somewhat jarring/jumpy behavior at times. But it is not as general and as you point out, it is not at all useful when there are outliers compressing the rest of the data. I actually switched a personal project over to highlighting to filtering just yesterday for this exact reason and I will update the PR do do the same thing here.

> Behavior Across Concatenated Views: How should filtering behave in multi-view charts? ...

This is a great insight and something I didn't even think about. Thanks for sharing it! My preference would be propagation across the entire spec and with filtering instead of opacity highlighting that would be more straightforward.

>  Longer term might make sense to roll up some of altair express's code to .interactive.

I'm very receptive to plan ahead for this if there are things you already have in mind and you think we should take into account when designing this PR (and you're more than welcome to help out implementing it too if you have time available).

---

What do you all think about keeping the highlighting behavior on control clicking the legend and having filtering on regular click? On the one hand I worry that we are overloading the legend interaction, but on the other hand, I don't think this would get in the way for people who only clicks the legend without holding control. I can't help but to really like the visual effect of highlighting...

If either of you have more ideas of what we could do here, I would be happy to adjust this PR further with the goal of giving the most helpful default interactivity for everyone who is not interested in setting up their own bindings. One aspect I think would be useful would be to have the cursor change to a "clickable hand" when hovering the legend to indicate that it is indeed. It seems as if this could be a simple change as marks in Vega/Vega-Lite already supports a `cursor` parameter, but after a brief glance at [the source](https://github.com/vega/vega-lite/blob/main/src/compile/selection/legends.ts), I can't say that it is immediately obvious what to do so maybe we have to wait for https://github.com/vega/vega-lite/issues/4155...

mattijn:
Given the extensive research that has been put in place on this topic by @dwootton, and his offer to investigate how we can integrate this into the `.interactive()` method. I think we should not rush this and come up with a deliberate approach that is in-line with the best practices discovered by altair-express.

joelostblom:
@dwootton I was working on this today and ran into the issue of how to keep the legend entries constant when the legend is used to filter the data. I opened https://github.com/vega/vega-lite/issues/9360 with more details, but thought I would also mention it here in case you have a clever way to solve this in altair express.

dwootton:
Ooh yeah I remember that. The crux of the issue is that in the data filtering occurs upstream from other options in the dataflow graph. As such, when the color scale is recomputed, it is recomputed over the filtered data, resulting in the behavior seen. Three ways to get around this: 

1.  [Compute color domain non-reactively](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAZgAY2AacaYsiygAlMiRoVYcAvpO50AhoTl4QpAE4AbFCDGEADpWQB6Q2DpQAdACtKdTOrhpV5qJkKGougLaG0mBHIBaeUVKV0oAATQARnMAJgBOczZDYLkTOVVKK0poEBkQXQy5T0oUAG1QKGLMLULVYoB9FhBuUPVMMHFUEEIAT10a5AKIEnFuTChIOhIEUuQynnUIVRAAXXyAIxI6LXaECZ2ZSuraos8AFhaQTAAPXRWhughPORIACgBySCXVD4BKPKcY6eQaLZYNJ4vEhXVSYOSdFAwOTqULjO4PApnS6SVbcQj1KCUGDLTzlUDwdSETAPUB1YqneqeJp5HHcF6qADWyj6A1qIyIVwp7R2yHxpEw+QmUxmym+y2U8DsIpAAHlVIhoXj+qCoM8SMirpQwMjBqBIa8oMpbvctHLVBDnhaWfkbgqsOplYJlqFdBBsNSrjzQQBHUhyIhwRTEXyAkC9N1KrQAWTg7UoDQG9oA4silpatbyhqHw8Qow4atJuAASI2MTAvLQ6fRGHx+OTmBCRxikDbmOAQQy1+tpXz+AL2Kk+ACs5iiAHYktlcpIgA). 
When you set `reactive=False` for a parameter, the parameter will not update in response to changes in their upstream dependencies. By adding two parameters and then setting the color domain to be driven off the non-reactive parameter, you can reference the initial color legend. 
{
         "name":"param4", // this parameter will grab the domain from the color scale. 
         "expr":"domain('color')" // note: may need fancier logic to handle cases where there are independent color scales.
      },
      {
         "name":"color_domain", // stores the domain of the original color scale at initial load 
         "react":false, 
         "expr":"param4"
      }
 
As this breaks VL's schema, I'm not sure this approach would be suitable for the altair project though. However, out of the other options, this felt the cleanest as it didn't require major modifications to the specification. 

2. [Create a layered visualization using their provided view with opacity=0 marks](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAZgAY2AacaYsiygAlMiRoVYcAvpO50AhoTl4QpAE4AbFCDGEADpWQB6Q2DpQAdACtKdTOrhpV5qJkKGougLaG0mBHIBaeUVKV0oAATQARnMAJgBOczZDYLkTOVVKK0poEBkQdTkAT0xVFABtUF0MuU9KCtAoWswtatVagH0WEG5Q9UwwcVQQQiLdFuQQXQgScW5MKEg6EgR65HKedQgygF18gCMSOi1+hAXjmUbm1prPABYekEwAD10yyboITzkSAAoAckgW1U-wAlHlOFdPBMprcAKyPNBydSkGFtWoPSQ7biEdpQSgwbaeBogeDqQilZTo4mTaldPJY7jfVQAa2Uo3GrRmREeZP6x2QuNR+QWSxWyiB22U8DsApAAHlVIgSI8OTCoF8SMiISBntKsOo5YJtqFpthKTixjCAI6kOREOCKYi+HVFfWyrQAWTg-UoHXGqg6AHFkVsoKqrVpbfbiE6HC1pJcQMy2cM1VzZryffyUELMCLFhBllAkMMINUwI63cMkSiJmx8pL3qAZYatIrleHLZzJhrPFrNPk9cNW0aTZgzRaRpHJtGHXGXflqy2DXLvb7-aVg6Hct2bXb5wp4wzGSAACSUMCMTDfLQ6fRGHx+OTmBCOxikfbmOAQQyX6-fE+-gBPYFI+HC5hRAA7Ek2S5JIQA).  

This will keep the color scale set to unfiltered data but hide any marks from appearing on the visualization. 

3. (Outside of VL) Precompute the domain in python and add it to the color encoding field. 

This is probably the easiest option, but from the chart object, you can grab the field from the color encoding and then precompute the unfiltered domain as the domain for the color encoding.

joelostblom:
Thank you for the detailed and helpful answer @dwootton ! Sorry it took me a while to get back to you, I wanted to find time to see if I could add your first option to Vega-Lite and finally managed to put together a PR in https://github.com/vega/vega-lite/pull/9374. I think your first option is the cleanest and it would be great to have this supported in the official Vega-Lite grammar. Since the data can be passed as an URL etc, we must make assumption if precomputing the color domain and also rely on a dataframe library, which we are trying to avoid (but this is what I use most of the time personally).

mattijn:
VL5.20 was released including https://github.com/vega/vega-lite/pull/9374🙌️. Thanks for releasing @kanitw!

altair/vegalite/v5/api.py
```diff
@@ -3816,33 +3816,140 @@ def add_selection(self, *params) -> Self:  # noqa: ANN002
...
+                    (
+                        enc
+                        for enc in possible_legend_encodings
+                        if not isinstance(
+                            interactive_chart.encoding[enc],
+                            utils.schemapi.UndefinedType,
+                        )
+                    ),
+                    None,  # type: ignore
+                )
```
joelostblom, line 3913: I'm unsure what the best way is to deal with this mypy error:

```
altair/vegalite/v5/api.py:3903: error: Generator has incompatible item type
"str"; expected
"Literal['radius', 'radius2', 'color', 'fill', 'stroke', 'opacity', 'fillOpacity', 'strokeOpacity', 'strokeWidth', 'strokeDash', 'size', 'angle', 'shape']"
 [misc]
                            enc
```

I could ignore it as for the default argument, but maybe there is a better way?

joelostblom (commented):
Rebased on https://github.com/vega/altair/pull/3525 and implemented the suggestion above with using `react: false` in the parameter setup to create the constant color domain. Tasks left to take care of:

- [ ] Sort the legend. I can't find a way to sort the array produced by`domain()`.
- [ ] Figure out how to support ordinal encodings as it seems like using `domain()` is incompatible with ordinal color schemes https://github.com/vega/vega/issues/3959. Converting to nominal and sorting would be one solution but the colorscale would then be the nominal one instead of the expected ordinal one.
- [ ] Support the `size` encoding. For some unknown reason, when the size encoding is used, no marks are produced in the graph. It works when it is quantitative but not nominal.
- [x] Decide if the `tooltip` parameter should take a list of encoding channels or just be a boolean
    - (edit) Thinking about this more, I believe that a boolean is fine since it is easy to just switch to the `tooltip` encoding when wanting to pass a list
- [ ] Write tests
- [ ] Update docs

altair/vegalite/v5/api.py
```diff
@@ -3816,33 +3816,140 @@ def add_selection(self, *params) -> Self:  # noqa: ANN002
...
+                }
+            else:
+                interactive_chart.mark.tooltip = tooltip
+
+        if legend:
+            if isinstance(legend, str):
+                legend_encoding = legend
+            else:
+                # Set the legend to commonly used encodings by default
+                possible_legend_encodings = [
```
dangotbanned, line 3889:
```suggestion
                possible_legend_encodings: SingleDefUnitChannel_T = [
```

altair/vegalite/v5/api.py
dangotbanned:
```suggestion
                legend_encoding = next(
                    (
                        enc
                        for enc in possible_legend_encodings
                        if not utils.is_undefined(
                            interactive_chart.encoding[enc]
                        )
                    ),
                    None,  # type: ignore
                )
```

Sorry I've hijacked this thread for this one, but I do know how else to do a multi-line suggestion.

altair/vegalite/v5/api.py
```diff
@@ -3816,33 +3816,140 @@ def add_selection(self, *params) -> Self:  # noqa: ANN002
         return self.add_params(*params)
 
     def interactive(
-        self, name: str | None = None, bind_x: bool = True, bind_y: bool = True
+        self,
+        name: Optional[str] | None = None,
```
dangotbanned, line 3820: `name` doesn't seem to be in-use yet, but unless you're planning to have different behaviour for `Undefined` vs `None` this could be simplified to:

```suggestion
        name: str | None = None,
```

When you wrote this, the it would have referred to `typing.Optional` and not the `altair`-specific type:

https://github.com/vega/altair/blob/f0c1e0a77bf8e3660fdabf484229a47a021d1f8c/altair/utils/schemapi.py#L787-L817

altair/vegalite/v5/api.py
```diff
@@ -3814,33 +3814,139 @@ def add_selection(self, *params) -> Self:  # noqa: ANN002
         return self.add_params(*params)
 
     def interactive(
```
joelostblom, line 3816:
```suggestion
    def interactive(  # noqa: C901
```

dangotbanned:
@joelostblom if you'd find it helpful, I could take a stab at resolving all the typing/lint issues?

https://github.com/vega/altair/actions/runs/10338393500/job/28616425584?pr=3394

Thought I'd ask, since they're usually easier to fix in an IDE

joelostblom:
@dangotbanned Thanks for offering, I think I just got the mypy checks to pass, but my solution of listing the types three times seems rather inelegant and I suspect there is a better way. Please have a look if you got the time.

And sorry for the noise here, I was trying to resolve these locally, but for some reason I see 40 unrelated mypy errors when I run the tests on my machine (after creating a new virtual env and making sure mypy is the latest version), so it becomes a bit hard to troubleshoot locally. Any insight into this would be helpful too, I'm still learning about using types and mypy effectively.

<details><summary>All the error I'm seeing</summary>

```
altair/vegalite/v5/schema/_typing.py:65: error: Type variable
"altair.vegalite.v5.schema._typing.T" is unbound  [valid-type]
    ...eq = TypeAliasType("OneOrSeq", Union[T, Sequence[T]], type_params=(T,)...
                                                        ^
altair/vegalite/v5/schema/_typing.py:65: note: (Hint: Use "Generic[T]" or "Protocol[T]" base class to bind "T" inside a class)
altair/vegalite/v5/schema/_typing.py:65: note: (Hint: Use "T" in function signature to bind "T" inside a function)
altair/vegalite/v5/schema/_typing.py:65: error: Argument "type_params" to
"TypeAliasType" has incompatible type "tuple[object]"; expected
"tuple[TypeVar | ParamSpec | TypeVarTuple, ...]"  [arg-type]
    ...q = TypeAliasType("OneOrSeq", Union[T, Sequence[T]], type_params=(T,))
                                                                        ^~~~
altair/utils/plugin_registry.py:6: error: Module "typing_extensions" has no
attribute "TypeIs"  [attr-defined]
    from typing_extensions import TypeAliasType, TypeIs, TypeVar
    ^
altair/utils/plugin_registry.py:15: error: Argument "type_params" to
"TypeAliasType" has incompatible type "tuple[object]"; expected
"tuple[TypeVar | ParamSpec | TypeVarTuple, ...]"  [arg-type]
    Plugin = TypeAliasType("Plugin", Callable[..., R], type_params=(R,))
                                                                   ^~~~
altair/utils/plugin_registry.py:16: error: Variable
"altair.utils.plugin_registry.Plugin" is not valid as a type  [valid-type]
    PluginT = TypeVar("PluginT", bound=Plugin[Any])
                                       ^
altair/utils/plugin_registry.py:16: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/utils/plugin_registry.py:129: error: Variable
"altair.utils.plugin_registry.Plugin" is not valid as a type  [valid-type]
            self._active: Plugin[R] | None = None
                          ^
altair/utils/plugin_registry.py:129: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/utils/plugin_registry.py:248: error: Variable
"altair.utils.plugin_registry.Plugin" is not valid as a type  [valid-type]
        def get(self) -> partial[R] | Plugin[R] | None:
                                      ^
altair/utils/plugin_registry.py:248: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/utils/schemapi.py:53: error: Module "typing_extensions" has no attribute
"TypeIs"  [attr-defined]
            from typing_extensions import TypeIs
            ^
altair/utils/schemapi.py:437: error: Unused "type: ignore" comment 
[unused-ignore]
            value_strings = [",".join(err.validator_value) for err in erro...
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~...
altair/utils/schemapi.py:1408: error: Item "list[dict[str, Any]]" of
"dict[str, Any] | list[dict[str, Any]]" has no attribute "items"  [union-attr]
                    for k, v in dct.items()
                                ^~~~~~~~~
altair/utils/core.py:658: error: No overload variant of "from_native" matches
argument types "DataFrame | DataFrameLike | None", "bool"  [call-overload]
            data_nw = nw.from_native(data, eager_or_interchange_only=True)
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
altair/utils/core.py:658: note: Possible overload variants:
altair/utils/core.py:658: note:     def from_native(native_dataframe: Any, *, strict: Literal[False], eager_only: None = ..., eager_or_interchange_only: Literal[True], series_only: None = ..., allow_series: Literal[True]) -> Any
altair/utils/core.py:658: note:     def from_native(native_dataframe: Any, *, strict: Literal[False], eager_only: Literal[True], eager_or_interchange_only: None = ..., series_only: None = ..., allow_series: Literal[True]) -> Any
altair/utils/core.py:658: note:     def [IntoDataFrameT, T] from_native(native_dataframe: IntoDataFrameT | T, *, strict: Literal[False], eager_only: None = ..., eager_or_interchange_only: Literal[True], series_only: None = ..., allow_series: None = ...) -> DataFrame[IntoDataFrameT] | T
altair/utils/core.py:658: note:     def [IntoDataFrameT, T] from_native(native_dataframe: IntoDataFrameT | T, *, strict: Literal[False], eager_only: Literal[True], eager_or_interchange_only: None = ..., series_only: None = ..., allow_series: None = ...) -> DataFrame[IntoDataFrameT] | T
altair/utils/core.py:658: note:     def from_native(native_dataframe: Any, *, strict: Literal[False], eager_only: None = ..., eager_or_interchange_only: None = ..., series_only: None = ..., allow_series: Literal[True]) -> Any
altair/utils/core.py:658: note:     def from_native(native_dataframe: Any, *, strict: Literal[False], eager_only: None = ..., eager_or_interchange_only: None = ..., series_only: Literal[True], allow_series: None = ...) -> Any
altair/utils/core.py:658: note:     def [IntoFrameT, T] from_native(native_dataframe: IntoFrameT | T, *, strict: Literal[False], eager_only: None = ..., eager_or_interchange_only: None = ..., series_only: None = ..., allow_series: None = ...) -> DataFrame[IntoFrameT] | LazyFrame[IntoFrameT] | T
altair/utils/core.py:658: note:     def [IntoDataFrameT] from_native(native_dataframe: IntoDataFrameT, *, strict: Literal[True] = ..., eager_only: None = ..., eager_or_interchange_only: Literal[True], series_only: None = ..., allow_series: None = ...) -> DataFrame[IntoDataFrameT]
altair/utils/core.py:658: note:     def [IntoDataFrameT] from_native(native_dataframe: IntoDataFrameT, *, strict: Literal[True] = ..., eager_only: Literal[True], eager_or_interchange_only: None = ..., series_only: None = ..., allow_series: None = ...) -> DataFrame[IntoDataFrameT]
altair/utils/core.py:658: note:     def from_native(native_dataframe: Any, *, strict: Literal[True] = ..., eager_only: None = ..., eager_or_interchange_only: None = ..., series_only: None = ..., allow_series: Literal[True]) -> DataFrame[Any] | LazyFrame[Any] | Series
altair/utils/core.py:658: note:     def from_native(native_dataframe: Any, *, strict: Literal[True] = ..., eager_only: None = ..., eager_or_interchange_only: None = ..., series_only: Literal[True], allow_series: None = ...) -> Series
altair/utils/core.py:658: note:     def [IntoFrameT] from_native(native_dataframe: IntoFrameT, *, strict: Literal[True] = ..., eager_only: None = ..., eager_or_interchange_only: None = ..., series_only: None = ..., allow_series: None = ...) -> DataFrame[IntoFrameT] | LazyFrame[IntoFrameT]
altair/utils/core.py:658: note:     def from_native(native_dataframe: Any, *, strict: bool, eager_only: bool | None, eager_or_interchange_only: bool | None = ..., series_only: bool | None, allow_series: bool | None) -> Any
altair/utils/data.py:44: error: Module "typing_extensions" has no attribute
"TypeIs"  [attr-defined]
        from typing_extensions import TypeIs
        ^
altair/vegalite/v5/schema/channels.py:30904: error: Variable
"altair.vegalite.v5.schema._typing.OneOrSeq" is not valid as a type 
[valid-type]
    ChannelDetail: TypeAlias = OneOrSeq[Union[str, Detail, Map]]
                               ^
altair/vegalite/v5/schema/channels.py:30904: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/vegalite/v5/schema/channels.py:30919: error: Variable
"altair.vegalite.v5.schema._typing.OneOrSeq" is not valid as a type 
[valid-type]
    ChannelOrder: TypeAlias = OneOrSeq[Union[str, Order, Map, OrderValue]]
                              ^
altair/vegalite/v5/schema/channels.py:30919: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/vegalite/v5/schema/channels.py:30938: error: Variable
"altair.vegalite.v5.schema._typing.OneOrSeq" is not valid as a type 
[valid-type]
    ChannelTooltip: TypeAlias = OneOrSeq[Union[str, Tooltip, Map, TooltipV...
                                ^
altair/vegalite/v5/schema/channels.py:30938: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/vegalite/v5/schema/channels.py:30960: error: Variable
"altair.vegalite.v5.schema._typing.OneOrSeq" is not valid as a type 
[valid-type]
            detail: Optional[OneOrSeq[str | Detail | Map]] = Undefined,
                             ^
altair/vegalite/v5/schema/channels.py:30960: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/vegalite/v5/schema/channels.py:30979: error: Variable
"altair.vegalite.v5.schema._typing.OneOrSeq" is not valid as a type 
[valid-type]
            order: Optional[OneOrSeq[str | Order | Map | OrderValue]] = Un...
                            ^
altair/vegalite/v5/schema/channels.py:30979: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/vegalite/v5/schema/channels.py:31000: error: Variable
"altair.vegalite.v5.schema._typing.OneOrSeq" is not valid as a type 
[valid-type]
            tooltip: Optional[OneOrSeq[str | Tooltip | Map | TooltipValue]...
                              ^
altair/vegalite/v5/schema/channels.py:31000: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/vegalite/v5/schema/channels.py:31450: error: Variable
"altair.vegalite.v5.schema._typing.OneOrSeq" is not valid as a type 
[valid-type]
        detail: OneOrSeq[str | Detail | Map]
                ^
altair/vegalite/v5/schema/channels.py:31450: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/vegalite/v5/schema/channels.py:31461: error: Variable
"altair.vegalite.v5.schema._typing.OneOrSeq" is not valid as a type 
[valid-type]
        order: OneOrSeq[str | Order | Map | OrderValue]
               ^
altair/vegalite/v5/schema/channels.py:31461: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/vegalite/v5/schema/channels.py:31474: error: Variable
"altair.vegalite.v5.schema._typing.OneOrSeq" is not valid as a type 
[valid-type]
        tooltip: OneOrSeq[str | Tooltip | Map | TooltipValue]
                 ^
altair/vegalite/v5/schema/channels.py:31474: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/vegalite/v5/api.py:61: error: Module "typing_extensions" has no
attribute "TypeIs"  [attr-defined]
            from typing_extensions import Required, TypeIs
            ^
altair/vegalite/v5/api.py:480: error: Item "UndefinedType" of
"TopLevelSelectionParameter | SelectionParameter | UndefinedType" has no
attribute "select"  [union-attr]
                if field_name in getattr(param.select, prop):
                                         ^~~~~~~~~~~~
altair/vegalite/v5/api.py:615: error: Unused "type: ignore" comment 
[unused-ignore]
    class _ConditionClosed(TypedDict, closed=True, total=False):  # type: ...
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~...
altair/vegalite/v5/api.py:625: error: Unused "type: ignore" comment 
[unused-ignore]
    class _ConditionExtra(TypedDict, closed=True, total=False):  # type: i...
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~...
altair/vegalite/v5/api.py:632: error: Variable
"altair.vegalite.v5.schema._typing.OneOrSeq" is not valid as a type 
[valid-type]
        __extra_items__: _StatementType | OneOrSeq[_LiteralValue]
                                          ^
altair/vegalite/v5/api.py:632: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
altair/vegalite/v5/api.py:649: error: Unused "type: ignore" comment 
[unused-ignore]
    class _Value(TypedDict, closed=True, total=False):  # type: ignore[cal...
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~...
altair/vegalite/v5/api.py:758: error: Incompatible types in assignment
(expression has type
"Parameter | Expr | Mapping[str, Any] | str | Expression | PredicateComposition | OperatorMixin | UndefinedType",
variable has type
"Parameter | Expr | Mapping[str, Any] | str | Expression | PredicateComposition | OperatorMixin")
 [assignment]
            composed = predicate
                       ^~~~~~~~~
altair/vegalite/v5/api.py:1794: error: Argument 1 to
"to_eager_narwhals_dataframe" has incompatible type "Any | UndefinedType";
expected "NativeFrame | DataFrame[Any] | DataFrameLike"  [arg-type]
                    data = _to_eager_narwhals_dataframe(original_data)
                                                        ^~~~~~~~~~~~~
altair/vegalite/v5/api.py:1797: error: Incompatible types in assignment
(expression has type "Any | UndefinedType", variable has type "DataFrame[Any]") 
[assignment]
                    data = original_data
                           ^~~~~~~~~~~~~
altair/vegalite/v5/api.py:4055: error: Argument 1 to "_check_if_valid_subspec"
has incompatible type
"Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart | UndefinedType";
expected
"Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart | FacetSpec | LayerSpec | RepeatSpec | FacetedUnitSpec | LayerRepeatSpec | NonNormalizedSpec | NonLayerRepeatSpec | ConcatSpecGenericSpec | ConcatSpecGenericSpec | HConcatSpecGenericSpec | VConcatSpecGenericSpec | Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart | UnitSpec | LayerSpec"
 [arg-type]
            _check_if_valid_subspec(spec, "RepeatChart")
                                    ^~~~
altair/vegalite/v5/api.py:4057: error: Incompatible types in assignment
(expression has type
"list[Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart]",
variable has type
"list[Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart | UndefinedType]")
 [assignment]
            params, _spec_as_list = _combine_subchart_params(params, _spec...
                                    ^
altair/vegalite/v5/api.py:4057: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
altair/vegalite/v5/api.py:4057: note: Consider using "Sequence" instead, which is covariant
altair/vegalite/v5/api.py:4057: error: Argument 2 to "_combine_subchart_params"
has incompatible type
"list[Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart | UndefinedType]";
expected
"list[Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart]"
 [arg-type]
    ...arams, _spec_as_list = _combine_subchart_params(params, _spec_as_list)
                                                               ^~~~~~~~~~~~~
altair/vegalite/v5/api.py:4063: error: Argument 2 to "_repeat_names" has
incompatible type
"list[str] | LayerRepeatMapping | RepeatMapping | UndefinedType"; expected
"list[str] | LayerRepeatMapping | RepeatMapping"  [arg-type]
                params = _repeat_names(params, repeat, spec)
                                               ^~~~~~
altair/vegalite/v5/api.py:4632: error: Argument 1 to "_check_if_valid_subspec"
has incompatible type
"Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart | UndefinedType";
expected
"Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart | FacetSpec | LayerSpec | RepeatSpec | FacetedUnitSpec | LayerRepeatSpec | NonNormalizedSpec | NonLayerRepeatSpec | ConcatSpecGenericSpec | ConcatSpecGenericSpec | HConcatSpecGenericSpec | VConcatSpecGenericSpec | Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart | UnitSpec | LayerSpec"
 [arg-type]
            _check_if_valid_subspec(spec, "FacetChart")
                                    ^~~~
altair/vegalite/v5/api.py:4634: error: Incompatible types in assignment
(expression has type
"list[Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart]",
variable has type
"list[Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart | UndefinedType]")
 [assignment]
            params, _spec_as_list = _combine_subchart_params(params, _spec...
                                    ^
altair/vegalite/v5/api.py:4634: note: "List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
altair/vegalite/v5/api.py:4634: note: Consider using "Sequence" instead, which is covariant
altair/vegalite/v5/api.py:4634: error: Argument 2 to "_combine_subchart_params"
has incompatible type
"list[Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart | UndefinedType]";
expected
"list[Chart | RepeatChart | ConcatChart | HConcatChart | VConcatChart | FacetChart | LayerChart]"
 [arg-type]
    ...arams, _spec_as_list = _combine_subchart_params(params, _spec_as_list)
                                                               ^~~~~~~~~~~~~
altair/vegalite/v5/api.py:4803: error: Item "UndefinedType" of
"Any | Sequence[VariableParameter | TopLevelSelectionParameter | SelectionParameter] | UndefinedType"
has no attribute "__iter__" (not iterable)  [union-attr]
            for param in subchart.params:
                         ^~~~~~~~~~~~~~~
altair/vegalite/v5/api.py:4833: error: Item "UndefinedType" of
"Sequence[VariableParameter | TopLevelSelectionParameter | SelectionParameter] | UndefinedType"
has no attribute "__iter__" (not iterable)  [union-attr]
        for param in params:
                     ^~~~~~
altair/vegalite/v5/api.py:4852: error: Item "UndefinedType" of
"Any | Sequence[VariableParameter | TopLevelSelectionParameter | SelectionParameter] | UndefinedType"
has no attribute "__iter__" (not iterable)  [union-attr]
            for param in subchart.params:
                         ^~~~~~~~~~~~~~~
altair/vegalite/v5/api.py:4935: error: Item "UndefinedType" of
"Sequence[VariableParameter | TopLevelSelectionParameter | SelectionParameter] | UndefinedType"
has no attribute "__iter__" (not iterable)  [union-attr]
        for param in params:
                     ^~~~~~
Found 41 errors in 7 files (checked 370 source files)
```

<details>

dangotbanned:
@joelostblom 
> And sorry for the noise here, I was trying to resolve these locally, but for some reason I see 40 unrelated mypy errors when I run the tests on my machine (after creating a new virtual env and making sure mypy is the latest version), so it becomes a bit hard to troubleshoot locally. Any insight into this would be helpful too, I'm still learning about using types and mypy effectively.
> All the error I'm seeing ...

So this one stood out to me, as it would appear your env is using an old `typing_extensions` version:

```py
altair/utils/plugin_registry.py:6: error: Module "typing_extensions" has no attribute "TypeIs"  [attr-defined]
    from typing_extensions import TypeAliasType, TypeIs, TypeVar
```

[TypeIs](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.TypeIs) was added in `v4.10.0` and this is listed as the lower bound in https://github.com/vega/altair/blob/0062e6228134f176bb9fbaa944724039bb6b043a/pyproject.toml#L14-L18

> mypy errors when I run the tests on my machine (after creating a new virtual env and making sure mypy is the latest version)

Apologies if this seems like an odd question, but are you managing envs manually or using `hatch`?
I would have expected this dependency to be updated by `hatch` (if needed) each time you run [`hatch test`](https://github.com/vega/altair/blob/main/CONTRIBUTING.md#testing-your-changes)

dangotbanned:
> @dangotbanned Thanks for offering, I think I just got the mypy checks to pass, but my solution of listing the types three times seems rather inelegant and I suspect there is a better way. Please have a look if you got the time.

No problem, having a look now

joelostblom (commented):
Thanks for the pointers @dangotbanned. I was running `hatch run mypy altair tests` directly because I didn't want to run the full test suite. When I run either `hatch test` or just `mypy altair tests` there are no mypy errors (the output is the same as that from github actions) so I guess for some reason, `hatch run ...` is picking up another environment on my system with an old version of `typing_extensions` installed. Not sure how that happens since `which -a hatch` is only pointing to my dedicated conda environment, but I will figure it out eventually (or just run `mypy altair tests` since that does what I want). 

> No problem, having a look now

Thanks!

dangotbanned:
@joelostblom everything should be passing in [`a8001da` (#3394)](https://github.com/vega/altair/pull/3394/commits/a8001da0245ad5d7515c2a90877c7b7cb2b91ce0)

I couldn't split the changes up across commits, so I'll add some comments w/ explanations shortly

altair/vegalite/v5/api.py
```diff
@@ -3865,101 +3850,81 @@ def interactive(  # noqa: C901
...
+                    initial_computed_domain,
+                    nonreactive_domain,
+                ).transform_filter(legend_selection)
+            else:
+                msg = f"Expected only 'nominal' legend type but got {legend_type!r}"
+                raise NotImplementedError(msg)
+        return chart
+
+
+LegendChannel_T: TypeAlias = Literal[
```
dangotbanned, line 3893: This is the singular definintion you can reuse.

If any of the *TODO*s get resolved, everything will stay in sync w/ this.

altair/vegalite/v5/api.py
```diff
@@ -3865,101 +3850,81 @@ def interactive(  # noqa: C901
...
+
+
+def _infer_legend_encoding(encoding: FacetedEncoding, /) -> LegendChannel_T:
+    """Set the legend to commonly used encodings by default."""
+    _channels = t.get_args(LegendChannel_T)
+    it = (ch for ch in _channels if not utils.is_undefined(encoding[ch]))
+    if legend := next(it, None):
+        return legend
+    else:
+        msg = f"Unable to infer target channel for 'legend'.\n\n{encoding!r}"
```
dangotbanned, line 3926: It wasn't clear to me what the intention was for this case. 

At a minimum I think there should be a warning displayed, rather than silently continuing.

altair/vegalite/v5/api.py
```diff
@@ -3865,101 +3850,81 @@ def interactive(  # noqa: C901
...
+    "strokeWidth",
+    "strokeDash",
+    "angle",  # TODO Untested
+    "radius",  # TODO Untested
+    "radius2",  # TODO Untested
+    # "size",  # TODO Currently size is not working, renders empty legend
+]
+
+
+def _add_tooltip(chart: _TChart, /) -> _TChart:
```
dangotbanned, line 3910: Mainly thinking ahead here for all the other `ChartType`s.

I imagine this would be preferred to simply repeating the code in each method.

altair/vegalite/v5/api.py
```diff
@@ -3813,29 +3813,14 @@ def add_selection(self, *params) -> Self:  # noqa: ANN002
...
-            "strokeWidth",
-            "strokeDash",
-            "angle",  # TODO Untested
-            "radius",  # TODO Untested
-            "radius2",  # TODO Untested
-            # "size",  # TODO Currently size is not working, renders empty legend
-        ] = True,
-    ) -> Self:
+        legend: bool | LegendChannel_T = False,
+    ) -> Chart:
```
dangotbanned, line 3823: Using `Self` here is mainly useful when this is a base class.

I know this wasn't added in this PR, but using `Self` would cause conflicts in any generic "helper" functions like [`a8001da` (#3394)](https://github.com/vega/altair/pull/3394/commits/a8001da0245ad5d7515c2a90877c7b7cb2b91ce0#r1712993394)

altair/vegalite/v5/api.py
```diff
@@ -3813,29 +3813,14 @@ def add_selection(self, *params) -> Self:  # noqa: ANN002
...
-            "strokeOpacity",
-            "strokeWidth",
-            "strokeDash",
-            "angle",  # TODO Untested
-            "radius",  # TODO Untested
-            "radius2",  # TODO Untested
-            # "size",  # TODO Currently size is not working, renders empty legend
-        ] = True,
-    ) -> Self:
+        legend: bool | LegendChannel_T = False,
```
dangotbanned, line 3822: Had to change the default to `False` (for now), as you'll need to adapt the existing tests to handle the new behaviour.

altair/vegalite/v5/api.py
```diff
@@ -3865,101 +3850,81 @@ def interactive(  # noqa: C901
             encodings.append("x")
         if bind_y:
             encodings.append("y")
-        interactive_chart = self.add_params(
+        chart: Chart = self.copy().add_params(
```
dangotbanned, line 3853: Moved the `.copy()` to occur first, otherwise the params are added to the original chart as well

altair/vegalite/v5/api.py
```diff
@@ -3865,101 +3850,81 @@ def interactive(  # noqa: C901
...
-                    ).transform_filter(legend_selection)
-        return interactive_chart
+                    scale["domain"] = nonreactive_domain
+                chart = chart.add_params(
+                    legend_selection,
+                    initial_computed_domain,
+                    nonreactive_domain,
+                ).transform_filter(legend_selection)
+            else:
+                msg = f"Expected only 'nominal' legend type but got {legend_type!r}"
```
dangotbanned, line 3888: Same reasoning as for [`a8001da` (#3394)](https://github.com/vega/altair/pull/3394/commits/a8001da0245ad5d7515c2a90877c7b7cb2b91ce0#r1712993209)

altair/vegalite/v5/api.py
joelostblom: I'm not sure what is the best approach here. I agree that a warning/error is more explicit than silently continuing, but since `.interactive()` is a sort of convenient method to quickly add basic interactivity to charts, I also want to keep it simple to use. To me, the least friction for the user is if legend interactivity is applied when it makes sense and not applied when it doesn't make sense without being too noisy with warnings and errors. I'm thinking that people would not expect a gradient color legend to be interactive in any way, but they would expect to click on discrete points/lines to select and unselect groups. This is also how e.g. plotly works, see the examples on this page https://plotly.com/python/plotly-express/#gallery. If we go this path, one case that might be confusing in altair is [quantitative scales in the size encoding](https://altair-viz.github.io/gallery/histogram_scatterplot.html) since these appear to have discrete categories in the legend but clicking them would not select anything (since there are rarely observations with those exact values). I think this is minor, and am still leaning towards the "apply interactivity where it makes sense" approach for `.interactive()` and I'm ok with this method being less explicitly verbose since I see it as more of a convenience.

joelostblom (commented):
Thanks for all the help here @dangotbanned, your changes are really neat!

altair/vegalite/v5/api.py
dangotbanned: > but didn't know of `typing.get_args()` , that's very helpful!

It is quite handy for situations like this, where during development it isn't known what the final set will be.

Since it is new to you, I'd say try not rely on it too much.
You might run into issues using it on types declared **outside of the current module** - see https://github.com/vega/altair/pull/3516#discussion_r1703151552 - which a lot of `Literal`(s) are.

I would suggest that once the final `LegendChannel_T` definition is decided, then add a either a `frozenset` (if the order *is not* important) or a `tuple` (if the order *is* important) containing those elements.
It would also be faster, but not sure that is much of a concern here

dangotbanned:
> Thanks for the pointers @dangotbanned. I was running `hatch run mypy altair tests` directly because I didn't want to run the full test suite. When I run either `hatch test` or just `mypy altair tests` there are no mypy errors (the output is the same as that from github actions) so I guess for some reason, `hatch run ...` is picking up another environment on my system with an old version of `typing_extensions` installed. Not sure how that happens since `which -a hatch` is only pointing to my dedicated conda environment, but I will figure it out eventually (or just run `mypy altair tests` since that does what I want).

So when you use `hatch run script_name` it will run in the `default` environment.

`hatch test` uses a special environment called `hatch-test`, with some options specific to testing.

# PRs
- https://github.com/vega/altair/pull/3461
- https://github.com/vega/altair/pull/3477

# Related
- https://hatch.pypa.io/latest/environment/
- https://hatch.pypa.io/latest/environment/#selection
- https://hatch.pypa.io/latest/environment/#removal
- https://hatch.pypa.io/latest/config/internal/testing/

altair/vegalite/v5/api.py
joelostblom: Yes, that sounds good. I think the only other type we will support it `'ordinal'`. I don't see how either `'quantitative'` or `'temporal'` would make use of interactivity (I guess one could technically imagine an interval selection on a quantitative legend but that is currently not possible in VegaLite).

altair/vegalite/v5/api.py
joelostblom: >  If I'm a new user and I call Chart.interactive(legend=True), but result doesn't contain a legend, how am I supposed to know what went wrong?

Good point, and it brings up an interesting distinction I didn't think of before. Whereas setting `tooltip=True` actually **creates** a tooltip, setting `legend=True` converts an existing legend into an interactive/clickable format (but doesn't create one from scratch as you pointed out). I will make this distinction explicit via the docstring and/or a more suitable parameter name. Then we can consider whether that is clear enough or we also need the warning/error; I appreciate your reasoning and suggestion with the nudging message, but I want to avoid that users need to pass `legend=False` just because there color scale is using a quantitative encoding (so eventually I will likely favor either a warning or an explicit docstring, but I will leave the error raise for now until next time I can work on this PR)

dangotbanned: That's fair!

> > If I'm a new user and I call Chart.interactive(legend=True), but **result** doesn't contain a legend, how am I supposed to know what went wrong?

> Good point, and it brings up an interesting distinction I didn't think of before. Whereas setting `tooltip=True` actually **creates** a tooltip, setting `legend=True` converts an existing legend into an interactive/clickable format (but doesn't create one from scratch as you pointed out)

My bad for writing the last comment without testing.
I was assuming it created a legend (forgetting that it would already be there).

I guess the starting point would be the same though, encountering that error currently would require `legend=True` and a `Chart` without a channel displaying a legend?

dangotbanned:
@joelostblom when you're available, could you add some of the cases you've tried as tests please?
Just ending on `chart.to_dict()` where you'd previously be displaying in a notebook would be enough.

Thinking that I may be able to dip into this again soon, but want to make sure everything you've tried out so far stays working 👍

joelostblom:
@dangotbanned Sorry it took me a few days to get back to this, but I just pushed a test suite draft together with a couple of minor fixes to the changes you made to ensure they work with the test cases I had in mind. If you find the time, feel free to modify as you see fit and thanks for helping out with this PR!

altair/vegalite/v5/api.py
```diff
@@ -3909,11 +3909,10 @@ def interactive(
 
 
 def _add_tooltip(chart: _TChart, /) -> _TChart:
-    mark = chart.mark
-    if isinstance(mark, str):
-        mark = {"type": mark, "tooltip": True}
+    if isinstance(chart.mark, str):
+        chart.mark = {"type": chart.mark, "tooltip": True}
```
dangotbanned, line 3913: 👍 

Glad to see you was spotted after testing @joelostblom .
I did a lot of renaming & aliasing, but this one was both invalid & didn't save any lines.

mattijn:
While testing this option with zoomable scales and fixed legends in VL I noticed that the selection of the legend is emptied of all values (resetting the domain) upon a mouseup event on pan. Raising here, as that seems undesirable? See following animated gif ([Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtBmx5CBKMtlSIyuMOLbKsuJgKycCXhX6r1ERxAACxACMlABMAJyUAAwEXkgmSNoQFnIgAL5KOAlICBBoANqgstlw+gBebGzITIJwAPrQSDW5Sj41UJhooJgAnjgl6DSymHDaxI2KIHCyUGzKg2S5qHkgAB4T3SAAukpQNQloAGaNPukgTIPK+g1NaQqFxfozgmza9dZw7ezyLe-tnSA9Pr6HBsQYdJRTGZzWQLfIgJ4vLanc6yS7oGpkKaXdL3BD9eGVHAMYbKWrKSpIQYTOArHDafTk5CDAAUAHIEdpWQBKW64-EcskUqlKbRwJB-VBHQQ+CG0+noGYIIkkwVM+SpbYA7RIWQQA4vBD5UAHGiCYby0CZbWGhVsZ6vVofTBfNIapTIbQAa3+gPxILBExNghqaMw2gYcCUmAqZpoODQYYjp0hs3m-w5-xN1jRIAA8toaLRvgDevjZJVBuMWlBGv1QIzKfJNJM5Y87S9VY3Xac1s2s4IcwAJF4+EEAdxGE19+gAjgwdc71M7SLcQBs+zRs-oALKmuAQWp9V4AcUaz2L0-Qc4XdDUNBXqUfQA)):

zoomable_react

And when trying the following Altair specification using this PR:
```python
import altair as alt

source = data.movies.url
predicate = (alt.datum.IMDB_Rating == None) | (alt.datum.Rotten_Tomatoes_Rating == None)

chart = alt.Chart(source).mark_point(invalid=None).encode(
    x="IMDB_Rating:Q",
    y="Rotten_Tomatoes_Rating:Q",
    color=alt.when(predicate).then(alt.value("grey"))
)
chart.interactive()   
```
It now will give the following error:

> ```python
> TypeError: Then.__init__() got an unexpected keyword argument 'condition'
> ```

I don't think this option is currently covered in the new tests added within this PR.

dangotbanned:
@mattijn I will take a look tomorrow, just added a commit to force a fail.

Seems to be related to https://github.com/vega/altair/pull/3543 more than `Chart.interactive` I think

dangotbanned:
@mattijn just pushed a fix in #3553

joelostblom:
@mattijn in response to your comment:

>  I noticed that the selection of the legend is emptied of all values (resetting the domain) upon a mouseup event on pan. Raising here, as that seems undesirable? 

I agree that it is undesireable, but it is a general issue in Vega-Lite rather than specific to this PR, [see for example this example.](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAZgAY2AacaYsiygAlMiRoVYcAvpO50AhoTl4QpAE4AbFCDGEADpWQB6Q2DpQAdACtKdTOrhpV5qJkKGougLaG0mBHIBaeUVKV0oAATQARnMAJgBOczZDYLkTOVVKK0poEBkQTwyAa2VCAE9dTC1dCBJxfMwoSDoSJFQedQhVZXg7Oi0AeVVEEhBucsqtKAhPEjlNfIhdOTA4cuUCFuJc9uXVOU9qjIOAfRYxkDR50irkNnMADnyr9RuUe9j8gA8erHV+5AgQRdUI1bCYbrjCq3EAAR1IciIawUDiq+TKvz6WgAsnB1JhKCdKqoTgBxeadKAXCYw+GI4iKYi+PL5PYHSgoADaoCgBxhbM8ZwuoXxYHE7Rp1VqRAuvX+HOQnJAQxGVIAuvkAEYkAEgfEIRr9GQ8vlHfaCgAswrsmDFpWhWjqEJeF0azVaCqVP24GI13G1UF1lDA8wJeTV3AAJMHGJhClodPojD4-HJzAg1oxSJrzHAIIYY3G0r5-AF7IRMD4AKxxe4xay5SRAA). I think the solution would be to fire the legend selection event only on `pointerup` events that are not part of `dragleave`. I don't know if it is possible to combine and negate selectors like that so I asked on slack if anyone knows.

dwootton:
Got this working in vegalite! This approach uses an event stream filtered to legend marks such that only click events on the legend are captured. It also adds a clear behavior on double click, similar to the clear behavior for interval selections. 



[Working example from above.](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtBmx5CBKMtlSIyuMOLbKsuJgKycCXhX6r1ERxAACxACMlABMAJyUAAwEXkgmSNoQFnIgAL5KOAlICBBoANqgstlw+gBebGzITIJwAPrQSDW5Sj41UJhooJgAnjgl6DSymHDaxI2KIHCyUGzKg2S5qHkgAB4T3SAAukpQNQloAGaNPukgTIPK+g1NaQqFxfozgmza9dZw7ezyLe-tnSA9Pr6HBsQYdJRTGZzWQLfIgJ4vLZKFKaEDIbQAayKCH6IAA8toaLRZLUamQpsp6t0EEw2IJmgDerjdjQoBjbvC9tp9Mpqiy2RzzrJLqiyRT-uisQ90ASiYNSeQKVSaXSGYDmcIBaltXcQNjmZUcAxhpTlJUkIMJnAVjhuegzchBgAKADkCO0LoAlLd7jjHnSXrUHRbviBtHAkH9UEd6XAITa7fDDca4KbzZbUtsAdokLIIAcXgh8qADjRBMM7aBMjmi+h3W82pgvmlM0pJf91cDQUMJqXBDURZhtAw4wCKuWaDg0EOR6dIbN5v93f9S9YRfjCcSJp30LJKoNxi0oI1+qBg5bUdbbf7nq9z-Jtac1qjV4J1wAJF4+EEAdxG26ZfQAEcGFzJt1CbUgOQ2F8aDXfQAFkyzgCBaj6V4AHFGmeUMdxAECwLoNQaCgx8gA)

dwootton:
You should be able to use Shift + click to do a multi-select! I'd recommend keeping this approach as its the same behavior that's instrumented across all point selections by default. 

If you wanted to get multi-select without shift+click, you can add the toggle property to allow for points to be added with just clicks. 




[Example](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtBmx5CBKMtlSIyuMOLbKsuJgKycCXhX6r1ERxAACxACMlABMAJyUAAwEXkgmSNoQFnIgAL5KOAlICBBoANqgstlw+gBebGzITIJwAPrQSDW5Sj41UJhooJgAnjgl6DSymHDaxI2KIHCyUGzKg2S5qHkgAB4T3SAAukpQNQloAGaNPukgTIPK+g1NaQqFxfozgmza9dZw7ezyLe-tnSA9Pr6HBsQYdJSYNhkMg1fRwUhDCZTGZzWQLfIgJ4vLZKFKaEDIbQAayKCH6IAA8toaLRZLUamQpsp6t0EEw2IJmgDeuTdjQoETbpi9tp9MpqnyBULzrJLviGUz-oSSQ90FSaYN6eQmSy2RyuYDecIpalTXcQKTeZUcAxhszlJUkIMkSscKL0A7kIMABQAcix2l9AEpbvcyY8OS9ap6nd8QNo4Eg-qgjpy4Eo4K73Zjrba4PbHc7UtsAdokLIIAcXgh8qADjRBMN3aBMmWa+gA282pgvmli0plf9DcDQYilPXBDU5ZhtAx0wCKo2aDg0DO56dkbN5v8A-969Y5ZTqbSJsP0LJKoNxi0oI1+qAY878Zm3RHnq9H-JTac1vj94JDwACReHwQQAdxGU8eX0ABHBhyx7dQe1IIUNj-GgD30ABZBs4AgWo+leABxRpnjjM8QDghC6DUGgUO-IA)

joelostblom:
That's odd, it doesn't work for me to shift + click with your example (in either Firefox of Chromium). If I remove the two occurrences of `"markname": "Origin_legend_symbols"`, shift + clicking works as expected

joelostblom:
So interesting... I don't see any errors or logs in the Vega console. The browser debug console just shows the following (and it shows the same for the working spec):

image

I'm on Linux, so that is a difference, but it doesn't seem like that should impact anything here. Let's hear from someone else if it is working for them or not.

mattijn:
Thanks for pushing this further @dwootton, pretty awesome! I also think I understand why you have mixed results. I noticed that @dwootton is clicking on the symbols in the legend and I think that @joelostblom is clicking on the labels in the legend. 

If you change `"markname": "Origin_legend_symbols"` to `"markname": "Origin_legend_labels"` (2x) than you can do multi-select with shift + click on the legend labels. Clicking on the legend symbols behaves as if there is a toggle property (no need to hold shift for multi-select in that case). 

[See the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtBmx5CBKMtlSIyuMOLbKsuJgKycCXhX6r1ERxAACxACMlABMAJyUAAwEXkgmSNoQFnIgAL5KOAlICBBoANqgstlw+gBebGzITIJwAPrQSDW5Sj41UJhooJgAnjgl6DSymHDaxI2KIHCyUGzKg2S5qHkgAB4T3SAAukpQNQn6ytW7NFAA1mlKTIPK+g1NF4XF+jOCbNr11nDt7PItn+2dEA9Pr6HBsQYdJRTGZzWQLfIgF5vLZKFKaEDIbSnIoIfogADy2hotFktRqZCmyjJSCY1maQN6eOOZwuiL22gOR2ELPSICushu6PJlMBmOxT3QhOJgzJ5Ep1NpgnpwKZ3POqQ1CkeuOelRwDGGVOUlSQgwmcBWOA56GNyEGAAoAORI7SOgCUDxAOKZbFe71tpt+IG0cCQANQADNGj4oZbrYi9Qa4EaTWbUtsgdokLIIBG3gh8qAIzRBMNraBMlmC+gXR82pgfml00oxYCVaDwUMJsXBDVBZhtAw4EpMBVSzQcGgB0PedDZvNAS7AcXrIKCUSSRN2+hZJVBuMWlBGv1QAGzeiLVbdX7amf5BreWt0SvBGuABJvHxggDuIy3jP0ABHBhswbdQG1IVkNmfGhV30ABZEs4AgWo+neABxRpXiDbcQGA0C6DUGhIIfIA)

I also changed `"clear": false` to `"clear": "dblclick"` for the `zoomable_scales` selection parameter. I think that is consistent with current behavior.

joelostblom:
Great catch @mattijn! That's it! Ideally we would want both the legend label and symbol to trigger the same selection behavior. I guess we can duplicate the color_selection parameter and legend binding once each for `"Origin_legend_symbols"` and `"Origin_legend_labels"`, but maybe there is a cleaner way?

I still see some unexpected behavior with the spec you posted @mattijn. I start by clicking the "USA" label to select only that to only and then shift+click the label of "Japan" so that I have those two labels selected. Now, if I **regular click** the "Europe" **label**, only "Europe" would be selected (as expected). However, if I instead would have **regular clicked** the "Europe" **symbol**, all three labels are selected (so it behaves as a shift click for some reason; I would expect only Europe to be selected in this case too, just as when the label is clicked)

> I also changed "clear": false to "clear": "dblclick" for the zoomable_scales selection parameter. I think that is consistent with current behavior.

+1

PR #3395, opened 2024-04-07
docs: Add example of reordering stacked bars

joelostblom:
This example shows how to reorder bar segments via a click interaction in the legend. Credit to @dwootton who came up with the spec in [this very neat tweet](https://twitter.com/WoottonDylan/status/1646596161548046346) which I modified slightly to include opacity changes and to allow for sorting multiple segments.

https://github.com/altair-viz/altair/assets/4560057/f4e55602-7c4b-4bc7-a981-4bc3264b3929

### DIFF
--- a/tests/examples_arguments_syntax/interactive_reorder_stacked_bars.py
+++ b/tests/examples_arguments_syntax/interactive_reorder_stacked_bars.py
@@ -1,0 +1,30 @@
+"""
+Reorder stacked bar segments
+============================
+This example uses a calculate transform
+to check the values of the "site" column
+vs the clicked values in the legend,
+and assigns a lower order (0)
+if there is a match.
+The use of "indexOf" checks for equality in an array,
+which here allows for mutliple segments to be reordered
+by holding down the shift key while clicking the legend.
+"""
+# category: interactive charts
+import altair as alt
+from vega_datasets import data
+
+
+selection = alt.selection_point(fields=['site'], bind='legend')
+
+alt.Chart(data.barley.url).mark_bar().transform_calculate(
+    site_order=f"if({selection.name}.site && indexof({selection.name}.site, datum.site) !== -1, 0, 1)"
+).encode(
+    x='sum(yield):Q',
+    y='variety:N',
+    color='site:N',
+    order='site_order:N',
+    opacity=alt.condition(selection, alt.value(0.9), alt.value(0.2))
+).add_params(
+    selection
+)
--- a/tests/examples_arguments_syntax/interactive_reorder_stacked_bars.py
+++ b/tests/examples_arguments_syntax/interactive_reorder_stacked_bars.py
@@ -7,17 +7,18 @@ vs the clicked values in the legend,
 and assigns a lower order (0)
 if there is a match.
 The use of "indexOf" checks for equality in an array,
-which here allows for mutliple segments to be reordered
+which here allows for multiple segments to be reordered
 by holding down the shift key while clicking the legend.
 """
 # category: interactive charts
 import altair as alt
 from vega_datasets import data
 
-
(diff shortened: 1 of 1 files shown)


tests/examples_arguments_syntax/interactive_reorder_stacked_bars.py
```diff
@@ -0,0 +1,30 @@
+"""
+Reorder stacked bar segments
+============================
+This example uses a calculate transform
+to check the values of the "site" column
+vs the clicked values in the legend,
+and assigns a lower order (0)
+if there is a match.
+The use of "indexOf" checks for equality in an array,
+which here allows for mutliple segments to be reordered
```
binste, line 10:
```suggestion
which here allows for multiple segments to be reordered
```

tests/examples_arguments_syntax/interactive_reorder_stacked_bars.py
```diff
@@ -0,0 +1,30 @@
...
+by holding down the shift key while clicking the legend.
+"""
+# category: interactive charts
+import altair as alt
+from vega_datasets import data
+
+
```
binste, line 17:
```suggestion
```
NIT: Most examples just have 1 line after the import statements

tests/examples_arguments_syntax/interactive_reorder_stacked_bars.py
```diff
@@ -0,0 +1,30 @@
...
+selection = alt.selection_point(fields=['site'], bind='legend')
+
+alt.Chart(data.barley.url).mark_bar().transform_calculate(
```
binste, line 20:
```suggestion
selection = alt.selection_point(fields=['site'], bind='legend')

source = data.barley.url

alt.Chart(source).mark_bar().transform_calculate(
```
To follow the instructions in `CONTRIBUTING.md`:

> For consistency all data used for a visualization should be assigned to the
  variable `source`. Then `source` is passed to the `alt.Chart` object.
  If the example requires multiple dataframes then this does not apply. See
  other examples for guidance.

binste (changes requested):
So simple and convenient! I'll definitely start using this :)

I have just some minor comments.

joelostblom:
Thanks @binste ! I updated with your suggested changes.

tests/examples_arguments_syntax/interactive_reorder_stacked_bars.py
```diff
@@ -0,0 +1,31 @@
...
+# category: interactive charts
+import altair as alt
+from vega_datasets import data
+
+selection = alt.selection_point(fields=['site'], bind='legend')
+
+source = data.barley.url
+
+alt.Chart(source).mark_bar().transform_calculate(
+    site_order=f"if({selection.name}.site && indexof({selection.name}.site, datum.site) !== -1, 0, 1)"
```
mattijn, line 22: Im fine with current approach, but an observation is that this is JavaScript syntax. I'm not sure if the Python syntax for this type of Vega expressions is more readable. 

While often underestimated in efficiency, we don't really have a good approach documented to work with these Vega expressions.

joelostblom: Yes, @jonmmease made a similar comment in https://github.com/altair-viz/altair/pull/3362#discussion_r1526139226 and I would also prefer to change to the Python syntax in general in the docs. However, I think we should solve https://github.com/altair-viz/altair/issues/3366 first, so that the Py syntax covers everything that the JS syntax can do (this instance is the only difference I'm aware of).

mattijn: Ah, yes, I remember being bitten by this recently too!

binste (approved):

PR #3405, opened 2024-04-14
refactor: Use Parameter class directly in type hints instead of _Parameter protocol

binste:
Turns out I was wrong in https://github.com/vega/altair/pull/3208#issuecomment-1797868027 which is nice in this case :) We can use the `Parameter` class directly in type hints without running into circular imports thanks to `typing.TYPE_CHECKING`. See https://mypy.readthedocs.io/en/stable/runtime_troubles.html#import-cycles for more details but in summary, type checkers can be much more flexible how they import stuff as they don't run actual code so they can also deal with circular imports.

There is no impact for users except that instead of `_Parameter` as type hints:

image

they now see the less confusing `Parameter`:
image


dangotbanned:
Fixed in: https://github.com/dangotbanned/altair/blob/3ec3e6abfeb742942d1b2bc5d1a0bd4c4c5fa7d2/altair/vegalite/v5/schema/mixins.py#L448

mattijn:
Is this PR still applicable after merging #3430?

dangotbanned:
> Is this PR still applicable after merging #3430?

@mattijn I think all of this was covered in that PR, which I could have conveyed better in https://github.com/vega/altair/pull/3405#issuecomment-2169284531

binste:
Good point! I'll remove some comments which are now outdated in #3405

PR #3411, opened 2024-04-26
docs: Use raw strings with escape sequences

### LINKED ISSUE #3410: Escaping special chars in col names with a single backslash raises warnings
### What happened?

When I run [the example with special chracters from the docs](https://altair-viz.github.io/user_guide/encodings/index.html#escaping-special-characters-in-column-names), I see a bunch of warnings:

image

Replacing the single `\` with double `\\` fixes the issue. We could just update the docs, but before doing so I'm curious if anyone knows if something changed (in altair/ipython/python) that makes this warning visible now and if others see it or it is something with my local env.

main

joelostblom:
@jonmmease I agree with your comment about using raw strings instead of double backslash to make it more readable.

close #3410

### DIFF
--- a/doc/user_guide/encodings/index.rst
+++ b/doc/user_guide/encodings/index.rst
@@ -280,9 +280,9 @@ in some data structures.
 The recommended thing to do when you have special characters in a column name
 is to rename your columns.
 For example, in pandas you could replace ``:`` with ``_``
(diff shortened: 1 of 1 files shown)


binste:
LGTM, thanks Joel!

PR #3413, opened 2024-05-02
fix: Secure Source of Randomness

pixeeai:
This codemod replaces all instances of functions in the `random` module (e.g. `random.random()` with their, much more secure, equivalents from the `secrets` module (e.g. `secrets.SystemRandom().random()`).

There is significant algorithmic complexity in getting computers to generate genuinely unguessable random bits. The `random.random()` function uses a method of pseudo-random number generation that unfortunately emits fairly predictable numbers.

If the numbers it emits are predictable, then it's obviously not safe to use in cryptographic operations, file name creation, token construction, password generation, and anything else that's related to security. In fact, it may affect security even if it's not directly obvious.

Switching to a more secure version is simple and the changes look something like this:

```diff
- import random
+ import secrets
  ...
- random.random()
+ secrets.SystemRandom().random()
```

<details>
  <summary>More reading</summary>

  * [https://owasp.org/www-community/vulnerabilities/Insecure_Randomness](https://owasp.org/www-community/vulnerabilities/Insecure_Randomness)
  * [https://docs.python.org/3/library/random.html](https://docs.python.org/3/library/random.html)
</details>

🧚🤖Powered by Pixeebot (codemod ID: [pixee:python/secure-random](https://docs.pixee.ai/codemods/python/pixee_python_secure-random))

### DIFF
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -1,6 +1,5 @@
 import json
 import os
-import random
 import hashlib
 import warnings
 from typing import Union, MutableMapping, Optional, Dict, Sequence, TYPE_CHECKING, List
@@ -17,6 +16,7 @@ from .plugin_registry import PluginRegistry
 
 
 from typing import Protocol, TypedDict, Literal
+import secrets
 
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


binste:
We don't use `random` in cryptographic contexts so this is not relevant for Altair.

PR #3414, opened 2024-05-05
fix: solve mypy errors which are due to same object names in core.py and channels.py

### LINKED ISSUE #3408: alt.Color fails mypy --strict with error: Call to untyped function "Color" in typed context  [no-untyped-call]
### What happened?

I use altair in my project, which is typedchecked with mypy strict mode. When I have a graph in that project which uses alt.Color, I get `error: Call to untyped function "Color" in typed context  [no-untyped-call]` from mypy --strict. Oddly, this does not happen to other alt.Whatever objects, only color. I can't be sure if this is an altair bug or a mypy bug, to be honest.

Here's a minimal example:

```python
import altair as alt
import pandas as pd
dicted_rows = {"a": [1, 2, 3], "b": [4, 5, 6], "c": ["red", "blue", "yellow"]}
chart = alt.Chart(pd.DataFrame(dicted_rows)).mark_circle(size=90).encode(alt.X("a"), alt.Y("b"), alt.Color("c"))
chart.save('chart.html') #note: this works fine
```
Results from mypy version 1.9.0 (compiled: yes):

```
$ mypy "minimal altair example.py"
Success: no issues found in 1 source file
```

```
$ mypy --strict "minimal altair example.py"
minimal altair example.py:4: error: Call to untyped function "Color" in typed context  [no-untyped-call]
Found 1 error in 1 file (checked 1 source file)
```

The types should be annotated appropriately so Color is deduced to be a typed function and therefore mypy strict would not report an error.

I'm not sure exactly what this would involve. https://peps.python.org/pep-0484/ says the return type of __init__ ought to be `None`, so that's somewhere to start, maybe, perhaps.

5.3.0

binste:
Fixes #3408. See [this comment](https://github.com/vega/altair/issues/3408#issuecomment-2091253221) for a description of the issue). This PR adds `__all__` to `core.py` which includes all objects which were previously exported except for `Color` and `Text` which are names which are also defined in `channels.py`. With this, mypy can correctly detect the type of the top-level `altair.Color` and `altair.Text` as coming from `channels.py` and not `core.py`.

I've tested against the main branch that all objects are still exported except for some objects from `typing` which are not needed anyway. On the top-level of Altair, you can also see that nothing changed as the auto-generated `_all__` in `altair/__init__.py` did not change. Hence, this is purely a fix for mypy and does not have any functional implications for users.


jonmmease (approved):
Makes sense. If mypy and ruff are happy, I'm happy 😄 
Thanks @binste!

binste:
Haha same here ;) Thanks for the review

Issue #3418, opened 2024-05-11
Tests fail on main due to a fiona/geopandas error


binste:
### What happened?

Tests fail, see e.g. https://github.com/vega/altair/actions/runs/9005770219/job/24850952007#step:9:67

I'll look into it. Can't replicate this on my Mac so far.

mattijn:
It's because of this line: https://github.com/vega/altair/blob/main/tests/examples_methods_syntax/interval_selection_map_quakes.py#L14

```python
gdf_quakies = gpd.read_file(data.earthquakes.url, driver="GeoJSON")
```

Maybe a glitch or something changed in geopandas?

binste:
No new version of geopandas and fiona in the last few days and it only happens for Python 3.11:

<img width="343" alt="image" src="https://github.com/vega/altair/assets/23366411/da05d1b0-c253-4aec-a46b-c4dd6ee7f608">

I can't replicate this with Python 3.11 on Debian and I struggle to install Fiona on the same Ubuntu version as the GitHub runners use.

Maybe some change in a system dependency such as gdal? As it's not related to Altair and it's not blocking any major developments, we could wait a few days to see if this gets resolved somewhere upstream.

mattijn:
Raised https://github.com/geopandas/geopandas/issues/3284 at the geopandas repo.

mattijn:
Temporarily fixed by https://github.com/vega/altair/pull/3421

binste:
Thanks @mattijn! Also for engaging in the geopandas discussion.

binste:
Reopening, see https://github.com/geopandas/geopandas/issues/3284#issuecomment-2118674542

joelostblom:
Closing as per https://github.com/geopandas/geopandas/pull/3311#issuecomment-3106064942, but please reopen if that's not correct

PR #3419, opened 2024-05-12
ci: Remove references to archived altair_viewer and altair_saver in ci, docs, and tests. Uninstall anywidget and vl-convert-python during one test run

binste:
Now that both altair_viewer and altair_saver are archived and not compatible with Altair anyway, I'd suggest to remove all references to them to clean up the code base a bit. In this PR, I:

* removed all references to altair_viewer
* removed all references to altair_saver except for 1 mention in the docs
* uninstall anywidget and vl-convert-python as well in the build workflow to make sure we don't introduce hard dependencies. This is not directly related to the rest of the PR but hope it's ok for efficiency reasons. Else, happy to take it out!
* added a deprecation warning if someone uses the `webdriver` argument to the `.save` method as it's not relevant for `vl-convert`. I didn't want to remove it just now as users might still have it set although they have switched to `vl-convert` in which case it has just been ignored so far
* run test_save_html again with inline=True as vl-convert supports it

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,17 +16,10 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-      # - name: Set Up Chromedriver
-      #   run: |
-      #     sudo apt-get update
-      #     sudo apt-get --only-upgrade install google-chrome-stable
-      #     sudo apt-get -yqq install chromium-chromedriver
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
           pip install ".[all, dev]"
-          # pip install "selenium<4.3.0"
-          # pip install altair_saver
       - name: Install specific jsonschema
         # Only have to execute this if we don't want the latest jsonschema version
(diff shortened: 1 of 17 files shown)


binste:
Need to wait for #3418 to be resolved before tests can run.

jonmmease (approved):
Looks great. Thanks for doing this cleanup @binste!

mattijn:
#3418 is resolved, but tests are not yet happy. It now raises an anywidget dependency issue, see https://github.com/vega/altair/actions/runs/9090334764/job/24983082085. Probably related to this sentence in OP: 

> uninstall anywidget and vl-convert-python as well in the build workflow to make sure we don't introduce hard dependencies.

Can you have a look to this @binste. Thanks!

binste:
Thank you both! I added `pytest.skip` statements to some more tests which relied on vl-convert or anywidget. Now, the tests can be run with only Altair's required dependencies. Any tests which require optional dependencies are skipped. Merging as it's only CI changes and Jon approved an earlier version but happy to do work in a follow-up PR if anyone has comments.

PR #3420, opened 2024-05-13
feat: Adds `ChartType` type and type guard `is_chart_type`. Change PurePath to Path type hints

dangotbanned:
This PR mostly relates to type hints and associated tools, from the perspective of a downstream library author.

## Note
Below was automatically added, but is only from the first commit. 
The others also have descriptions, not sure if you'd like them all collected into the PR as well.

However, it was the motivation behind the contributions generally so I've left it in.

<details>
<summary>First commit description</summary>

I'm working on a library that uses `altair`, and looking to make it an optional dependency. During this process, I thought `ChartType` and `is_chart_type` might be a good fit in `altair` itself.

I don't believe any existing `altair` code would directly benefit from these additions - but you could likely reduce some `isinstance` checks with [type narrowing](https://typing.readthedocs.io/en/latest/spec/narrowing.html) functions like `is_chart_type`.
For example `altair/vegalite/v5/api.py` has **63** occurrences of `isinstance`.

---

per [CONTRIBUTING.md](https://github.com/vega/altair/blob/main/CONTRIBUTING.md)
> In particular, we welcome companion efforts from other visualization libraries to render the Vega-Lite specifications output by Altair. We see this portion of the effort as much bigger than Altair itself

Towards this aim, it could be beneficial to provide `typing` constructs/tools for downstream use.

Regarding my specfic use case, the below [Protocol](https://typing.readthedocs.io/en/latest/spec/protocol.html#runtime-checkable-decorator-and-narrowing-types-by-isinstance) - I believe - describes the concept of a `_/Chart`, for the purposes of writing to file:

```py
from typing import Any, Protocol, runtime_checkable

@runtime_checkable
class AltairChart(Protocol):
    data: Any

    def copy(self, *args: Any, **kwargs: Any) -> Any: ...
    def save(self, *args: Any, **kwargs: Any) -> None: ...
    def to_dict(self, *args: Any, **kwargs: Any) -> dict[Any, Any]: ...
    def to_html(self, *args: Any, **kwargs: Any) -> str: ...
```

Due to the number of symbols in `altair` and heavy use of mixins, coming to this conclusion can be a stumbling block.

However, exporting things like `ChartType`, `is_chart_type`, `AltairChart` could be a way to expose these shared behaviours - without requiring any changes to the core API.

</details>

---

I'd likely have written this differently if `altair` was already using [PEP 563](https://peps.python.org/pep-0563/), but imagine introducing that could impact the autogenerated code. 
Did want to note it though, as it might help with some circular references.


dangotbanned:
I'm unsure if I've followed the contributing guide correctly. 

I expected running this locally, without errors, would be safe to push:
```
hatch run test
```

I'll hold of on any more commits for now, please let me know if I've missed something

mattijn:
Thanks for the PR! Without giving a detailed review I can say that the CI errors originate from the auto-generated code. 
You will likely need to run the following to have these generated:
```python
hatch run python tools/generate_schema_wrapper.py
```
See also: https://github.com/vega/altair/blob/main/NOTES_FOR_MAINTAINERS.md#auto-generating-the-python-code.

dangotbanned:
> Thanks for the PR! Without giving a detailed review I can say that the CI errors originate from the auto-generated code. You will likely need to run the following to have these generated:
> 
> ```python
> hatch run python tools/generate_schema_wrapper.py
> ```
> 
> See also: [`main`/NOTES_FOR_MAINTAINERS.md#auto-generating-the-python-code](https://github.com/vega/altair/blob/main/NOTES_FOR_MAINTAINERS.md?rgh-link-date=2024-05-13T14%3A18%3A11Z#auto-generating-the-python-code).

Thanks, will give this a try.

dangotbanned:
Looks like the same as error as #3418

dangotbanned:
@mattijn still getting the same error, not `anywidget` as in https://github.com/vega/altair/pull/3419#issuecomment-2111607849

Is there anything I can do to help here?

altair/utils/save.py
```diff
@@ -160,12 +158,9 @@ def perform_save():
...
             )
             write_file_or_filename(
-                fp,
-                mimebundle["text/html"],
-                mode="w",
-                encoding=kwargs.get("encoding", "utf-8"),
+                fp, mimebundle["text/html"], mode="w", encoding=encoding
             )
-        elif format in ["png", "svg", "pdf", "vega"]:
+        elif format in {"png", "svg", "pdf", "vega"}:
```
dangotbanned, line 163: I've only made the change here, but this lint is from [ruff PLR6201](https://docs.astral.sh/ruff/rules/literal-membership/), if you wanted to add to `pyproject.toml` for the quick fix everywhere.

altair/utils/save.py
```diff
@@ -9,28 +9,26 @@
...
-    fp: Union[str, pathlib.PurePath, IO],
+    fp: Union[pathlib.Path, IO],
     content: Union[str, bytes],
     mode: str = "w",
     encoding: Optional[str] = None,
 ) -> None:
     """Write content to fp, whether fp is a string, a pathlib Path or a
     file-like object"""
-    if isinstance(fp, str) or isinstance(fp, pathlib.PurePath):
+    if isinstance(fp, pathlib.Path):
```
dangotbanned, line 19: If you keep the original, I'd recommend adding [ruff SIM101](https://docs.astral.sh/ruff/rules/duplicate-isinstance-call/) to quick fix multiple isinstance calls

binste:
Thanks @dangotbanned! I'll try to have a detailed look at this soon.

altair/utils/save.py
```diff
@@ -9,28 +9,26 @@
 
 
 def write_file_or_filename(
-    fp: Union[str, pathlib.PurePath, IO],
+    fp: Union[pathlib.Path, IO],
```
binste, line 12: Let's keep `str` around as a type hint as the function actually works for strings as well. I don't think we need to remove this functionality. Also, in the Altair codebase there is unfortunately no clear separation between public and private functions/classes, i.e. most things need to be considered public. Removing `str` here could be a breaking change for someone who uses this function directly (although I'm not sure why anyone would).

altair/utils/save.py
binste: Based on the comment above, let's keep the `str` isinstance check but I like the idea of adding SIM101! Could you add it to pyproject.toml and autofix everything it finds in a new commit?

altair/utils/save.py
```diff
@@ -9,28 +9,26 @@
...
+    if isinstance(fp, pathlib.Path):
         with open(file=fp, mode=mode, encoding=encoding) as f:
             f.write(content)
     else:
         fp.write(content)
 
 
 def set_inspect_format_argument(
-    format: Optional[str], fp: Union[str, pathlib.PurePath, IO], inline: bool
+    format: Optional[str], fp: Union[pathlib.Path, IO], inline: bool
```
binste, line 27: Same reasoning for this function as above -> Let's keep `str` around and switch back to `PurePath`.

altair/utils/save.py
binste, line 12: What's the reason to use `Path` instead of `PurePath`? In my understanding, `PurePath` is more generic as `Path` is a subclass of it.

altair/utils/save.py
```diff
@@ -70,7 +68,7 @@ def set_inspect_mode_argument(
 
 def save(
     chart,
-    fp: Union[str, pathlib.PurePath, IO],
+    fp: Union[str, pathlib.Path, IO],
```
binste, line 71: Same question here regarding `PurePath` vs. `Path`

altair/utils/save.py
binste: Let's keep the original list here. It's quite a common codestyle throughout the codebase and we almost never use sets. We could of course change but I'd prefer consistency for now.

altair/utils/save.py
```diff
@@ -130,6 +128,8 @@ def save(
     """
     if json_kwds is None:
         json_kwds = {}
+    fp = pathlib.Path(fp) if isinstance(fp, str) else fp
```
binste, line 131: Having `encoding` up here makes sense to me, thanks! As I'd suggest to add back the functionality for `str` in the other functions, the other line here with the conversion to a Path is no longer necessary.

binste (changes requested):
Thanks again for the PR! Agree that this is a helpful utility to have in Altair.

I've made `ChartType` and `is_chart_type` public and bumped typing_extensions so it includes `TypeIs`. I've added some other smaller comments after which I think this is good to go.

altair/utils/save.py
dangotbanned: I spent some time thinking about this, because it was my first time seeing `PurePath` usage outside of the python docs. So I did a re-read to try and understand why `PurePath` may have originally been used here.

Near the top of the [pathlib docs](https://docs.python.org/3/library/pathlib.html) it states:
> Pure paths are useful in some special cases; for example:
> 1. If you want to manipulate Windows paths on a Unix machine (or vice versa). You cannot instantiate a [WindowsPath](https://docs.python.org/3/library/pathlib.html#pathlib.WindowsPath) when running on Unix, but you can instantiate [PureWindowsPath](https://docs.python.org/3/library/pathlib.html#pathlib.PureWindowsPath).
> 2. You want to make sure that your code only manipulates paths without actually accessing the OS. In this case, instantiating one of the pure classes may be useful since those simply don’t have any OS-accessing operations.

I don't believe either of those cases are relevant here.

From my understanding the core difference between `PurePath` and `Path` is that the latter introduces methods for I/O. Either type can be passed to `open`, as they are [PathLike](https://docs.python.org/3/library/os.html#os.PathLike) - but I believe using `Path` more clearly communicates that it will be used for I/O. 

> In my understanding, PurePath is more generic as Path is a subclass of it. @binste 

If the choice is based on wanting more generic option, `Path` already handles this internally based on the host OS: 
pathlib-inheritance

I will note, there are changes [scheduled for python 3.13](https://discuss.python.org/t/make-pathlib-extensible/3428/168) that will implement a more traditional ABC hierarchy, that is documented [here](https://pathlib-abc.readthedocs.io/en/latest/).

dangotbanned: @binste as the remaining comments all tie into `Path`/`PurePath`, I'll wait to hear back from you on this and then make the remaining changes in one commit. 

Really appreciate you taking the time to review

altair/utils/save.py
binste: Thanks for that! Looks good. Personally, I like autofixable ruff rules which harmonize the code base so I'd be open to discuss this further. Thanks for bringing it up!

altair/utils/save.py
binste: And I appreciate that you take the time to thoroughly research it and provide such an explanation! :)

I see your point about `Path` better communicating that it will be used for I/O. I've also never seen PurePath before, when adding type hints to this method I think i took it over from some other part of the library but I can't see a good reason for it anymore and maybe it's easier to just stick to what people know.

-> I'm ok with using `Path` instead of `PurePath` as you have already implemented. Would still be great to keep `str` around as an input argument for the functions where it was possible to use before.

binste (approved):
Thanks again! Looks good to me.

PR #3426, opened 2024-05-26
refactor(DRAFT): Remove `toolz` dependency

dangotbanned:
This PR is a proof-of-concept, entirely* removing the hard dependency on `toolz`.
*Excluding usage in tests and twice for a deprecation warning.

I've tried to provide reasoning on each commit, but generally, it seemed to me that the behaviour `altair` needs is easily replicated within `stdlib`. 

### Additional Notes

- `pipe` and `curry` were the only imports
  - `pipe` is trivial to replace.
  - `curry` seems overkill when the only non-default argument is always `data`
    - Also has the drawback of [less type safety](https://github.com/pytoolz/toolz/issues/496)
- While this is documented under [Advanced Usage](https://altair-viz.github.io/user_guide/data_transformers.html#piping), I haven't come across any more complex examples.
  - I'd imagine users could see a benefit to performance performance, given the complexity of [curry](https://github.com/pytoolz/toolz/blob/7a0382e326eb2549e4dc7b833dc666a877feca38/toolz/functoolz.py#L168C5-L168C6).
  - If anyone has some sample code that heavily uses data transformers, it could be helpful to benchmark any changes.

If there was any desire to pursue this further, in b9dc070c30a23f09adac2d3686114ac3fc9c5cc9 I mentioned that this could be implemented as a decorator. 
I stopped short of that for now, but would be happy to go down that route to simplify the `@overload` usage.

### DIFF
--- a/altair/utils/_vegafusion_data.py
+++ b/altair/utils/_vegafusion_data.py
@@ -1,8 +1,10 @@
-from toolz import curried
 import uuid
 from weakref import WeakValueDictionary
 
 from typing import (
+    Any,
+    Literal,
+    Optional,
     Union,
     Dict,
     Set,
@@ -10,11 +12,18 @@ from typing import (
     TypedDict,
     Final,
     TYPE_CHECKING,
+    overload,
 )
+from collections.abc import Callable
 
 from altair.utils._importers import import_vegafusion
 from altair.utils.core import DataFrameLike
-from altair.utils.data import DataType, ToValuesReturnType, MaxRowsError
+from altair.utils.data import (
+    DataType,
+    ToValuesReturnType,
+    MaxRowsError,
+    NonLikeDataType,
+)
 from altair.vegalite.data import default_data_transformer
 
 if TYPE_CHECKING:
@@ -36,21 +45,41 @@ class _ToVegaFusionReturnUrlDict(TypedDict):
     url: str
 
 
-@curried.curry
+_VegaFusionReturnType = Union[_ToVegaFusionReturnUrlDict, ToValuesReturnType]
+
+
+@overload
 def vegafusion_data_transformer(
-    data: DataType, max_rows: int = 100000
-) -> Union[_ToVegaFusionReturnUrlDict, ToValuesReturnType]:
+    data: Literal[None] = ..., max_rows: int = ...
+) -> Callable[..., Any]: ...
+
+
+@overload
+def vegafusion_data_transformer(
+    data: DataFrameLike, max_rows: int
+) -> ToValuesReturnType: ...
+
+
+@overload
+def vegafusion_data_transformer(
+    data: NonLikeDataType, max_rows: int
+) -> _VegaFusionReturnType: ...
+
+
+def vegafusion_data_transformer(
+    data: Optional[DataType] = None, max_rows: int = 100000
+) -> Union[Callable[..., Any], _VegaFusionReturnType]:
     """VegaFusion Data Transformer"""
-    if hasattr(data, "__geo_interface__"):
-        # Use default transformer for geo interface objects
-        # # (e.g. a geopandas GeoDataFrame)
-        return default_data_transformer(data)
+    if data is None:
+        return vegafusion_data_transformer
     elif isinstance(data, DataFrameLike):
         table_name = f"table_{uuid.uuid4()}".replace("-", "_")
         extracted_inline_tables[table_name] = data
         return {"url": VEGAFUSION_PREFIX + table_name}
     else:
-        # Use default transformer if we don't recognize data type
+        # Use default transformer for geo interface objects
+        # # (e.g. a geopandas GeoDataFrame)
+        # Or if we don't recognize data type
         return default_data_transformer(data)
 
 
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -3,7 +3,22 @@ import os
 import random
 import hashlib
 import warnings
-from typing import Union, MutableMapping, Optional, Dict, Sequence, TYPE_CHECKING, List
+from typing import (
+    Union,
+    MutableMapping,
+    Optional,
+    Dict,
+    Sequence,
+    TYPE_CHECKING,
+    List,
+    TypeVar,
+    Protocol,
+    TypedDict,
+    Literal,
+    overload,
+    runtime_checkable,
+    Any,
+)
 
 import pandas as pd
 from toolz import curried
@@ -15,14 +30,16 @@ from .core import sanitize_geo_interface
 from .deprecation import AltairDeprecationWarning
 from .plugin_registry import PluginRegistry
 
-
-from typing import Protocol, TypedDict, Literal
-
+if sys.version_info >= (3, 13):
+    from typing import TypeIs
+else:
+    from typing_extensions import TypeIs
 
 if TYPE_CHECKING:
     import pyarrow.lib
 
 
+@runtime_checkable
 class SupportsGeoInterface(Protocol):
     __geo_interface__: MutableMapping
 
@@ -32,6 +49,8 @@ TDataType = TypeVar("TDataType", bound=DataType)
 
 VegaLiteDataDict = Dict[str, Union[str, dict, List[dict]]]
 ToValuesReturnType = Dict[str, Union[dict, List[dict]]]
+def is_data_type(obj: Any) -> TypeIs[DataType]:
+    return isinstance(obj, (dict, pd.DataFrame, DataFrameLike, SupportsGeoInterface))
 
 
 # ==============================================================================
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -72,8 +72,15 @@ def is_data_type(obj: Any) -> TypeIs[DataType]:
 # form.
 # ==============================================================================
 class DataTransformerType(Protocol):
-    def __call__(self, data: DataType, **kwargs) -> VegaLiteDataDict:
-        pass
+    @overload
+    def __call__(
+        self, data: Literal[None] = None, **kwargs
+    ) -> "DataTransformerType": ...
+    @overload
+    def __call__(self, data: DataType, **kwargs) -> VegaLiteDataDict: ...
+    def __call__(
+        self, data: Optional[DataType] = None, **kwargs
+    ) -> Union["DataTransformerType", VegaLiteDataDict]: ...
 
 
 class DataTransformerRegistry(PluginRegistry[DataTransformerType]):
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -23,8 +23,6 @@ from typing import (
 )
 
 import pandas as pd
-from toolz import curried
-from typing import TypeVar
 
 from ._importers import import_pyarrow_interchange
 from .core import sanitize_dataframe, sanitize_arrow_table, DataFrameLike
@@ -414,7 +412,9 @@ def pipe(data, *funcs):
         AltairDeprecationWarning,
         stacklevel=1,
     )
-    return curried.pipe(data, *funcs)
+    from toolz.curried import pipe
+
+    return pipe(data, *funcs)
 
 
 def curry(*args, **kwargs):
@@ -428,7 +428,9 @@ def curry(*args, **kwargs):
         AltairDeprecationWarning,
         stacklevel=1,
     )
-    return curried.curry(*args, **kwargs)
+    from toolz.curried import curry
+
+    return curry(*args, **kwargs)
 
 
 def arrow_table_from_dfi_dataframe(dfi_df: DataFrameLike) -> "pyarrow.lib.Table":
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -233,7 +233,7 @@ def to_csv(
 def to_values(data: DataType) -> ToValuesReturnType:
     """Replace a DataFrame by a data model with values."""
     check_data_type(data)
-    if hasattr(data, "__geo_interface__"):
+    if isinstance(data, SupportsGeoInterface):
         if isinstance(data, pd.DataFrame):
             data = sanitize_dataframe(data)
         # Maybe the type could be further clarified here that it is
@@ -276,7 +276,7 @@ def _compute_data_hash(data_str: str) -> str:
 def _data_to_json_string(data: DataType) -> str:
     """Return a JSON string representation of the input data"""
     check_data_type(data)
-    if hasattr(data, "__geo_interface__"):
+    if isinstance(data, SupportsGeoInterface):
         if isinstance(data, pd.DataFrame):
             data = sanitize_dataframe(data)
         # Maybe the type could be further clarified here that it is
@@ -302,7 +302,7 @@ def _data_to_json_string(data: DataType) -> str:
 def _data_to_csv_string(data: Union[dict, pd.DataFrame, DataFrameLike]) -> str:
     """return a CSV string representation of the input data"""
     check_data_type(data)
-    if hasattr(data, "__geo_interface__"):
+    if isinstance(data, SupportsGeoInterface):
         raise NotImplementedError(
             "to_csv does not work with data that "
             "contains the __geo_interface__ attribute"
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -51,6 +51,11 @@ TDataType = TypeVar("TDataType", bound=DataType)
 
 VegaLiteDataDict = Dict[str, Union[str, dict, List[dict]]]
 ToValuesReturnType = Dict[str, Union[dict, List[dict]]]
+SampleReturnType = Optional[
+    Union[pd.DataFrame, Dict[str, Sequence], "pyarrow.lib.Table"]
+]
+
+
 def is_data_type(obj: Any) -> TypeIs[DataType]:
     return isinstance(obj, (dict, pd.DataFrame, DataFrameLike, SupportsGeoInterface))
 
@@ -145,11 +150,22 @@ def limit_rows(
     return data
 
 
-@curried.curry
+@overload
 def sample(
-    data: DataType, n: Optional[int] = None, frac: Optional[float] = None
-) -> Optional[Union[pd.DataFrame, Dict[str, Sequence], "pyarrow.lib.Table"]]:
+    data: Literal[None] = ..., n: Optional[int] = ..., frac: Optional[float] = ...
+) -> partial: ...
+@overload
+def sample(
+    data: DataType, n: Optional[int], frac: Optional[float]
+) -> SampleReturnType: ...
+def sample(
+    data: Optional[DataType] = None,
+    n: Optional[int] = None,
+    frac: Optional[float] = None,
+) -> Union[partial, SampleReturnType]:
     """Reduce the size of the data model by sampling without replacement."""
+    if data is None:
+        return partial(sample, n=n, frac=frac)
     check_data_type(data)
     if isinstance(data, pd.DataFrame):
         return data.sample(n=n, frac=frac)
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -1,10 +1,10 @@
 from functools import partial
 import json
-import os
 import random
 import hashlib
 import warnings
 import sys
+from pathlib import Path
 from typing import (
     Union,
     MutableMapping,
@@ -48,6 +48,7 @@ class SupportsGeoInterface(Protocol):
 
 DataType = Union[dict, pd.DataFrame, SupportsGeoInterface, DataFrameLike]
 TDataType = TypeVar("TDataType", bound=DataType)
+NonGeoDataType = Union[dict, pd.DataFrame, DataFrameLike]
 
 VegaLiteDataDict = Dict[str, Union[str, dict, List[dict]]]
 ToValuesReturnType = Dict[str, Union[dict, List[dict]]]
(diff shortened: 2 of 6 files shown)


mattijn:
Hi @dangotbanned, this looks good! Thanks for working on this. Reducing the hard dependencies where possible is much welcome. 
If types and tests are happy than we are very far in the progress already. 
Could you explain why this PR requires the introduction of a new `._to_text()` function, see here: https://github.com/vega/altair/pull/3426/files#diff-478a9046db15a358587cf77454cd2cf9cc4723435276cad0864c268d17e07650R295. It seems to be introduced during the development process?

mattijn:
Thanks for the explanation! That makes sense, leave it in please. 

I'm in favor of removing the functions that have been officially deprecated for such a long time.

dangotbanned:
@mattijn No rush on the review, just wanted to update that all traces of `toolz` are gone 😅

mattijn:
Hi @dangotbanned. Since you have multiple PRs open, do you have a certain order in mind how we should review and merge these?

dangotbanned:
> Hi @dangotbanned. Since you have multiple PRs open, do you have a certain order in mind how we should review and merge these?

# #3426 

Hi @mattijn this one is ready to go. 

> I'm in favor of removing the functions that have been officially deprecated for such a long time.

I'm not sure if you wanted me to remove `pipe`, `curry`, the new optional import and tests - or if the removal needs any more input from other maintainers. 

Personally, I agree and think they should be removed.

# Other PRs
## #3431 
- I'm pretty happy with where it's at now, but the changes I've made in the [tools](https://github.com/dangotbanned/altair/tree/05ae13f50b5ebe8cb2652ed55bd475ebdcf72760/tools) package relating to `Optional`, `Parameter`, `SchemaBase` likely need further review. 
- They were somewhat "hacky" so as to not alter the surrounding code, and someone more familiar with the code generation may be able to improve upon it.
- This one (I think) has addressed all the points from @binste original [review](https://github.com/vega/altair/pull/3431#pullrequestreview-2102376479)

## #3427 
- I've left this on draft, not ready yet, but would be good to hear your thoughts on a couple of things
1. `alt.when` [signature](https://github.com/dangotbanned/altair/blob/8c0f45ae22bc829a500a495b2a978a2faedcb749/altair/vegalite/v5/api.py#L882)
	- To preserve compatibility with `alt.condition`, the first `predicate` arg behaves identically - but `*more_predicates` is more restrictive in the types allowed - as they all must implement `&`.
	- I've left this to bake before writing docs, as I'm not sure whether the trade-off in added complexity (for the user) is worth preserving identical semantics as `alt.condition` for a single argument.
2. The new parts added for `alt.expr` [comment](https://github.com/vega/altair/pull/3427#discussion_r1626275778) which are "an" implementation, but not sure it would be the right way to go.

mattijn:
I would be I favor of removing pipe and curry as well and instead of a deprecation warning use a deprecation error. 

Good point to have the opinion and/or concerns of another maintainer as well! Let's wait for that👍.

altair/utils/data.py
```diff
@@ -69,12 +100,19 @@ class MaxRowsError(Exception):
     pass
 
 
-@curried.curry
-def limit_rows(data: TDataType, max_rows: Optional[int] = 5000) -> TDataType:
+@overload
+def limit_rows(data: Literal[None] = ..., max_rows: Optional[int] = ...) -> partial: ...
```
joelostblom, line 104: What's the purpose of setting the default parameter values to elipsis here (`Literal[None] = ...`)?

altair/utils/data.py
```diff
@@ -157,65 +204,115 @@ def sample(
...
+
+
+def to_json(
+    data: Optional[DataType] = None,
     prefix: str = "altair-data",
     extension: str = "json",
     filename: str = "{prefix}-{hash}.{extension}",
     urlpath: str = "",
-) -> _ToJsonReturnUrlDict:
+) -> Union[partial, _ToFormatReturnUrlDict]:
```
joelostblom, line 245: Could you elaborate on what the `@overload` decorator does here. I thought its main purpose was to help type checking understand functions that support different combinations of arguments/types, but in this case it seems like there is only one combination of arguments?

altair/utils/data.py
```diff
@@ -1,37 +1,61 @@
...
 
+@runtime_checkable
 class SupportsGeoInterface(Protocol):
     __geo_interface__: MutableMapping
 
 
 DataType = Union[dict, pd.DataFrame, SupportsGeoInterface, DataFrameLike]
 TDataType = TypeVar("TDataType", bound=DataType)
+NonGeoDataType = Union[dict, pd.DataFrame, DataFrameLike]
+NonLikeDataType = Union[dict, pd.DataFrame, SupportsGeoInterface]
```
joelostblom, line 48: Could we use a more descriptive name than `NonLikeDataType`? Maybe `GeoDataType` if this is only used for geo data (I see that is accepts a regular dataframe too, but in the vegafusion method where it is used, it seems like it only replaces the `__geo_interface__` case)?

joelostblom (commented):
Thanks a lot for all your work here and on the other PRs @dangotbanned ! I am also  in favor of removing pipe and curry if we can replicate the functionality with stlib. I haven't tested the code and I'm not very familiar with type checking, but from skimming the diff it looks good overall. I left comments on a few sections.

altair/utils/data.py
dangotbanned: There's two parts to this.

# Replicating the existing behaviour provided by `@curried.curry`
This isn't immediately obvious when seeing the original signature as: 
```py
@curried.curry
def limit_rows(data: TDataType, max_rows: Optional[int] = 5000) -> TDataType:...
```
but the [curry](https://toolz.readthedocs.io/en/latest/curry.html#id1) [decorator](https://toolz.readthedocs.io/en/latest/_modules/toolz/functoolz.html#curry) roughly* alters it to:
```py
def limit_rows(data: Optional[TDataType] = None, max_rows: Optional[int] = 5000) -> TDataType:...
```
*with the caveats that:
- `TDataType` is only returned once all non-optional arguments have been provided
- Until that condition is met, other arguments can be provided again and alter the function

With this in mind, the correctly typed definition has this default `None` already there - as this was implicit before and hidden behind the decorator.

# Representing default arguments within `@typing.overload`
`...` in an `@overload` will match without the user providing an argument. Otherwise, they would need to write the below for the a match:
```py
limit_rows(max_rows=20) # fails
limit_rows(None, 5000) # matches
```
With the current `@overload` defintion, both are valid and match the same `@overload` def.

## Side note
While I was searching through the [typing spec](https://typing.readthedocs.io/en/latest/source/stubs.html#functions-and-methods) for use of `...` I came across [Literal[None] conversion](https://typing.readthedocs.io/en/latest/spec/literal.html#legal-parameters-for-literal-at-type-check-time). 
Which was news to me, so I'm going to simplify that part - and some places in #3431.

joelostblom: Thanks for elaborating, I think I understand now... something along the lines of that `...` is a "typeless" default argument that is useful where `None` would not satisfy the type constraint, such as for `overload` decorators.

The simplification for `Literal[None]` is welcome, good find.

altair/utils/data.py
dangotbanned: @joelostblom does the first part of https://github.com/vega/altair/pull/3426#discussion_r1638279005 help with this at all? 

The reason for `@overload` and the pattern used are the same. 

In short, when you call these functions:
- *without* `data` they return a function, [functools.partial](https://docs.python.org/3/library/functools.html#functools.partial)
- *with* `data` they return the type specified by the matching `@overload`

altair/utils/data.py
dangotbanned: > Could we use a more descriptive name than `NonLikeDataType`? Maybe `GeoDataType` if this is only used for geo data (I see that is accepts a regular dataframe too, but in the vegafusion method where it is used, it seems like it only replaces the `__geo_interface__` case)?

Absolutely agree it needs a better name!

I was confused seeing `NonLikeDataType` there, but in [this context](https://github.com/dangotbanned/altair/blob/60b8231c4ff785c1fd61918097caff58f8be2cc2/altair/utils/_vegafusion_data.py#L63) it made slightly more sense as its distiguishing two `@overload` defs, that have different return types:

```py
class _ToVegaFusionReturnUrlDict(TypedDict):
    url: str


_VegaFusionReturnType = Union[_ToVegaFusionReturnUrlDict, ToValuesReturnType]


@overload
def vegafusion_data_transformer(
    data: Literal[None] = ..., max_rows: int = ...
) -> Callable[..., Any]: ...


@overload
def vegafusion_data_transformer(
    data: DataFrameLike, max_rows: int # <-------------------------
) -> ToValuesReturnType: ...


@overload
def vegafusion_data_transformer(
    data: NonLikeDataType, max_rows: int # <-------------------------
) -> _VegaFusionReturnType: ...


def vegafusion_data_transformer(
    data: Optional[DataType] = None, max_rows: int = 100000
) -> Union[Callable[..., Any], _VegaFusionReturnType]:
    """VegaFusion Data Transformer"""
    if data is None:
        return vegafusion_data_transformer
    elif isinstance(data, DataFrameLike):
        table_name = f"table_{uuid.uuid4()}".replace("-", "_")
        extracted_inline_tables[table_name] = data
        return {"url": VEGAFUSION_PREFIX + table_name}
    else:
        # Use default transformer for geo interface objects
        # # (e.g. a geopandas GeoDataFrame)
        # Or if we don't recognize data type
        return default_data_transformer(data)
```

dangotbanned:
> Thanks a lot for all your work here and on the other PRs @dangotbanned ! I am also in favor of removing pipe and curry if we can replicate the functionality with stlib. I haven't tested the code and I'm not very familiar with type checking, but from skimming the diff it looks good overall. I left comments on a few sections.

@joelostblom thanks I really appreciate it. Has been fun to poke around `altair` after having enjoyed using it for a while now.

altair/utils/data.py
dangotbanned: @joelostblom Yeah you're pretty much there. 

If you're familiar with [pandas.DataFrame.rename](https://github.com/pandas-dev/pandas/blob/de5d7323cf6fcdd6fcb1643a11c248440787d960/pandas/core/frame.py#L5440) you could try that out in your IDE to see how different `overload`'s get matched. 

I picked that one as it has arguments without `None` - as that part of it isn't directly linked to `...` - it is more about matching the already defined default. 

So for `errors` the `...` represents *'ignore'* by default, or either *'ignore'*, *'raise'* if provided explicitly:
```py
class DataFrame:
    @overload
	def rename(
        self,
        errors: Literal['ignore', 'raise'] = ...,
    ) -> DataFrame | None:
        ...

    def rename(
        self,
        errors: Literal['ignore', 'raise'] = "ignore",
    ) -> DataFrame | None:
```

altair/utils/data.py
dangotbanned: Fixed in: b0d0e5c6607339db655ef1c5387910adce6a2c46

altair/utils/data.py
dangotbanned: > The simplification for `Literal[None]` is welcome, good find.

Added in: a8e537f1547a9547e1dfc75e6848574829fcdd8a

altair/utils/data.py
joelostblom: Great, thank you! I agree that it is clearer to write out the three possibilities instead of grouping them under `NonLikeDataType` as in your update..

altair/utils/data.py
joelostblom: Thanks for elaborating, I will look into it more to fully understand it more deeply, but it's not blocking in any way.

altair/utils/data.py
```diff
@@ -317,27 +409,15 @@ def pipe(data, *funcs):
 
     Deprecated: use toolz.curried.pipe() instead.
     """
-    warnings.warn(
```
dangotbanned, line 320: Marked for removal: `altair.utils.data.pipe`

altair/utils/data.py
```diff
@@ -317,27 +409,15 @@ def pipe(data, *funcs):
...
-        "alt.pipe() is deprecated, and will be removed in a future release. "
-        "Use toolz.curried.pipe() instead.",
-        AltairDeprecationWarning,
-        stacklevel=1,
-    )
-    return curried.pipe(data, *funcs)
+    return import_toolz_function("pipe")(data, *funcs)
 
 
 def curry(*args, **kwargs):
```
dangotbanned, line 415: Marked for removal: `altair.utils.data.curry`

altair/utils/_importers.py
```diff
@@ -95,3 +101,24 @@ def pyarrow_available() -> bool:
         return True
     except (ImportError, RuntimeError):
         return False
+
+
+def import_toolz_function(
```
dangotbanned, line 106: Marked for removal: `altair.utils._importers.import_toolz_function`.

Alternatively, could be made generic, if this functionality is useful for other modules/packages?

tests/utils/test_data.py
```diff
@@ -135,12 +146,33 @@ def test_dict_to_csv():
...
         filename = result1["url"]
         output = pd.read_csv(filename).to_dict(orient="records")
     finally:
         os.remove(filename)
 
     assert result1 == result2
     assert data == {"values": output}
+
+
+def test_toolz():
```
dangotbanned, line 160: Marked for removal: `tests.utils.test_data.test_toolz`

Dependent on resolution to comment in https://github.com/vega/altair/pull/3426/files#r1639506036

dangotbanned:
@mattijn 
> I would be I favor of removing pipe and curry as well and instead of a deprecation warning use a deprecation error.

@joelostblom 
> I am also in favor of removing pipe and curry if we can replicate the functionality with stlib.

Thank you both for reviewing. All of your comments should be resolved now.

I have added comments suggesting the parts to remove. This excludes the imports, which would be highlighted by any IDE anyway afterwards.

If either of you would like to do the removal, feel free, otherwise let me know and I'll finish this PR off with that.

altair/utils/_importers.py
joelostblom: I don't know if this is useful for any other module, @mattijn do you know?

mattijn: Nope. No idea.

dangotbanned: Realising now that comment hasn't highlighted the contents of the function. 

Also the deprecation part might be distracting from the part that could be useful.
```py
...
try:
    module = import_module("toolz.curried")
except ImportError as err:
    msg = f"{msg} {err.args[0]}"
    raise ImportError(msg) from err
...
``` 

The other functions in `_importers`:
- `import_pyarrow_interchange`
- `import_vl_convert`
- `import_vegafusion`

They all do fairly similar things, although based on version constraints and use regular imports, like `import vl_convert as vlc` instead of `import_module("vl_convert")`:

<details>
<summary>import_vl_convert</summary>

```py
def import_vl_convert() -> ModuleType:
    min_version = "1.3.0"
    try:
        version = importlib_version("vl-convert-python")
        if Version(version) < Version(min_version):
            raise RuntimeError(
                f"The vl-convert-python package must be version {min_version} or greater. "
                f"Found version {version}"
            )
        import vl_convert as vlc

        return vlc
    except ImportError as err:
        raise ImportError(
            f"The vl-convert Vega-Lite compiler and file export feature requires\n"
            f"version {min_version} or greater of the 'vl-convert-python' package. \n"
            f"This can be installed with pip using:\n"
            f'   pip install "vl-convert-python>={min_version}"\n'
            "or conda:\n"
            f'   conda install -c conda-forge "vl-convert-python>={min_version}"\n\n'
            f"ImportError: {err.args[0]}"
        ) from err
```

</details>


So rather than having a **dedicated function per import** to maintain and keep consistent - you could do something like below. 
I've taken this from my own code I was working on during #3420, as it shows a larger number of optional dependencies: 

```py
from types import ModuleType
from typing import NamedTuple

class ModuleAvailable(NamedTuple):
    module: ModuleType | None = None
    available: bool = True


_init_modules = (
    "msgspec",
    "tomli_w",
    "aiofiles",
    "xlsxwriter",
    "plotly",
    "altair",
    "lxml",
    "ftfy",
    "playwright",
    "keyring",
)
_available_modules: dict[str, ModuleAvailable] = {
    name: ModuleAvailable() for name in _init_modules
}

def import_optional(
    module_name: str,
    /,
    scope: _UserDefinedScope | str | Literal["ignore"],
    extra: str = "",
) -> ModuleType:
    if item := _available_modules.get(module_name):
        if module := item.module:
            return module
        elif not item.available:
            msg = _optional_module_msg(scope, module_name, extra) # defined elsewhere
            raise ModuleNotFoundError(msg)
    return _import(module_name, scope, extra)

def _import(
    module_name: str, /, scope: _UserDefinedScope | str, extra: str
) -> ModuleType:
    global _available_modules
    lib_name = module_name.split(".", 1)[0]
    try:
        module = import_module(module_name)
        _available_modules[module_name] = ModuleAvailable(module=module)
        if module_name != lib_name:
            _available_modules[lib_name] = ModuleAvailable(module=sys.modules[lib_name])
    except ImportError:
        unavailable = ModuleAvailable(available=False)
        _available_modules[module_name] = unavailable
        if module_name != lib_name:
            _available_modules[lib_name] = unavailable
        msg = _optional_module_msg(scope, module_name, extra) # defined elsewhere
        raise ModuleNotFoundError(msg) from None
    return module
```

Something like [polars.dependencies](https://github.com/pola-rs/polars/blob/687dee3095ae00137ff7fdb31533ecf960b324e7/py-polars/polars/dependencies.py) is a much more developed spin on this - particularly [dependencies.import_optional](https://github.com/pola-rs/polars/blob/687dee3095ae00137ff7fdb31533ecf960b324e7/py-polars/polars/dependencies.py#L231) which has version bounds.

dangotbanned: Just to be clear, this is more me floating an idea - not something that needs to be tied to this PR

mattijn:
I think we need something else if we want a `NonGeoDataType` type, since the following supports the `SupportsGeoInterface` and also is postive as a `NonGeoDataType` being:

```python
from altair.utils.data import NonGeoDataType

from shapely import geometry
import geopandas as gpd

data_geoms = [
    {"color": "#F3C14F", "geometry": geometry.Polygon([[1.45, 3.75], [1.45, 0], [0, 0], [1.45, 3.75]])},
    {"color": "#4098D7", "geometry": geometry.Polygon([[1.45, 0], [1.45, 3.75], [2.57, 3.75], [2.57, 0], [2.33, 0], [1.45, 0]])},
    {"color": "#66B4E2", "geometry": geometry.Polygon([[2.33, 0], [2.33, 2.5], [3.47, 2.5], [3.47, 0], [3.2, 0], [2.57, 0], [2.33, 0]])},
    {"color": "#A9CDE0", "geometry": geometry.Polygon([[3.2, 0], [3.2, 1.25], [4.32, 1.25], [4.32, 0], [3.47, 0], [3.2, 0]])},
]

gdf_geoms = gpd.GeoDataFrame(data_geoms)
isinstance(gdf_geoms, SupportsGeoInterface), isinstance(gdf_geoms, NonGeoDataType)
```

> ```cmd
> (True, True)
> ```

Example data taken from docs here: https://altair-viz.github.io/user_guide/data.html#spatial-data-gdf

altair/utils/_importers.py
mattijn: Thanks for mentioning this suggestions! Can we remove this `import_toolz_function()` function? I don't think it is used anywhere anymore.

dangotbanned:
> I think we need something else if we want a `NonGeoDataType` type, since the following supports the `SupportsGeoInterface` and also is postive as a `NonGeoDataType` being:

@mattijn I think this would be unavoidable given that [GeoDataFrame](https://github.com/geopandas/geopandas/blob/9e3f6dbadfcbe103cea46aef64507c49fa3b9020/geopandas/geodataframe.py#L78) is a subclass of `pd.DataFrame` and python's typing doesn't support [discriminated unions](https://github.com/python/typing/issues/1467).

So it would pass `NonGeoDataType = Union[dict, pd.DataFrame, DataFrameLike]`
- `DataFrameLike` as it has `.__dataframe__`
- `pd.DataFrame` as it is a subclass

I'm only finding this out now btw, in 8a9d59389df6d1a270f389346cbe2dbabf2e67ad I just attached a name to an existing type annotation.

For the actual usage in `to_csv`, `SupportsGeoInterface` is rejected in https://github.com/dangotbanned/altair/blob/ed13af6d4598c96d1f5e5a816b95a85ff82f81f3/altair/utils/data.py#L376. So the impact of this would be static type checking doesn't pick up the incompatibility - but at runtime the existing code does.

altair/utils/_importers.py
dangotbanned: Sure thing @mattijn

Fixed in 88fbd1141b39ca74b47106d254969661e77bdabf

mattijn:
Thanks for responding! Let me think about it too. Just wondering, do we really need this `NonGeoDataType`? Maybe this constrain within `to_csv` is too conservative?

dangotbanned:
> Just wondering, do we really need this `NonGeoDataType`? 

I've no strong feelings either way @mattijn. 

The only thing I'd add to https://github.com/vega/altair/pull/3426#issuecomment-2175794353 is that having an alias here at least hints the user towards something that we can't express as a type - if that makes sense?

I wouldn't expect passing a `GeoDataFrame` would work as a parameter labelled `NonGeo...`.

Looking back to the original signature, there is no indication that `to_csv` doesn't accept `SupportsGeoInterface`:
 https://github.com/vega/altair/blob/596b28224b36a72b7f32808d2c00e85ba5446533/altair/utils/data.py#L198-L205

mattijn:
I've opened https://github.com/vega/altair/issues/3441. Maybe we can rephrase the `NotImplementedError` with a link tot that github issue. Then we don't need the newly introduced `NonGeoDataType` type in this PR. Thanks for still working on this @dangotbanned, almost there!

altair/utils/_vegafusion_data.py
```diff
@@ -36,21 +45,41 @@ class _ToVegaFusionReturnUrlDict(TypedDict):
...
-) -> Union[_ToVegaFusionReturnUrlDict, ToValuesReturnType]:
+    data: Union[dict, pd.DataFrame, SupportsGeoInterface], max_rows: int
+) -> _VegaFusionReturnType: ...
+
+
+def vegafusion_data_transformer(
+    data: Optional[DataType] = None, max_rows: int = 100000
+) -> Union[Callable[..., Any], _VegaFusionReturnType]:
     """VegaFusion Data Transformer"""
-    if hasattr(data, "__geo_interface__"):
```
mattijn, line 44: I just noticed this won't go well. Moving this check here will break vegafusion, as it will try to parse a GeoDataFrame as a normal DataFrame, which it shouldn't do.

dangotbanned: @mattijn would this be better?:

```py
def vegafusion_data_transformer(
    data: Optional[DataType] = None, max_rows: int = 100000
) -> Union[Callable[..., Any], _VegaFusionReturnType]:
    """VegaFusion Data Transformer"""
    if data is None:
        return vegafusion_data_transformer
    elif isinstance(data, DataFrameLike) and not isinstance(data, SupportsGeoInterface):
        table_name = f"table_{uuid.uuid4()}".replace("-", "_")
        extracted_inline_tables[table_name] = data
        return {"url": VEGAFUSION_PREFIX + table_name}
    else:
        # Use default transformer for geo interface objects
        # # (e.g. a geopandas GeoDataFrame)
        # Or if we don't recognize data type
        return default_data_transformer(data)
```

mattijn: Yes! I think that will do.

mattijn:
Tests are happy too after synching with main. Merging! Thanks @dangotbanned!

Issue #3301, opened 2023-12-24
Adapting Vega-Lite example "Bar Chart with Highlighting on Hover and Selection on Click"


ChristopherDavisUCI:
This is a follow-up to https://github.com/altair-viz/altair/discussions/3295

@mattijn Thanks for linking this [Vega-Lite example](https://vega.github.io/vega-lite/examples/interactive_bar_select_highlight.html) which I didn't know existed.

In that example, `condition` seems to take an array, but I don't see any mention of that as an option in the [condition documentation](https://vega.github.io/vega-lite/docs/condition.html), nor does our code seem to allow using lists for a condition, as far as I can tell:
https://github.com/altair-viz/altair/blob/d38409f8f50282d09507f90e490e65c0f0247b25/altair/vegalite/v5/api.py#L787-L796

Does that seem accurate to you?  How would you suggest we approach that?

Needless to say, I'm happy for comments from anyone!

mattijn:
Hi @ChristopherDavisUCI! This was discussed in this issue: https://github.com/altair-viz/altair/issues/2759. 

It would be nice if we can extend and improve Altair by allowing for multiple predicates as part of a condition.

ChristopherDavisUCI:
Thanks @mattijn!  Do you know anywhere that this array functionality is described on the Vega-Lite side?

mattijn:
Is this page of help: https://vega.github.io/vega-lite/docs/condition.html?

ChristopherDavisUCI:
Not that I know of?  I've tried searching that page for the word "array", and have skimmed through the examples and haven't seen anything that looks similar to the list/array usage in the [Bar Chart example](https://vega.github.io/vega-lite/examples/interactive_bar_select_highlight.html), but I could definitely believe that it's there and I'm missing it...

mattijn:
You are right, it's not really clear from there.
Maybe a bit from the Vega docs at this page: https://vega.github.io/vega/docs/marks/#encode. Scroll down to the section on _Production Rules_

PR #3427, opened 2024-05-29
feat: Improve the syntax for conditions with multiple predicates

### LINKED ISSUE #2759: Improve the syntax for conditions with multiple predicates
Questioned in [SO: How to create an altair condition with three or more choices [duplicate]](https://stackoverflow.com/questions/72190774/how-to-create-an-altair-condition-with-three-or-more-choices) and [SO: Combine hover and click selections in Altair](https://stackoverflow.com/questions/66108224/combine-hover-and-click-selections-in-altair) and surfaced in https://github.com/altair-viz/altair/pull/2756

It is currently possible to write a `condition` with single predicate:
```python
import altair as alt

cars = data.cars.url

base = alt.Chart(cars).mark_tick().encode(x='Miles_per_Gallon:Q')

# condition with single predicate
base.encode(color=alt.condition(
    alt.datum.Origin == 'Europe', 
    alt.value('green'), # when true
    alt.value('red') # when false
    )
)
```
image

And one can combine multiple selections into a single predicate as is written here: https://altair-viz.github.io/user_guide/interactions.html#composing-multiple-selections

To combine multiple predicates within a single condition one can define a `dict` with a list of conditions as such:
```python
# condition with multiple predicates
color_condition = {
    "condition": [
        {"test": "datum.Origin === 'Europe'", "value": "green"},
        {"test": "datum.Origin === 'Japan'", "value": "blue"},
    ],
    "value": "red",
}
base.encode(
    color=color_condition
)
```
image

But using an `alt.condition()` this is not possible. 

Maybe something as such should be supported:?
```python
base.encode(color=alt.condition(
    predicate=[
        alt.pred(expr=alt.datum.Origin == "Europe", if_true=alt.value("green")),
        alt.pred(expr=alt.datum.Origin == "Japan", if_true=alt.value("blue")),
    ],
    if_false=alt.value("red")
    )
)
```

dangotbanned:
Fixes #2759 

Throughout this PR I frequently reference [`polars.when`](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.when.html), and I want to make sure any reviewers have the [origin handy](https://github.com/pola-rs/polars/blob/227b350419030677839f469fa38c5ed91c021432/py-polars/polars/expr/whenthen.py#L18).

# Tasks

## Review
- [x] Investigate strange `SchemaValidationError` https://github.com/vega/altair/pull/3427#issuecomment-2210664643
	- [x] Redesigned `_Conditions` -> `_Conditional`
	- See commit 12929625ad4c02b834b21366e7d7b045e6915861
	- Generic w/ constraint on the type of condition allowed, aligning with [vega-lite](https://vega.github.io/vega-lite/docs/condition.html)
	
	> In addition, there are two ways to encode the data that satisfy the specified condition:
	>	1. Combining one [conditional field definition](https://vega.github.io/vega-lite/docs/condition.html#field) with a single base value or datum definition.
	>	2. Combining one or more [conditional datum or value definitions](https://vega.github.io/vega-lite/docs/condition.html#value) with a field, datum, or value definition.
- [x] Investigate @mattijn proposed `str_as: Literal["value", "datum", "field", ...]` as an alternative to `str_as_lit` https://github.com/vega/altair/pull/3427#issuecomment-2209536374
	- [x] `str_as_lit=False` -> `str_as="shorthand"`
	- See commit https://github.com/vega/altair/pull/3427/commits/d0784f8bf9f8ee1b80d04a429e1efc00804f18e5
	- See thread https://github.com/vega/altair/pull/3427#discussion_r1667443802

## Core

- [x] Add more clearly defined types
	- [x] Update `alt.condition` signature with overloads
- [x] Extract existing multi-stage parsing logic into new functions
- [x] Define utility classes for `when-then-otherwise` syntax
- [x] Reuse/adapt parsing functions for utility classes
- [x] Investigate existing function/method(s) that would accept the new conditions
	- [x] Return of `otherwise` 
		- [x] Valid anywhere `alt.condition` is used
	- [x] Return of `then`
		- [x] Valid as encoding input [`dfc14de` (#3427)](https://github.com/vega/altair/pull/3427/commits/dfc14deb47501f2c4d48e1b7fdf92cbb92125a04)
		- [x] Identify other cases https://github.com/vega/altair/pull/3427#discussion_r1626275778
- [x] Add top-level `when` function as entry-point
	- [x] Rename `_when` -> `when`
- [x] Refactor to reduce code duplication [`38d425f` (#3427)](https://github.com/vega/altair/commit/38d425f6479035d30aafad8eb337e86be532a2f8)
- [x] Add docs for `when-then-otherwise`
	- [x] Initial docs, for: classes, most new types and functions
	- [x] `when`
	- [x] `When.then`
	- [x] `Then.when`
	- [x] `Then.otherwise`
	- [x] `ChainedWhen.then`
- [x] Add tests for `when-then-otherwise`
	- [x] for [passing](https://github.com/dangotbanned/altair/blob/1ad06f4806cdc317ecdec815ea0352642e94c4ca/tests/vegalite/v5/test_api.py#L300) case identified in #3301 
	- [x] for `_Then` ending conditions
	- [x] for mixed predicate input

### Stretch
These would be nice to have and were demonstrated in https://github.com/vega/altair/issues/2759#issuecomment-2137384675. Each idea comes from [polars.when](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.when.html).

However, they need further discussion due to deviating from `alt.condition`'s original behaviour.

- [x] Wrap literals passed to `then`, `otherwise` in `alt.value`. [`38d425f` (#3427)](https://github.com/vega/altair/commit/38d425f6479035d30aafad8eb337e86be532a2f8)
- [x] Wrap kwargs passed to `when` as `alt.datum.key.__eq__(value)`. [`38d425f` (#3427)](https://github.com/vega/altair/commit/38d425f6479035d30aafad8eb337e86be532a2f8)
	- I found only 2 examples that used `alt.condition(**kwargs)`, for something other than `empty`:
		- [Selection Detail](https://altair-viz.github.io/gallery/select_detail.html) passes `legend=None`.
		- [Interactive Scatter Plot and Linked Layered Histogram](https://altair-viz.github.io/gallery/scatter_with_layered_histogram.html) passes `scale=alt.Scale(...)`.
	- Solved by moving `**kwargs` from `when` -> `then`, `otherwise`.
- [x] Modify `when`'s predicate argument to be variadic, reducing all predicates as an `AND` reduction. [`38d425f` (#3427)](https://github.com/vega/altair/commit/38d425f6479035d30aafad8eb337e86be532a2f8)
	- Probably the least contentious
		- `alt.condition` already [accepts the result](https://altair-viz.github.io/user_guide/interactions.html#parameter-composition) of this.
		- `when(predicate, **kwargs)` -> `when(*predicate, **kwargs)` doesn't change the meaning of arguments other than `predicate`
			- A concern that [I raised](https://github.com/vega/altair/issues/2759#issuecomment-2135528133) when proposing making this change within `alt.condition`


altair/utils/schemapi.py
```diff
@@ -893,7 +893,7 @@ def to_dict(
         *,
         ignore: Optional[List[str]] = None,
         context: Optional[Dict[str, Any]] = None,
-    ) -> dict:
+    ) -> Dict[str, Any]:
```
dangotbanned, line 896: I think this is the blocker:

```log
...
Updating API docs
  ->/home/runner/work/altair/altair/doc/user_guide/api.rst
The code generation modified the following files:
altair/utils/schemapi.py

```

altair/vegalite/v5/api.py
```diff
@@ -345,6 +361,260 @@ def check_fields_and_encodings(parameter: Parameter, field_name: str) -> bool:
...
 
+# -------------------------------------------------------------------------
+# Tools for working with conditions
+_TestPredicateType = Union[str, _expr_core.Expression, core.PredicateComposition]
+"""TODO
+
+Item [(2)](https://vega.github.io/vega-lite/docs/condition.html) Specifying a `test` predicate: ...
+"""
+
+_PredicateType = Union[Parameter, core.Expr, TypingDict[str, Any], _TestPredicateType]
```
dangotbanned, line 372: `core.Expr` was in the union of `alt.condition(predicate)`, but the internal logic would raise a `NotImplementedError.
Keeping it around for now, but either the annotation or internals should be updated.

altair/vegalite/v5/api.py
```diff
@@ -19,7 +31,7 @@
 from .schema import core, channels, mixins, Undefined, UndefinedType, SCHEMA_URL
 
 from .data import data_transformers
-from ... import utils, expr
+from ... import utils, expr  # noqa: F401 [No longer needed here, but keeping to prevent breakage elsewhere]
 from ...expr import core as _expr_core
```
dangotbanned, line 35: No longer needed internally, but keeping to prevent breakage elsewhere.

mattijn:
Awesome @dangotbanned, I tried a few things that were mentioned in #2759 which were working already, yeah! 

I tried also two examples of which I was not expecting it to work, but since we are pushing boundaries here.. let me raise them anyways. 

I tried to use the `_alt._when()` method as a replacement for `alt.expr.if_()`. 

Eg in this example https://altair-viz.github.io/user_guide/marks/text.html#labels-position-based-on-condition:
```python
import numpy as np
import pandas as pd
import altair as alt

df = pd.DataFrame({
    'xval': range(100),
})

bind_range = alt.binding_range(min=100, max=300, name='Slider value:  ')
param_width = alt.param(bind=bind_range)

# Examples of how to write both js and python expressions

# possible?:
param_color_py_when = alt.param(expr=_alt._when(param_width < 200).then('red').otherwise('black'))

    alt.X('xval').axis(titleColor=param_color_js_expr),
    alt.Y('yval').axis(titleColor=param_color_py_expr)
).add_params(
    param_width,
    param_color_js_expr,
    param_color_py_expr
)
chart
```
This:
```python
_alt._when(param_width < 200).then('red').otherwise('black')
```
Gives:
```cmd
NotImplementedError: condition predicate of type 'ParameterExpression'
```

and this example: https://altair-viz.github.io/user_guide/interactions.html#expressions-inside-parameters
```python
import altair as alt
import pandas as pd

source = pd.DataFrame({
    "a": ["A", "B", "C"],
    "b": [28, -5, 10]
})

    y="a:N",
    x=alt.X("b:Q").scale(domain=[-10, 35])
)

text_conditioned = bar.mark_text(
    align="left",
    baseline="middle",
    dx=alt.expr.if_(alt.datum.b >= 0, 10, -20))
    # dx=alt.expr(_alt._when(alt.datum.b >= 0).then(10).otherwise(-20))
).encode(text="b")

bar + text_conditioned
```
This:
```python
alt.expr(_alt._when(alt.datum.b >= 0).then(10).otherwise(-20))
```
Gives:
```cmd
SchemaValidationError: '{'condition': [{'test': '(datum.b >= 0)', 'value': 10}], 'value': -20}' is an invalid value for `expr`. Valid values are of type 'string'.
```

And I completely understand this is not expecting to work, but I was just trying.

mattijn:
Hmm, maybe we should add  `alt.expr.when()` for this?

dangotbanned:
Thanks for such a quick response @mattijn. 
These are just my first thoughts after adding the two tests. 

I'm going to read up on the vega-lite spec some more to see how this could be mapped over.

## NotImplementedError: condition predicate of type 'ParameterExpression'

I rewrote this as an `alt.condition` to confirm:
```py
# possible?:
# param_color_py_when = alt.param(
#    expr=_alt._when(param_width < 200).then("red").otherwise("black")
# )
param_color_py_condition = alt.param(
    expr=alt.condition(param_width < 200, alt.value("red"), alt.value("black"))
)
```
That error is part of the [original](https://github.com/dangotbanned/altair/blob/7b671db85d344fd6e0ca2d2a25a862d84118988c/altair/vegalite/v5/api.py#L487) `alt.condition` code, I left the message unchanged other than using f-strings. 

I also added [this comment](https://github.com/vega/altair/pull/3427#discussion_r1623389174), which I originally thought was the issue here. However, there's no relation between `core.Expr` and `ParameterExpression`. 

Essentially, the `predicate` argument for `alt.condition` and `alt._when` are identical in what they allow - both using the same `_PredicateType` annotation - derived from the [original signature](https://github.com/vega/altair/blob/18a2c3c237014591d172284560546a2f0ac1a883/altair/vegalite/v5/api.py#L795).

I'm more than happy to extend what is allowed here, but I do not know why the original restriction was in place. 

## SchemaValidationError: '{'condition': [{'test': '(datum.b >= 0)', 'value': 10}], 'value': -20}' is an invalid value for `expr`. Valid values are of type 'string'.

I just arrived here as well after experimenting with the other one, after adding a new clause to [_predicate_to_condition](https://github.com/dangotbanned/altair/blob/7b671db85d344fd6e0ca2d2a25a862d84118988c/altair/vegalite/v5/api.py#L468)

```py
def _predicate_to_condition(...):
	...
    elif _is_test_predicate(predicate):
        condition = {"test": predicate}
    elif isinstance(predicate, dict):
        condition = predicate
	# new clause that passes the rest of `alt.condition`, but triggers the same `SchemaValidationError`
    elif isinstance(predicate, _expr_core.OperatorMixin):
        condition = {"test": predicate._to_expr()}
    else:
        msg = f"condition predicate of type {type(predicate).__name__!r}"
        raise NotImplementedError(msg)
    return condition
```

altair/vegalite/v5/api.py
```diff
@@ -345,11 +384,596 @@ def check_fields_and_encodings(parameter: Parameter, field_name: str) -> bool:
...
+    condition = _parse_when(predicate, *more_predicates, empty=empty, **constraints)
+    return _When(condition)
+
+
+def _condition_to_expr_str(c: Union[_SelectionType, _Conditions], /) -> str:
+    """Convert the output of `alt.condition`/`when-then-otherwise` to an expression string."""
+    return _condition_to_expr_ref(c).expr
+
+
+def _condition_to_expr_ref(c: Union[_SelectionType, _Conditions], /) -> core.ExprRef:
```
dangotbanned, line 925: @mattijn [comment](https://github.com/vega/altair/pull/3427#issuecomment-2144027795)
> Hmm, maybe we should add  `alt.expr.when()` for this?

Although this wasn't part of the original issue, this was where I landed on this idea.

From my understanding, `alt.expr.if_` and vega-lite [if](https://vega.github.io/vega/docs/expressions/#if) are more limited in that they always require 3 arguments. 
So while `when-then-otherwise` seems a good fit for this case, `alt.expr.if_` couldn't accept:
- `when-then`
- `when-then-when-then-otherwise`
- etc

Personally, I'd expect the same behaviour from both and find it frustrating to encounter the limitation.

An option could be accepting a subset of possible outputs from `alt.condition`/`when-then-otherwise`. At least for the two examples you provided, this solution could work. 

I must admit though I'm not familiar with `alt.expr` in general, and found it difficult to follow outside of a notebook environment.

altair/vegalite/v5/api.py
```diff
@@ -345,11 +384,596 @@ def check_fields_and_encodings(parameter: Parameter, field_name: str) -> bool:
...
+        raise TypeError(if_false)
+    return selection
+
+
+class _Conditions(_TypedDict, total=False):
+    condition: _Required[List[_ConditionType]]
+    value: Any
+
+
+class _LiteralConfig(typing.NamedTuple):
```
dangotbanned, line 547: Wrapped as a `NamedTuple` to avoid conflicts with `then`, `otherwise` accepting `**kwargs`.
Also keeps the parameters easy to move around without [FBT001](https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/) issues.

dangotbanned:
Adding a summary I wrote in #3426 on the status of this PR, as I'd rather keep it updated here if I have any more thoughts.

I've left this on draft, not ready yet, but would be good to hear your thoughts on a couple of things:

1. `alt.when` [signature](https://github.com/dangotbanned/altair/blob/8c0f45ae22bc829a500a495b2a978a2faedcb749/altair/vegalite/v5/api.py#L882)
	- To preserve compatibility with `alt.condition`, the first `predicate` arg behaves identically - but `*more_predicates` is more restrictive in the types allowed - as they all must implement `&`.
	- I've left this to bake before writing docs, as I'm not sure whether the trade-off in added complexity (for the user) is worth preserving identical semantics as `alt.condition` for a single argument.
2. The new parts added for `alt.expr` [comment](https://github.com/vega/altair/pull/3427#discussion_r1626275778) which are "an" implementation, but not sure it would be the right way to go.

altair/vegalite/v5/api.py
```diff
@@ -345,11 +384,641 @@ def check_fields_and_encodings(parameter: Parameter, field_name: str) -> bool:
...
+        **kwargs: Any,
+    ) -> _Then:
+        lit = _LiteralConfig(str_as_lit, seq_as_lit)
+        condition = self._when_then(statement, kwargs, lit)
+        conditions = self._conditions.copy()
+        conditions["condition"].append(condition)
+        return _Then(conditions)
+
+
+def when(
```
dangotbanned, line 889: RFC on `alt.when` docstring.

The passing commit hasn't got the messages, so for context I've added those below. 

Of note, `alt.when`'s doc will be very similar to [`alt._Then.when`](https://github.com/dangotbanned/altair/blob/80a08122ad1e1decdd0986e25a852f928f743e1d/altair/vegalite/v5/api.py#L825) - so I'll be leaving that until after this has been agreed.


`_FieldEqualType` in 989b1f50ee428d4b7de7f591a24731276cbbea1c
> - Provides static type checking instead of `Any` 
> - Alias name hints towards usage
> - Alias doc gives further context, without requiring repeating & syncing wherever it is used

`alt.when` in 7d412198720170a1e4ab46717df931b800e1dee6

> I'm happy with the content in the doc, hopefully this can be refined during review.
> 
> - It adapts parts of `alt.condition`, but not verbatim as there was a `# TODO: update the docstring`
> - Contains links to other sources of inspo
> - Deliberately avoided the word "condition" as much as possible
>   -  `when`'s result can become a condition, but only after calling `result.then(...)`.
>   - Stressing that `when` != `condition`

doc/user_guide/api.rst
```diff
@@ -166,6 +166,7 @@ API Functions
    topo_feature
    value
    vconcat
+   when
```
dangotbanned, line 169: I expected `When` would also show up after 58e005406e42268ec0ffd63c38d077384685375a but there are 2 issues. 

1. It would never be selected under the current conditions of `generate_api_docs` https://github.com/vega/altair/blob/76a9ce1e27a21359e7b0ef9e6cb865f8d95fc298/tools/generate_api_docs.py#L122-L133
2. If this was altered to include classes from `api`, there would be a collision with `when`/`When` on the url.

I'm still fairly new to `sphinx` so there may be a simple solution to the latter. Due to how `api` is defined, moving these classes to another module would require some less than desirable import special casing.

altair/__init__.py
```diff
@@ -617,11 +611,13 @@ def __dir__():
     return __all__
 
 
-from .vegalite import *
-from .jupyter import JupyterChart
+from altair.vegalite import *
+from altair.vegalite.v5.schema.core import Dict
```
dangotbanned, line 615: This explicit, absolute import is to ensure mypy emits an `[assignment]` error if a child package/module misbehaves in the future. 

E.g.
```
altair\__init__.py:615: error: Incompatible import of "Dict" (imported name has type "type[altair.vegalite.v5.schema.core.Dict]", local name has type "type[dict[Any, Any]]") [assignment]
```

tools/generate_api_docs.py
```diff
@@ -64,6 +64,16 @@
    :nosignatures:
 
    {lowlevel_wrappers}
+
+API Utility Classes
```
dangotbanned, line 68: This was my solution to https://github.com/vega/altair/pull/3427#discussion_r1649699382

Ideally, I'd want to have `when`, `then`, `otherwise` *methods* documented fully and hide `__init__`, `to_dict`.

Currently only the `when` *function* pulls through the docs, which I think isn't helpful alone. 

I'm not sure how to achieve this without changing https://github.com/dangotbanned/altair/blob/54a68a310f48ba9d2201da1820c4abaf459154aa/doc/_templates/class.rst and that would affect all classes (which I'm trying to avoid).

The `narwhals` [docs](https://narwhals-dev.github.io/narwhals/api-reference/expressions/) are an example of what I'd like, but with the 3 classes sharing a single page.

doc/user_guide/api.rst
dangotbanned: See https://github.com/vega/altair/pull/3427#discussion_r1659965947

altair/vegalite/v5/api.py
```diff
@@ -380,24 +422,625 @@ def check_fields_and_encodings(parameter: Parameter, field_name: str) -> bool:
...
+class When(_BaseWhen):
+    """Utility class for ``when-then-otherwise`` conditions.
+
+    Represents the state after calling :func:`.when()`.
+
+    This partial state requires calling :meth:`When.then()` to finish the condition.
+
+    References
+    ----------
+    `polars.expr.whenthen <https://github.com/pola-rs/polars/blob/b85c5e0502ca99c77742ee25ba177e6cd11cf100/py-polars/polars/expr/whenthen.py>`__
```
dangotbanned, line 742: Marking this for discussion of all `polars` references. 

I feel there is a need to reference `polars`, but I'm not sure the "best" way to do so. 
There is [intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html), but I think that might only solve for the user guide/ api reference links.

altair/vegalite/v5/api.py
```diff
@@ -380,24 +422,625 @@ def check_fields_and_encodings(parameter: Parameter, field_name: str) -> bool:
...
+        if isinstance(if_false, str):
+            if_false = utils.parse_shorthand(if_false)
+            if_false.update(kwargs)
+        selection = dict(condition=condition, **if_false)
+    else:
+        raise TypeError(if_false)
+    return selection
+
+
+class _Conditions(TypedDict, total=False):
```
dangotbanned, line 583: This is helpful for making sense of the internal structure. 

An issue it poses can be seen in `alt.Chart` during [this test](https://github.com/dangotbanned/altair/blob/25ac2fa4601a02c340ba20be1d6dc858e3e18aad/tests/vegalite/v5/test_api.py#L375-L376)

Using a `TypedDict` makes `mypy` unhappy, as it recognizes `dict` as an accepted type - but not the kind produced by the `TypedDict` metaclass [alt.Chart.encode](https://github.com/vega/altair/blob/7f05ecbf999766ae1398dcbeb6144c128413026c/altair/vegalite/v5/schema/channels.py#L34206-L34216)

altair/vegalite/v5/api.py
dangotbanned: Removed in 801cccc80e7764df3d46cd0db0343f095ee78d01

dangotbanned:
@binste was closing this intentional?

This branch is attached to a roadmap issue. 

- #2759 


The link to #3458 was only addressing the comment I added in https://github.com/vega/altair/pull/3427#discussion_r1662542866

binste:
Nope, not intentional. Somehow GitHub closed it when I merged the PR :)

dangotbanned:
> Nope, not intentional. Somehow GitHub closed it when I merged the PR :)

Thank you

altair/vegalite/v5/api.py
dangotbanned: Fixed in https://github.com/vega/altair/commit/91a1de17f97d0b1de20724f3cfa771f07741dcc7

altair/vegalite/v5/api.py
mattijn: I’m fine with leaving any work done on `alt.expr.when()` out of this PR @dangotbanned! We can open a new issue on this topic, based on the experience of this PR. I’m positive we can figure out a solution, since the work done in this PR is more advanced than what is needed for `alt.expr.if_`.

dangotbanned: Removed in [`f89990f` (#3427)](https://github.com/vega/altair/pull/3427/commits/f89990f512e12bd11d3f511900410840f139f081)

altair/vegalite/v5/api.py
```diff
@@ -380,24 +422,573 @@ def check_fields_and_encodings(parameter: Parameter, field_name: str) -> bool:
...
```
dangotbanned, line 583: A better name for this could be `Conditional`, to align with the other `condition` related classes

mattijn:
I noticed you have marked this PR as review ready. Nice! 
I tried the following examples:

```python
import altair as alt

cars = data.cars.url

base = alt.Chart(cars).mark_tick().encode(x='Miles_per_Gallon:Q')

# condition with single predicate
base.encode(color=alt.when(alt.datum.Origin == 'Europe')
            .then(alt.value('green'))
            .otherwise(alt.value('red'))
)
```
image

```python
# condition with multiple predicates
base.encode(color=alt.when(alt.datum.Origin == 'Europe')
            .then(alt.value('green'))
            .when(alt.datum.Origin == 'Japan')
            .then(alt.value('blue'))
            .otherwise(alt.value('red'))
)
```
image

and
```python
import altair as alt
import polars as pl

df = pl.DataFrame([
      {"a": "A", "b": 28}, {"a": "B", "b": 55}, {"a": "C", "b": 43},
      {"a": "D", "b": 91}, {"a": "E", "b": 81}, {"a": "F", "b": 53},
      {"a": "G", "b": 19}, {"a": "H", "b": 87}, {"a": "I", "b": 52}
])

highlight = alt.selection_point(on='mouseover', empty=False)
select = alt.selection_point(on='click', empty=False)

# condition with multiple predicates using params
alt.Chart(df).mark_bar(fill="#4C78A8", stroke="black", cursor="pointer").encode(
    x="a:N",
    y="b:Q",
    fillOpacity=alt.when(highlight)
        .then(alt.value(1))
        .otherwise(alt.value(0.3)),
    strokeWidth=alt.when(select)
        .then(alt.value(2))
        .when(highlight)
        .then(alt.value(1))
        .otherwise(alt.value(0))
).configure_scale(
    bandPaddingInner=0.2
).add_params(highlight, select)
```
image

I noticed that
```python
alt.when(alt.datum.Origin == 'Europe').then(alt.value('green')).otherwise(alt.value('red'))
```
returns
```cmd
{'condition': [{'test': (datum.Origin === 'Europe'), 'value': 'green'}],
 'value': 'red'}
```
And that:
```python
alt.when(alt.datum.Origin == 'Europe').then('green').otherwise('red')
```
Returns
```cmd
{'condition': [{'test': (datum.Origin === 'Europe'), 'field': 'green'}],
 'field': 'red'}
```
Can you provide some description which option you choose from your options in this comment here: https://github.com/vega/altair/issues/2759#issuecomment-2137384675

Thanks again for this work, its really neat!

dangotbanned:
Thanks for getting to this so quickly @mattijn 

# Main

> Can you provide some description which option you choose from your options in this comment here: [#2759 (comment)](https://github.com/vega/altair/issues/2759#issuecomment-2137384675)
> 
> Thanks again for this work, its really neat!

Sure thing! Maybe I need to work on the docstrings, I added an example on `alt.when` but didn't for each method due to https://github.com/vega/altair/pull/3427#discussion_r1659965947

The three below are all equivalent (labelled `a`, `b`, `c`):
```py
alt.when(alt.datum.Origin == 'Europe').then(alt.value('green')).otherwise(alt.value('red')) # a
alt.when(Origin="Europe").then(alt.value("green")).otherwise(alt.value("red"))				# b
alt.when(Origin="Europe").then("green", str_as_lit=True).otherwise("red", str_as_lit=True)	# c
```

The keyword `str_as_lit`'s name comes from [polars](https://github.com/pola-rs/polars/blob/f803053fb73abda45bd11ad25fc0a6e642e1038e/py-polars/polars/_utils/parse/expr.py#L22), which behaves in a very similar way - but they use `pl.lit` instead of `alt.value`.

I spent quite some time considering the name, the other candidates were `value`, `shorthand`, `lit`, `field`:
- It should not obscure the `**kwargs`, inherited from `alt.condition`
	- All except `lit` have other meanings in `altair` & `vega-lite`
- As it is only required for `str`, using either `value=True`/`shorthand=False` could lead to confusion
	- `alt.when(Origin="Europe").then(1).otherwise(2)` doesn't require the keyword
	- `lit` would be shorter but probably cause the same issue

You have probably noticed `b` is **longer** than `c`. 
I initially thought this was a deal-breaker, but despite the additional length - this is clearly documented and autocomplete-able. 
Compared to [alt.condition](https://altair-viz.github.io/user_guide/generated/api/altair.condition.html#altair.condition), which does not indicate how an `str` is special-cased.

I'm quite interested in your opinion on my compromise

# Example rewrites

The results will be the same, but just to utilise the new stuff:
```py
# condition with multiple predicates
base.encode(
    color=alt.when(Origin="Europe")
    .then("green", str_as_lit=True)
    .when(Origin="Japan")
    .then("blue", str_as_lit=True)
    .otherwise("red", str_as_lit=True)
)
```

```py
# condition with multiple predicates using params
alt.Chart(df).mark_bar(fill="#4C78A8", stroke="black", cursor="pointer").encode(
    x="a:N",
    y="b:Q",
    fillOpacity=alt.when(highlight).then(1).otherwise(0.3),
    strokeWidth=alt.when(select).then(2).when(highlight).then(1).otherwise(0),
).configure_scale(bandPaddingInner=0.2).add_params(highlight, select)
```

Thank you again for giving this a whirl. 
I'm hopeful that there's enough to work with here, even if all of it doesn't make it in

mattijn:
Thanks for the explanation. Just documenting my experiences here. I'm intrigued by the `str_as_lit`, that is an interesting approach.  I think for altair it makes more sense if there was not `str_as_lit`, but the following items:
- `str_as_value`
- `str_as_datum`
- `str_as_field`

See also https://vega.github.io/vega-lite/docs/condition.html. Maybe the list should be extended with:
- `str_as_aggregate`
- `str_as_bin`
- `str_as_timeunit`
- `str_as_scale`
- `str_as_legend` 
- `str_as_format`

And I think it makes sense it returns the output with the behavior similar to `str_as_value=True`
***
When I do:
```python
import altair as alt
import pandas as pd

df = pd.DataFrame({'local': ['2018-01-01T00:00:00'],
                   'utc': ['2018-01-01T00:00:00Z']})

alt.Chart(df).transform_calculate(
    compliant="hours(datum.local) != hours(datum.utc) ? true : false",
).mark_text(size=20, baseline='middle').encode(
    text=alt.when('datum.compliant').then('OK', str_as_lit=True).otherwise('not OK', str_as_lit=True),
    color=alt.when('datum.compliant').then('OK', str_as_lit=True).otherwise('not OK', str_as_lit=True)
).properties(width=80, height=50)
```
It returns an empty chart.
image

When comparing the output of
```python
alt.when('datum.compliant').then('OK', str_as_lit=True).otherwise('not OK', str_as_lit=True)
```
> ```python
> {'condition': [{'test': 'datum.compliant', 'field': 'OK'}], 'field': 'not OK'}
> ```
with
```python
alt.condition('datum.compliant', alt.value('OK'), alt.value('not OK'))
```
> ```python
> {'condition': {'test': 'datum.compliant', 'value': 'OK'}, 'value': 'not OK'}
> ```
I noticed that the `condition` is nested within `[]` for the `alt.when()`. That is different.
***
It would be great if the following example would work:
```python
import altair as alt

cars = data.cars()

alt.Chart(cars).mark_rect().encode(
    x='Cylinders:N',
    y='Origin:N',
    color=alt.when(brush).then('count()').otherwise('grey')
    #color=alt.condition(brush, if_true='count()', if_false=alt.value('grey'))
).add_params(brush)
```
This currently gives a `SchemaValiditionError` and the output of:
```python
alt.when(brush).then('count()').otherwise('grey')
```
gives:

> ```python
> {'condition': [{'param': 'param_14',
>    'aggregate': 'count',
>    'type': 'quantitative'}],
>  'field': 'grey'}
> ```

Almost good already! Its nice to see that `count()` is translated into a transformation function (`aggregate`).

***
This returns a condition with `value`s
```python
alt.when(Origin="Europe").then(1).otherwise(0.3)
```

> ```python
> {'condition': [{'test': (datum.Origin === 'Europe'), 'value': 1}],
>  'value': 0.3}
> ```
And this with `field`s
```python
alt.when(Origin='Europe').then('green').otherwise('red')
```

> ```python
> {'condition': [{'test': (datum.Origin === 'Europe'), 'field': 'green'}],
>  'field': 'red'}
> ```
***
If I don't specify a `.otherwise()`, I get the following output
```python
alt.when(brush).then('count()')#.otherwise('grey')
```
```python
<altair.vegalite.v5.api.Then at 0x2c24399d960>
```
But at the above mentioned VL page on condition it is mentioned that:
> Note that if the “else” case value is not specified, default mark color will be applied.

I noticed it work correctly within an `alt.Chart()` defintion, as in, it doesnt error when `.otherwise()` is not included.

dangotbanned:
> Edit: maybe easier to have `str_as` and then a `Literal` with the allowed options `['value', 'datum', 'field', ...]`

@mattijn I really like this! Absolutely wouldn't have thought of it myself.
I'll experiment with the idea and come back with any quirks I find.

> And I think it makes sense if it returns the default output with the behavior similar to `str_as_value=True`

Syntactically, I agree this would be the cleanest. 
Something that I've tried to do is **maintain the existing semantics** from `alt.condition` by default as seen in https://github.com/dangotbanned/altair/blob/8f66b7a8ec11b9be917d1374edcfaa20dc7de941/tests/vegalite/v5/test_api.py#L318.

So these should both result in the same spec (despite the `list` in the latter):

```py
import altair as alt

alt.condition(select, "column_1", alt.value("blue"))
alt.when(select).then("column_1").otherwise(alt.value("blue"))
```
> I noticed that the `condition` is nested within `[]` for the `alt.when()`. That is different.

Yeah that hasn't appeared to impact the usage in a chart for the tests I had so far. Also I thought this was expected after your proposal in https://github.com/vega/altair/issues/2759#issue-1501529331

> This currently gives a `SchemaValiditionError` and the output of:
> ```py
> alt.when(brush).then('count()').otherwise('grey')
> ```

This again is trying to preserve semantics from `alt.condition`

So in your example:

```py
import altair as alt

cars = data.cars()

alt.Chart(cars).mark_rect().encode(
    x='Cylinders:N',
    y='Origin:N',
	# color=alt.when(brush).then('count()').otherwise('grey')
    color=alt.when(brush).then('count()').otherwise(alt.value('grey'))
    # color=alt.condition(brush, if_true='count()', if_false=alt.value('grey'))
).add_params(brush)
```

What I'm finding interesting here is the `SchemaValidationError` itself:
```
SchemaValidationError: `Color` has no parameter named 'aggregate'

Existing parameter names are:

See the help for `Color` to read the full description of these parameters
```

This is strange to me as it was passed:
```py
{'condition': [{'param': 'param_7',
   'aggregate': 'count',
   'type': 'quantitative'}],
 'value': 'grey'}
```

I'm going to add this one to the tests and try and debug what is going on here, because it clearly recognizes the nested `aggregate` but being within a list has produced a nonsensical error.

> If I don't specify a .otherwise(), I get the following output
> But at the above mentioned VL page on condition it is mentioned that:
> 
> > Note that if the “else” case value is not specified, default mark color will be applied.

Thanks I'm glad this was noticed, see these for context:
- https://github.com/dangotbanned/altair/blob/8f66b7a8ec11b9be917d1374edcfaa20dc7de941/altair/utils/core.py#L55
- https://github.com/dangotbanned/altair/blob/8f66b7a8ec11b9be917d1374edcfaa20dc7de941/altair/utils/core.py#L837
- https://github.com/dangotbanned/altair/blob/8f66b7a8ec11b9be917d1374edcfaa20dc7de941/altair/vegalite/v5/api.py#L779
- https://github.com/dangotbanned/altair/blob/8f66b7a8ec11b9be917d1374edcfaa20dc7de941/altair/vegalite/v5/api.py#L862

I would prefer if we could work out a *less* special-cased integration.
The challenge is these classes aren't part of the schema, but `Then` needs to be maintained - so have access to its methods.

> Also, altair is able to understand that count() is an aggregate transformation function and yearmonthdate() a timeunit transformation function. Does that conflict with the str_as_xx() approach?

I'm not sure I fully understand what you mean. 
For reference, the default behaviour of `.then(str)`, `.otherwise(str)` in https://github.com/dangotbanned/altair/blob/8f66b7a8ec11b9be917d1374edcfaa20dc7de941/altair/vegalite/v5/api.py#L672 simply defers to https://github.com/vega/altair/blob/c0f85cd1d1676f5fcdfda337d4990e24e93a9c39/altair/utils/core.py#L458-L650

altair/vegalite/v5/api.py
```diff
@@ -732,15 +732,27 @@ def _parse_when(
...
         raise TypeError(msg)
 
 
+def _str_as(val: str, to: Literal["datum", "field", "value"], /):
+    if to == "value":
+        return value(val)
+    elif to == "field":
+        return utils.parse_shorthand(val)
+    elif to == "datum":
+        return _ConditionClosed(test=_expr_core.GetAttrExpression(to, val))
```
dangotbanned, line 750: Not accessible yet. 
Purely guessing that this would be the correct shape. 

The returned type **must** support `.update()` and `**`, but cannot call `to_dict()` as `Expression` overrides it to return an `str` https://github.com/vega/altair/blob/c0f85cd1d1676f5fcdfda337d4990e24e93a9c39/altair/expr/core.py#L175-L176

dangotbanned: After searching for `alt.condition` in `r"./altair/tests/examples_arguments_syntax/*.py"`, I've found no examples of `datum` being used in this way. 

# Examples
All of these use some form of datum-based expression as the **predicate** only.

<details>
<summary>7 Examples</summary>


https://github.com/vega/altair/blob/c0f85cd1d1676f5fcdfda337d4990e24e93a9c39/tests/examples_arguments_syntax/bar_chart_with_highlighted_bar.py#L16-L20

https://github.com/vega/altair/blob/c0f85cd1d1676f5fcdfda337d4990e24e93a9c39/tests/examples_arguments_syntax/bar_chart_with_negatives.py#L15-L19

https://github.com/vega/altair/blob/c0f85cd1d1676f5fcdfda337d4990e24e93a9c39/tests/examples_arguments_syntax/candlestick_chart.py#L15-L17

https://github.com/vega/altair/blob/c0f85cd1d1676f5fcdfda337d4990e24e93a9c39/tests/examples_arguments_syntax/lasagna_plot.py#L11-L15

https://github.com/vega/altair/blob/c0f85cd1d1676f5fcdfda337d4990e24e93a9c39/tests/examples_arguments_syntax/layered_heatmap_text.py#L34-L38

https://github.com/vega/altair/blob/c0f85cd1d1676f5fcdfda337d4990e24e93a9c39/tests/examples_arguments_syntax/slider_cutoff.py#L24-L27

https://github.com/vega/altair/blob/c0f85cd1d1676f5fcdfda337d4990e24e93a9c39/tests/examples_arguments_syntax/us_employment.py#L31-L35

</details>

Comparing to where the `str_as`/`str_as_lit` argument in `alt.when` applies to `alt.condition`:

```py
import altair as alt

alt.condition(predicate, "maybe_str_1", "maybe_str_2")

alt.when(predicate).then("maybe_str_1").otherwise("maybe_str_2")
```
>  I think for altair it makes more sense if there was not str_as_lit, but the following items:
> - `str_as_value`
> - `str_as_datum`
> - `str_as_field`

> Edit: maybe easier to have str_as and then a Literal with the allowed options ['value', 'datum', 'field', ...]

@mattijn is there something I'm missing here?

If I am understanding this correctly, I think there is only a need for `str_as: Literal["shorthand", "value"]`.
The more specific options are covered already by default in [alt.utils.core.parse_shorthand](https://github.com/vega/altair/blob/c0f85cd1d1676f5fcdfda337d4990e24e93a9c39/altair/utils/core.py#L458-L650).

If this were a performance concern, I would focus on `parse_shorthand` first - with changes like this earlier commit https://github.com/vega/altair/commit/14f6ea162b50a1cd2cf0c9678deb46f96c03b1a1

dangotbanned: > If this were a performance concern, I would focus on `parse_shorthand` first ...

I was curious, but even with the changes below I wasn't able to observe any performance improvement when running `hatch test --all`:

<details>
<summary>Code block <b>altair/altair/utils/core.py</b></summary>

```py
# altair/altair/utils/core.py

...

_InferOption = Literal["aggregates", "window_ops", "timeunits", "types"]
_InferConfig: TypeAlias = Tuple[Tuple[_InferOption, bool], ...]

_SHORTHAND_PATTERN_CACHE: dict[_InferConfig, list[re.Pattern]] = {}

# I also tried adding a 2nd layer of caching here with `@functools.lru_cache(None)`
# Still saw no performance benefit
def _find_matches(
    shorthand: str, *, aggregates: bool, window_ops: bool, timeunits: bool, types: bool
) -> dict[str, str | Any]:
    # NOTE:
    # - Result of regexps is the cache target
    # - keyed to the `parse_` args
    # - `TopLevelMixin.transform_[timeunit|window]`
    #  - are only library code that uses non-default
    global _SHORTHAND_PATTERN_CACHE
    config: _InferConfig = (
        ("aggregates", aggregates),
        ("window_ops", window_ops),
        ("timeunits", timeunits),
        ("types", types),
    )

    def _generate() -> list[re.Pattern]:
        fmts: list[str] = []
        if aggregates:
            fmts.extend([r"{agg_count}\(\)", r"{aggregate}\({field}\)"])
        if window_ops:
            fmts.extend([r"{op_count}\(\)", r"{window_op}\({field}\)"])
        if timeunits:
            fmts.append(r"{timeUnit}\({field}\)")
        fmts.append(r"{field}")

        it = chain.from_iterable((p + ":{type}", p) for p in fmts) if types else fmts
        patterns = [
            re.compile(rf"\A{p.format(**SHORTHAND_UNITS)}\Z", re.DOTALL) for p in it
        ]
        _SHORTHAND_PATTERN_CACHE[config] = patterns
        return patterns

    patterns = _SHORTHAND_PATTERN_CACHE.get(config, _generate())
    return next(m.groupdict() for pat in patterns if (m := pat.match(shorthand)))


def parse_shorthand(
    shorthand: dict[str, Any] | str,
    data: pd.DataFrame | DataFrameLike | None = None,
    parse_aggregates: bool = True,
    parse_window_ops: bool = False,
    parse_timeunits: bool = True,
    parse_types: bool = True,
) -> dict[str, Any]:
    """General tool to parse shorthand values ..."""
    from altair.utils._importers import pyarrow_available

    if not shorthand:
        return {}
    # find matches depending on valid fields passed
    elif isinstance(shorthand, dict):
        # NOTE: This branch previously did all the same initial work,
        # but then used `shorthand` every time
        attrs = shorthand
    else:
        attrs = _find_matches(
            shorthand,
            aggregates=parse_aggregates,
            window_ops=parse_window_ops,
            timeunits=parse_timeunits,
            types=parse_types,
        )

    # Handle short form of the type expression
    if "type" in attrs:
        attrs["type"] = INV_TYPECODE_MAP.get(attrs["type"], attrs["type"])

    # counts are quantitative by default
    if attrs == {"aggregate": "count"}:
        attrs["type"] = "quantitative"

    # times are temporal by default
    if "timeUnit" in attrs and "type" not in attrs:
        attrs["type"] = "temporal"

    # if data is specified and type is not, infer type from data
    if "type" not in attrs:
        if pyarrow_available() and data is not None and isinstance(data, DataFrameLike):
            dfi = data.__dataframe__()
            if "field" in attrs:
                unescaped_field = attrs["field"].replace("\\", "")
                if unescaped_field in dfi.column_names():
                    column = dfi.get_column_by_name(unescaped_field)
                    try:
                        attrs["type"] = infer_vegalite_type_for_dfi_column(column)
                    except (NotImplementedError, AttributeError, ValueError):
                        # Fall back to pandas-based inference.
                        # Note: The AttributeError catch is a workaround for
                        # https://github.com/pandas-dev/pandas/issues/55332
                        if isinstance(data, pd.DataFrame):
                            attrs["type"] = infer_vegalite_type(data[unescaped_field])
                        else:
                            raise

                    if isinstance(attrs["type"], tuple):
                        attrs["sort"] = attrs["type"][1]
                        attrs["type"] = attrs["type"][0]
        elif isinstance(data, pd.DataFrame):
            # Fallback if pyarrow is not installed or if pandas is older than 1.5
            #
            # Remove escape sequences so that types can be inferred for columns with special characters
            if "field" in attrs and attrs["field"].replace("\\", "") in data.columns:
                attrs["type"] = infer_vegalite_type(
                    data[attrs["field"].replace("\\", "")]
                )
                # ordered categorical dataframe columns return the type and sort order as a tuple
                if isinstance(attrs["type"], tuple):
                    attrs["sort"] = attrs["type"][1]
                    attrs["type"] = attrs["type"][0]

    # If an unescaped colon is still present, it's often due to an incorrect data type specification
    # but could also be due to using a column name with ":" in it.
    if (
        "field" in attrs
        and ":" in attrs["field"]
        and attrs["field"][attrs["field"].rfind(":") - 1] != "\\"
    ):
        raise ValueError(
            '"{}" '.format(attrs["field"].split(":")[-1])
            + "is not one of the valid encoding data types: {}.".format(
                ", ".join(TYPECODE_MAP.values())
            )
            + "\nFor more details, see https://altair-viz.github.io/user_guide/encodings/index.html#encoding-data-types. "
            + "If you are trying to use a column name that contains a colon, "
            + 'prefix it with a backslash; for example "column\\:name" instead of "column:name".'
        )
    return attrs

...

```

</details>

Thought I'd leave this here, since I won't be committing but it may be helpful to know this part isn't a performance bottleneck.

dangotbanned:
@mattijn 
> Regarding:
> 
> > Does that conflict with the str_as_xx() approach?
> 
> Maybe it is better to explain what I meant with an example:
> 
> ```python
> alt.when(predicate).then('count()', str_as='value')
> ```
> 
> If the predicate satisfies, will this evaluate to a condition that contains an `aggregate` or will it evaluate to a condition containing a `value`?

So I'll follow on from your example with what the intention is with `str_as`.


```py

class When:
	...
	def then(
        self,
        statement: _StatementOrLiteralType,
        *,
        str_as: Literal["shorthand", "value"] = "shorthand", # This is equivalent to `str_as_lit=False`
        **kwargs: Any,
    ) -> Then[Any]: ...

when = alt.when(predicate)
```

This is all that is happening when passing the keyword, https://github.com/vega/altair/pull/3427#discussion_r1667443802
```py
when.then("count()", str_as="value") == when.then(alt.value("count()"))

when.then("count()", str_as="shorthand") == when.then(alt.utils.parse_shorthand("count()"))
```

altair/vegalite/v5/api.py
```diff
@@ -380,26 +425,727 @@ def check_fields_and_encodings(parameter: Parameter, field_name: str) -> bool:
...
+        :class:`Then`
+        """
+        condition = self._when_then(statement, kwds, str_as)
+        if _is_condition_extra(condition, statement, kwds=kwds, str_as=str_as):
+            return Then(_Conditional(condition=condition))
+        else:
+            return Then(_Conditional(condition=[condition]))
+
+
+class Then(core.SchemaBase, t.Generic[_C]):
```
dangotbanned, line 861: Following https://github.com/vega/altair/commit/d12a74117ec201fc096f5e27328d4e50ea493f84 `Then` inherits from `SchemaBase`.

This addresses https://github.com/vega/altair/pull/3427#issuecomment-2209536374

@mattijn 
> If I don't specify a `.otherwise()`, I get the following output
> ```py
> >>> alt.when(brush).then('count()')#.otherwise('grey')
> ```
> <altair.vegalite.v5.api.Then at 0x2c24399d960>
> 
> 
> But at the above mentioned VL page on condition it is mentioned that:

> > Note that if the “else” case value is not specified, default mark color will be applied.

> I noticed it work correctly within an `alt.Chart()` defintion, as in, it doesnt error when `.otherwise()` is not included.

A drawback to this is the `Then.[when|otherwise]` methods are more difficult to find with auto-complete:

## When

image


## Then

image


## ChainedWhen

image

mattijn:
Thanks for the changes and explanations! 
When I try the following example:
```python
import altair as alt

cars = data.cars()

alt.Chart(cars).mark_rect().encode(
    x='Cylinders:N',
    y='Origin:N',
    color=alt.when(brush).then('count()').otherwise('grey')
).add_params(brush)
```
It returns an empty chart
image

That was a bit confusing for me. After inspecting
```python
alt.when(brush).then('count()').otherwise('grey')
```
I notices it gives:
```python
{'condition': [{'param': 'param_1', 'value': 'count()'}], 'value': 'grey'}
```

But after changing it to:
```python
alt.when(brush).then('count()', str_as='shorthand').otherwise('grey')
```
It returns the correct condition:
```
{'condition': {'param': 'param_1',
  'aggregate': 'count',
  'type': 'quantitative'},
 'value': 'grey'}
```
But if the options for `str_as` are solely `['shorthand', 'value']`, then it probably makes more sense to check if the `str` should very much likely resolve as a `shorthand` and otherwise treat it as a value? And if a user requires that the `str` resolves to a `value`, then the user can just force it using `alt.value()`. Then there is also no need for the `str_as` parameter.

I have been experimenting a lot to get an example working with an `alt.datum()` in the `.then()` or `.otherwise()`, but I could not get that to work, not even within the vega editor. Even though I think the docs on condition ([link](https://vega.github.io/vega-lite/docs/condition.html)) states that that should be possible. Since this is not working within VL, it also is not necessary to anticipate for this here.

I cannot find an extensive list of examples utilizing a condition, so let me add some more examples here which I tried.

I tried this and it works great
```
import altair as alt

cars = data.cars.url

hover = alt.selection_point(on='pointerover', nearest=True, empty=False)

chart = alt.Chart(cars, title='Selection obscured by other points').mark_circle(opacity=1).encode(
    x='Horsepower:Q',
    y='Miles_per_Gallon:Q',
    color=alt.when(hover).then('coral').otherwise('lightgray'),
    size=alt.when(hover).then(300).otherwise(30)
).add_params(
    hover
)

chart | chart.encode(
    order=alt.when(hover).then(1).otherwise(0)
).properties(
    title='Selection brought to front'
)
```

I tried this and it doesn't work great:
```python
    x='Horsepower:Q',
    y='Miles_per_Gallon:Q',
    color=alt.when(brush).then('Origin:N').otherwise('lightgray')
).add_params(
    brush
)
```
From a user perspective, it is clear that `'Origin:N'` is meant as a field definition and not as a `value`. Where in 
 `alt.when(hover).then('coral').otherwise('lightgray')`, it is clear that `'coral'` is meant as a value definition.

dangotbanned:
Thanks for getting back to me on this @mattijn, really appreciate you taking the time to try out some examples.

> From a user perspective, it is clear that `'Origin:N'` is meant as a field definition and not as a value. Where in
`alt.when(hover).then('coral').otherwise('lightgray')`, it is clear that `'coral'` is meant as a value definition.

I do have more to say on this point, and will follow up in another comment discussing potential issues to address.
*If you are reading this, you've been to quick for me to finish typing up the next comment* 😎 

---


## Prev examples using `alt.condition`
Re-writing the two problematic `alt.when`s as `alt.condition`s produces the following:

```py
import altair as alt

# alt.when(brush).then('count()').otherwise('grey')
>>> alt.condition(brush, "count()", "grey")
ValueError: A field cannot be used for both the `if_true` and `if_false` values of a condition. One of them has to specify a `value` or `datum` definition.
```

```py
# alt.when(brush).then('Origin:N').otherwise('lightgray')
>>> alt.condition(brush, "Origin:N", "grey")
ValueError: A field cannot be used for both the `if_true` and `if_false` values of a condition. One of them has to specify a `value` or `datum` definition.
```

So from the user's perspective, `alt.when` fails silently. I don't see this as a positive UX.
However, I don't think the error or [docstring](https://altair-viz.github.io/user_guide/generated/api/altair.condition.html#altair.condition) for `alt.condition` clearly illustrates the issue either. 

Nowhere was the term `field` specified until the error, and an identical error is presented when changing the definition to:

```py
>>> alt.condition(brush, "coral", "lightgrey")
ValueError: A field cannot be used for both the `if_true` and `if_false` values of a condition. One of them has to specify a `value` or `datum` definition.
```

Related, but not the same issue, trying this example:

```py
# alt.when(hover).then(300).otherwise(30)
>>> alt.condition(brush, 300, 30)
TypeError: 'int' object is not iterable
```

---

For all of these examples, `alt.when` *can* accept the inputs - albeit sometimes requiring the documented `str_as="shorthand"`, whereas `alt.condition` will not and IMO is less intuitive in hinting towards how to resolve the error.

mattijn:
Its quite difficult to have auto inference of `str`ings to either a field, aggregate, value or datum. Maybe we have been too ambitious on this aspect? A potential way out is to start with the explicit variant within the `alt.when().then().otherwise()` and continue from there to see if an implicit variant also can be included?

dangotbanned:
@mattijn 
> Here I use an `alt.value('coral')`, `alt.datum('foo')` and `count()` for the `x` or `color` encoding channel.
> 
> If we like to have a clever implicit variant, that knows what this is:
> 
> ```python
> brush = alt.selection_interval()
> alt.when(brush).then('count()').otherwise('coral')
> alt.when(brush).then('count()').otherwise(300)
> ```
> 
> Will that also mean one should expect this to work?:
> 
> ```python
> tick_value = alt.Chart(df).mark_bar().encode(color='coral', y=alt.Y('a:N'))
> tick_datum = alt.Chart(df).mark_bar().encode(x='foo', y=alt.Y('a:N'))
> tick_agg = alt.Chart(df).mark_bar().encode(x='count()', y=alt.Y('a:N'))
> tick_value | tick_datum | tick_agg
> ```

Yeah this was the issue I was thinking of when writing 

> To me though, it seems this could be unreliable and be introducing deviating behaviour from what user's may expect from a shorthand string

I guess my main concern is introducing some new inference logic that applies only in certain places - unless this has a strong reason for doing so. 
I still think the change in default mentioned in https://github.com/vega/altair/pull/3427#issuecomment-2212473301 qualifies as this, but I'm not happy about the silent passthrough you discovered in https://github.com/vega/altair/pull/3427#issuecomment-2227448475 that I didn't account for in testing

dangotbanned:
@binste @joelostblom @jonmmease 

Do you guys have any thoughts/opinions on the discussions here?

mattijn:
I'm not sure if you can be certain that a non-`str` is a value and not a datum.

Example without using a condition
```python
point = alt.Chart(df).mark_point(filled=True, size=150, cursor='pointer').encode(x='a', y='a')
rule = alt.Chart().mark_rule(strokeDash=[12, 6], size=4).encode(
    x=alt.datum(0.5)  
)
point + rule
```
image

And now I like to add a condition that changes the location of the x-rule based on when I select a certain point. So I add an `alt.condition(select, alt.datum(0.5), alt.datum(1.5))` as such:
```python
select = alt.selection_point()

point = alt.Chart(df).mark_point(filled=True, size=150, cursor='pointer').encode(x='a', y='a')
rule = alt.Chart().mark_rule(strokeDash=[12, 6], size=4).encode(
    x=alt.condition(select, alt.datum(0.5), alt.datum(1.5))
    # x=alt.when(select).then(0.5).otherwise(1.5)    
)
(point + rule).add_params(select)
```
It currently errors in Altair, probably due to errors in VL I think.

But if I simplify this example a bit more in Vega-Lite there is a warning in the editor and the predicate is always false, but the x-rule is present at `"datum": 1.5`: ([Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAZgAY2AacaYsiygAlMiRoVYcAvpO4AbAIYBPTACcUAbVB15heXhBR5AW0woQ23QFoAnDABsdgKyYAjABYA7C0xsATPJY7MEwWfzY3OgAjMEi3Oxc7FhAZECN5FQBrfUJFAAdTZBBciBJxbnhZWUw6FEIVUkwUzChIOhIkVBAAD314TFkawr1uHPyzAEdSeSI4XWI0UxTFXqwBs2GQUYKQSeniObgF5JTDEzMMHAB9R2TOLR09TtPtzCNcnNvU9KzOrbN6qogbiUOAALwKbmBdQgGUwABF5JRmMh1C5fJw7ABdJotCBtKAdUA9X55ba7GYHI7cCykIwoNgAOkc0kkmO4uXSxkoGlAzzMHJUxkuaKBIEo-UwYHEJLGhWKpVF0DMRgEmAgCzUKQu2G5KJA2uuIGxbJAABJKGBGK9HiAxIRcpRkAB6J0LBDyBkIWaMUiRBlwCBOi1WtKuzDuyyyWaYV2OBkuDwMxkAK0oSupD3FhF19ystgczncXh8-kCwVC8nCURicQSSRRoEeLhkjZQzbuIEevlZ3Fe72WDdZ0iAA))

```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.17.0.json",
  "layer": [
    {
      "params": [
        {"name": "CHANGE_RULE_ON_SELECT", "select": {"type": "point"}}
      ],
      "data": {"values": [{"a": 1}, {"a": 1}, {"a": 2}]},
      "mark": {
        "type": "point",
        "filled": true,
        "cursor": "pointer",
        "size": 150
      },
      "encoding": {
        "x": {"field": "a", "type": "quantitative"},
        "y": {"field": "a", "type": "quantitative"}
      },
      "name": "LAYER_VIEW"
    },
    {
      "data": {"values": [{"a": 1}, {"a": 1}, {"a": 2}]},
      "mark": {"type": "rule", "size": 4, "strokeDash": [12, 6]},
      "encoding": {
        "x": {
          "condition": {"param": "CHANGE_RULE_ON_SELECT", "datum": 0.5},
          "datum": 1.5
        }
      }
    }
  ]
}
```
image

(I wish there was a predicate viewer in the editor, where I can test and evaluate predicates).

So in this case, when treating non-`str` as values would have been wrong within `x=alt.when(select).then(0.5).otherwise(1.5)`. 

Other thoughts are much welcome here!

mattijn:
Completely agree that that error is not helpful at all! Maybe that could be improved?

Your suggestion helped figuring out an example that doesn't error. Not sure if it actually makes sense, but it does something and has an `alt.datum()` in the `if_true` argument (would be `alt.then()`).
```python
import altair as alt
import polars as pl

select = alt.selection_point(empty=False)

df = pl.DataFrame({'a': [1, 1, 2]})
point = alt.Chart(df).mark_point(filled=True, size=150, cursor='pointer').encode(x='a', y='a')
rule = alt.Chart(df).mark_rule(strokeDash=[12, 6], size=4).encode(
    x='a',
    color=alt.condition(select, alt.datum(2), alt.value("1.5"))
    #color=alt.when(select).then(2).otherwise("1.5")
)

(point + rule).add_params(select)
```

mattijn:
See https://altair-viz.github.io/gallery/line_chart_with_datum.html and https://altair-viz.github.io/gallery/line_chart_with_color_datum.html for usage of datum (examples are without conditions)

jonmmease:
> Do you guys have any thoughts/opinions on the discussions here?

Hi @dangotbanned, I haven't dug into this PR, but I do appreciate your work on it!  I'll defer to @mattijn on the API decisions here

joelostblom:
I echo what John wrote above. My general thoughts on the API are still captured in https://github.com/vega/altair/issues/2759#issuecomment-2139018131, and I'm a bit short of time at the moment to dive into the specifics that you and Mattijn have uncovered. From my side, I'm ok if the first implementation is explicit with datum and value, if finding a solution to the string inference is deemed to time-consuming/ambiguous.

mattijn:
My suggestion would be to keep this PR to solely introducing the explicit input variants for `alt.when().then().otherwise()` methods, where this approach support chaining as such; `alt.when().then().when().then().otherwise()` or to use Altair defaults instead of the `.otherwise()`, by just leaving it out, like `alt.when().then()`. This is all great and awesome and deserves some more superlatives. Thanks for all the work here @dangotbanned!

joelostblom:
> or to use Altair defaults instead of the `.otherwise()`, by just leaving it out, like `alt.when().then()`

I didn't realize this was possible, how convenient!

dangotbanned:
> My suggestion would be to keep this PR to solely introducing the explicit input variants for `alt.when().then().otherwise()` methods, where this approach support chaining as such; `alt.when().then().when().then().otherwise()` or to use Altair defaults instead of the `.otherwise()`, by just leaving it out, like `alt.when().then()`. This is all great and awesome and deserves some more superlatives. Thanks for all the work here @dangotbanned!

Thank you @mattijn and @joelostblom, @jonmmease for the feedback. 

No qualms from me, having explicit input for `.then()`, `.otherwise()` is definitely the simplest option - and the changes I'd need to make should be easy enough to do tomorrow. 

---

There have been some other points I raised (unrelated to most of what we've discussed) that may be good to resolve in the meantime. 

Most of these relate to documentation, so they might be more approachable/easier to weigh in on:

- https://github.com/vega/altair/pull/3427#discussion_r1668638763
- https://github.com/vega/altair/pull/3427#discussion_r1623389174
- https://github.com/vega/altair/pull/3427#discussion_r1659965947
- https://github.com/vega/altair/pull/3427#discussion_r1662514935
- https://github.com/vega/altair/pull/3427#discussion_r1642900469

dangotbanned:
@mattijn I've just finished scoping out what I anticipate are the requested changes. 

I wanted to do this ahead of time to ensure I've understood correctly. 
Also I think it will make the next review easier than studying the diff alone.

If this all sounds right to you, a simple 👍 is all I need, `.otherwise()` please follow this up with any corrections/alternatives/etc.

---

# Core Changes

## Removing `str_as`
Always assumed that `str` is shorthand, defering `statement` to https://github.com/vega/altair/blob/f8c9776a72afa433f3e6b311bdaf5133d38f127b/altair/utils/core.py#L492-L493

## Drop support for these `statement` inputs to `.then()`, `.otherwise()`
```py

NoLongerSupported: TypeAlias = bool | float | int | Sequence[str | bool | float | int]

```
## For both
- Removing related `@overload`s
- Removing helper functions, that are now unused
- Removing any `TypeAlias` definitions, that are now unused
- Updating/removing tests
- Updating docs
- Updating error messages, which refer to these features

# Impact on examples
Current (as of [`0f862ad` (#3427)](https://github.com/vega/altair/pull/3427/commits/0f862adf64d153a1a6892adaac5b0dc75ed01495)) is commented out vs next

## https://github.com/vega/altair/pull/3427#issuecomment-2227448475

```py
import altair as alt

cars = data.cars()
hover = alt.selection_point(on="pointerover", nearest=True, empty=False)

# color_1 = alt.when(brush).then('count()', str_as='shorthand').otherwise('grey')
color_1 = alt.when(brush).then("count()").otherwise(alt.value("grey"))

chart_1 = (
    alt.Chart(cars)
    .mark_rect()
    .encode(x="Cylinders:N", y="Origin:N", color=color_1)
    .add_params(brush)
)

# --------

# color_2 = alt.when(hover).then("coral").otherwise("lightgray")
color_2 = alt.when(hover).then(alt.value("coral")).otherwise(alt.value("lightgray"))

# size = alt.when(hover).then(300).otherwise(30)
size = alt.when(hover).then(alt.value(300)).otherwise(alt.value(30))

# order = alt.when(hover).then(1).otherwise(0)
order = alt.when(hover).then(alt.value(1)).otherwise(alt.value(0))

chart_2 = (
    alt.Chart(cars, title="Selection obscured by other points")
    .mark_circle(opacity=1)
    .encode(x="Horsepower:Q", y="Miles_per_Gallon:Q", color=color_2, size=size)
    .add_params(hover)
)

chart_2 | chart_2.encode(order=order).properties(title="Selection brought to front")
```

## [#3301](https://github.com/vega/altair/issues/3301)

```py
import altair as alt

select = alt.selection_point(name="select", on="click")
highlight = alt.selection_point(name="highlight", on="pointerover")

# fill_opacity = alt.when(select).then(1).otherwise(0.3)
fill_opacity = alt.when(select).then(alt.value(1)).otherwise(alt.value(0.3))

# stroke_width = (
#     alt.when(select)
#     .then(2, empty=False)
#     .when(highlight)
#     .then(1, empty=False)
#     .otherwise(0)
# )
stroke_width = (
	alt.when(select)
    .then(alt.value(2, empty=False))
    .when(highlight)
    .then(alt.value(1, empty=False))
	.otherwise(alt.value(0))
)
data = ...
chart = (
    alt.Chart(data)
    .mark_bar(fill="#4C78A8", stroke="black", cursor="pointer")
    .encode(x="a:O", y="b:Q", fillOpacity=fill_opacity, strokeWidth=actual_stroke)
    .configure_scale(bandPaddingInner=0.2)
    .add_params(select, highlight)
)
```

mattijn:
👍😊

altair/vegalite/v5/api.py
```diff
@@ -132,7 +132,19 @@
 _TSchemaBase = TypeVar("_TSchemaBase", bound=core.SchemaBase)
 _T = TypeVar("_T")
 _OneOrSeq = TypeAliasType("_OneOrSeq", Union[_T, Sequence[_T]], type_params=(_T,))
-"""One of ``_T`` specified type(s), or a `Sequence` of such."""
+"""One of ``_T`` specified type(s), or a `Sequence` of such.
```
dangotbanned, line 135: I'll be moving this into `_typing.py` eventually, but wanted to leave the trail of the origin in [`3fa4a3f` (#3427)](https://github.com/vega/altair/pull/3427/commits/3fa4a3f7baa26e7ea4c6eacac778b91a9801bd58)

altair/vegalite/v5/api.py
```diff
@@ -96,7 +127,24 @@
...
+
+Examples
+--------
+The parameters ``short``, ``long`` accept the same range of types::
+
+    # ruff: noqa: UP006, UP007
+
+    def func(
+        short: _OneOrSeq[str | bool | float],
```
dangotbanned, line 144: I'll be moving this into `_typing.py` eventually, but wanted to leave the trail of the origin in [`3fa4a3f` (#3427)](https://github.com/vega/altair/pull/3427/commits/3fa4a3f7baa26e7ea4c6eacac778b91a9801bd58)

altair/vegalite/v5/api.py
```diff
@@ -379,26 +427,662 @@ def check_fields_and_encodings(parameter: Parameter, field_name: str) -> bool:
...
+    -----
+    - Directly inspired by the ``when-then-otherwise`` syntax used in ``polars.when``.
+
+    References
+    ----------
+    `polars.when <https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.when.html>`__
+
+    Examples
+    --------
+    Using keyword-argument ``constraints`` can simplify compositions like::
```
dangotbanned, line 1043: Extracted from a comment:

```py
# TODO
# ----
# - Previously a note in the docstring
# - but probably a better fit in user guide
#
# Similar to `alt.condition` but:
# Allows specifiying a predicate, without a default
# ```py
# condition = (predicate, if_true)
# ```
# Allows multiple precise predicates, optionally ending with a default:
# ```py
# conditions = [
#     (predicate_1, if_true_1),
#     (predicate_2, if_true_2),
#     (predicate_3, if_true_3),
#     ...,
#     if_false,
# ]
# ```
```

dangotbanned:
> 👍😊

@mattijn All of https://github.com/vega/altair/pull/3427#issuecomment-2236376458 is finished and ready for review.
Did a little bit of tidying up as well.

If you'd like to defer the [docs](https://github.com/vega/altair/pull/3427#issuecomment-2234096501) part to another PR, I'd be happy with merging as-is.

mattijn:
Thanks! Looks great! A few question for clarifications.

I was confused about these two new tests:
```python
    when_constraint = alt.when(Origin="Europe")
    when_constraints = alt.when(
        Name="Name_1", Color="Green", Age=25, StartDate="2000-10-01"
    )
    expected_constraint = alt.datum.Origin == "Europe"
    expected_constraints = (
        (alt.datum.Name == "Name_1")
        & (alt.datum.Color == "Green")
        & (alt.datum.Age == 25)
        & (alt.datum.StartDate == "2000-10-01")
    )
```

This is syntax I have not seen before. I noticed this is mentioned in the polars docs, that mentions this:

```
    constraints
        Apply conditions as `col_name = value` keyword arguments that are treated as
        equality matches, such as `x = 123`. As with the predicates parameter, multiple
        conditions are implicitly combined using `&`.
```
And
```
    Pass conditions as keyword arguments:

    >>> df.with_columns(val=pl.when(foo=4, bar=0).then(99).otherwise(-1))
    shape: (3, 3)
    ┌─────┬─────┬─────┐
    │ foo ┆ bar ┆ val │
    │ --- ┆ --- ┆ --- │
    │ i64 ┆ i64 ┆ i32 │
    ╞═════╪═════╪═════╡
    │ 1   ┆ 3   ┆ -1  │
    │ 3   ┆ 4   ┆ -1  │
    │ 4   ┆ 0   ┆ 99  │
    └─────┴─────┴─────┘
```
Do you know where this type of syntax comes from? Is it unique for polars or also used elsewhere? 

This only functions with a single `=` operator? Ie. a predicate like `foo<4` is not supported? And what about `alt.when('my column has spaces'=4)`

dangotbanned:
> Thanks! Looks great! A few question for clarifications.
> 
> I was confused about these two new tests:
> 
> ```python
>     when_constraint = alt.when(Origin="Europe")
>     when_constraints = alt.when(
>         Name="Name_1", Color="Green", Age=25, StartDate="2000-10-01"
>     )
>     expected_constraint = alt.datum.Origin == "Europe"
>     expected_constraints = (
>         (alt.datum.Name == "Name_1")
>         & (alt.datum.Color == "Green")
>         & (alt.datum.Age == 25)
>         & (alt.datum.StartDate == "2000-10-01")
>     )
> ```
> 
> This is syntax I have not seen before. I noticed this is mentioned in the polars docs, that mentions this:
> 
> ```
>     constraints
>         Apply conditions as `col_name = value` keyword arguments that are treated as
>         equality matches, such as `x = 123`. As with the predicates parameter, multiple
>         conditions are implicitly combined using `&`.
> ```
> 
> And
> 
> ```
>     Pass conditions as keyword arguments:
> 
>     >>> df.with_columns(val=pl.when(foo=4, bar=0).then(99).otherwise(-1))
>     shape: (3, 3)
>     ┌─────┬─────┬─────┐
>     │ foo ┆ bar ┆ val │
>     │ --- ┆ --- ┆ --- │
>     │ i64 ┆ i64 ┆ i32 │
>     ╞═════╪═════╪═════╡
>     │ 1   ┆ 3   ┆ -1  │
>     │ 3   ┆ 4   ┆ -1  │
>     │ 4   ┆ 0   ┆ 99  │
>     └─────┴─────┴─────┘
> ```
> 
> Do you know where this type of syntax comes from? Is it unique for polars or also used elsewhere?
> 
> This only functions with a single `=` operator? Ie. a predicate like `foo<4` is not supported? And what about `alt.when('my column has spaces'=4)`

@mattijn I'm confused also but for a different reason. 
This has been in here for a while and it is described with an example in the doc for `alt.when` https://github.com/vega/altair/blob/87cd9bd1d4d7c9de192f83ef68da8235cfb2cb95/altair/vegalite/v5/api.py#L1035

It utilises `**kwargs`, so you can technically do it for columns with spaces like this:

```py
alt.when(**{'my column has spaces':4})
```

I would maybe note this in the *User Guide* as an option, but I'd probably lean on a different feature for those cases. 

Apologies, I'm doing this from my phone - but I'm pretty sure one of the later tests has this syntax of unpacking. 

I'll update this tomorrow with a more complete response, but appreciate you checking this out so quickly

mattijn:
Great, this docstring you refer to is very clear. Thanks for linking. 

It would be great if we can extend the documentation as well in a follow up PR on this new feature's scope to manage user expectations.

Maybe without realizing, but there have been many new things in this PR! Thanks for all these efforts. 

I'm approving, but please comment when you think this PR can be merged!

mattijn (approved):

dangotbanned:
> Great, this docstring you refer to is very clear. Thanks for linking.
> 
> It would be great if we can extend the documentation as well in a follow up PR on this new feature's scope to manage user expectations.
> 
> Maybe without realizing, but there have been many new things in this PR! Thanks for all these efforts.
> 
> I'm approving, but please comment when you think this PR can be merged!

@mattijn thank you, I'm ready for merging 

Will absolutely be following up with another PR w/ more docs, especially examples on the methods (not just `alt.when`)

mattijn:
Thank you @dangotbanned!

PR #3428, opened 2024-05-30
docs: Add example of how to create polar bar charts

joelostblom:
I was playing around a bit and thought this was a potentially useful example of how to do something that is not immediately obvious when reading the arc_mark docs.[ In many other contexts these are referred to as "polar bar charts"](https://duckduckgo.com/?t=ffab&q=polar+bar+chart&atb=v310-1&iax=images&ia=images) so I chose that name to show that it is possible to create these in altair, although we are not technically supporting a polar coordinate system.

image

### DIFF
--- a/tests/examples_arguments_syntax/polar_bar_chart.py
+++ b/tests/examples_arguments_syntax/polar_bar_chart.py
@@ -1,0 +1,62 @@
+"""
+Polar Bar Chart
+---------------
+This example shows how to make a polar bar chart using ``mark_arc``.
+This could also have been called a "pie chart with axis labels",
+but is more commonly referred to as a polar bar chart.
+The axis lines are created using pie charts with only the stroke visible.
+"""
+# category: circular plots
+
+import math
+import altair as alt
+import pandas as pd
+
+source = pd.DataFrame({
+    "hour": range(24),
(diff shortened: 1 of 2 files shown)


mattijn:
Maybe add the name "rose plot" to it as well. 
I did a section on this at the geopython conference last year, see also onwards from here: https://mattijn.github.io/posts/2023-03-07-talk-geopython-2023/#interaction

joelostblom:
Nice post @mattijn ! Could you link me to where you defined the function `utils_chart_rose()`? Maybe I can borrow something from there to improve this example as well

jonmmease:
Great example @joelostblom! Someday it would be great to have the logic for creating the polar guide included in a package somewhere. Either in Vega-Altair itself or an `altair-polar` package or something.  We could probably also support points and lines in polar coordinates as well.

mattijn:
> Could you link me to where you defined the function `utils_chart_rose()`?

Its here: https://github.com/mattijn/talks/blob/main/docs/utils_geoconf_23.py#L35. Totally agree with @jonmmease that better support for this is welcome as there is currently no any automated axis positioning. 

eg. sometimes these polar bars are not starting at zero, but has the middle line of the bar at zero, then you can use a theta offset as such `alt.Chart().mark_arc(thetaOffset=np.deg2rad(-7.5)).encode(..)` to accommodate for this.

Issue #3430, opened 2024-06-04
Add additional `ruff` rules to `pyproject.toml`


dangotbanned:
Following a comment in [#3420 review](https://github.com/vega/altair/pull/3420#discussion_r1610436425), I've collected a range of `ruff` rules, to propose adopting in `altair`. 

This is an opinionated list, with a focus on rules providing an autofix (`ruff check . --fix`) - to limit any extra work removing warnings.

I think the `ruff` docs do a great job explaining each rule, so I've kept my additional notes brief. Strongly recommend checking out rules marked with a 🛠️ for examples of the change they would make.

Also, if accepted, I would suggest introducing any new rules close to a new release - since this would very likely create a large diff, especially if all of these were added at once.

## [flake8-simplify (SIM)](https://docs.astral.sh/ruff/rules/#flake8-simplify-sim)
- Majority have autofixes
- First mentioned in [#3420 review](https://github.com/vega/altair/pull/3420#discussion_r1610436425)

## [pydocstyle (D)](https://docs.astral.sh/ruff/rules/#pydocstyle-d)
  - Majority have autofixes
  - I haven't come across any `altair` specific doc conventions
  - A starting point could be the config below and then tweaking `DXXX` rules from there
  - See [docstring-formatting](https://docs.astral.sh/ruff/formatter/#docstring-formatting) for sample output

```toml
[tool.ruff.format]
docstring-code-format=true
docstring-code-line-length=72

[tool.ruff.lint]
pycodestyle={ max-doc-length=88 }
pydocstyle={ convention="numpy" }
```

## [pyupgrade (UP)](https://docs.astral.sh/ruff/rules/#pyupgrade-up)
- All have autofixes and respect `tool.ruff.target-version`
- Really helpful considering [altair 1.0.0](https://pypi.org/project/altair/1.0.0/#description) targeted `python>=2.7,<3.6`
- When paired with [FA100](https://docs.astral.sh/ruff/rules/future-rewritable-type-annotation/#future-rewritable-type-annotation-fa100), the autofixes can help improve readability of annotations - some of which are particularly verbose on `python<3.9`.

## [flake8-type-checking (TCH)](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tch)
- Majority have autofixes
- Related to [#3405](https://github.com/vega/altair/pull/3405) @binste
- Also relates to `UP`, `FA100`

[...]

PR #3431, opened 2024-06-05
refactor: Add `ruff` rules, improve type annotations, improve ci performance

dangotbanned:
Fixes: https://github.com/vega/altair/issues/3430

- Adds most of the autofixable rules
- Tweak `pyproject.toml`
- Made some manual fixes, due to comment placement blocking the autofix
- Resolved 2 issues relating to `win32` and `pyright`
- Rerun lint, format, schemagen

# Stats from the first round of rules
`ruff check . --statistics`
Found **313** errors, **293** fixable `[*]`.

<details>
<summary>Details</summary>

```md
91      UP032   [*] Use f-string instead of `format` call
62      EM101   [*] Exception must not use a string literal, assign to variable first
28      EM103   [*] Exception must not use a `.format()` string directly, assign to variable first
24      EM102   [*] Exception must not use an f-string literal, assign to variable first
19      UP008   [*] Use `super()` instead of `super(__class__, self)`
11      RUF002  [ ] Docstring contains ambiguous `–` (EN DASH). Did you mean `-` (HYPHEN-MINUS)?
10      RUF100  [*] Unused `noqa` directive (non-enabled: `C901`)
 8      PIE790  [*] Unnecessary `pass` statement
 8      UP030   [*] Use implicit references for positional format fields
 6      SIM102  [ ] Use a single `if` statement instead of nested `if` statements
 6      SIM118  [*] Use `key in dict` instead of `key in dict.keys()`
 5      SIM108  [*] Use ternary operator `cls = matches[0] if matches else default_class` instead of `if`-`else`-block
 4      SIM117  [ ] Use a single `with` statement with multiple contexts instead of nested `with` statements
 4      SIM300  [*] Yoda conditions are discouraged, use `Version("1.0.0") <= PANDAS_VERSION` instead
 4      RUF005  [*] Consider `(*scope, group_index)` instead of concatenation
 3      SIM910  [*] Use `channel_to_name.get(type_)` instead of `channel_to_name.get(type_, None)`
 3      UP015   [*] Unnecessary open mode parameters
 3      RUF015  [*] Prefer `next(iter(dct["datasets"].values()))` over single element slice
 2      PIE800  [*] Unnecessary spread `**`
 2      PIE810  [*] Call `startswith` once with a `tuple`
 2      SIM105  [ ] Use `contextlib.suppress(AttributeError)` instead of `try`-`except`-`pass`
 2      UP004   [*] Class `ParameterExpression` inherits from `object`
 2      UP027   [*] Replace unpacked list comprehension with a generator expression
 1      PIE807  [*] Prefer `dict` over useless lambda
 1      UP028   [*] Replace `yield` over `for` loop with `yield from`
 1      UP031   [*] Use format specifiers instead of percent format
 1      RUF010  [*] Use explicit conversion flag
```

</details>

# Tasks
- [x] Find source of `RUF002` characters
	- [x] 10 cases [like](https://github.com/dangotbanned/altair/blob/b6190f27235700b03052aafcb7f5ee5fa0f25d75/altair/vegalite/v5/api.py#L1416)
	> Docstring contains ambiguous `’` (RIGHT SINGLE QUOTATION MARK). Did you mean ``` (GRAVE ACCE...
	- [x] 1 case [like](https://github.com/dangotbanned/altair/blob/b6190f27235700b03052aafcb7f5ee5fa0f25d75/altair/vegalite/v5/api.py#L1454)
	> Docstring contains ambiguous `–` (EN DASH). Did you mean `-` (HYPHEN-MINUS)?
- [x] [flake8-use-pathlib (PTH)](https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth) (all manual)
- [x] [Preview rules](https://docs.astral.sh/ruff/preview/)
	- [x] [pylint (PLC, E, R, W)](https://docs.astral.sh/ruff/rules/#pylint-pl)
	- [x] [refurb (FURB)](https://docs.astral.sh/ruff/rules/#refurb-furb)
- [x] Resolve [cases](https://github.com/vega/altair/pull/3431#discussion_r1628051704) of comments being lost
	- Seems to occur when a comment is nested within an `elif/else`, but is collapsed after linting
- [x] [flake8-type-checking (TCH)](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tch)/[FA100](https://docs.astral.sh/ruff/rules/future-rewritable-type-annotation)/[pyupgrade (UP)](https://docs.astral.sh/ruff/rules/#pyupgrade-up) manual component
	- For many of these fixes, you first need a `from __future__ import annotations` declaration at the top of each file.
	- I want to be sure all the maintainers (particularly @binste) understand the impact this can have on any code that relies on `typing` constructs storing class objects - which would then be "stringified".
	- It would still be a huge positive in simplifying imports, but there may be scenarios where the new behaviour is a source of unexpected issues
	- See [PEP 563](https://peps.python.org/pep-0563/)
	- Also [PEP 649](https://peps.python.org/pep-0649/) which is due for `python=3.14`
 suggesting requires further discussion
- [x] Library-specific (mostly manual)
	- [x] [flake8-pytest-style (PT)](https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt)

# Future Rule Considerations
- [pydocstyle (D)](https://docs.astral.sh/ruff/rules/#pydocstyle-d)
	- See [comment](https://github.com/vega/altair/issues/3430#issuecomment-2148981414)
- Library-specific (mostly manual)
	- [NumPy-specific rules (NPY)](https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy)
	- [pandas-vet (PD)](https://docs.astral.sh/ruff/rules/#pandas-vet-pd)


sphinxext/altairgallery.py
```diff
@@ -355,7 +355,7 @@ def main(app):
             example["prev_ref"] = "gallery_{name}".format(**prev_ex)
         if next_ex:
             example["next_ref"] = "gallery_{name}".format(**next_ex)
-        Path(target_dir, f"{example["name"]}.rst").write_text(
+        Path(target_dir, example["name"] + ".rst").write_text(
```
dangotbanned, line 358: Thinking this would've been possible if [docbuild](https://github.com/vega/altair/blob/18a2c3c237014591d172284560546a2f0ac1a883/.github/workflows/docbuild.yml#L13) was using the `python-version: "3.10"`.

Would there be a specific reason for using `3.10` here?
I'd understand `3.8` or `3.12`, but seems unlikely there's a dependency on `3.10`.

tools/schemapi/schemapi.py
```diff
@@ -218,20 +218,15 @@ def _prepare_refs(d: Dict[str, Any]) -> Dict[str, Any]:
         for key, value in d.items():
             if key == "$ref":
                 d[key] = _VEGA_LITE_ROOT_URI + d[key]
-            else:
-                # $ref values can only be nested in dictionaries or lists
```
dangotbanned, line 222: This comment block seems to have been dropped

altair/utils/schemapi.py
```diff
@@ -220,20 +220,15 @@ def _prepare_refs(d: Dict[str, Any]) -> Dict[str, Any]:
         for key, value in d.items():
             if key == "$ref":
                 d[key] = _VEGA_LITE_ROOT_URI + d[key]
-            else:
-                # $ref values can only be nested in dictionaries or lists
```
dangotbanned, line 224: Comment block lost

altair/vegalite/v5/api.py
```diff
@@ -4065,17 +4043,16 @@ def remove_prop(subchart, prop):
...
-                raise ValueError(f"There are inconsistent values {values} for {prop}")
+                msg = f"There are inconsistent values {values} for {prop}"
+                raise ValueError(msg)
+        elif all(
+            getattr(c, prop, Undefined) is Undefined or c[prop] == chart[prop]
+            for c in subcharts
+        ):
+            output_dict[prop] = chart[prop]
         else:
-            # Top level has this prop; subchart must either not have the prop
```
dangotbanned, line 4070: Comment block lost

sphinxext/altairgallery.py
binste: Yeah we just have to use a version but it can be any of the supported ones by Altair, I'm also not aware of any specific dependency on 3.10 for the documentation. Feel free to bump it to 3.12 and if `docbuild` still runs through that should be fine.

binste:
Regarding the futures type annotation import, I'm not too familiar with the consequences of enabling that but it seems to be a highly debated topic in the Python community, see our own previous discussion at https://github.com/vega/altair/pull/2950#issuecomment-1473660066 and https://github.com/python/cpython/issues/82786 where it's mentioned that there are now other PEPs to treat this differently. I'd rather not introduce it if we're not fully aware of the consequences, especially if it's only for cosmetic reasons.

binste (changes requested):
It's very satisfying to see all the harmonization done in this PR! And great to see that you seem already very familiar with the automated code generation :) Thanks for putting in the work to do this! Especially on the pathlib stuff. I'm sure that was not so easy but good if we can move away from relying on `os` so much.

Personally, I'd not use quite all the rules that are enabled here but I think the beauty of it is, that we don't have to discuss these things going forward so I tried to only comment on rules where I have a stronger preference. After these comments, it'd be good to go from my side. If any other maintainer has strong preferences on the rule set which is applied, please chime in!

dangotbanned:
@binste thanks for getting to this so quickly. 

I've only skimmed through, will go over each fully tomorrow.

I should've mentioned in the description, but I separated the rule changes & the effect into sequential commits. 

For example new rules a16b2920dea9f9257ee86cfb4c8744bd24d40ce6 is followed by fixes 7fff541d8aeb816ff12ddfa5a9a8b5369c91e665.


I think viewing from [`de50aa9` (#3431)](https://github.com/vega/altair/pull/3431/commits/de50aa934fe620574227ec2fedb8bded00bd88a9) and clicking "next" will help identify rules that aren't to taste.

image

> Could you instead set the encoding to `utf-8`? I think we always want that and `getpreferredencoding` is not a well-known function.
> 
> We should have done that before. Good to introduce it now.

For example, that was introduced by [PLW1514](https://docs.astral.sh/ruff/rules/unspecified-encoding/) autofix
[`007b8b3` (#3431)](https://github.com/vega/altair/pull/3431/commits/007b8b3bf9104a175924e83ef11b78d441023b7f)

I'd also never heard of it, which I viewed as a positive - since nobody would ever need to remember/write it. 

Anyway, thanks again really appreciate your time

sphinxext/altairgallery.py
dangotbanned: Fixed in [`e935ad9` (#3431)](https://github.com/vega/altair/pull/3431/commits/e935ad9a6553828d7bd58bc0b5a6021afa579858)

altair/vegalite/v5/api.py
dangotbanned: Fixed in e1ab7667cf29943fd0e38bae14e214b546bf91ba

altair/vegalite/v5/api.py
```diff
@@ -3830,11 +3814,8 @@ def remove_data(subchart):
         if subdata is not Undefined and all(c.data is subdata for c in subcharts):
             data = subdata
             subcharts = [remove_data(c) for c in subcharts]
-    else:
-        # Top level has data; subchart data must be either
```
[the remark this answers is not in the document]
dangotbanned: Fixed in e1ab7667cf29943fd0e38bae14e214b546bf91ba

tools/schemapi/schemapi.py
dangotbanned: > All comments should be preserved. Can you please readd them in all of these cases? Maybe just above the `elif` statement that got added.

This case I handled differently to the others in e1ab7667cf29943fd0e38bae14e214b546bf91ba, so I just wanted to add the context before reverting. 


Before [_prepare_refs](https://github.com/vega/altair/blob/18a2c3c237014591d172284560546a2f0ac1a883/tools/schemapi/schemapi.py#L215)
<details>
<summary>Code Block</summary>

```py
def _prepare_references_in_schema(schema: Dict[str, Any]) -> Dict[str, Any]:
    # Create a copy so that $ref is not modified in the original schema in case
    # that it would still reference a dictionary which might be attached to
    # an Altair class _schema attribute
    schema = copy.deepcopy(schema)

    def _prepare_refs(d: Dict[str, Any]) -> Dict[str, Any]:
        """Add _VEGA_LITE_ROOT_URI in front of all $ref values. This function
        recursively iterates through the whole dictionary."""
        for key, value in d.items():
            if key == "$ref":
                d[key] = _VEGA_LITE_ROOT_URI + d[key]
            else:
                # $ref values can only be nested in dictionaries or lists
                # as the passed in `d` dictionary comes from the Vega-Lite json schema
                # and in json we only have arrays (-> lists in Python) and objects
                # (-> dictionaries in Python) which we need to iterate through.
                if isinstance(value, dict):
                    d[key] = _prepare_refs(value)
                elif isinstance(value, list):
                    prepared_values = []
                    for v in value:
                        if isinstance(v, dict):
                            v = _prepare_refs(v)
                        prepared_values.append(v)
                    d[key] = prepared_values
        return d

    schema = _prepare_refs(schema)
    return schema
```

</details>

After [_prepare_refs](https://github.com/dangotbanned/altair/blob/e935ad9a6553828d7bd58bc0b5a6021afa579858/tools/schemapi/schemapi.py#L215)

<details>
<summary>Code Block</summary>

```py
def _prepare_references_in_schema(schema: Dict[str, Any]) -> Dict[str, Any]:
    # Create a copy so that $ref is not modified in the original schema in case
    # that it would still reference a dictionary which might be attached to
    # an Altair class _schema attribute
    schema = copy.deepcopy(schema)

    def _prepare_refs(d: Dict[str, Any]) -> Dict[str, Any]:
        """Add _VEGA_LITE_ROOT_URI in front of all $ref values.

        This function recursively iterates through the whole dictionary.

        $ref values can only be nested in dictionaries or lists
        as the passed in `d` dictionary comes from the Vega-Lite json schema
        and in json we only have arrays (-> lists in Python) and objects
        (-> dictionaries in Python) which we need to iterate through.
        """
        for key, value in d.items():
            if key == "$ref":
                d[key] = _VEGA_LITE_ROOT_URI + d[key]
            elif isinstance(value, dict):
                d[key] = _prepare_refs(value)
            elif isinstance(value, list):
                prepared_values = []
                for v in value:
                    if isinstance(v, dict):
                        v = _prepare_refs(v)
                    prepared_values.append(v)
                d[key] = prepared_values
        return d

    schema = _prepare_refs(schema)
    return schema
```

</details>

I was weighing up the following:
1. The original location of the comment is no longer a distinct `else` branch
	- Whereas the comment describes both `list` and `dict` together
2. This was the only comment within a nested function
	- Unlike a lot of the other comments, this opened the possibility of moving it into the docstring
	- There doesn't seem to be a clear pattern of using comments instead of docstrings for "private"/ `_functions` in `altair`
		- However, if there were, I would assume it wouldn't extend to inner functions that are not exposed as objects to the user 
3. The code is copied from `tools.schemaapi.schemaapi.py` to [altair.utils.schemaapi.py](https://github.com/dangotbanned/altair/blob/e935ad9a6553828d7bd58bc0b5a6021afa579858/altair/utils/schemapi.py#L217)
	- I think at the time I wasn't sure if this was a literal copy of file contents, or something like [ast](https://docs.python.org/3/library/ast.html) - where the latter would have an understanding of the link between the docstring and function
	- After reading through more of `altair` during this PR, I don't believe there's anything relating to abstract syntax trees - so this may be a moot point

altair/utils/schemapi.py
dangotbanned: See https://github.com/vega/altair/pull/3431#discussion_r1631089964 as that change produces this code.

altair/utils/plugin_registry.py
```diff
@@ -207,8 +207,9 @@ def get(self) -> Optional[PluginType]:
             return self._active
 
     def __repr__(self) -> str:
-        return "{}(active={!r}, registered={!r})" "".format(
-            self.__class__.__name__, self._active_name, list(self.names())
+        return (
+            f"{self.__class__.__name__}(active={self._active_name!r}, registered={list(self.names())!r})"
+            ""
```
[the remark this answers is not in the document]
dangotbanned: > Maybe I'm missing something but I think the syntax was wrong before with the double quotes in front of `.format` as that was just an empty string? So we could now remove the second empty string that got created here. What do you think?

Well spotted @binste!

This was two python quirks in combination:
1. Implicit string concatenation
	- [Related rule](https://docs.astral.sh/ruff/rules/single-line-implicit-string-concatenation/#single-line-implicit-string-concatenation-isc001) but not currently part of `altair`'s rules
2. The empty string effectively being dropped during concatenation

Fixed in [`9073a1f` (#3431)](https://github.com/vega/altair/pull/3431/commits/9073a1f23a0a9d4e779fc066ef721735e090a948)

I also simplified it, temporarily adding these tests to be sure all were equivalent:

<details>
<summary>Code block</summary>

```py
def _original_repr(self: PluginRegistry) -> str:
    return "{}(active={!r}, registered={!r})" "".format(  # noqa: UP032
        self.__class__.__name__, self._active_name, list(self.names())
    )


def _ruff_repr(self: PluginRegistry) -> str:
    return (
        f"{self.__class__.__name__}(active={self._active_name!r}, registered={list(self.names())!r})"
        ""
    )


def _manual_rewrite_repr(self: PluginRegistry) -> str:
    return f"{type(self).__name__}(active={self.active!r}, registered={self.names()!r})"


def test_plugin_registry_repr() -> None:
    plugins = TypedCallableRegistry()
    original = _original_repr(plugins)
    ruff = _ruff_repr(plugins)
    manual_rewrite = _manual_rewrite_repr(plugins)
    assert original == ruff
    assert original == manual_rewrite

```

</details>

sphinxext/altairgallery.py
```diff
@@ -148,67 +149,70 @@
...
 
 
-def save_example_pngs(examples, image_dir, make_thumbnails=True):
+def save_example_pngs(
+    examples: List[Dict[str, Any]], image_dir: Path, make_thumbnails=True
+):
     """Save example pngs and (optionally) thumbnails"""
-    if not os.path.exists(image_dir):
-        os.makedirs(image_dir)
+    encoding = locale.getpreferredencoding(False)
```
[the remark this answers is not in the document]
dangotbanned: Fixed in [`88ec5c8` (#3431)](https://github.com/vega/altair/pull/3431/commits/88ec5c8b6e4b8234ca90c0a0e80d9f7dfe41dd44)

altair/utils/data.py
```diff
@@ -189,9 +187,8 @@ def to_json(
     data_json = _data_to_json_string(data)
     data_hash = _compute_data_hash(data_json)
     filename = filename.format(prefix=prefix, hash=data_hash, extension=extension)
-    with open(filename, "w") as f:
-        f.write(data_json)
-    return {"url": os.path.join(urlpath, filename), "format": {"type": "json"}}
+    Path(filename).write_text(data_json, encoding=locale.getpreferredencoding(False))
```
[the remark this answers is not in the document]
dangotbanned: Fixed in [`88ec5c8` (#3431)](https://github.com/vega/altair/pull/3431/commits/88ec5c8b6e4b8234ca90c0a0e80d9f7dfe41dd44)

altair/utils/_vegafusion_data.py
```diff
@@ -125,11 +126,9 @@ def get_inline_tables(vega_spec: dict) -> Dict[str, DataFrameLike]:
     table_names = get_inline_table_names(vega_spec)
     tables = {}
     for table_name in table_names:
-        try:
+        # otherwise, named dataset that was provided by the user
+        with contextlib.suppress(KeyError):
```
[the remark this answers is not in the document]
dangotbanned: @binste Not a problem!

Fixes: 
- [`de01aa5` (#3431)](https://github.com/vega/altair/pull/3431/commits/de01aa5fae14f77af853e822951b7f1f670a7547)
- [`80a7a22` (#3431)](https://github.com/vega/altair/pull/3431/commits/80a7a223fa30157acd92928bf061460a750dae67)
- [`b96c4e9` (#3431)](https://github.com/vega/altair/pull/3431/commits/b96c4e938e67d508229ad85517bf630d91c73dd0)

Apologies if I went overboard with the descriptions 😅

tests/utils/test_utils.py
```diff
@@ -121,7 +122,10 @@ def test_sanitize_dataframe_arrow_columns():
 
 
 @pytest.mark.skipif(pa is None, reason="pyarrow not installed")
-def test_sanitize_pyarrow_table_columns():
+@pytest.mark.skipif(
+    sys.platform == "win32", reason="Timezone database is not installed on Windows"
```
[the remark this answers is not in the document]
mattijn: Windows requires a user-provided database for timezones and it should be located in the proper place. I had issues with this in the past in relation to pyarrow and had a small discussion on this topic in this issue: https://github.com/apache/arrow/issues/35600. The situation is improved and is now documented within the pyarrow docs here: https://arrow.apache.org/docs/python/install.html#tzdata-on-windows.

To install the timezone on windows, use `pyarrow.util.download_tzdata_on_windows()`, by default it is placed in `%USERPROFILE%\Downloads\tzdata` which can be changed by using `pa.set_timezone_db_path("custom_path")` (`import pyarrow as pa`).

mattijn:
> If any other maintainer has strong preferences on the rule set which is applied, please chime in!

No strong opinion, I think it is a very good thing to make the code base more readable and these auto-format fixes are very useful for this. I don't have a good understanding of the details of each rule, but checking the final code-diff these changes are sensible. 

One comment. I notice this pattern change:
```python
foo in ("bar", "baz")
```
Or 
```python
foo in ["bar", "baz"]
```
To this:
```python
foo in {"bar", "baz"}
```

My personal preference would be to use a list here. Is there more documentation on this, why this rule prefers `{`, `}`?

dangotbanned:
> > If any other maintainer has strong preferences on the rule set which is applied, please chime in!
> 
> No strong opinion, I think it is a very good thing to make the code base more readable and these auto-format fixes are very useful for this. I don't have a good understanding of the details of each rule, but checking the final code-diff these changes are sensible.
> 
> One comment. I notice this pattern change:
> 
> ```python
> foo in ("bar", "baz")
> ```
> 
> Or
> 
> ```python
> foo in ["bar", "baz"]
> ```
> 
> To this:
> 
> ```python
> foo in {"bar", "baz"}
> ```
> 
> My personal preference would be to use a list here. Is there more documentation on this, why this rule prefers `{`, `}`?

@mattijn This comes from [PLR6201](https://docs.astral.sh/ruff/rules/literal-membership/), which is an optimization available since python 3.2.
Generally, when all the elements are: known, unique and hashable - `set` would be preferred.

Cases for other `Sequence` types more generally:
- `tuple` should be used when there are no mutability requirements: elements may not be hashable; there is a need preserve ordering
- `list` should be used when you need mutability and access to arbitrary indexes
- `deque` should be used when you need mutability, but only access from the start/end

pyproject.toml
```diff
@@ -104,8 +105,20 @@ artifacts = ["altair/jupyter/js/index.js"]
 
 [tool.hatch.envs.default]
 features = ["all", "dev"]
+# https://hatch.pypa.io/latest/how-to/environment/select-installer/#enabling-uv
+installer = "uv"
 
 [tool.hatch.envs.default.scripts]
+generate-schema-wrapper = [
```
dangotbanned, line 112: TODO
----
Adapt this into `generate_schema_wrapper.py` so the test can pass CI.

dangotbanned:
Most recent [`9582971` (#3431)](https://github.com/vega/altair/pull/3431/commits/958297128bb3912ad5b730477b9fd48d380dfcd1)

```
The code generation modified the following files:
altair/vegalite/v5/schema/channels.py
```
Previous [`36da1c7` (#3431)](https://github.com/vega/altair/pull/3431/commits/36da1c79ba5aacfa4a8e176668b22ce7d647b92e)

```
The code generation modified the following files:
altair/vegalite/v5/schema/channels.py altair/vegalite/v5/schema/core.py altair/vegalite/v5/schema/mixins.py
```

Seems to have resolved all but `channels.py`.

I've tried to reproduce locally, but currently not showing any diff:

<details>
<summary>Local run</summary>

```ps
>>> hatch run generate-schema-wrapper
cmd [1] | mypy tools/schemapi/schemapi.py
Success: no issues found in 1 source file
cmd [2] | python tools/generate_schema_wrapper.py
Copying
 C:\Users\.\Documents\GitHub\altair\tools\schemapi\schemapi.py
  -> C:\Users\.\Documents\GitHub\altair\tools\..\altair\utils\schemapi.py
1 file reformatted
Writing C:\Users\.\Documents\GitHub\altair\altair\vegalite\v5\schema\__init__.py
All checks passed!
All checks passed!
1 file reformatted
Generating
 C:\Users\.\Documents\GitHub\altair\altair\vegalite\v5\schema\vega-lite-schema.json
  ->C:\Users\.\Documents\GitHub\altair\altair\vegalite\v5\schema\core.py
Found 6061 errors (6061 fixed, 0 remaining).
Found 1 error (1 fixed, 0 remaining).
1 file reformatted
Generating
 C:\Users\.\Documents\GitHub\altair\altair\vegalite\v5\schema\vega-lite-schema.json
  ->C:\Users\.\Documents\GitHub\altair\altair\vegalite\v5\schema\channels.py
Found 5810 errors (5810 fixed, 0 remaining).
Found 1 error (1 fixed, 0 remaining).
1 file reformatted
Generating
 C:\Users\.\Documents\GitHub\altair\altair\vegalite\v5\schema\vega-lite-schema.json
  ->C:\Users\.\Documents\GitHub\altair\altair\vegalite\v5\schema\mixins.py
Found 3340 errors (3340 fixed, 0 remaining).
Found 1 error (1 fixed, 0 remaining).
1 file reformatted
Updating API docs
  ->C:\Users\.\Documents\GitHub\altair\doc\user_guide\api.rst
1 file reformatted
cmd [3] | ruff check .
All checks passed!
cmd [4] | ruff format --diff --check .
83 files already formatted
cmd [5] | mypy altair tests
Success: no issues found in 360 source files
cmd [6] | python -m pytest --pyargs --doctest-modules tests altair
====================================== test session starts ================================
platform win32 -- Python 3.11.9, pytest-8.2.0, pluggy-1.5.0
rootdir: C:\Users\.\Documents\GitHub\altair
configfile: pyproject.toml
plugins: anyio-4.3.0, cov-5.0.0, xdist-3.6.1
collected 410 items

...
========== 406 passed, 2 skipped, 2 xpassed, 55 warnings in 137.34s (0:02:17) =============
```

</details>

pyproject.toml
dangotbanned: See:
- https://github.com/vega/altair/pull/3431#issuecomment-2160849568
- [`9582971` (#3431)](https://github.com/vega/altair/pull/3431/commits/958297128bb3912ad5b730477b9fd48d380dfcd1)

pyproject.toml
```diff
@@ -164,22 +164,85 @@ exclude = [
 ]
 
 [tool.ruff.lint]
-extend-safe-fixes=["SIM101"]
+preview = true
```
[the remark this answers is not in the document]
dangotbanned: # Autofix enabled with preview
- 3171a3d12d440f5894573db864ed2d1223bd9647
- 5ba210d53d68955e95625f04bd3f736ee3f89673
- 8ef6184d710aec28788cf0a940eee27c9d18eb8a

# All/mostly preview rules
- a16b2920dea9f9257ee86cfb4c8744bd24d40ce6
- 7fff541d8aeb816ff12ddfa5a9a8b5369c91e665
- 1e175338ecd2b6e322baf92ed4080a622443a27c
- 007b8b3bf9104a175924e83ef11b78d441023b7f
- 10eb5d5161a4c7edee192510350c55cc81d6b536

# Current
[pyproject.toml](https://github.com/dangotbanned/altair/blob/33056a2a85b1126c6bbb26e4ca22dab2118fbf8f/pyproject.toml#L198)

I've linked each of the related commits and the most recent config, which now contains links to the relevant `ruff` docs. 
The main setting to note is `extend-select`. Over the course of this PR I've reduced the total number of preview rules to 5 from [`pylint`](https://docs.astral.sh/ruff/rules/#pylint-pl) and all of [`refurb`](https://docs.astral.sh/ruff/rules/#refurb-furb)

dangotbanned:
> Regarding the futures type annotation import, I'm not too familiar with the consequences of enabling that but it seems to be a highly debated topic in the Python community, see our own previous discussion at [#2950 (comment)](https://github.com/vega/altair/pull/2950#issuecomment-1473660066) and [python/cpython#82786](https://github.com/python/cpython/issues/82786) where it's mentioned that there are now other PEPs to treat this differently. I'd rather not introduce it if we're not fully aware of the consequences, especially if it's only for cosmetic reasons.

@binste just in case you see this before I've finished writing up my overview: commits [`c378b65` (#3431)](https://github.com/vega/altair/pull/3431/commits/c378b65edd93265dbc13fadd11d695bbf9c4744e) through to [`33056a2` (#3431)](https://github.com/vega/altair/pull/3431/commits/33056a2a85b1126c6bbb26e4ca22dab2118fbf8f) document the implementation of `from __future__ import annotations`. 

I started out trying to find examples of incompatible code that would prevent `altair` from using it, but found lots of cases where it would be beneficial and remove the need for working around circular imports. 

The main takeaway is due to the code generation working entirely with strings (and not the types themselves) this was achievable without losing any functionality.

**[Placeholder while writing up overview ...]**

mattijn:
I was reading:
- https://peps.python.org/pep-0649/

And was wondering, what is the impact of [this paragraph](https://peps.python.org/pep-0649/#changes-to-allowable-annotations-syntax) for us?

> `__annotate__` now delays the evaluation of annotations until `__annotations__` is referenced in the future. It also means annotations are evaluated in a new function, rather than in the original context where the object they were defined on was bound. There are four operators with significant runtime side-effects that were permitted in stock semantics, but are disallowed when `from __future__ import annotations` is active, and will have to be disallowed when this PEP is active:
> 
> - `:=`
> - `yield`
> - `yield from`
> - `await`


And after reading:
- https://discuss.python.org/t/pep649-means-that-pep563-will-see-more-usage-not-less-and-will-break-runtime-typecheckers/47658/10 and 
- https://peps.python.org/pep-0749/

I hope we can align ourselves with PEP 649, and not so much with PEP 563, but is that even possible with minimal python support of 3.9?

tools/schemapi/schemapi.py
```diff
@@ -741,6 +742,18 @@ def __repr__(self) -> str:
 
 
 Undefined = UndefinedType()
+T = TypeVar("T")
+Optional: TypeAlias = Union[T, UndefinedType]
```
mattijn, line 746: This is nice!

dangotbanned:
> I hope we can align ourselves with PEP 649, and not so much with PEP 563, but is that even possible with minimal python support of 3.9?

@mattijn I was confused reading that just now, but the end of [PEP 749](https://peps.python.org/pep-0749/#which-expressions-can-be-stringified) states:
> The following are disallowed in annotation scopes and therefore not relevant:
    NamedExpr (:=)
    Await
    Yield
    YieldFrom

I believe the more recent one makes it clearer, but this is stating what is allowed within the annotation itself.

So something like this is not valid:
```py
ComplexAlias: (tp := getattr(variable, "type_from_somewhere")) = value
```

[PEP 695](https://peps.python.org/pep-0695/#type-parameter-scopes), although using syntax from 3.12, gives some examples of invalid expressions in annotation scope:

```py
class ClassA[T]((x := Sequence[T])): ...  # Syntax error: assignment expression not allowed

def func1[T](val: (x := int)): ...  # Syntax error: assignment expression not allowed

def func2[T]() -> (x := Sequence[T]): ...  # Syntax error: assignment expression not allowed

type Alias1[T] = (x := list[T])  # Syntax error: assignment expression not allowed
```

I hope that all makes sense. 
The main point I wanted to get across by linking PEP 749, is that introducing `from __future__ import annotations` now will have predicatable behaviour [until October 2029](https://peps.python.org/pep-0719/#lifespan) - as it seemed there was some hesitancy based on the number of PEPs/potential conflict between them.

tools/schemapi/schemapi.py
dangotbanned: As a comparison, see the gradual improvements in `MarkMethodMixin.mark_arc`:

- [main](https://github.com/vega/altair/blob/18a2c3c237014591d172284560546a2f0ac1a883/altair/vegalite/v5/schema/mixins.py#L22)
- [`|` Operator introduced](https://github.com/vega/altair/blob/36da1c79ba5aacfa4a8e176668b22ce7d647b92e/altair/vegalite/v5/schema/mixins.py#L27)
- [`Optional` introduced](https://github.com/vega/altair/blob/fbc02e2d577a46509436ae6f40b9b56035e996c8/altair/vegalite/v5/schema/mixins.py#L31)

There's still some low-hanging fruit with a future addition aliases for all the long `Literal` types, but even as-is, these changes have reduced ~9000 lines.

mattijn:
Understood, also what I realized. 
If we add a comment in the code that this annotations import from `__future__` can be removed when Python 3.14 is the minimum supported version, I'm fine with this!

Also reassuring that we can still use these other mentioned operators. Your clarification was helpful for this.

tools/schemapi/schemapi.py
dangotbanned: > There's still some low-hanging fruit with a future addition aliases for all the long `Literal` types, but even as-is, these changes have reduced ~9000 lines.

`Literal` -> `TypeAlias` added in [`a7b0ab5` (#3431)](https://github.com/vega/altair/pull/3431/commits/a7b0ab58b5f0e24b996d44f3e81b5235600defa1)

dangotbanned:
> Edit: Or
> 
> ```python
> import sys 
> if sys.version_info < (3, 14):
>     from __future__ import annotations
> ```

@mattijn I'll try to find the link tomorrow, but this wouldn't be possible because the `from __future_ ...` is only valid as the first (non-comment/doc) line in the file. 

iirc, in PEP749 somewhere it has a mention. 

I think a comment as a warning could work, albeit quite forward-thinking. 
In 5 years `ruff` might not be the tool, but this sort of thing would eventually be part of the `UP`, `TCH` or `FA` groups I'd imagine

tests/utils/test_utils.py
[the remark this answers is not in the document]
dangotbanned: Possibly resolved in 65e3279487e0f625b74b620e60e8d4e612030286

dangotbanned:
Some very marginal reductions in `lint` workflow duration as of 330f638f934ef121d044ddce4a4e5ec84e75b2e3

Just scrolling through [previous runs](https://github.com/vega/altair/actions/workflows/lint.yml?query=is%3Asuccess) all seem to be 60-80s with this first down to 43s.

What I'm aiming to for is utilising parallism provided by https://github.com/dangotbanned/altair/blob/330f638f934ef121d044ddce4a4e5ec84e75b2e3/pyproject.toml#L135, as this alone speeds up `pytest` runs a lot. 

I've done some work locally on distributing the existing tests better - for comparison: 

# Prior to using `pytest-xdist`
```log
=========================== test session starts ===============================
platform win32 -- Python 3.11.9, pytest-8.2.0, pluggy-1.5.0
rootdir: C:\Users\*\*\altair
configfile: pyproject.toml
plugins: anyio-4.3.0, cov-5.0.0
collected 403 items

tests\expr\test_expr.py .............                                    [  3%]
...
=========== 401 passed, 2 skipped, 55 warnings in 138.85s (0:02:18) ===========
```

# With `pytest-xdist` + optimizing slow tests
```log
=========================== test session starts ===============================
platform win32 -- Python 3.11.9, pytest-8.2.0, pluggy-1.5.0
rootdir: C:\Users\*\*\altair
configfile: pyproject.toml
plugins: anyio-4.3.0, cov-5.0.0, xdist-3.6.1
24 workers [1244 items]   
........................................................................ [ 13%]

=========== 1240 passed, 1 skipped, 2 xfailed, 1 xpassed in 29.48s ============
```

**138.85s -> 29.48s**: CPU: AMD Ryzen 9 3900X 12-Core/24-Thread, 3.8/4.1 GHz

Not expecting as much on a 4-Core workflow, but should speed things up somewhat

dangotbanned:
Not sure what `pip` is up to here during install https://github.com/vega/altair/actions/runs/9521446923/job/26248868264

But `pytest` is now running in parallel at least 😄

tools/schemapi/utils.py
```diff
@@ -22,6 +28,133 @@
 }
 
 
+class _TypeAliasTracer:
```
dangotbanned, line 31: Thinking this will be the main topic for review, due to its contribution to the diff.

The obvious choices for a postfix on the `TypeAlias` name would create some odd results. In [this selection](https://github.com/dangotbanned/altair/blob/d0c01eb6259c4c79cce42add2bcc8bb3d1ec41ed/altair/vegalite/v5/schema/_typing.py#L502-L546) you can see some examples that make it difficult to apply a common postfix, without causing a collision. 

Below I've included 4 others which were an issue. 
An alternative argument to `_TypeAliasTracer(fmt=...)` would need to make sense across all of these: 

| Input | Output |
|--------|--------|
| `MarkType` | `MarkType_T` |
| `Mark` | `Mark_T` |
| `RangeEnum` | `RangeEnum_T`  |
| `ProjectionType` | `ProjectionType_T` |
| `ResolveMode` | `ResolveMode_T` | 
| `ScaleInterpolateEnum` | `ScaleInterpolateEnum_T` | 
| `ScaleType` | `ScaleType_T` | 
| `SelectionResolution` | `SelectionResolution_T` | 
| `SelectionType` | `SelectionType_T` | 
| `TypeForShape` | `TypeForShape_T` | 
| `Type` | `Type_T` |

dangotbanned:
@mattijn would @binste be the only maintainer who could review this PR?

I ask because I have 2 (more focused) topics in mind that would make sense as separate PRs. 
However I'm hesitant to open any more, beyond this and #3427, to avoid generating a lot of merge conflicts. 

As of d0c01eb6259c4c79cce42add2bcc8bb3d1ec41ed this is up-to-date with `main`, but I'm going to need to manually resolve conflicts and re-run with the new `ruff` config for each new commit before this is merged. 

Would really appreciate any advice, if there is a way I can manage this better

mattijn:
@binste is the most capable as he spend a lot of time to make Altair a typed library. 

If you keep the PRs small then they are easier to review and has higher potential to be accepted sooner.

Is this PR ready for review then?

dangotbanned:
@mattijn 
> @binste is the most capable as he spend a lot of time to make Altair a typed library.
> If you keep the PRs small then they are easier to review and has higher potential to be accepted sooner.

Yeah that makes sense thank you

> Is this PR ready for review then?

Definitely ready. I did [re-request a review](https://github.com/vega/altair/pull/3431#event-13137880436) last week, but wanted to check first before pinging anyone else - as I suspected I may have to wait until @binste is available. 

No problem with waiting, just wanted to check in and make sure I understood the status. Thanks again for the feedback on all my PRs so far.

binste:
Sorry, just moved apartments which created a lot of new tasks which fill my spare time :) But I'll try to give this a review soon. Would appreciate the opportunity to have a look before merging and thanks for your patience and work on this!

dangotbanned:
> Sorry, just moved apartments which created a lot of new tasks which fill my spare time :) But I'll try to give this a review soon. Would appreciate the opportunity to have a look before merging and thanks for your patience and work on this! Fully understand that you ask the question, good to be aligned.

@binste Good to hear from you, but absolutely prioritze real-life over this 😄. Happy to continue with this one when you're free, thanks again

tools/generate_schema_wrapper.py
```diff
@@ -193,16 +160,19 @@ class DatumChannelMixin:
...
     ) -> dict:
         context = context or {}
         ignore = ignore or []
-        datum = self._get("datum", Undefined)  # type: ignore[attr-defined]
+        datum = self._get("datum", Undefined)  # type: ignore[attr-defined] # noqa
         copy = self  # don't copy unless we need to
-        if datum is not Undefined:
-            if isinstance(datum, core.SchemaBase):
-                pass
+        # NOTE: The block below does nothing?
```
binste, line 170: This is a PR comment and should not be part of the code.

And fully agree, that code can be removed.

tools/schemapi/schemapi.py
binste: I do like the use of a `TypeVar` for this but I don't know if we should name it `Optional` as it already has a different meaning which is well known to users. If I read `Optional` in my IDE, I expect that I can pass `None` and I might write code for it. What do you think? I can't come up with a better name for it right now...

I really like the literals to reduce the code!

binste (changes requested):
This PR grew quite big and I struggle to review it fully so I just did some spotchecks. I think that's fine given the test coverage and linters. I'm sometimes guilty of the same crime and let PRs get quite big... ;) That being said, I really appreciate all the work you put into this, I'm aware that it was probably challenging to keep all PRs in sync, and I think it's a great improvement to the quality of the code base.

I've only added some remarks which should be quick to resolve. The most important one might be the naming of `Optional` which worries me a bit and I hope we can find a better name there. After that, I'm ok to get this into the code base. Also ok with the `__future__` import, thanks for the research on it.

.github/workflows/build.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         run: |
           python tools/generate_schema_wrapper.py
           # This gets the paths of all files which were either deleted, modified
-          # or are not yet tracked by Git
+          # or are not yet tracked by Git
```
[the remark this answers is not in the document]
dangotbanned: @binste thanks for spotting

I can't remember the exact issue, but I needed to force a diff to happen here to resolve some conflict locally.

dangotbanned:
@binste I can't seem to be able to reply to this [comment](https://github.com/vega/altair/pull/3431#discussion_r1657123165)

> I do like the use of a `TypeVar` for this but I don't know if we should name it `Optional` as it already has a different meaning which is well known to users. If I read `Optional` in my IDE, I expect that I can pass `None` and I might write code for it. What do you think? I can't come up with a better name for it right now...

I do agree with this sentiment, and when I first added the alias in https://github.com/vega/altair/commit/f1d9dc2cf35a1aaebc19415182dfe7393fd82a90 I used the name `_AltOptional` for this exact reason. 

By the time I added it to this branch, I'd changed my mind and added the reasoning to the [commit message](https://github.com/vega/altair/commit/fbc02e2d577a46509436ae6f40b9b56035e996c8):
> `typing.Optional` is deprecated and is no longer needed in `altair`. The new alias improves the readability of generated signatures, by reducing noise (repeating `UndefinedType`) and clearly grouping together the options for an argument.
https://typing.readthedocs.io/en/latest/spec/historical.html#union-and-optional

I'm not sure about all IDEs, but for VSCode the "dangling" docstring will show on hover. 

Just spotted I need to update the example since https://github.com/microsoft/pylance-release/releases/tag/2024.6.100 added sphinx support. 

image

> I expect that I can pass `None`

That is fair. Although say for the example above, the arguments accepting `None` are already differentiated. 
A user with a type-checker will see an error for `None` on `aggregate`, `bandPosition` - but not `bin`, `field`, `header`.


I am open to an alternative, but the semantics of this alias and `typing.Optional` are so similar that I felt we could reuse the name - as long as `typing.Optional` appeared nowhere in `altair` - to avoid any confusion.


> I really like the literals to reduce the code!

Thank you. I was really suprised at how much more readable it made the signatures, especially those with multiple literals.

binste:
Valid points. I can't come up with a better name either so let's give it a try like this.

I'll merge this in to free up other PRs. Refinements such as updating examples can happen in a follow-up PR.

binste (approved):

dangotbanned:
Just seen this https://github.com/astral-sh/ruff/releases/tag/0.5.0

@binste thanks for reviewing & merging. 

I'll continue in another PR as suggested

PR #3444, opened 2024-06-20
perf: Fix issues with `Chart|LayerChart.encode`, 1.32x speedup to `infer_encoding_types`

dangotbanned:
## Background
I stumbled across this issue in the docs during 80a08122ad1e1decdd0986e25a852f928f743e1d and went down a bit of a rabbit hole

### Before
image

### After
image

## Fixes
- [Sphinx warning](https://altair-viz.github.io/user_guide/generated/toplevel/altair.Chart.html#altair.Chart) on `Chart.encode`. Also incorrectly under `Attributes` section
- Preserve static typing previously found in `_encode_signature` but lost after `_EncodingMixin.encode`
  - Re-running `mypy` output 'Found 63 errors in 47 files (checked 360 source files)', tests/examples

## Perf
- This was a response to the `TODO` left at the top of `infer_encoding_types`

https://github.com/vega/altair/blob/76a9ce1e27a21359e7b0ef9e6cb865f8d95fc298/altair/utils/core.py#L785-L795

### Benchmark

<details>
<summary>Code block</summary>

```py
import altair as alt
from altair import Undefined
from altair.utils.core import infer_encoding_types
import random

def args_kwds():
	# Generating 40 mock channels
    tps = [
        alt.Angle,
        alt.Color,
        alt.Column,
        alt.Description,
        alt.Detail,
        alt.Facet,
        alt.Fill,
        alt.FillOpacity,
        alt.Href,
        alt.Key,
        alt.Latitude,
        alt.Latitude2,
        alt.Longitude,
        alt.Longitude2,
        alt.Opacity,
        alt.Order,
        alt.Radius,
        alt.Radius2,
        alt.Row,
        alt.Shape,
        alt.Size,
        alt.Stroke,
        alt.StrokeDash,
        alt.StrokeOpacity,
        alt.StrokeWidth,
        alt.Text,
        alt.Theta,
        alt.Theta2,
        alt.Tooltip,
        alt.Url,
        alt.X,
        alt.X2,
        alt.XError,
        alt.XError2,
        alt.XOffset,
        alt.Y,
        alt.Y2,
        alt.YError,
        alt.YError2,
        alt.YOffset,
    ]
    random.shuffle(tps)
    split = random.randint(0, 40)
	# when only keyword-args are passed, a slightly faster branch can be used
    pos_only = tps[:split]
    kwd_only = tps[split:]
    args = [arg("field_name") for arg in pos_only]
    kwargs = {kwd._encoding_name: kwd("field_name") for kwd in kwd_only}
    return args, kwargs


def _infer_new(
    *args,
    self=None,  # Just to include the time it takes to pop
				# Would be the implicit self of Chart.encode
    angle=Undefined,
    color=Undefined,
    column=Undefined,
    description=Undefined,
    detail=Undefined,
    facet=Undefined,
    fill=Undefined,
    fillOpacity=Undefined,
    href=Undefined,
    key=Undefined,
    latitude=Undefined,
    latitude2=Undefined,
    longitude=Undefined,
    longitude2=Undefined,
    opacity=Undefined,
    order=Undefined,
    radius=Undefined,
    radius2=Undefined,
    row=Undefined,
    shape=Undefined,
    size=Undefined,
    stroke=Undefined,
    strokeDash=Undefined,
    strokeOpacity=Undefined,
    strokeWidth=Undefined,
    text=Undefined,
    theta=Undefined,
    theta2=Undefined,
    tooltip=Undefined,
    url=Undefined,
    x=Undefined,
    x2=Undefined,
    xError=Undefined,
    xError2=Undefined,
    xOffset=Undefined,
    y=Undefined,
    y2=Undefined,
    yError=Undefined,
    yError2=Undefined,
    yOffset=Undefined,
):
    kwargs = locals()
    kwargs.pop("self")
    args = kwargs.pop("args")
    if args:
		# now required to remove any collisions 
		# as all have a default
        kwargs = {k: v for k, v in kwargs.items() if v is not Undefined}
    return infer_encoding_types(args, kwargs)


def test_infer():
    args, kwds = args_kwds()
    return _infer_new(*args, **kwds)

def test_infer_old():
	# This is still benefitting from all non-cache changes to `infer_encoding_types`
    args, kwds = args_kwds()
    return infer_encoding_types(args, kwds, alt.channels)

```

</details>

```py
# %%timeit -n 100000
>>> test_infer()
154 µs ± 165 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

# %%timeit -n 100000
>>> test_infer_old()
204 µs ± 256 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
*1.32x speedup*


altair/utils/core.py
```diff
@@ -767,7 +769,132 @@ def display_traceback(in_ipython: bool = True):
...
+            cached.name_to_channel = _invert_group_channels(cached.channel_to_name)
+            _CHANNEL_CACHE = cached
+        return _CHANNEL_CACHE
+
+    def get_encoding(self, tp: Type[Any], /) -> str:
+        if encoding := self.channel_to_name.get(tp):
+            return encoding
+        raise NotImplementedError(f"positional of type {type(tp).__name__!r}")
+
+    def _wrap_in_channel(self, obj: Any, encoding: str, /):
```
mattijn, line 817: It is the first time I see a forward slash, `/` as argument within a function. Can you explain what that does?

mattijn:
Much appreciated for this PR @dangotbanned! I cannot really oversee the implications of this PR, but if it solves a todo within the code, resolves the docs issue and introduces a nice performance bump, than I'm really happy you went into that rabbit hole... and came out back again:)

All tests are happy already, but if you have a suggestion how I could review this better than looking carefully to the code-diff, that would be much appreciated.

Thanks again for this PR!

altair/utils/core.py
dangotbanned: Sure @mattijn 

`/` is a marker for [positional-only parameters](https://peps.python.org/pep-0570/#benefits-of-positional-only-parameters). Any parameter to the left of `/` is positional-only, and may not be used by name.
In this case, calling `self._wrap_in_channel(encoding="enc", obj=[1,2,3])` raises a `TypeError`. The [python tutorial docs](https://docs.python.org/3/tutorial/controlflow.html#special-parameters) may be helpful for an overview as well. 

Personally, I like to use `/` in cases like:
- The function/method is not part of the public API, to leave more flexibility in renaming parameters in the future, without introducing a breaking change
- There are a 1-3 parameters
- The function is *currently* used in one specific way and the function name and parameter order are clear
	- In this case, `_wrap_in_channel` could logically be thought of as having parameters `_wrap_in_channel(wrappee, wrapper)`.

Looking back at PEP570, I see that this was introduced in `python3.8`, which may explain the feature's absence in `altair` until now.

Hope all of that was helpful

altair/utils/core.py
```diff
@@ -767,7 +769,132 @@ def display_traceback(in_ipython: bool = True):
...
+Initialized on first use.
+"""
+
+
+class _ChannelCache:
+    channel_to_name: Dict[Type[SchemaBase], str]
+    name_to_channel: Dict[str, Dict[_ChannelType, Type[SchemaBase]]]
+
+    @classmethod
+    def from_channels(cls, channels: ModuleType, /) -> "_ChannelCache":
```
dangotbanned, line 785: As stated in the comments, I've kept this for an existing test https://github.com/dangotbanned/altair/blob/e4ab7052e9a1e62ff1fd80379864489c69a1e020/tests/utils/test_core.py#L263

I would prefer it if this was removed and `test_infer_encoding_types` was re-written using actual objects from `channels`.

altair/utils/core.py
```diff
@@ -767,7 +769,132 @@ def display_traceback(in_ipython: bool = True):
...
+
+
+def _init_channel_to_name():
+    """
+    Construct a dictionary of channel type to encoding name.
+
+    Note
+    ----
+    The return type is not expressible using annotations, but is used
+    internally by `mypy`/`pyright` and avoid the need for type ignores.
```
dangotbanned, line 852:
```suggestion
    internally by `mypy`/`pyright` and avoids the need for type ignores.
```

altair/vegalite/v5/api.py
```diff
@@ -786,15 +799,33 @@ def binding_range(**kwargs):
     return core.BindRange(input="range", **kwargs)
 
 
+_TSchemaBase = typing.TypeVar("_TSchemaBase", bound=core.SchemaBase)
+
+
+@typing.overload
```
dangotbanned, line 805: These overloads originate from https://github.com/dangotbanned/altair/blob/e159ea08f970e3fd388044cf07cb670987f6f967/altair/vegalite/v5/api.py#L1479

They fully encpasualte the complex internal logic of `condition`, which on this branch is now required due to `Chart.encode` being understood by type checkers.

A very simple example of `encode` not being understood, is the correction required from `Color` -> `Fill` here https://github.com/dangotbanned/altair/blob/e4ab7052e9a1e62ff1fd80379864489c69a1e020/tests/examples_arguments_syntax/hexbins.py#L33

Relating back to `condition`, this message:
> Re-running `mypy` output 'Found 63 errors in 47 files (checked 360 source files)', tests/examples

Almost all of these were `condition`, which are now **not** emitting errors - entirely due to the new overloads. I didn't need to add any "# type: ignore" comments in the examples, which would have been a less desirable alternative.

altair/vegalite/v5/api.py
```diff
@@ -3629,20 +3643,20 @@ def transformed_data(
 
         return transformed_data(self, row_limit=row_limit, exclude=exclude)
 
-    def __iadd__(self, other: Union[core.LayerSpec, core.UnitSpec]) -> Self:
+    def __iadd__(self, other: Union["LayerChart", Chart]) -> Self:
```
dangotbanned, line 3646: All of these seem to have been ignored by type checkers not understanding `.encode`, which like `condition` is now fixed

dangotbanned:
> Much appreciated for this PR @dangotbanned! I cannot really oversee the implications of this PR, but if it solves a todo within the code, resolves the docs issue and introduces a nice performance bump, than I'm really happy you went into that rabbit hole... and came out back again:)
> 
> All tests are happy already, but if you have a suggestion how I could review this better than looking carefully to the code-diff, that would be much appreciated.
> 
> Thanks again for this PR!

Thanks @mattijn 

I've marked up some comments now with some additional info that may be helpful. 
Other than that, if you go through the commits sequentially - you can see specific `mypy` fixes, which help explain some changes that may have appeared unrelated.

tests/utils/test_schemapi.py
```diff
@@ -648,6 +633,11 @@ def chart_error_example__four_errors():
     )
 
 
+# NOTE: These 6 `TypeError`s are due to the removal of `kwargs` in Chart.encode
+# That is, only declared keywords can now be used in Chart.encode,
+# so these are not conditions worth checking against.
```
binste, line 638: I believe this NOTE is only relevant for the PR review so it can be removed from the code.

binste (changes requested):
Nice catch! Wasn't aware that the signature for `encode` does not work and the new implementation is definitely more robust with the explicit declaration of the kwargs. I'm also glad you went into that rabbit hole ;)

I added some smaller comments. Afterwards, I think this is ready to be merged.

altair/utils/core.py
```diff
@@ -767,7 +769,132 @@ def display_traceback(in_ipython: bool = True):
...
+
+    Returns
+    -------
+        mapping: dict[type[`<subclass of FieldChannelMixin and SchemaBase>`] | type[`<subclass of ValueChannelMixin and SchemaBase>`] | type[`<subclass of DatumChannelMixin and SchemaBase>`], str]
+    """
+    from altair.vegalite.v5.schema import channels as ch
+
+    mixins = ch.FieldChannelMixin, ch.ValueChannelMixin, ch.DatumChannelMixin
+
+    return {
```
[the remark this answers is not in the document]
dangotbanned: @binste I'll be commenting on 788+ after, but this one is a bit quicker. 

In [generate_schema_wrapper](https://github.com/dangotbanned/altair/blob/e4ab7052e9a1e62ff1fd80379864489c69a1e020/tools/generate_schema_wrapper.py#L102) I added the common class attribute `_encoding_name` to the Mixin classes. 

The [previous implementation](https://github.com/vega/altair/blob/6b5a867983c3f8fb0407bc44778e0b9c1ee21d76/altair/utils/core.py#L795) couldn't be typed accurately because `SchemaBase` does not have `_encoding_name` - but all descendents `mixins = ch.FieldChannelMixin, ch.ValueChannelMixin, ch.DatumChannelMixin` do.

I'll go into the distinction between `from_cache`/`from_channels` in more detail on the other comment.

tests/utils/test_core.py
```diff
@@ -260,6 +260,7 @@ def _getargs(*args, **kwargs):
     return args, kwargs
 
 
+# NOTE: Dependent on a no longer needed implementation detail
```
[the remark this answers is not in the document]
dangotbanned: @binste I wouldn't expect this to work for `channels=None`.

The default option would be to use the definitions from `channels.py` which wouldn't include those defined in `FAKE_CHANNELS_MODULE`.

I suppose we could test that this should fail under the condition of:
- objects external to `channels.py` are used
- they cannot be wrapped using existing logic

altair/utils/core.py
```diff
@@ -785,68 +912,18 @@ def infer_encoding_types(args: Sequence, kwargs: MutableMapping, channels: Modul
         All args and kwargs in a single dict, with keys and types
         based on the channels mapping.
     """
-    # Construct a dictionary of channel type to encoding name
-    # TODO: cache this somehow?
-    channel_objs = (getattr(channels, name) for name in dir(channels))
-    channel_objs = (
-        c for c in channel_objs if isinstance(c, type) and issubclass(c, SchemaBase)
+    cache = (
```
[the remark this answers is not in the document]
dangotbanned: From my understanding, the only instance where `infer_encoding_types` is ever called (excluding tests) is within the `.encode` method.

In this PR, I have updated that call to not pass a `channels` argument https://github.com/dangotbanned/altair/blob/e4ab7052e9a1e62ff1fd80379864489c69a1e020/tools/generate_schema_wrapper.py#L258

> Or did you decide on making channels optional so that it's easier to cache as you then no longer call it with channels passed but to maintain backwards compatibility, if someone passes from_channels it does not use the cached version but instead determines it based on the provided channels

To simplify all of this, there are two cases:

1. Calling without `channels` == default == caching of `channels.py`
2. Calling with some module == no caching == identicial behaviour as before

I suppose it would be possible to cache any module passed in, but I think it would add a lot of new complexities.
- What guarantees can me make about other modules?
- Would this be scoped to modules in `altair`, or third-party?
- Will the cached metadata remain correct?

tests/utils/test_schemapi.py
```diff
@@ -557,10 +542,10 @@ def chart_error_example__wrong_tooltip_type_in_layered_chart():
 
 def chart_error_example__two_errors_in_layered_chart():
     # Error 1: Wrong data type to pass to tooltip
-    # Error 2: invalidChannel is not a valid encoding channel
+    # Error 2: `Color` has no parameter named 'invalidChannel'
```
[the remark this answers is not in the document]
dangotbanned: Fixed in [`fe736f9` (#3444)](https://github.com/vega/altair/pull/3444/commits/fe736f98dba106da717073f541893c748d8136b7)

Well spotted @binste

tests/utils/test_schemapi.py
dangotbanned: Fixed in [`0e0e167` (#3444)](https://github.com/vega/altair/pull/3444/commits/0e0e167e11b298dd971707dd2f41d2ad5e54ad16)

altair/utils/core.py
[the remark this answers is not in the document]
binste: Agree, ok for me to not cache any other modules as this was never the intended goal and this allows us to still have backwards compatibility in case anyone has ever used this function with another module (which would be unexpected)

tests/utils/test_core.py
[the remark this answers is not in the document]
binste: You're right, my comment doesn't make sense as I did not properly read the test and assumed that the channels module is passed in. No need to add additional logic. Looks good to me as-is

binste (approved):
Looks good to me, thanks again! I leave it to @mattijn to merge in case he has any additional comments.

dangotbanned:
Thanks @binste appreciate the review

binste:
`api.LookupData` is already the one which you get from top-level `altair` which is correct:

image

But indeed, mypy does get confused by this and thinks it's `core.LookupData`. We already had issues with all the `from ... import *` statements in the package which are used to expose almost all objects on the top-level. Ideally, we could tell mypy that it's `api.LookupData` on the top-level. Unfortunately, I could not convince mypy with a simple type hint after the import statements. I now pushed a commit which excludes `LookupData` from being exported. We already do this for 2 other classes which clash with classes defined in other modules. Downside is that there is now no more `LookupData` defined on `altair.vegalite.v5.schema` but I think that's ok as I don't think this was a popular way to import it.

dangotbanned:
https://github.com/vega/altair/pull/3444#issuecomment-2195482320

Thanks @binste for the explaination and quick fix

tools/update_init_file.py
```diff
@@ -92,7 +92,7 @@ def relevant_attributes(namespace: dict[str, Any], /) -> list[str]:
     it = (
         name
         for name, attr in namespace.items()
-        if (not name.startswith("_")) and _is_relevant(attr)
+        if (not name.startswith("_")) and _is_relevant(attr, name)
```
dangotbanned, line 95: @binste this could also work, without re-introducing `name` to the attribute checking

```py
it = (
	name
	for name, attr in namespace.items()
	if (not name.startswith(("_", "pd", "jsonschema")) and _is_relevant(attr)
)
```

Another option could be adding a global tuple of banned names, like with `_TYPING_CONSTRUCTS`.

What do you think?

dangotbanned (commented):
Very minor tweak, but think it would be more maintainable

tools/update_init_file.py
binste: I don't see a downside of introducing name as an argument and it's more explicit to me if the function handles all checks if something is relevant. startswith is a bit obscure for a new code reader as we don't expect multiple entries to start with "pd".

As it's only used once, I don't think we need a global constant but no real preference.

dangotbanned: > As it's only used once, I don't think we need a global constant but no real preference

@binste that's fair, with how few names there are now, I'm fine with it for now! 

For context, I'm trying to avoid slipping back into the earlier version:
https://github.com/vega/altair/blob/18a2c3c237014591d172284560546a2f0ac1a883/tools/update_init_file.py#L86-L119

dangotbanned:
> Nice catch! Wasn't aware that the signature for `encode` does not work 

@binste just to explain this a bit further, the issue (I think) stems from `@utils.use_signature`:
- The assumption that `Obj` is always a class, whereas `channels._encode_signature` was a function.
- It appears to be acting like `@functools.wraps`, but the above is quite different to [`functools.update_wrapper`](https://github.com/python/cpython/blob/1a2e7a747540f74414e7c50556bcb2cc127e9d1c/Lib/functools.py#L36-L64).

https://github.com/vega/altair/blob/c9106f0f8511a18551a9346f6587ce9799e1532a/altair/utils/core.py#L689-L696




Comparing to `@deprecation.deprecated`, which does use `@functools.wraps`: 
- Avoids the need for *# type: ignore* comments
- Can be fully annotated
	- I added this during #3431
	- [Previous version](https://github.com/vega/altair/blob/51ea29198a851936554a9d92fb3bbc4ad248c73e/altair/utils/deprecation.py)
- Understood by type checkers
https://github.com/vega/altair/blob/c9106f0f8511a18551a9346f6587ce9799e1532a/altair/utils/deprecation.py#L1-L94

mattijn:
Thanks for the PR and review, LGTM!

PR #3446, opened 2024-06-27
chore: avoid pandas warning for `freq='H'` in test_utils.py

MarcoGorelli:
pandas deprecated `freq='H'` in favour of `freq='h'` to avoid case-sensitivity issues (`'ms'` means millisecond, but `'MS' month-start), `'h'` works on both older and newer versions


binste (approved):
LGTM, thanks!

Issue #1027, opened 2018-07-17
Monthly data and ordinal encoding create an "off by one" error


palewire:
I uncovered this "bug" during some data exploration for development of a [open-source library](https://github.com/datadesk/cpi/blob/master/notebooks/analysis.ipynb) tracking the U.S. Consumer Price Index. 

The source data is monthly, with the values linked with the first date of each month. 

```python
import pandas as pd
import altair as alt
df = pd.read_json("https://raw.githubusercontent.com/datadesk/cpi/master/notebooks/last_13.json", dtype={"date_label": pd.np.datetime64})
df.head(13)[['date',  'pct_change_rounded']]
```

![screenshot from 2018-07-16 18-33-09](https://user-images.githubusercontent.com/9993/42791686-b7d773e0-8926-11e8-8c19-e7adaabb1e09.png)

My aim was to format the dates in the x-axis labels in the same manner as the government's sample chart.

![screenshot from 2018-07-16 18-36-58](https://user-images.githubusercontent.com/9993/42791803-42d466c4-8927-11e8-9ca1-2310cc895e2e.png)

I tried to do that with `timeUnit` and the `format` option to `axis`.

```python
alt.Chart(df).mark_bar().encode(
    x=alt.X("date:O", timeUnit="yearmonth", axis=alt.Axis(format="%b %y")),
    y="pct_change_rounded:Q"
)
```

Here's what I got:

![visualization 39](https://user-images.githubusercontent.com/9993/42791838-6427bbb4-8927-11e8-8207-4177b104c0a4.png)

Look closely and you can see that the latest value, June 1, is rendered as May.

What's up with that?

jakevdp:
Looks like something to do with Vega-Lite. I've opened an issue there.

jakevdp:
Per response in https://github.com/vega/vega-lite/issues/4044, t looks like the issue is that Javascript parses dates in UTC. you can fix this here by using ``timeUnit='utcyearmonth'`` rather than ``timeUnit='yearmonth'``

palewire:
I'm not sure I totally get it. If I submit "2018-06-01," how is that parsed by VegaLite? What would the Python datetime object come back as?

jakevdp:
My understanding is that it is parsed (in UTC) as 00:00 on 2018-06-01, which is then interpreted (in ISO) as midnight on 2018-05-31, leading to the issue you see.

The solution is to parse the month and year in UTC.

[...]

PR #3452, opened 2024-06-30
feat: make pandas and NumPy optional dependencies, don't require PyArrow for plotting with Polars/Modin/cuDF

### LINKED ISSUE #3445: Remove PyArrow dependency for Polars support
### What is your suggestion?

Currently, PyArrow is required by Altair for Polars support. I think it shouldn't be too hard to remove it, given that Polars implements the dataframe interchange protocol natively (without depending on PyArrow)

If #3384 can make it in, then Altair would actually support plotting Polars dataframe natively without any extra heavy dependencies. That'd be...pretty amazing? I'd suggest using Altair for `polars.DataFrame.plot` if that was the case

I think what would need doing is:
- don't require `pyarrow` to be installed for the `dfi = data.__dataframe__` part
- instead of using `sanitize_arrow_table`, for Polars, just select date/datetime columns and call `.dt.to_string()`
- instead of using `to_pylist` from PyArrow, just use `DataFrame.rows(named=True)` for Polars
- for categoricals, find a non-pyarrow workaround for Polars in `infer_vegalite_type_for_dfi_column`. I haven't tried this yet, but it looks straightforward-ish

Would you open to considering this? Happy to work on a PR if so

Just keep the status-quo :) But, I think Altair is the only plotting library that gets close to native Polars support without extra large dependencies, and it doesn't look like a large stretch to go all the way there, so I'm hoping we can do it 💪 

---

Demo from having tried this locally:

image

### LINKED ISSUE #3456: `sample` doesn't preserve original object type
### What happened?

```
import polars as pl
import pandas as pd
import altair as alt

df_pd = pd.DataFrame({'a': [1,2,3]})
print(alt.sample(df_pd, n=2))
df_pl = pl.DataFrame({'a': [1,2,3]})
print(alt.sample(df_pl, n=2))
```

This outputs (exact values differ between runs due to the nature of `sample`, but the points is - Polars.DataFrame gets transformed to PyArrow)
```
   a
0  1
1  2
pyarrow.Table
a: int64
----
a: [[1,2]]
```

The docs for `alt.sample` say
```
Reduce the size of the data model by sampling without replacement.
```
so I'd have expected the input library to be preserved, and to get something like:
```python
   a
1  1
2  2
shape: (2, 1)
┌─────┐
│ a   │
│ --- │
│ i64 │
╞═════╡
│ 1   │
│ 2   │
└─────┘
```

5.3.0

### LINKED ISSUE #3471: Plotting ibis.table converts to pyarrow twice
### What happened?

```
import polars as pl
import altair as alt

import ibis
ibis.set_backend('polars')

df = pl.DataFrame({'a': [1,2,3], 'b': [4,5,6]})
tbl = ibis.memtable(df)

alt.Chart(tbl).mark_point().encode(x='a', y='b').to_json()
```

If I put a breakpoint in `arrow_table_from_dfi_dataframe`, then I see the conversion from ibis table to arrow table happening twice

I think it should be possible to convert at most once?

5.4.0dev (latest commit on main)

### LINKED ISSUE #3473: ibis plot fails with dtype date32
### What happened?

```python
import polars as pl
import altair as alt
from datetime import date

import ibis
df = pl.DataFrame(
    {
        "a": [1,2,3],
        "b": [date(2020, 1, 1), date(2020, 1, 2), date(2020, 1, 3)],
    }
)
tbl = ibis.memtable(df)
print(tbl.to_pyarrow())
alt.Chart(tbl).mark_line().encode(x="a", y="b").to_json()
```

raises:
```
ValueError: Data type date32[day] not supported by interchange protocol
```

Plot it without raising

5.3.0

MarcoGorelli:
Suggested in https://github.com/vega/altair/issues/3445#issuecomment-2194585068

closes #3456
closes #3445
closes #3473 
~closes #3471~ (this wasn't an issue to begin with)

### Polars can work without PyArrow nor pandas installed (EDIT: and without NumPy either!):

Screenshot 2024-06-30 153556

### PyArrow table is still supported

(notice the `.to_arrow`)

Screenshot 2024-06-30 150956

### Ibis is still supported, and now Date dtype works too!

(previously, it would have been necessary to manually cast 'date' to 'datetime' type before plotting)

image

## Performance

There's also a perf improvement to staying Polars-native where possible - for example, if I run
```python
import polars as pl
import altair as alt
from datetime import date

# dataset: https://github.com/Mcompetitions/M6-methods/blob/main/assets_m6.csv
df = pl.read_csv('../scratch/assets.csv', try_parse_dates=True)
df = df.filter(pl.col('date')>date(2022, 11, 15))
def func():
    alt.Chart(df).mark_line().encode(
        x='date', y='price', color='symbol',
    ).to_json()
results = %timeit -o -q func()
results.best
```
in IPython, then I see:
- Altair `main` branch: 0.033584889399935494
- Altair `MarcoGorelli:narwhalify` branch: 0.024283270300657023

That's for a (4475, 3)-shaped dataframe - for dataframe with many columns (esp. with strings / categoricals) I'd expect the difference to grow


dangotbanned:
@MarcoGorelli if all the tests are passing locally for you, try running these two:
```bash
hatch run generate-schema-wrapper
hatch run update-init-file
```

If there are any changes other than whitespace, you'll get a more informative error than https://github.com/vega/altair/actions/runs/9733301918/job/26860053484?pr=3452

Otherwise those two should clean everything up

tools/generate_schema_wrapper.py
```diff
@@ -564,7 +565,7 @@ def generate_vegalite_channel_wrappers(
     imports = imports or [
         "from __future__ import annotations\n",
         "from typing import Any, overload, Sequence, List, Literal, Union, TYPE_CHECKING",
-        "import pandas as pd",
+        "from narwhals.dependencies import is_pandas_dataframe",
```
dangotbanned, line 568:
```suggestion
        "from narwhals.dependencies import is_pandas_dataframe as _is_pandas_dataframe",
```
Avoiding adding to `__all__` (1/2)
https://github.com/vega/altair/commit/7ab395bd10732ecf7bf3c7ffd969ac2eedc20feb

altair/vegalite/v5/schema/channels.py
```diff
@@ -13,7 +13,7 @@
 
 from typing import TYPE_CHECKING, Any, Literal, Sequence, overload
 
-import pandas as pd
+from narwhals.dependencies import is_pandas_dataframe
```
dangotbanned, line 16:
```suggestion
from narwhals.dependencies import is_pandas_dataframe as _is_pandas_dataframe
```
Avoiding adding to `__all__` (2/2)

altair/vegalite/v5/schema/channels.py
```diff
@@ -72,7 +72,7 @@ def to_dict(
                 # We still parse it out of the shorthand, but drop it here.
                 parsed.pop("type", None)
             elif not (type_in_shorthand or type_defined_explicitly):
-                if isinstance(context.get("data", None), pd.DataFrame):
+                if is_pandas_dataframe(context.get("data", None)):
```
dangotbanned, line 75:
```suggestion
                if _is_pandas_dataframe(context.get("data", None)):
```
Avoiding adding to `__all__` (3/3)

tools/generate_schema_wrapper.py
```diff
@@ -108,7 +108,7 @@ def to_dict(
                 # We still parse it out of the shorthand, but drop it here.
                 parsed.pop("type", None)
             elif not (type_in_shorthand or type_defined_explicitly):
-                if isinstance(context.get("data", None), pd.DataFrame):
+                if is_pandas_dataframe(context.get("data", None)):
```
dangotbanned, line 111:
```suggestion
                if _is_pandas_dataframe(context.get("data", None)):
```
Avoiding adding to `__all__` (3/4)

tools/generate_schema_wrapper.py
dangotbanned: @MarcoGorelli there is quite a lot of implicit handling of imports in `altair`.

Certain things are [special-cased](https://github.com/dangotbanned/altair/blob/801cccc80e7764df3d46cd0db0343f095ee78d01/tools/update_init_file.py) like `typing`, but generally the safest option is just aliasing w/ `_` prefix. 

I think you'd need to run those changes locally first though to remove `is_pandas_dataframe` from `__all__`

MarcoGorelli: aah thanks, that stumped me 😄

dangotbanned:
https://github.com/vega/altair/actions/runs/9734437115/job/26862450764?pr=3452
@MarcoGorelli try this to run the tests on the matrix

```bash
hatch test --all
```

dangotbanned:
fix: add missing `pandas` import to doctest

```log
519     Examples 520     -------- 521     >>> data = pd.DataFrame({'foo': ['A', 'B', 'A', 'B'], UNEXPECTED EXCEPTION: NameError("name 'pd' is not defined") Traceback (most recent call last):   File "..\Local\hatch\env\virtual\.pythons\3.12\python\Lib\doctest.py", line 1361, in __run           exec(compile(example.source, filename, "single",   File "<doctest altair.utils.core.parse_shorthand[0]>", line 1, in <module> NameError: name 'pd' is not defined C:\Users\danie\Documents\GitHub\altair\altair\utils\core.py:521: UnexpectedException ============================================== short test summary info ==============================================  FAILED altair/utils/core.py::altair.utils.core.parse_shorthand ========================= 1 failed, 1250 passed, 1 skipped, 2 xfailed, 1 xpassed in 30.64s ==========================
```
```diff
diff --git a/altair/utils/core.py b/altair/utils/core.py
index e6620146..d5d753db 100644
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -520,6 +520,7 @@ def parse_shorthand(
 
     Examples
     --------
+    >>> import pandas as pd
     >>> data = pd.DataFrame({'foo': ['A', 'B', 'A', 'B'],
     ...                      'bar': [1, 2, 3, 4]})
 

```

dangotbanned:
refactor(typing): Reuse `_typing` alias as `InferredVegaLiteType`
```diff
diff --git a/altair/utils/core.py b/altair/utils/core.py
index e6620146..f3a5469a 100644
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -42,6 +42,7 @@ else:
 if TYPE_CHECKING:
     from types import ModuleType
     import typing as t
+    from altair.vegalite.v5.schema._typing import StandardType_T as InferredVegaLiteType
     from altair.utils._dfi_types import DataFrame as DfiDataFrame
     from altair.utils.data import DataType
     from narwhals.typing import IntoExpr
@@ -199,9 +200,6 @@ TIMEUNITS = [
 ]
 
 
-InferredVegaLiteType = Literal["ordinal", "nominal", "quantitative", "temporal"]
-
-
 def infer_vegalite_type_for_pandas(
     data: object,
 ) -> InferredVegaLiteType | tuple[InferredVegaLiteType, list[Any]]:

```

dangotbanned:
@MarcoGorelli these were two minor things. I couldn't figure out a way to request these another way.

#### fix: add missing `pandas` import to doctest

```log
519     Examples 520     -------- 521     >>> data = pd.DataFrame({'foo': ['A', 'B', 'A', 'B'], UNEXPECTED EXCEPTION: NameError("name 'pd' is not defined") Traceback (most recent call last):   File "..\Local\hatch\env\virtual\.pythons\3.12\python\Lib\doctest.py", line 1361, in __run           exec(compile(example.source, filename, "single",   File "<doctest altair.utils.core.parse_shorthand[0]>", line 1, in <module> NameError: name 'pd' is not defined C:\Users\danie\Documents\GitHub\altair\altair\utils\core.py:521: UnexpectedException ============================================== short test summary info ==============================================  FAILED altair/utils/core.py::altair.utils.core.parse_shorthand ========================= 1 failed, 1250 passed, 1 skipped, 2 xfailed, 1 xpassed in 30.64s ==========================
```
```diff
diff --git a/altair/utils/core.py b/altair/utils/core.py
index e6620146..d5d753db 100644
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -520,6 +520,7 @@ def parse_shorthand(
 
     Examples
     --------
+    >>> import pandas as pd
     >>> data = pd.DataFrame({'foo': ['A', 'B', 'A', 'B'],
     ...                      'bar': [1, 2, 3, 4]})
 

```


#### refactor(typing): Reuse `_typing` alias as `InferredVegaLiteType`
```diff
diff --git a/altair/utils/core.py b/altair/utils/core.py
index e6620146..f3a5469a 100644
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -42,6 +42,7 @@ else:
 if TYPE_CHECKING:
     from types import ModuleType
     import typing as t
+    from altair.vegalite.v5.schema._typing import StandardType_T as InferredVegaLiteType
     from altair.utils._dfi_types import DataFrame as DfiDataFrame
     from altair.utils.data import DataType
     from narwhals.typing import IntoExpr
@@ -199,9 +200,6 @@ TIMEUNITS = [
 ]
 
 
-InferredVegaLiteType = Literal["ordinal", "nominal", "quantitative", "temporal"]
-
-
 def infer_vegalite_type_for_pandas(
     data: object,
 ) -> InferredVegaLiteType | tuple[InferredVegaLiteType, list[Any]]:

```

mattijn:
Thank you very much for making this Pull Request. That is very much appreciated!

This PR introduces a new dependency on [narwhals](https://github.com/narwhals-dev/narwhals). Narwhals is described as a lightweight and extensible compatibility layer between multiple dataframe libraries.
Narwhals can replace our hard dependency on pandas, the size of pandas is ~11 MB (without mentioning its subdependencies). This by itself is a very interesting development already. 

Next to this it enables support for native dataframes (pandas still included) without converting these to a pyarrow Tables through the dataframe interchange protocol.

The latest release of Altair is ~850 kB and after this PR, Altair will have the following hard dependencies:

- jinja2 : ~130 kB
- jsonschema : ~90 kB
- typing_extensions: ~35 kB
- packaging : ~55 kB
- narwhals : ~80 kB
- numpy : ~18 MB

When using Altair with only remote endpoints (like APIs that are returning JSON-data), I've the feeling that we do not need to rely on numpy. Therefor I'm still hopefull that we can reduce this hard dependency on numpy as well one day.

I've a few questions regarding this PR and narwhals.

1. Just curious, why is the lower limit of pandas in narwhals version 1.1.5? As the supported lower limit of pandas in Altair before this PR was version 0.25.

2. You serialize the `nw.Date` type to:
```python
if dtype == nw.Date:
	columns.append(nw.col(name).dt.to_string("%Y-%m-%d"))
```
	
And change a test from:
```python
"a": {"date": "2019-01-01T00:00:00"},
```
To:
```python
"a": {"date": "2019-01-01"},
```
Upon first sight, I think we cannot introduce this breaking change. Does this change still adhere to full ISO-8601 dates representation as is required by Javascript? See also past issues like this: https://github.com/vega/altair/issues/1027.

3. Can you explain the behavior of `nw.from_native()`?
When I look to this code:
```python
data = nw.from_native(data, eager_only=True, strict=False)
if isinstance(data, nw.DataFrame):
	return data
if isinstance(data, DataFrameLike):
	from altair.utils.data import arrow_table_from_dfi_dataframe

	pa_table = arrow_table_from_dfi_dataframe(data)
	return nw.from_native(pa_table, eager_only=True)
```
I'm a bit surprised that the `data` object, as a result of passing it into `nw.from_native` can be a non-narwhals dataframe. If it is not supported by narwhals it returns the original data object?

4. Maybe a naive question, but do you have any intention to support dataframes that support the dataframe interface protocol within narwhals?

5. Not being a type expert, but I cannot really understand why there needs to be quotes around the type `"IntoDataFrame"`, like in:
```python
DataType: TypeAlias = Union[
    Dict[Any, Any], "IntoDataFrame", SupportsGeoInterface, DataFrameLike
]
```
I remember I have discussed this with @dangotbanned recently in https://github.com/vega/altair/pull/3431, see https://github.com/vega/altair/pull/3431#issuecomment-2162877321 onwards, but I can't find a clear answer anymore on this topic.

Thanks again for this PR! Really promising! (also this PR will supersede https://github.com/vega/altair/pull/3384)

dangotbanned:
>     5. Not being a type expert, but I cannot really understand why there needs to be quotes around the type `"IntoDataFrame"`, like in:
> 
> 
> ```python
> DataType: TypeAlias = Union[
>     Dict[Any, Any], "IntoDataFrame", SupportsGeoInterface, DataFrameLike
> ]
> ```
> 
> I remember I have discussed this with @dangotbanned recently in #3431, see [#3431 (comment)](https://github.com/vega/altair/pull/3431#issuecomment-2162877321) onwards, but I can't find a clear answer anymore on this topic.

Thanks for the ping @mattijn 

The good news is you can trust the `ruff` config now for managing cases like this (particularly [`TCH`](https://docs.astral.sh/ruff/rules/#flake8-type-checking-tch)), what @MarcoGorelli has written is correct.

https://github.com/MarcoGorelli/altair/blob/2df9422dca75a26eb16fd0c8030b6e89b4d4fa0f/altair/utils/data.py#L44-L66

All of the imports within the `TYPE_CHECKING` block must be quoted when used in a runtime alias definition. `IntoDataFrame` is used in: `DataType`, `TIntoDataFrame`, `SampleReturnType`.

AFAIK, this should be fine to move to a regular import though
- `alt.utils.data` requires explicit exports in `alt.utils.__init__.__all__`
- `narwhals` is already imported at runtime in https://github.com/MarcoGorelli/altair/blob/2df9422dca75a26eb16fd0c8030b6e89b4d4fa0f/altair/utils/data.py#L26

altair/utils/data.py
```diff
@@ -23,10 +23,16 @@
 from functools import partial
 import sys
 
-import pandas as pd
+import narwhals as nw
+from narwhals.dependencies import is_pandas_dataframe as _is_pandas_dataframe
```
dangotbanned, line 27:
```suggestion
from narwhals.dependencies import is_pandas_dataframe as _is_pandas_dataframe
from narwhals.typing import IntoDataFrame
```

altair/utils/data.py
```diff
@@ -37,6 +43,8 @@
 
 if TYPE_CHECKING:
     import pyarrow as pa
+    import pandas as pd
+    from narwhals.typing import IntoDataFrame
```
dangotbanned, line 47:
```suggestion

```

altair/utils/data.py
```diff
@@ -45,20 +53,23 @@ class SupportsGeoInterface(Protocol):
 
 
 DataType: TypeAlias = Union[
-    Dict[Any, Any], pd.DataFrame, SupportsGeoInterface, DataFrameLike
+    Dict[Any, Any], "IntoDataFrame", SupportsGeoInterface, DataFrameLike
```
dangotbanned, line 56:
```suggestion
    Dict[Any, Any], IntoDataFrame, SupportsGeoInterface, DataFrameLike
```

altair/utils/data.py
```diff
@@ -45,20 +53,23 @@ class SupportsGeoInterface(Protocol):
...
 ]
 
 TDataType = TypeVar("TDataType", bound=DataType)
+TIntoDataFrame = TypeVar("TIntoDataFrame", bound="IntoDataFrame")
```
dangotbanned, line 60:
```suggestion
TIntoDataFrame = TypeVar("TIntoDataFrame", bound=IntoDataFrame)
```

altair/utils/data.py
```diff
@@ -45,20 +53,23 @@ class SupportsGeoInterface(Protocol):
...
 
 VegaLiteDataDict: TypeAlias = Dict[
     str, Union[str, Dict[Any, Any], List[Dict[Any, Any]]]
 ]
 ToValuesReturnType: TypeAlias = Dict[str, Union[Dict[Any, Any], List[Dict[Any, Any]]]]
-SampleReturnType = Optional[Union[pd.DataFrame, Dict[str, Sequence], "pa.lib.Table"]]
+SampleReturnType = Optional[Union["IntoDataFrame", Dict[str, Sequence]]]
```
dangotbanned, line 66:
```suggestion
SampleReturnType = Optional[Union[IntoDataFrame, Dict[str, Sequence]]]
```

MarcoGorelli:
Thanks so much for your quick and careful reviews @mattijn and @dangotbanned  🙏 ! Much appreciated

1. No reason :) I just lowered the minimum version of pandas in Narwhals - I set it to `0.25.3` rather than `0.25` though, as when I tried using `0.25` in CI it just froze indefinitely. `0.25.3` is also the one that Altair is currently testing against in CI - is it OK to set that as the minimum?
2. Interesting, thanks, and I'm sorry for not having picked up on that! Quoting the issue
   >  If you were east of London when running this code, you wouldn't see this issue 😄.

   that might be why I didn't see it when testing it out 😳 .
   Polars doesn't let us format `Date`s with time directives (because the underlying Rust crate `chrono` doesn't either) - but, we can just cast to datetime, I've added a comment about why this is done. On my laptop, `pyarrow.compute.strftime` is about 200 times slower than casting Date to Datetime - so, the extra cast to ensure consistency, in the grand scheme of the things, is quite negligible, and not the bottleneck of the `sanitize` function
3. That's because of `strict=False` in the first line of the function https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.from_native, the idea is:
   - `strict=False`: if the object isn't supported by Narwhals, then it just gets passed through as-is
   - `strict=True`: if the object isn't supported by Narwhals, then raise
4. We'd like to have [2 levels of supported dataframes](https://github.com/narwhals-dev/narwhals/issues/388) in Narwhals:
   - full API
   - basic. I.e. an extended and more user-friendly version of the dataframe interchange protocol. Anything which already supports the interchange protocol would be in scope. My main gripe with the interchange protocol is that it's bundled with dataframe libraries themselves, which means is _super slow_ to change - for example, Jon reported https://github.com/pandas-dev/pandas/issues/54239, and it got fixed, but it's going to be several years until the minimum pandas version gets bumped up enough that it can be used... 😭 
5. thanks @dangotbanned for helping out with this one!

>  I've the feeling that we do not need to rely on numpy

"in for a penny, in for a pound" - I've added a commit to do this. NumPy is used very little, mostly isinstance checks and `np.ceil`, so I think it can already be made optional. Then Altair would truly be the lightest visualisation library in the ecosystem?

I've caught up with Polars devs, and they're on board with using Altair in `polars.DataFrame.plot` if the plots can be done directly without going via pandas, so we may want to have a conversation later about how that should look (personally, I think `df.plot.line(x='x', y='y')` as a shorthand for `alt.Chart(df).mark_line().encode(x='x', y='y').interactive()` would be really nice, and users would ❤️ love ❤️  it)

dangotbanned:
Thanks @MarcoGorelli 

When I [originally suggested](https://github.com/vega/altair/pull/3384#issuecomment-2188349276) `narwhals` I didn't expect you'd take an active role - and this PR is certainly far beyond that!
I get the impression that this is beneficial for both projects, as any issues you've encountered have been quickly resolved in subsequent [narwhals releases](https://github.com/narwhals-dev/narwhals/releases)

> I've caught up with Polars devs, and they're on board with using Altair in `polars.DataFrame.plot` if the plots can be done directly without going via pandas, so we may want to have a conversation later about how that should look (personally, I think `df.plot.line(x='x', y='y')` as a shorthand for `alt.Chart(df).mark_line().encode(x='x', y='y').interactive()` would be really nice, and users would ❤️ love ❤️ it)

I think most people would agree on the value of this, but probably best for another issue.
@joelostblom would probably be able to chime in, there was also this discussion https://github.com/vega/altair/discussions/2789#discussioncomment-4579799

I would expect there would be more support if the implementation was different to the existing [hvPlot](https://github.com/pola-rs/polars/blob/1fb6eb6d4a8762072a462d1755daed291e0882f5/py-polars/polars/dataframe/frame.py#L591) accessor.
Currently, the burden would be on `altair` to define what kinds of charts (e.g. line, scatter, etc), but this isn't a 1-to-1 with how many of the other plotting libraries work. 

My suggestion would be for `polars` to define a protocol with some chart methods, with `altair` implementing those in an accessor.
That way, these methods/functions may have less of a chance of leaking out into the rest of the `altair` api.

tools/schemapi/schemapi.py
```diff
@@ -477,17 +477,26 @@ def _subclasses(cls: type[Any]) -> Iterator[type[Any]]:
 
 def _todict(obj: Any, context: dict[str, Any] | None) -> Any:
```
dangotbanned, line 478: Is there any performance hit with `get_numpy`, `get_pandas` introduced here?

There's a whole bunch of recursion going on and those checks might add up.

dangotbanned: @MarcoGorelli if changing this complicates things, (saw some tests fail) don't worry.
I was just curious *if* it had a performance impact due to an earlier PR https://github.com/vega/altair/pull/3383

MarcoGorelli:
Thanks @dangotbanned !

> I would expect there would be more support if the implementation was different to the existing [hvPlot (polars)](https://github.com/pola-rs/polars/blob/1fb6eb6d4a8762072a462d1755daed291e0882f5/py-polars/polars/dataframe/frame.py#L591) / [hvPlot](https://github.com/holoviz/hvplot/blob/667aaabedae4378a3fbd8ef765c581f4809f617b/hvplot/plotting/core.py#L171) accessor.

Definitely - `plot` is marked as `unstable` in Polars 😎 What I had in mind was just something like
```
class PlotNameSpace:
    def __init__(self, df: DataFrame) -> None:
        self.df = df
    def line(self, *args, **kwargs):
        return alt.Chart(self.df).mark_line().encode(*args, **kwargs).interactive()
    def point(self, *args, **kwargs):
        return alt.Chart(self.df).mark_point().encode(*args, **kwargs).interactive()
    ...
```
Anyway, we can figure out the details later, whether or not Altair decides to go ahead with Narwhals

(for the record, I think hvplot is amazing and its developers are really friendly, so I'd still advocating for it to be advertised in the Polars docs - but if Altair could provide a more lightweight alternative which doesn't require other heavy libraries, I think it'd be a more suitable default)

tools/schemapi/schemapi.py
MarcoGorelli: interesting, thanks - locally, I'm seeing that `get_pandas` / `get_numpy` take about 4 microseconds to run. They're just doing a dictionary lookup, so it should be really fast

nonetheless, it's not much effort to define them one level up, so I can do that (and just use `sys.modules`)

mattijn:
Really awesome to see you were able to remove numpy as a hard dependency too! Its great to see this happening!

A few more comments while you might still have this PR open.

In my opinion, there is no need to serialize microseconds for type `Date` (also reducing characters in the vega-lite specification). If we define `local_iso_fmt_string = "%Y-%m-%dT%H:%M:%S"` and use that when `nw.Date` applies and when `nw.Datetime` applies we can use `f"{local_iso_fmt_string}%.f"` or something similar.

A single comment like ``# using `strict=False` will return the `data` as-is if the object cannot be converted.`` would be appreciated for my future-self or other people who will be reading the following line:
https://github.com/vega/altair/blob/7312f25eea31b6c86e273d591641b6c2fb494e3b/altair/utils/core.py#L477

This line can be removed:
https://github.com/vega/altair/blob/7312f25eea31b6c86e273d591641b6c2fb494e3b/altair/utils/core.py#L451

Thanks again for pushing this!

mattijn:
A bit off topic, but regarding `polars.DataFrame.plot` functionality: (1) you also might learn some lessons from this archived repo: https://github.com/altair-viz/pdvega, (2) there is also some work underway to extend the `.interactive()` method within altair https://github.com/vega/altair/pull/3394.

mattijn (approved):
For me this is a +1, it would be great if there is another maintainer doing a review.

binste:
This is exciting! :) I'd also like to have a quick look. Hopefully get to it this weekend. @jonmmease if you have a moment you might want to have a look as well as you've worked on the interchange protocol implementation and hence might be aware of edge cases we need to consider.

pyproject.toml
```diff
@@ -77,6 +77,7 @@ dev = [
     "types-jsonschema",
     "types-setuptools",
     "geopandas",
+    "polars",
```
binste, line 80: The original goal of the `all` dependeny group was to provide an easy way for users to set up all dependencies they are likely to use together with Altair. See https://github.com/vega/altair/pull/3354. I'm not sure if we want to include pandas/numpy/polars here or instead only in `dev` for the tests. To me, there is a good case for pandas and numpy as almost all examples in the Altair docs use Pandas. We want one install command which can be used in workshops which gets everything up and running. As there is no content so far for Polars, maybe we'd just move that one to `dev` for now?

If Altair and Polars get more integrated in the future, we should also have comprehensive examples with Polars in the docs and could reconsider moving it to `all`.

Any thoughts on this? Pinging @jonmmease @mattijn @joelostblom as all of you have been involved in this discussion at one point.

binste (commented):
I only added one comment regarding dependency groups. LGTM otherwise!

pyproject.toml
mattijn: It's probably good to mention `pandas` explicitly in the `dev` group too. 

I don't think it is necessary to have a dataframe library directly included within `all`. Meanwhile, pandas is still a dependency of `vega-datasets` and `vegafusion`, so it still will be installed when installing the full `altair-all` (just like `pyarrow` btw).

dangotbanned: @mattijn @binste 

Are you aware of the test dependencies feature in `hatch`?
https://hatch.pypa.io/latest/config/internal/testing/#dependencies

mattijn: Sounds interesting! Especially in combination with https://hatch.pypa.io/latest/how-to/integrate/vscode/, so it is possible to debug/run single tests.

jonmmease:
I'm very interested in this PR, but I'm on holiday and won't have time to look at it until mid next week

joelostblom (commented):
Exciting PR, thank for the contribution! Narwhal looks like a neat library and super helpful for removing heavier dependencies like pandas etc. I have not looked through the code in detail, but I'm in favor of moving in this direction.


>> I've caught up with Polars devs, and they're on board with using Altair in polars.DataFrame.plot if the plots can be done directly without going via pandas, so we may want to have a conversation later about how that should look (personally, I think df.plot.line(x='x', y='y') as a shorthand for alt.Chart(df).mark_line().encode(x='x', y='y').interactive() would be really nice, and users would ❤️ love ❤️ it)
>> 
> I think most people would agree on the value of this, but probably best for another issue.
> @joelostblom would probably be able to chime in, there was also this discussion https://github.com/vega/altair/discussions/2789#discussioncomment-4579799

This is an interesting idea and it seems like Altair would lend itself well to being a lightweight library that's easy to bundle with dataframe plotting APIs. I will say that I personally don't find built-in plotting in dataframes very convenient and I have always found it easier to work directly with the different viz libraries instead. But since these approaches already exist in both pandas and polars, I think it would be great if Altair can help make it more convenient and your suggestion with `df.plot.line()` seems like a natural mapping to `chart(df).line().encode()` @MarcoGorelli . Regarding https://github.com/vega/altair/discussions/2789#discussioncomment-4579799, I don't think that's too relevant here, at least not at the moment, but it could be in the future if polars want to incorporate more advanced EDA charts like splooms (and I find time to work on altair-ally more before then).

MarcoGorelli:
thanks for your lovely reviews!

I've updated to:
- use the `narwhals.stable.v1` (see [here](https://narwhals-dev.github.io/narwhals/backcompat/) for the stable api policy)
- Polars is in the `dev` section of `pyproject.toml`, whilst pandas and NumPy in `all` (but pandas is also in `dev` as per https://github.com/vega/altair/pull/3452#discussion_r1664492208)

altair/utils/data.py
```diff
@@ -413,7 +426,7 @@ def arrow_table_from_dfi_dataframe(dfi_df: DataFrameLike) -> pa.Table:
     # has more control over the conversion, and may have broader compatibility.
     # This is the case for Polars, which supports Date32 columns in direct conversion
     # while pyarrow does not yet support this type in from_dataframe
-    for convert_method_name in ("arrow", "to_arrow", "to_arrow_table"):
+    for convert_method_name in ("arrow", "to_arrow", "to_arrow_table", "to_pyarrow"):
```
MarcoGorelli, line 429: this change by itself isn't enough to address https://github.com/vega/altair/issues/3473 on main, because the interchange protocol is used directly - there could be an extra pyarrow conversion, but then that would worsen https://github.com/vega/altair/issues/3471

tests/vegalite/v5/test_api.py
```diff
@@ -1065,3 +1072,35 @@ def test_validate_dataset():
...
+    # Check pandas and PyArrow weren't imported anywhere along the way,
+    # confirming that the plot above would work without pandas no PyArrow
+    # installed.
+    assert "pandas" not in sys.modules
+    assert "pyarrow" not in sys.modules
+    assert "numpy" not in sys.modules
+
+
+@pytest.mark.skipif(
+    int(pd.__version__.split(".")[0]) < 2,
```
dangotbanned, line 1093:
```suggestion
    int(version("pandas")[0]) < 2,
```

Might be a little bit more readable, would require `from importlib.metadata import version` as well

MarcoGorelli: thanks for your review! I think that would still return a string though? so `version('pandas')[0]` would be `'1'` when pandas hits version 10 (granted, that's a while away 😄 )

dangotbanned: @MarcoGorelli you're absolutely right there

I suppose even with `version` you'd still need this for v10

```py
int(version('pandas').split(".")[0]) < 2
```

tests/vegalite/v5/test_api.py
```diff
@@ -1065,3 +1072,35 @@ def test_validate_dataset():
...
+    df = pl.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
+    _ = alt.Chart(df).mark_line().encode(x="a", y="b").to_json()
+    # Check pandas and PyArrow weren't imported anywhere along the way,
+    # confirming that the plot above would work without pandas no PyArrow
+    # installed.
+    assert "pandas" not in sys.modules
+    assert "pyarrow" not in sys.modules
+    assert "numpy" not in sys.modules
+
+
```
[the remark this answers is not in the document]
dangotbanned:
```
E   pyarrow.lib.ArrowInvalid: Cannot locate timezone 'UTC': Timezone database not found at "C:\Users\..\Downloads\tzdata"

pyarrow\\error.pxi:91: ArrowInvalid
======================================================================================== short test summary info ========================================================================================= 
FAILED tests/vegalite/v5/test_api.py::test_ibis_with_date_32 - pyarrow.lib.ArrowInvalid: Cannot locate timezone 'UTC': Timezone database not found at "C:\Users\d..\Downloads\tzdata"
==================================================================== 1 failed, 1258 passed, 1 skipped, 1 xfailed, 2 xpassed in 32.94s ==================================================================== 
```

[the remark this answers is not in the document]
MarcoGorelli: thanks!

tests/vegalite/v5/test_api.py
MarcoGorelli: sure, done, it is more readable 👍

tests/vegalite/v5/test_api.py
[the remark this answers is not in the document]
dangotbanned: Sorry for adding this right after you updated the other one 😓 

Just tried testing locally

tools/schemapi/schemapi.py
```diff
@@ -475,20 +472,34 @@ def _subclasses(cls: type[Any]) -> Iterator[type[Any]]:
             yield cls
 
 
-def _todict(obj: Any, context: dict[str, Any] | None) -> Any:
+def _todict(obj: Any, context: dict[str, Any] | None, np_opt: Any, pd_opt: Any) -> Any:
     """Convert an object to a dict representation."""
+    if np_opt is not None:
```
dangotbanned, line 477: Not suggesting a change here, but if anyone else was curious (as I was).

Reordering this branch (re-prioritising `SchemaBase`) **doesn't** appear to affect the performance of the test suite.

dangotbanned (approved):
Certainly doesn't carry the same weight as from a maintainer, but from someone who has read through a lot of `altair` code lately - this is great @MarcoGorelli

Thank you for the PR

mattijn:
@dangotbanned, I totally agree agree that you have been going very deep in the altair code recently! Which is very much appreciated. If being a maintainer would make it more easy for you to contribute more to altair, I would like to propose you to become an official maintainer.

dangotbanned:
@MarcoGorelli 

FYI, I think you should hold off on updating this branch until https://github.com/vega/altair/pull/3475 is resolved, to avoid getting an unrelated `mypy` error

altair/utils/core.py
```diff
@@ -597,39 +634,20 @@ def parse_shorthand(
...
-                attrs["type"] = infer_vegalite_type(
-                    data[attrs["field"].replace("\\", "")]
-                )
-                # ordered categorical dataframe columns return the type and sort order as a tuple
-                if isinstance(attrs["type"], tuple):
-                    attrs["sort"] = attrs["type"][1]
-                    attrs["type"] = attrs["type"][0]
+    if "type" not in attrs and is_data_type(data):
+        unescaped_field = attrs["field"].replace("\\", "")
+        data_nw = narwhalify(data)
```
jonmmease, line 639: Something important that I think we're losing here is that previously we were checking the types of columns in DataFrame Interchange Protocol DataFrames without loading them all into memory as pyarrow (which is what narwhalify does for types that narwhals doesn't support). Ibis has [an optimization](https://github.com/ibis-project/ibis/pull/6733) where calling `dfi.get_column_by_name` is a metadata-only operation that doesn't trigger loading all columns into memory.

When used with plain Altair, the full dataset is loaded into memory anyway during the `to_values()` calculation. But when the [`"vegafusion"` data transformer](https://altair-viz.github.io/user_guide/large_datasets.html#vegafusion-data-transformer) is enabled, VegaFusion is able to detect which columns are needed and it will request only the required columns when loading into pyarrow (See [here](https://github.com/vega/vegafusion/blob/007bd44188676de7259bc02a61693b3dc7586072/python/vegafusion/vegafusion/datasource/dfi_datasource.py#L110)).

So unless/until narwhals can support wrapping DataFrame Interchange Protocol DataFrames directly, I think we need to avoid the arrow conversion here and fall back to the legacy `infer_vegalite_type_for_dfi_column` behavior for this case.

jonmmease (changes requested):
Thanks a lot for this PR @MarcoGorelli, it's a really nice step forward for minimizing dependencies while opening up support for more DataFrame libraries. I made a few small suggestions, but the only blocking issue is what I commented on in the `parse_shorthand` function.

To avoid regressing performance in the VegaFusion+Ibis case, I think we'll need to avoid loading the full DataFrame into memory as pyarrow at this point, and instead use the `__dataframe__` protocol directly to introspect the column types.

This is the intended behavior at least, but based on what you observed in #3471, perhaps something else is causing a double evaluation with Ibis. Even if the double evaluation persists (and isn't a regression), I think this PR should preserve the behavior of the `parse_shorthand` function for DataFrame Interchange Protocol DataFrames not supported by narwhals.  I can dig into the double evaluation separately after this PR merges if needed (you've already gone above and beyond with this PR!).

Thanks again!

altair/utils/core.py
```diff
@@ -429,30 +438,57 @@ def to_list_if_array(val):
...
+        # Early return if already a Narwhals DataFrame
+        return data
+    # Using `strict=False` will return `data` as-is if the object cannot be converted.
+    data = nw.from_native(data, eager_only=True, strict=False)
+    if isinstance(data, nw.DataFrame):
+        return data
+    if isinstance(data, DataFrameLike):
+        from altair.utils.data import arrow_table_from_dfi_dataframe
+
+        pa_table = arrow_table_from_dfi_dataframe(data)
```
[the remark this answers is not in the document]
MarcoGorelli: It is possible, although pyarrow is a required dependency of ibis anyway

MarcoGorelli:
Thanks @jonmmease for your careful review, really appreciate it! 🙏 

> But when the ["vegafusion" data transformer](https://altair-viz.github.io/user_guide/large_datasets.html#vegafusion-data-transformer) is enabled, VegaFusion is able to detect which columns are needed and it will request only the required columns when loading into pyarrow (See [here](https://github.com/vega/vegafusion/blob/007bd44188676de7259bc02a61693b3dc7586072/python/vegafusion/vegafusion/datasource/dfi_datasource.py#L110)).

Aah, that's amazing! I'd missed that. Maybe polars.LazyFrame should also implement the interchange protocol just at the metadata level. That's one for a separate discussion anyway

To address this here, we've made a new release of Narwhals which includes interchange-level support - that is to say, if the input object implements the interchange protocol, then we let you inspect the schema

Like this, the code complexity goes down, _and_ we fix the bug Ibis users would have encountered whereby Date dtype wasn't supported (unless they use the vegafusion transformer, in which case, things are the same for them as they are with the latest Altair release). So, this PR improves things for Ibis users using the default data transformers, and preserves the status-quo for those using the vegafusion data transformer 😇

altair/utils/core.py
```diff
@@ -429,30 +438,57 @@ def to_list_if_array(val):
...
+
+    If `data` is not supported by Narwhals, but it is convertible
+    to a PyArrow table, then first convert to a PyArrow Table,
+    and then wrap in `narwhals.DataFrame`.
+    """
+    if isinstance(data, nw.DataFrame):
+        # Early return if already a Narwhals DataFrame
+        return data
+    # Using `strict=False` will return `data` as-is if the object cannot be converted.
+    data = nw.from_native(data, eager_only=True, strict=False)
```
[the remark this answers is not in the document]
MarcoGorelli: For now they're going down different paths:

https://github.com/MarcoGorelli/altair/blob/5d54bc4263dd5bdba600819b316dda8c592bdd8f/altair/utils/data.py#L316-L334

Maybe this can all be unified further, although the pandas/geointerface paths are doing some very library-specific things

tests/utils/test_dataframe_interchange.py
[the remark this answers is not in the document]
MarcoGorelli: Ah I can't see the original line you'd commented on any more...any chance you remember which test it was? Sorry about this

mattijn:
Thanks for the changes @MarcoGorelli! 

I still have one question which I cannot really understand. Given the following example:

```python
import datetime
import polars as pl
import pyarrow as pa
import narwhals as nw

data_type_date = [{"date": datetime.date(2024,7,15)}]
local_iso_fmt_string = "%Y-%m-%dT%H:%M:%S"
nw_selector = [nw.col('date').cast(nw.Datetime).dt.to_string(local_iso_fmt_string)]

pl_df = pl.DataFrame(data_type_date)
nw_pl_df = nw.from_native(pl_df, eager_only=True)
nw_pl_df.select(nw_selector).rows(named=True)
```
This returns

> ```python
> [{'date': '2024-07-15T00:00:00'}]
> ```

But if I do:
```python
py_tb = pa.Table.from_pylist(data_type_date)
nw_py_tb = nw.from_native(py_tb, eager_only=True)
nw_py_tb.select(nw_selector).rows(named=True)
``` 
It returns

> ```python
> [{'date': '2024-07-15T00:00:00.000000'}]
> ``` 

I expected these output to be equal.

mattijn:
Yes for type `nw.Datetime` it works correctly. The question I have is about the `nw.Date`  type.

One change with your latest comment compared to my comment [above](https://github.com/vega/altair/pull/3452#issuecomment-2228062681) is that for me the `local_iso_fmt_string` does not contain a `%.f`. 

I thought it was an issue with pyarrow, but if I do:
```python
import pyarrow.compute as pc
pc.strftime(py_tb['date'], format=local_iso_fmt_string).to_pylist()
``` 
It returns 

> ```python
> ['2024-07-15T00:00:00']
> ```

That seems right. And in the narwhals docs you link to (thanks for the link!) it is mentioned that:

> for PyArrow, we replace `"%S.%f"` with `"%S"`.

That makes me even more confused with your statement:

> Practically speaking, if `'2024-07-15'` gets transformed to
> - `'2024-07-15T00:00:00'` for Polars and pandas-like libraries, and
> - `'2024-07-15T00:00:00.000000'` for PyArrow

Thanks for answering so promptly! Appreciated.

MarcoGorelli:
The difference is the `.cast(nw.Datetime)` 😉 If you start with a `datetime`, then PyArrow's `strftime` outputs fractional seconds even if we just specify `'%S'`:
```python

In [1]: import pyarrow as pa

In [2]: from datetime import date, datetime

In [3]: import pyarrow.compute as pc

In [4]: pc.strftime(pa.array([datetime(2020, 1, 1)]), '%Y-%m-%dT%H%:%M%:%S')
Out[4]:
<pyarrow.lib.StringArray object at 0x7fd513149cc0>
[
  "2020-01-01T00%:00%:00.000000"
]
```

mattijn:
Interesting... Let me see if I can find something on this at the pyarrow library. Thanks!

mattijn:
We have to cast the `nw.Date` type to `nw.Datetime` only because polars does not support time directives isn't it? Maybe we check upfront if we have to cast or not. @jonmmease, how important is this? Do you have some suggestions here. Its about this piece:
https://github.com/vega/altair/blob/8b4b3dbeadbe4459eecbe0ad4459e6fb701dcdfe/altair/utils/core.py#L451-L456

MarcoGorelli:
Sure - if you're OK with a teeny-tiny bit of Polars-specific logic, this can work 👍 
```python
        if dtype == nw.Date and nw.get_native_namespace(data) is get_polars():
            # Polars doesn't allow formatting `Date` with time directives.
            # The date -> datetime cast is extremely fast compared with `to_string`
            columns.append(
                nw.col(name).cast(nw.Datetime).dt.to_string(local_iso_fmt_string)
            )
        elif dtype == nw.Date:
            columns.append(nw.col(name).dt.to_string(local_iso_fmt_string))
        elif dtype == nw.Datetime:
            columns.append(nw.col(name).dt.to_string(f"{local_iso_fmt_string}%.f"))
```

altair/utils/core.py
```diff
@@ -650,41 +668,23 @@ def parse_shorthand(
...
+def infer_vegalite_type_for_narwhals(
+    column: nw.Series,
+) -> InferredVegaLiteType | tuple[InferredVegaLiteType, list]:
+    dtype = column.dtype
     if (
-        kind == DtypeKind.CATEGORICAL
-        and column.describe_categorical["is_ordered"]
-        and column.describe_categorical["categories"] is not None
+        nw.is_ordered_categorical(column)
+        and not (categories := column.cat.get_categories()).is_empty()
```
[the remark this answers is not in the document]
MarcoGorelli: 😄 there's a few others if you're brave enough to try out [auto-walrus](https://github.com/MarcoGorelli/auto-walrus)

altair/utils/core.py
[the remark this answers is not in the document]
MarcoGorelli: that's right!

it's a (practically) free operation, `narwhals.DataFrame` just holds a reference to the original dataframe, there's no copying nor conversion involved - getting the original dataframe out again is just a matter of accessing 1-2 properties

https://github.com/narwhals-dev/narwhals/blob/a5276cd6e80781c61143c71041db81bd700a0e12/narwhals/translate.py#L41-L75

altair/utils/core.py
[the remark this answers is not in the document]
MarcoGorelli: niiice! looking forwards to being able to use ibis as a duckdb frontend without pandas+pyarrow being required 🤞

jonmmease (approved):

MarcoGorelli:
thanks @jonmmease ! review comments look great to me, much appreciated 🙏 ✅

binste:
Thanks to everyone involved! This is very exciting and we should cut a release soon :) I'll open a new discussion for it.

PR #3453, opened 2024-06-30
fix: remove remapped `ruff` rule `PLR1701`

dangotbanned:
Fixes warnings here https://github.com/vega/altair/actions/runs/9733301918/job/26860053484?pr=3452

This was due to `ruff` updating around the same time as https://github.com/vega/altair/pull/3431#issuecomment-2195071162
The `SIM101` rule that it is remapped to is already part of our config


mattijn:
Thanks @dangotbanned!

Issue #3454, opened 2024-07-01
Deprecation/breaking changes policy & process


dangotbanned:
# Note
This issue is focused on *how* future changes are made and *how* they are communicated both on release and at runtime.
Any specific reference is used only as an example. 

# Motivation

I've been thinking about this since #3426, which dealt with removing a 4-year old deprecation, but there are some other cases where it could have been helpful:
- https://github.com/vega/altair/pull/3444#discussion_r1656998100
	- Here @binste and I discuss not making a breaking change to `alt.utils.core.infer_encoding_types`
- https://github.com/vega/altair/pull/3420#discussion_r1606779689
	- Here @binste pretty much summing up this issue as a whole for `alt.utils.save.write_file_or_filename`
- https://github.com/vega/altair/pull/3480#discussion_r1681449076
	- Removal of unused private `_wrapper_classes` argument in `SchemaBase.from_dict`
- https://github.com/vega/altair/pull/3482#issuecomment-2241551416
	-  Potential (future) candidates for removal from top-level
- https://github.com/vega/altair/pull/3487
	- Deprecation to type of unused `plugin_type` argument in `PluginRegistry`

I believe this would benefit both users and contributors:
- Stronger guarantees that a set of features can be relied upon
- Greater freedom for contributors to make changes, without concern that they may break code not covered by the test suite

# Open Questions
There are likely many topics that could be discussed, but to get the ball rolling:

- How many major/minor releases should support for deprecated functionality remain?
- Should deprecated features be documented to the same level as their replacements?
	- Deprecated: https://altair-viz.github.io/user_guide/generated/api/altair.selection_single.html#altair-selection-single 
	- Replacement: https://altair-viz.github.io/user_guide/generated/api/altair.selection_point.html#altair.selection_point
- What functions/classes/modules are considered *public API*?
	- Possibly those listed at https://altair-viz.github.io/user_guide/api.html 
- Would changes to *private API* be considered breaking/require deprecation?

# PRs
- #3455 
	- ~~I have a draft~~ PR merged with some ~~ideas on~~ changes that ~~could~~ support versioning & static typing
	- This has some visual examples of how the end user ~~could~~ will be informed at import time
- #3488
	- Early draft PR implementing a policy based on discussions here
- #3509

# Examples from other projects

## `polars`
- [policy](https://docs.pola.rs/development/versioning/#policy-for-breaking-changes)
	- Pretty comprehensive, leaves very little room for ambiguity
[...]

PR #3455, opened 2024-07-01
feat(typing): `@deprecated` versioning, IDE highlighting

dangotbanned:
Supporting PR to attach to #3454 

# Notes

Deprecation should be handled using a decorator recognized by static type checkers:
  - [`@warnings.deprecated`](https://docs.python.org/3.13/library/warnings.html#warnings.deprecated) for `sys.version_info >= (3, 13)`
  - [`@typing_extensions.deprecated`](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.deprecated) for earlier versions
  - ~~*Note: need to see how these handle subclassing*~~ subclassing seems to be unrecognized statically
    - Ideally want to have a `version` keyword-only, required argument
    - Indicating the `altair==x.y.z` this was introduced
  - Consistently use the same decorator to mark all deprecations
    - Updating existing `warnings.warn` & `@utils.deprecation.deprecated`
    - These should provide a reference for any edge cases, that required extending `@typing_extensions.deprecated`

# Demo

The screenshot below shows the differences in behaviour of each decorator:

image

This is the ideal situation, where a type checker can identify the deprecation :

image

https://github.com/dangotbanned/altair/blob/fe1917e3d4986dfa431f428146819622b83b6546/altair/vegalite/v5/api.py#L796-L805

## Console output

<details>
<summary>Sample code block</summary>

```py
import typing_extensions as te
from altair.utils.deprecation import deprecated, AltairDeprecationWarning, deprecate
from altair.vegalite.v5.api import (
    selection,
    selection_point,
    selection_multi,
    selection_single,
)


@deprecated("", version="5.4.0", alternative="new_function")
def old_function(*args):
    return args


@te.deprecated("Deprecated in `altair=3.2.1`.", category=AltairDeprecationWarning)
def old_function_2(*args):
    return args


@deprecate(version="5.4.0", alternative="new_function")
def old_function_3(*args):
    return args


old_function()
old_function_2()
selection("point")
selection_multi()
old_function_3()
selection_point("not deprecated")
```

</details>

At runtime, *Sample code block* outputs:

<details>
<summary>Output</summary>

```log
./../.py:26: AltairDeprecationWarning: alt.old_function was deprecated in `altair=5.4.0`. Use alt.new_function instead.
  old_function()
./../.py:27: AltairDeprecationWarning: Deprecated in `altair=3.2.1`.
  old_function_2()
./../.py:28: DeprecationWarning: Deprecated in `altair=5.0.0`. Use 'selection_point()' or 'selection_interval()' instead.

These functions also include more helpful docstrings.
  selection("point")
./../.py:29: AltairDeprecationWarning: alt.selection_multi was deprecated in `altair=5.0.0`. Use alt.selection_point instead.
  selection_multi()
./../.py:30: AltairDeprecationWarning: Deprecated in `altair=5.4.0`. Use new_function instead.
  old_function_3()

Parameter('not deprecated', SelectionParameter({
  name: 'not deprecated',
  select: PointSelectionConfig({
    type: 'point'
  })
}))
```

</details>

# Summary
- [x] Use a single decorator, with the implementation of `@deprecate` but renaming to the prior `@deprecated`. 
	- `@deprecate` and the screenshot demo are referencing fe1917e3d4986dfa431f428146819622b83b6546 which also has a class implementation
- [x] Add an equivalent (in terms of parameters) warning version for deprecated code paths. 
	- This wouldn't have the same static typing benefits, but would issue consistent messages and reduce repetition like in <details><summary>`alt.param`</summary>
https://github.com/vega/altair/blob/7f05ecbf999766ae1398dcbeb6144c128413026c/altair/vegalite/v5/api.py#L448-L482
</details>

# Rejected Ideas
## Applying `@deprecated` to individual overloads
This is a pretty interesting feature, but it comes with too many downsides.

<details>
<summary>Happy overload</summary>

image

</details>

<details>
<summary>Unhappy overload</summary>

image

</details>

<details>
<summary>Implementation</summary>

image

</details>

- Only applicable to static type checkers, doesn't emit a runtime warning
	- Working around this would be possible, but not trivial
	- E.g. calling `typing.get_overloads`, writing some handler logic
- Having the message displayed **requires** using the `typing_extensions` decorator
	- Which means manually maintaining consistent messages
	- Cannot enforce `version` included
- Even with those conditions met, it still starts with *The function "save" is deprecated ...* - which is misleading
	- The deprecated part is the combination of arguments
	- In this case, the `webdriver` argument
- It adds a lot of extra `@overload` code, which I don't think is warranted given the other issues


altair/utils/deprecation.py
```diff
@@ -1,94 +1,113 @@
...
+    message: LiteralString | None,
+    /,
+) -> LiteralString:
+    output = f"Deprecated in `altair={version}`."
+    if alternative:
+        output = f"{output} Use {alternative} instead."
+    return f"{output}\n{message}" if message else output
 
+
+# NOTE: Annotating the return type breaks `pyright` detecting [reportDeprecated]
```
dangotbanned, line 35: I hope my comments here make sense. 

There is some strange behaviour but this specific combination of quirks makes `pyright` happy

altair/vegalite/v5/api.py
```diff
@@ -23,8 +23,8 @@
     using_vegafusion as _using_vegafusion,
     compile_with_vegafusion as _compile_with_vegafusion,
 )
-from ...utils.data import DataType, is_data_type as _is_data_type
-from ...utils.deprecation import AltairDeprecationWarning
+from altair.utils.data import DataType, is_data_type as _is_data_type
+from altair.utils.deprecation import AltairDeprecationWarning  # noqa: F401
```
dangotbanned, line 27: I've kept this import to avoid merge conflicts. 

Ideally, `AltairDeprecationWarning` should be imported at the top level only, for end-users to add to filters.

altair/utils/deprecation.py
```diff
@@ -1,94 +1,113 @@
...
+# NOTE: `LiteralString` requirement is introduced by stubs
```
dangotbanned, line 36: https://github.com/python/cpython/blob/94f50f8ee6872007d46c385f7af253497273255a/Lib/warnings.py#L559

https://github.com/python/typeshed/blob/e67340b98cda4d5fa207e2ffa1a3af8a03d42501/stdlib/typing_extensions.pyi#L440

The `typeshed` signature must be matched for the visual part

dangotbanned:
@binste 

This isn't a direct fix for [#2947](https://github.com/vega/altair/pull/2927), but I think its an improvement for VSCode users.

image

altair/vegalite/v5/api.py
binste: Can we now remove this line? I can't follow why it needs to be kept here

binste (changes requested):
Nice one! :) Thanks for another great PR.

I only added 1 question. Also, I can't replicate that VS Code strikes through the deprecated functions:

image

I'm getting the deprecation warnings at runtime which is good but is there any way how I can check if pyright correctly identifies the `deprecated` function as being a wrapper around `te.deprecated`? And does mypy support this as well?

altair/vegalite/v5/api.py
dangotbanned: > Can we now remove this line? I can't follow why it needs to be kept here

IIRC, removing this import also removed it from `altair.__init__.py` as it was the only explicit import of it. 

Its an easy fix, but I wanted to avoid messing with import across multiple files at the time (I think)

dangotbanned:
> Nice one! :) Thanks for another great PR.

Thanks @binste 

> I'm getting the deprecation warnings at runtime which is good but is there any way how I can check if pyright correctly identifies the `deprecated` function as being a wrapper around `te.deprecated`? And does mypy support this as well?

In the top-level I've got a workspace (locally) with these settings:

<details>
<summary>Image</summary>

image

</details>




```jsonc
{
	"settings": {
		"python.analysis.diagnosticSeverityOverrides": {
            //"reportDeprecated": "warning",
            "reportInvalidTypeForm": "none",
            "reportMissingImports": "none",
            "reportMissingModuleSource": "none"
            //"reportUndefinedVariable": "none"
        },
		"python.analysis.inlayHints.functionReturnTypes": true,
    	"python.analysis.supportRestructuredText": true,
		"python.analysis.typeCheckingMode": "basic"
	}
}

```

I think `"python.analysis.typeCheckingMode": "basic"` is all that is needed for the ~~strikethrough~~.

The actual diagnostic is `"reportDeprecated"`, so you could try changing that.
When I was testing it, setting that option at all un-special-cases it into a regular warning/error/etc

altair/vegalite/v5/api.py
dangotbanned: > IIRC, removing this import also removed it from altair.__init__.py as it was the only explicit import of it.

Yeah it triggers this failure, I'll resolve with an explicit one further up

```
    def test_completeness_of__all__():
        relevant_attributes = update_init_file.relevant_attributes(alt.__dict__)

        # If the assert statement fails below, there are probably either new objects
        # in the top-level Altair namespace or some were removed.
        # In that case, run `hatch run update-init-file` to update __all__
>       assert alt.__all__ == relevant_attributes
E       AssertionError: assert ['Aggregate',...tString', ...] == ['Aggregate',...tString', ...]
E
E         At index 6 diff: 'AltairDeprecationWarning' != 'Angle'
E         Left contains one more item: 'with_property_setters'
E         Use -v to get more diff

tests\test_toplevel.py:12: AssertionError
========================== 1 failed, 1257 passed, 1 skipped, 3 xfailed, 1 xpassed, 8 warnings in 45.23s ===========================
```

binste (approved):
Thanks for the explanations and the link to the mypy PR! LGTM :)

PR #3458, opened 2024-07-02
feat(typing): adds `Map` alias for `Mapping[str, Any]`

dangotbanned:
This allows a `TypedDict` to be used, where `mypy` previously would have required exactly a `dict`.

The name does not conflict with any existing types/classes in `altair` and has the benefit of being short like `dict`, but more permissive.

I have only added this to `Chart.encode` as I know this does not require the mutability of `dict`, which I cannot confidently say for elsewhere in `altair`.
A future PR could investigate other generated wrappers that use `dict` and either reuse `Map` or add another alias to the `EXTRA_ALIASES` literal.

Fixes https://github.com/vega/altair/pull/3427#discussion_r1662542866

```
cmd [6] | mypy altair tests
tests\vegalite\v5\test_api.py:375: error: Unused "type: ignore" comment  [unused-ignore]
tests\vegalite\v5\test_api.py:376: error: Unused "type: ignore" comment  [unused-ignore]
```

Related https://mypy.readthedocs.io/en/stable/typed_dict.html#typeddict

### DIFF
--- a/altair/vegalite/v5/schema/_typing.py
+++ b/altair/vegalite/v5/schema/_typing.py
@@ -4,9 +4,11 @@
 
 from __future__ import annotations
 
-from typing import Literal
+from typing import Any, Literal, Mapping
 
 from typing_extensions import TypeAlias
+
+Map: TypeAlias = Mapping[str, Any]
 
 AggregateOp_T: TypeAlias = Literal[
     "argmax",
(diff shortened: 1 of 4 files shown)


binste:
Looks good to me, thanks! Agree that it would make sense to relax this type hint also in other places.

dangotbanned:
> Looks good to me, thanks! Agree that it would make sense to relax this type hint also in other places.

thanks @binste 

Could you reopen #3427 please?

Issue #2980, opened 2023-03-21
Include gallery example of point paths on hover from Vega gallery


joelostblom:
I was reviewing the vega examples and think [this one is pretty cool](https://vega.github.io/vega/examples/global-development/) and could be useful to include in the Altair gallery as well, possibly with a search box as mentioned in https://github.com/altair-viz/altair/issues/2760#issuecomment-1477186125

dsmedia:
I've posted a [question](https://stackoverflow.com/questions/78686169/implementing-dimpvis-like-time-based-interaction-in-vega-lite) on Stack Overflow about how this kind of interaction might be implemented in Vega-Lite.

It wasn't immediately clear to me what capabilities Altair might have for this kind of interaction, which I agree is very cool. I figured that understanding the possibilities and limitations in Vega-Lite would be a good first step to determine potential Altair implementations, whether partial or complete.

dsmedia:
I've made progress on porting the Gapminder visualization [example](https://vega.github.io/vega/examples/global-development/) from Vega to Altair, focusing on the hover path and adding a search box. I'd like to share my current implementation and discuss the remaining issues.

### Progress Made
- Implemented a slider for selecting the year
- Added a search box for filtering countries
- Implemented filtering based on both conditions
- Created hover effects for displaying country trajectories and a single country name
- Added a large background year label based on the slider selection

### Remaining Issues
- Hover behavior on invisible points: Currently, hovering over invisible points (i.e., points not selected by the year slider or search box) still results in labels and lines appearing. This is probably not desired.
- Color encoding: The current implementation encodes color by country, not by region as specified in the original Vega example. This makes it harder to discern regional patterns in the data. My attempt to encode color by region ended up distorting the lines on hover.
- There are likely more direct or efficient ways to implement this than the approach I took.

### Next Steps

- Address the hover behavior issue to ensure that only visible points respond to hover events.
- Modify the color encoding to use regions instead of countries, while maintaining the ability to highlight individual country trajectories.
- Add legend back once regional color encoding is fixed.

### Current Implementation
Here's the code for the current state of the visualization:
```python

import altair as alt
from vega_datasets import data

# Data source
source = data.gapminder.url

# X-value slider
x_slider = alt.binding_range(min=1955, max=2005, step=5, name='Year: ')
x_select = alt.selection_point(name="x_select", fields=['year'], bind=x_slider, value=1980)

# Search box for country name
search_box = alt.param(
    value='',
[...]

PR #3459, opened 2024-07-03
docs: Add example with hover path and search box

dsmedia:
- point paths on hover
- search box to filter by country name

Relates to [#2980](https://github.com/vega/altair/issues/2980)

TODO:
- Investigate and potentially correct North Korea and South Korea data in vega_datasets' gapminder json, which appears to have swapped labels.

Thanks for contributing to Altair!

Please follow these guidelines when submitting your PR:

- Describe the purpose of the changes in PR, so that it is easy to understand the implication of the suggested changes.
    - Here is a [helpful article about writing effective PR descriptions](https://medium.com/@greenberg/writing-pull-requests-your-coworkers-might-enjoy-reading-9d0307e93da3).
- Include unit tests and documentation for new features
- Ensure that the title is a concise [semantic commit message](https://www.conventionalcommits.org/) (e.g. "feat: Add `embed_options` to charts").
    - Append `!` if the change is breaking (e.g. "fix!: Raise error when `embed_options` is `None`")


mattijn:
Thanks for this PR! Looks great, but a little bit big for my screen and adoption in the example page.

Can you add `nearest=True` within the following line?:
```python
hover = alt.selection_point(on='mouseover', fields=['country'], empty=False)
```

Another thing, it is recommended to use the approach as is documented in https://github.com/vega/altair/blob/main/CONTRIBUTING.md#adding-examples, especially:

> Once you have an example you would like to add there are a few guide lines to follow. Every example should:
> - have a `arguments_syntax` and `methods_syntax` implementation. Each implementation must be saved as a stand alone script in the `tests/examples_arguments_syntax` and `tests/examples_methods_syntax` directories.

Thanks again for this PR and hopefully you are able to improve this PR using these these suggestions.

dsmedia:
The addition of nearest=True adds a nice visual effect of keeping one path visible at all times once the mouse enters the chart. This serves as a reminder that there is a time element that can be interacted with, not only a cross-section of countries. Without this, a user may not realize there is the option to see both panel and time-series data at the same time, unless they happen to mouse over directly on a mark.

One concern about adding nearest=True is that it somewhat alters the experience of mousing over a country and then trying to scroll through the year slider while keeping in focus that country's path over time, as the movement of the mouse toward the slider tends to trigger *another* country's path to become highlighted as it passes near to another country.

Any thoughts on this? One thought is to keep nearest as False but then to have one country's path shown in the chart's initial state by default. I'm not sure how to implement that, though.

mattijn:
Good point. Maybe like this?
```python
hover = alt.selection_point(on='click', fields=['country'], empty=False, value='Egypt')
```
Its not really a `hover` anymore like this though..

tests/examples_methods_syntax/scatter_point_paths_hover.py
```diff
@@ -0,0 +1,95 @@
...
+
+# Data source
+source = data.gapminder.url
+
+# X-value slider
+x_slider = alt.binding_range(min=1955, max=2005, step=5, name='Year: ')
+x_select = alt.selection_point(name="x_select", fields=['year'], bind=x_slider, value=1980)
+
+# Hover selection
+hover = alt.selection_point(on='mouseover', fields=['country'], empty=False)
```
joelostblom, line 24:
```suggestion
# Hover selection
hover = alt.selection_point(on='mouseover', fields=['country'], empty=False)
# A separate hover for the points since these need empty=True
hover_point_opacity = alt.selection_point(on='mouseover', fields=['country'])
```

I think having the points fade out on hover makes it easier to see the line. See change below for where this is used.

tests/examples_methods_syntax/scatter_point_paths_hover.py
```diff
@@ -0,0 +1,95 @@
...
+)
+
+# Base chart
+base = alt.Chart(source).encode(
+    x=alt.X('fertility:Q', scale=alt.Scale(zero=False), title='Fertility'),
+    y=alt.Y('life_expect:Q', scale=alt.Scale(zero=False), title='Life Expectancy'),
+    color=alt.Color('region_name:N').legend(title='Region'),
+    detail='country:N'
+).transform_calculate(
+    region_name="{'0': 'South Asia', '1': 'Europe & Central Asia', '2': 'Sub-Saharan Africa', '3': 'America', '4': 'East Asia & Pacific', '5': 'Middle East & North Africa'}[datum.cluster]"
```
joelostblom, line 39:
```suggestion
    color=alt.Color('region:N', title='Region'),
    detail='country:N'
).transform_calculate(
    region="""{
        '0': 'South Asia',
        '1': 'Europe & Central Asia',
        '2': 'Sub-Saharan Africa',
        '3': 'America',
        '4': 'East Asia & Pacific',
        '5': 'Middle East & North Africa'
    }[datum.cluster]"""

```

Setting title the same way as for x and y

tests/examples_methods_syntax/scatter_point_paths_hover.py
```diff
@@ -0,0 +1,95 @@
...
+
+background_year = alt.Chart(source).mark_text(
+    align='center',
+    baseline='middle',
+    fontSize=96,
+    opacity=0.2,
+    x='width',
+    y='height',
+    dx=-400,
+    dy=-300
```
joelostblom, line 74:
```suggestion
    x=alt.expr('width/2'),
    y=alt.expr('height/2'),
```

joelostblom (changes requested):
This is great, thank you @dsmedia ! I made a few inline comments with suggestions. Regarding `nearest` I think the current behavior where the country line is not highlighted when the slider is moved is totally fine. I see that as a drill down action to see the entire trajectory of a single (or a few) countries, whereas the slider is more for seeing how all countries evolve over time as a group.

dsmedia:
Thanks for all the feedback @mattijn and @joelostblom. I've incorporated as best I could, while also fixing an issue with hover lines connecting out of sequential order, switched mark_line to mark_trail (which I think gives an intuitive sense of the sequential order) and made some other stylistic modifications, such as setting the country label on hover to appear above the point with the largest y-value for the country. I wasn't sure if there was a consensus about adding nearest=True but chose to leave it out as it seemed to make the hovering experience a bit too bouncy. But I'm happy to add it back in if that's the consensus. I will add the other syntax once the iteration on this vesion is complete. 

@joelostblom could you check the behavior when the enter key is hit on the search box? I'm not sure what's expected but it seems to clear the visualization when I tested it in an interactive window in vscode. Also on the search box, do there happen to be any configuration options (autocomplete?) for the search box that are worth demonstrating here, since I think it'll be its first appearance in an example?

tests/examples_methods_syntax/scatter_point_paths_hover.py
```diff
@@ -0,0 +1,149 @@
...
+    x=alt.X('width:Q', axis=None, scale=None),
+    y=alt.Y('height:Q', axis=None, scale=None)
+).transform_filter(
+    x_select
+).transform_aggregate(
+    year='max(year)'
+).transform_calculate(
+    width='width/2',
+    height='height/2'
+)
```
joelostblom, line 135:
```suggestion
background_year = alt.Chart(source).mark_text(
    align='center',
    baseline='middle',
    fontSize=96,
    opacity=0.2
).encode(
    text='year:O',
).transform_filter(
    x_select
).transform_aggregate(
    year='max(year)'
)
```

The year text appear centered for me by default, is there another reason this is needed?

tests/examples_methods_syntax/scatter_point_paths_hover.py
```diff
@@ -0,0 +1,149 @@
...
+    ),
+    # Point layer
+    base.mark_point(size=50).encode(
+        opacity=alt.condition(hover, alt.value(0.8), alt.value(0)),
+    )
+)
+
+# Year labels
+year_labels = base.mark_text(align='left', dx=5, dy=-5, fontSize=14).encode(
+    text='year:O', color=alt.value('#222222')
```
joelostblom, line 90:
```suggestion
    text='year:O',
    color=alt.value('#222222')
```

tests/examples_methods_syntax/scatter_point_paths_hover.py
```diff
@@ -0,0 +1,149 @@
...
+    }[datum.cluster]"""
+)
+
+# Points that are always visible (filtered by slider and search)
+visible_points = base.mark_circle(size=100).encode(
+    opacity=alt.condition(
+            hover_point_opacity & 
+        alt.expr.test(alt.expr.regexp(search_box, 'i'), alt.datum.country),
+        alt.value(0.8),
+        alt.value(0.1))
```
joelostblom, line 57:
```suggestion
    opacity=alt.condition(
        hover_point_opacity
        & alt.expr.test(alt.expr.regexp(search_box, 'i'), alt.datum.country),
        alt.value(0.8),
        alt.value(0.1)
    )
```

tests/examples_methods_syntax/scatter_point_paths_hover.py
```diff
@@ -0,0 +1,149 @@
...
+search_box = alt.param(
+    value='',
+    bind=alt.binding(input='search', placeholder="Country", name='Search ')
+)
+
+# Base chart
+base = alt.Chart(source).encode(
+    x=alt.X('fertility:Q', scale=alt.Scale(zero=False, nice=False), title='Fertility'),
+    y=alt.Y('life_expect:Q', scale=alt.Scale(zero=False, nice=False), title='Life Expectancy'),
+    color=alt.Color('region:N', title='Region').legend(orient='bottom-left', titleFontSize=14, labelFontSize=12).scale(scheme='dark2'),
```
joelostblom, line 38: Use `legend=` and `scale=` here to be consistent with the argument style used in the rest of the file

tests/examples_methods_syntax/scatter_point_paths_hover.py
```diff
@@ -0,0 +1,149 @@
...
+# Search box for country name
+search_box = alt.param(
+    value='',
+    bind=alt.binding(input='search', placeholder="Country", name='Search ')
+)
+
+# Base chart
+base = alt.Chart(source).encode(
+    x=alt.X('fertility:Q', scale=alt.Scale(zero=False, nice=False), title='Fertility'),
+    y=alt.Y('life_expect:Q', scale=alt.Scale(zero=False, nice=False), title='Life Expectancy'),
```
joelostblom, line 37: I think the default behavior looks better than `nice=False` here. It make the dots less crammed in the top left corner:

nice=False
image


Default:
image

tests/examples_methods_syntax/scatter_point_paths_hover.py
```diff
@@ -0,0 +1,149 @@
...
+    height='height/2'
+)
+
+# Combine all layers
+chart = alt.layer(
+    visible_points, year_labels, country_labels, hover_line, background_year
+).properties(
+    width=500,
+    height=500,
+    padding=35  # Padding ensures labels fit
```
joelostblom, line 143:
```suggestion
    padding=10  # Padding ensures labels fit
```

If we remove `nice=False` as per my comment above, I think we are fine with smaller padding here:

image

joelostblom (changes requested):
Great idea with the trail to indicate direction! I only have a few minor comments, so feel free to address those as you see fit and add the methods syntax version of the example also and then I think we can merge.

Regarding the behavior of the search box when enter is hit, I see the same as you both in VS Code and JupyterLab. I'm not sure how to change it, but we don't need to worry about it here, should probably be fixed on the Vega side of things.

Regarding autocomplete, Vega would need to add support for the `datalist` parameter and a way to pass a datalist to search input. I had opened an issue for it in the past here https://github.com/vega/vega/issues/3702, but didn't get any insights

dsmedia:
I see above "Some checks were not successful" with "lint / ruff-mypy (pull_request)" failing. Is there something I should do to fix that?

dangotbanned:
> I see above "Some checks were not successful" with "lint / ruff-mypy (pull_request)" failing. Is there something I should do to fix that?

Will be fixed once https://github.com/vega/altair/pull/3463 is merged

tests/examples_arguments_syntax/scatter_point_paths_hover.py
```diff
@@ -0,0 +1,148 @@
...
+    fontSize=18,
+    fontWeight='bold'
+).encode(
+    x='fertility:Q',
+    y='life_expect:Q',
+    text='country:N',
+    color=alt.value('black'),
+    opacity=alt.condition(hover, alt.value(1), alt.value(0))
+).transform_window(
+    rank='rank(life_expect)', #
```
joelostblom, line 113:
```suggestion
    rank='rank(life_expect)',
```

tests/examples_methods_syntax/scatter_point_paths_hover.py
```diff
@@ -0,0 +1,141 @@
...
+    fontSize=18,
+    fontWeight='bold'
+).encode(
+    x='fertility:Q',
+    y='life_expect:Q',
+    text='country:N',
+    color=alt.value('black'),
+    opacity=alt.condition(hover, alt.value(1), alt.value(0))
+).transform_window(
+    rank='rank(life_expect)', #
```
joelostblom, line 106:
```suggestion
    rank='rank(life_expect)',
```

joelostblom (approved):
This is looking great! Thanks for all your work in putting this example together and unveiling the data error as well @dsmedia ! Tests should pass now so I will go ahead and merge when they finish running!

PR #3464, opened 2024-07-08
refactor(typing): Reuse generated `Literal` aliases in `api`

dangotbanned:
Following https://github.com/vega/altair/pull/3431 a number of these are now importable.

Additionally, I spotted the `encodings` parameter (e.g. in `selection_point`) was annotated with `str`, but should be constrained to only `SingleDefUnitChannel_T`.

### DIFF
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -85,6 +85,13 @@ if TYPE_CHECKING:
         InlineDataset,
     )
     from altair.expr.core import Expression, GetAttrExpression
+    from .schema._typing import (
+        ImputeMethod_T,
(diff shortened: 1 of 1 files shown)


altair/vegalite/v5/api.py
```diff
@@ -3061,7 +3062,7 @@ def interactive(
             copy of self, with interactive axes added
 
         """
-        encodings = []
+        encodings: list[SingleDefUnitChannel_T] = []
```
dangotbanned, line 3065: I've only annotated the 4x of these to satisfy `mypy`.

binste:
Thanks for further simplifying the maintenance of the type hints!

PR #3465, opened 2024-07-09
docs: Fix camel case of fillOpacity channel

timtroendle:
The spelling of the `fillOpacity` channel was incorrect and ignored the camel case.


binste (approved):
Thanks!

PR #3469, opened 2024-07-10
feat(typing): Further simplify generated `Literal` aliases

dangotbanned:
Follow-up to: https://github.com/vega/altair/pull/3464 and https://github.com/vega/altair/pull/3431

- `Literal`s that only appear as part of a `Union` are now replaced by the `Union` (in all but 1 case)
- `Literal`s in docstrings & `TypeAlias`'s in annotations now appear *after* all other types
- `_typing.py` now has an autogenerated `__all__`
- A nice reduction of **5.5k lines of code**


altair/vegalite/v5/schema/_typing.py
```diff
@@ -38,9 +88,67 @@
...
+    "-shape",
+    "-fillOpacity",
+    "-strokeOpacity",
+    "-opacity",
+    "-text",
+]
 AutosizeType_T: TypeAlias = Literal["pad", "none", "fit", "fit-x", "fit-y"]
 AxisOrient_T: TypeAlias = Literal["top", "bottom", "left", "right"]
 Baseline_T: TypeAlias = Literal["top", "middle", "bottom"]
+BinnedTimeUnit_T: TypeAlias = Literal[
```
dangotbanned, line 122: This was the alias that I was aiming for when I opened this and below is an example of where it would appear.

https://github.com/vega/altair/blob/8b9f6978c6f1b27fd50a4e3a509aecb9b4a94236/altair/vegalite/v5/schema/core.py#L4525-L4563

Everything else was a bonus

tools/schemapi/utils.py
```diff
@@ -101,23 +101,47 @@ def add_literal(
                 tp = alias
         elif (alias := self._literals_invert.get(tp)) and replace:
             tp = alias
+        elif replace and info.is_union_enum():
+            # Handles one very specific edge case `WindowFieldDef`
```
dangotbanned, line 105: Special case referenced in comment

https://github.com/dangotbanned/altair/blob/ac23fd51460e69b1a7189da28b9a10e28eda2ac7/altair/vegalite/v5/schema/core.py#L26482

binste (approved):
Very nice, thanks! :)

PR #3475, opened 2024-07-12
fix(typing): Ignore `[arg-type]` error in `_deduplicate_enum_errors`

dangotbanned:
Fixes `mypy` error that appeared during a merge https://github.com/vega/altair/actions/runs/9905108163/job/27364368641

```
altair/utils/schemapi.py:423: error: Argument 1 to "join" of "str" has incompatible type "Any | Unset"; expected "Iterable[str]"  [arg-type]
Found 1 error in 1 file (checked 367 source files)
```

# Edit
This is ready to merge and is blocking for all other PRs

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -409,20 +409,6 @@ def _group_errors_by_validator(errors: ValidationErrorList) -> GroupedValidation
     return dict(errors_by_validator)
 
 
-def _safe_validator_value_iter(errors: ValidationErrorList, /) -> Iterator[str]:
-    for err in errors:
-        if isinstance(err.validator_value, Iterable):
-            yield ",".join(err.validator_value)
-        else:
-            msg = (
-                "It is assumed that values (and therefore `validator_value`) of an enum are always arrays.\n"
-                "Therefore, this error should be unreachable.\n"
-                "see https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values\n\n"
-                f"{type(err.validator_value).__name__!r} is not `Iterable`."
-            )
-            raise TypeError(msg)
-
-
 def _deduplicate_enum_errors(errors: ValidationErrorList) -> ValidationErrorList:
     """Deduplicate enum errors by removing the errors where the allowed values
     are a subset of another error. For example, if `enum` contains two errors
@@ -431,7 +417,10 @@ def _deduplicate_enum_errors(errors: ValidationErrorList) -> ValidationErrorList
     `enum` list only contains the error with ["A", "B", "C"].
     """
     if len(errors) > 1:
-        value_strings = list(_safe_validator_value_iter(errors))
+        # Values (and therefore `validator_value`) of an enum are always arrays,
+        # see https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values
+        # which is why we can use join below
+        value_strings = [",".join(err.validator_value) for err in errors]  # type: ignore
         longest_enums: ValidationErrorList = []
         for value_str, err in zip(value_strings, errors):
             if not _contained_at_start_of_one_of_other_values(value_str, value_strings):
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -409,6 +409,20 @@ def _group_errors_by_validator(errors: ValidationErrorList) -> GroupedValidation
     return dict(errors_by_validator)
 
 
+def _safe_validator_value_iter(errors: ValidationErrorList, /) -> Iterator[str]:
+    for err in errors:
+        if isinstance(err.validator_value, Iterable):
+            yield ",".join(err.validator_value)
+        else:
+            msg = (
+                "It is assumed that values (and therefore `validator_value`) of an enum are always arrays.\n"
+                "Therefore, this error should be unreachable.\n"
+                "see https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values\n\n"
+                f"{type(err.validator_value).__name__!r} is not `Iterable`."
+            )
+            raise TypeError(msg)
+
+
 def _deduplicate_enum_errors(errors: ValidationErrorList) -> ValidationErrorList:
     """Deduplicate enum errors by removing the errors where the allowed values
     are a subset of another error. For example, if `enum` contains two errors
@@ -417,10 +431,7 @@ def _deduplicate_enum_errors(errors: ValidationErrorList) -> ValidationErrorList
     `enum` list only contains the error with ["A", "B", "C"].
     """
     if len(errors) > 1:
-        # Values (and therefore `validator_value`) of an enum are always arrays,
-        # see https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values
-        # which is why we can use join below
-        value_strings = [",".join(err.validator_value) for err in errors]
+        value_strings = list(_safe_validator_value_iter(errors))
         longest_enums: ValidationErrorList = []
         for value_str, err in zip(value_strings, errors):
             if not _contained_at_start_of_one_of_other_values(value_str, value_strings):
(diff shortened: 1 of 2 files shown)


mattijn:
For my understanding, this is required because of a new version release of jsonschema ([v4.23.0](https://github.com/python-jsonschema/jsonschema/releases/tag/v4.23.0)) or mypy ([changelog](https://github.com/python/mypy/blob/master/CHANGELOG.md))?

dangotbanned:
> For my understanding, this is required because of a new version release of jsonschema ([v4.23.0](https://github.com/python-jsonschema/jsonschema/releases/tag/v4.23.0)) or mypy ([changelog](https://github.com/python/mypy/blob/master/CHANGELOG.md?rgh-link-date=2024-07-12T09%3A45%3A18Z))?

I'm fairly certain it is `jsonschema`, but I don't understand why the un-annotated `validator_value=_unset` in https://github.com/python-jsonschema/jsonschema/pull/1019/files would trigger this

mattijn:
@Julian, quick question: is this something to expect to introduce in packages that are dependent on jsonschema?

dangotbanned:
> @Julian, quick question: is this something to expect to introduce in packages that are dependent on jsonschema?

From a glance at the discussion, I think `Unset` appearing was something they were trying to avoid.

If `mypy` inferred this type to be `Any`, there wouldn't be a need to check it is not `Unset`

dangotbanned:
Okay so I'm more confident this was caused by the stubs being updated (11 hours ago)


https://github.com/python/typeshed/blob/eb0422195eb89810a4bf6f7cc5689c22d72248dd/stubs/jsonschema/jsonschema/exceptions.pyi#L36

This does have the annotation:

```py
validator_value: Any | Unset
```

Which `mypy` is referencing in 
> error: Argument 1 to "join" of "str" has incompatible type "Any | Unset"; expected "Iterable[str]"  [arg-type]

dangotbanned:
@mattijn really appreciate you questioning this in https://github.com/vega/altair/pull/3475#issuecomment-2225215229

There were no runtime changes introduced by `jsonschema`, so I'm simply ignoring this "error" as it is simply a correctness vs library intention issue.

If the `typeshed` stubs are updated in the future, as a result of https://github.com/python-jsonschema/jsonschema/pull/1019#issuecomment-2225773286 - we'll get a `[unused-ignore]` and be able to revert back to the original code.

mattijn:
I'm a bit concerned that this means that the latest released altair version will not pass mypy checks anymore when using the latest released version of jsonschema. Or I'm wrong in thinking this? Does this need to be raised at the typeshed repository? @binste, do you have some insights here, also on mypy_primer. I don't feel technically skilled in regards to type optimizations.

dangotbanned:
> I'm a bit concerned that this means that the latest released altair version will not pass mypy checks anymore when using the latest released version of jsonschema. Or I'm wrong in thinking this? Does this need to be raised at the typeshed repository? @binste, do you have some insights here, also on mypy_primer. I don't feel technically skilled in regards to type optimizations.

@mattijn apologies if I've misunderstood the question, but to break this down:



# `jsonschema`
- The new release of `jsonschema` adds type hints to `_Error`.
- The only parameter that is relevant to this PR is `validator_value`.
	- All that changed was adding the hint of `Any`.
- Nothing was introduced in this release which would change the actual value returned by `_Error.validator_value`

https://github.com/python-jsonschema/jsonschema/pull/1019#issuecomment-2225773286
> From an end-user perspective as I tried to say in this PR, it's just plain incorrect -- the public type of that attribute is `str`, will always be `str`, and it's an implementation detail that in the course of assembling the object we still do this temporary `_Unset` nonsense which we'd never do if it were authored today. But no one downstream using the attribute should be seeing anything but `str`, if they did it would be a bug in this library.

<table>
  <thead>
    <tr>
      <th width="500px">v4.22.0</th>
      <th width="500px">v4.23.0</th>
    </tr>
  </thead>
  <tbody>
  <tr width="600px">
      <td>

```py
class _Error(Exception):
    def __init__(
        self,
        message: str,
        validator=_unset,
        path=(),
        cause=None,
        context=(),
        validator_value=_unset, # <---------------------
        instance=_unset,
        schema=_unset,
        schema_path=(),
        parent=None,
        type_checker=_unset,
    ) -> None: ...

```

</td>
<td>


```py

class _Error(Exception):
    def __init__(
        self,
        message: str,
        validator: str = _unset,  
        path: Iterable[str | int] = (),
        cause: Exception | None = None,
        context=(),
        validator_value: Any = _unset, # <----------------------
        instance: Any = _unset,
        schema: Mapping[str, Any] | bool = _unset, 
        schema_path: Iterable[str | int] = (),
        parent: _Error | None = None,
        type_checker: _types.TypeChecker = _unset, 
    ) -> None: ...

```

</td>
</tr>

  </tbody>
</table>

# `typeshed`
- The corresponding update to `typeshed.stubs.jsonschema` is not equivalent
- To satisfy `mypy` as a result, you would need to check *every* parameter *every* time to confirm it is not `Unset`
- This is not something that the new release of `jsonschema` wanted to achieve, and is mentioned multiple times throughout the PR and 

<table>
  <thead>
    <tr>
      <th width="500px">v4.22.0</th>
      <th width="500px">v4.23.0</th>
    </tr>
  </thead>
  <tbody>
  <tr width="600px">
      <td>

```py
class _Error(Exception):
    def __init__(
        self,
        message: str,
        validator: _utils.Unset | None | protocols.Validator = ...,
        path: Sequence[str | int] = (),
        cause: Incomplete | None = None,
        context: Sequence[ValidationError] = (),
        validator_value=..., # <--------------------
        instance=...,
        schema=...,
        schema_path: Sequence[str | int] = (),
        parent: _Error | None = None,
        type_checker: _utils.Unset | TypeChecker = ...,
    ) -> None: ...

```

</td>
<td>


```py

class _Error(Exception):
    def __init__(
        self,
        message: str,
        validator: str | Unset = sentinel,
        path: Iterable[str | int] = (),
        cause: Exception | None = None,
        context: Sequence[ValidationError] = (),
        validator_value: Any | Unset = sentinel, # <--------------------
        instance: Any | Unset = sentinel,
        schema: Mapping[str, Any] | bool | Unset = sentinel,
        schema_path: Iterable[str | int] = (),
        parent: _Error | None = None,
        type_checker: TypeChecker | Unset = sentinel,
    ) -> None: ...

```

</td>
</tr>

  </tbody>
</table>

----

> I'm a bit concerned that this means that the latest released altair version will not pass mypy checks anymore 

Specifically on this point, `altair` does pass `mypy` checks on this branch.
To the same extent that any usage of *# type: ignore* can be considered passing `mypy`.


> Does this need to be raised at the `typeshed` repository?

I think so, but I'm unsure whether this would need to be raised by @Julian / @DanielNoord

binste (approved):
It's a joy to review PRs which are as well documented as this one! Thanks @dangotbanned for figuring out where the issue came from and explaining it, even with diffs :)

As Dan mentioned, mypy is happy if errors are ignored so that's fine + the error is due to a wrong type hint and not due to something that would cause issues at runtime -> Ignoring the error until type hints are fixed makes sense to me. Especially as Julian has mentioned that it's always `str` despite the different type hints.

I'm not familiar with mypy_primer but if it's easy to get Altair added, I think it would be great if PRs in mypy get tested against Altair code (if I understand it correctly).

dangotbanned:
> It's a joy to review PRs which are as well documented as this one! Thanks @dangotbanned for figuring out where the issue came from and explaining it, even with diffs :)

Thanks @binste 

I wasn't happy with my earlier uncertainty of what caused the issue, but it was an interesting trail to follow

> I'm not familiar with mypy_primer but if it's easy to get Altair added, I think it would be great if PRs in mypy get tested against Altair code (if I understand it correctly).

I'd seen it mentioned before, but it was specifically this https://github.com/python/typeshed/pull/12315#issuecomment-2221519020 which I thought was worth mentioning. 

If I've understood correctly, this bot would've flagged the change in `typeshed` as casuing `altair` to fail - possibly leading to a different set of stubs being released. 

This issue https://github.com/hauntsaninja/mypy_primer/issues/42 seems to be tracking projects to add

PR #3461, opened 2024-07-04
docs: Update `hatch` guidance

dangotbanned:
- Update references to `hatch` commands
- Adds two testing sub-sections in `CONTRIBUTING.md`

This is following hiccups in https://github.com/vega/altair/pull/3452 that can be avoided in the future.

Also a very minor change, but used `bash` for the fenced code blocks to highlight comments

PR #3477, opened 2024-07-13
docs: fix `CONTRIBUTING.md` phrasing

dangotbanned:
The 2 later commits are a follow-up to https://github.com/vega/altair/pull/3461

# Commits
- docs: fix phrasing in *Testing your Changes*
- docs: Use more consistent comments in *Failures on specific python version(s)*
  - These commands are referring to `hatch` environments named like `hatch-test.py3.8`, but not `doc` or `default`.
    - It is simpler to just describe this as python versions.
  - Also the [hatch test](https://hatch.pypa.io/latest/cli/reference/#hatch-test) docs refer to `--python` as accepting multiple versions, but do not give an example with more than one. 
    - I've not figured out how this would work, so this keeps it simple.
- docs: wrap long line


dangotbanned:
https://github.com/vega/altair/actions/runs/9919793684/job/27406003297?pr=3477 
`mypy` error will be resolved following https://github.com/vega/altair/pull/3475

jonmmease (approved):
Thanks!

PR #3480, opened 2024-07-15
refactor(typing): reduce type ignores in `api.py`

dangotbanned:
This PR is mainly working through *type: ignore* / *TODO* comments and fixing what I can. 

My primary motivation for this was to be able to use `pyright` within `altair` more often. 

Despite the existing *type: ignore* comments, I was getting 46 errors in `api` alone.

Anywhere that it appears I've made a change unrelated to existing comments, please know it was to solve one of these:

<details>
<summary>Large screenshot</summary>

image

</details>


tools/schemapi/schemapi.py
```diff
@@ -1085,10 +1085,7 @@ def _default_wrapper_classes(cls) -> Iterator[type[SchemaBase]]:
 
     @classmethod
     def from_dict(
-        cls: type[TSchemaBase],
-        dct: dict[str, Any],
-        validate: bool = True,
-        _wrapper_classes: Iterable[type[SchemaBase]] | None = None,
+        cls: type[TSchemaBase], dct: dict[str, Any], validate: bool = True, **kwds: Any
```
dangotbanned, line 1088: This to to align `Chart.from_dict`, without exposing the `_wrapper_classes` arg over there https://github.com/dangotbanned/altair/blob/419a4e944f231026322e2c4f137e7a3eb94735e8/altair/vegalite/v5/api.py#L2918

pyproject.toml
```diff
@@ -342,6 +342,7 @@ addopts = ["--numprocesses=logical"]
 
 [tool.mypy]
 warn_unused_ignores = true
+pretty = true
```
dangotbanned, line 345: Only recently discovered this https://mypy.readthedocs.io/en/stable/config_file.html#confval-pretty

It underlines the code that caused the error, which is pretty handy IMO

altair/utils/schemapi.py
```diff
@@ -1087,10 +1087,7 @@ def _default_wrapper_classes(cls) -> Iterator[type[SchemaBase]]:
 
     @classmethod
     def from_dict(
-        cls: type[TSchemaBase],
-        dct: dict[str, Any],
-        validate: bool = True,
-        _wrapper_classes: Iterable[type[SchemaBase]] | None = None,
+        cls: type[TSchemaBase], dct: dict[str, Any], validate: bool = True, **kwds: Any
```
dangotbanned, line 1090: See [comment](https://github.com/vega/altair/pull/3480/files#r1679062039)

altair/vegalite/v5/api.py
```diff
@@ -913,10 +933,23 @@ def condition(
 # Top-level objects
 
 
+def _top_schema_base(obj: Any, /):  # -> <subclass of SchemaBase and TopLevelMixin>
```
dangotbanned, line 936: This is similar to a recent change in `utils.core.py` 

I haven't seen this behaviour documented for either `mypy` or `pyright`, but until https://github.com/python/typing/issues/213 has been formalised - this is the best we can do without changing the class hierarchy

https://github.com/vega/altair/blob/9afd374a83f954e79a5fd6e4f8a8a9c4e5381c04/altair/utils/core.py#L855-L876

altair/vegalite/v5/api.py
```diff
@@ -913,10 +933,23 @@ def condition(
...
+    else:
+        msg = f"{type(obj).__name__!r} does not derive from {type(core.SchemaBase).__name__!r}"
+        raise TypeError(msg)
+
+
 class TopLevelMixin(mixins.ConfigMethodMixin):
     """Mixin for top-level chart objects such as Chart, LayeredChart, etc."""
 
     _class_is_valid_at_instantiation: bool = False
+    data: Any
```
dangotbanned, line 952: Fixes all of these, with no runtime changes:
https://github.com/vega/altair/blob/9afd374a83f954e79a5fd6e4f8a8a9c4e5381c04/altair/vegalite/v5/api.py#L2964
https://github.com/vega/altair/blob/9afd374a83f954e79a5fd6e4f8a8a9c4e5381c04/altair/vegalite/v5/api.py#L2870
https://github.com/vega/altair/blob/9afd374a83f954e79a5fd6e4f8a8a9c4e5381c04/altair/vegalite/v5/api.py#L2777
https://github.com/vega/altair/blob/9afd374a83f954e79a5fd6e4f8a8a9c4e5381c04/altair/vegalite/v5/api.py#L2787

altair/vegalite/v5/api.py
```diff
@@ -391,6 +397,20 @@ def check_fields_and_encodings(parameter: Parameter, field_name: str) -> bool:
 _DictOrSchema = Union[core.SchemaBase, typing.Dict[str, Any]]
 _StatementType = Union[core.SchemaBase, _DictOrStr]
 
+
+def _is_undefined(obj: Any) -> TypeIs[UndefinedType]:
```
dangotbanned, line 401: From the commit message:

> Direct copy from another PR https://github.com/dangotbanned/altair/blob/d607c70824860ef3478d7d3996d26dc516d69369/altair/vegalite/v5/api.py#L504

If I were to rewrite this now, I'd probably define it in `schemapi.py`

https://github.com/vega/altair/blob/9afd374a83f954e79a5fd6e4f8a8a9c4e5381c04/tools/schemapi/schemapi.py#L755-L758

I imagine it could be useful in other modules, but for simplicity I've just copied *as-is* from:
- #3427

altair/vegalite/v5/api.py
```diff
@@ -1011,11 +1041,14 @@ def to_dict(
...
 
             # apply theme from theme registry
-            the_theme = themes.get()
-            # Use assert to tell type checkers that it is not None. Holds true
-            # as there is always a default theme set when importing Altair
-            assert the_theme is not None
-            vegalite_spec = utils.update_nested(the_theme(), vegalite_spec, copy=True)
+            if theme := themes.get():
+                vegalite_spec = utils.update_nested(theme(), vegalite_spec, copy=True)
+            else:
```
dangotbanned, line 1046: The previous comment doesn't consider the two following scenarios:

1. Assertions may not be present at runtime https://docs.astral.sh/ruff/rules/assert/
2. How a user would be expected to resolve the issue of there being no active theme. I imagine this assertion failing wouldn't be helpful on its own

altair/utils/schemapi.py
binste: I'm not using pyright so not sure what error was raised here but could we just ignore it? At least looking at the code, we lose the information of what type `_wrapper_classes` can be and there is no need to be able to pass other keyword arguments. I'd prioritize readability over helping out pyright but maybe I'm missing something?

pyproject.toml
binste: That's nice! :)

tools/schemapi/schemapi.py
binste: See comment in corresponding `altair` file

altair/vegalite/v5/api.py
```diff
@@ -1998,13 +2030,13 @@ def transform_filter(
             returns chart to allow for chaining
         """
         if isinstance(filter, Parameter):
-            new_filter: dict[str, bool | str] = {"param": filter.name}
+            new_filter: dict[str, Any] = {"param": filter.name}
```
binste, line 2033: Why does this have to be an Any type? Can FilterTransform actually accept anything other than `bool` or `str`?

altair/vegalite/v5/api.py
binste: Wasn't aware of intersections being discussed, that's going to be very useful with the various mixins used in the Altair code base :) Nice find!

altair/vegalite/v5/api.py
```diff
@@ -2883,7 +2915,9 @@ def _get_name(cls) -> str:
         return f"view_{cls._counter}"
 
     @classmethod
-    def from_dict(cls, dct: dict, validate: bool = True) -> SchemaBase:  # type: ignore[override]  # Not the same signature as SchemaBase.from_dict. Would ideally be aligned in the future
+    def from_dict(
+        cls: type[_TSchemaBase], dct: dict[str, Any], validate: bool = True, **kwds: Any
```
binste, line 2919: Same comment regarding `kwds` here as in `schemapi.py`

altair/vegalite/v5/api.py
```diff
@@ -3357,14 +3380,14 @@ def __init__(self, data=Undefined, hconcat=(), **kwargs):
         self.data, self.hconcat = _combine_subchart_data(self.data, self.hconcat)
         self.params, self.hconcat = _combine_subchart_params(self.params, self.hconcat)
 
-    def __ior__(self, other: NonNormalizedSpec) -> Self:
```
binste, line 3360: Why do we have to remove this? Do we ever pass anything other than a `NonNormalizedSpec`?

binste (changes requested):
Great improvements! I especially like the type intersection :) Added some questions.

altair/utils/schemapi.py
dangotbanned: > I'm not using pyright so not sure what error was raised here but could we just ignore it? At least looking at the code, we lose the information of what type `_wrapper_classes` can be and there is no need to be able to pass other keyword arguments. I'd prioritize readability over helping out pyright but maybe I'm missing something?

So the original error was a `mypy` one in `Chart.from_dict`. 
I've confused myself trying to follow the links I created (sorry), for reference this is what I was fixing:



https://github.com/vega/altair/blob/9afd374a83f954e79a5fd6e4f8a8a9c4e5381c04/altair/vegalite/v5/api.py#L2885-L2890

The problem is defined in the mypy docs [incompatible-overrides](https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides)

Dropping `_wrapper_classes` made sense to me as a solution:
- I cannot find any `altair` code using the argument
- The only modification to the default occurs in `VegaLiteSchema` https://github.com/vega/altair/blob/9afd374a83f954e79a5fd6e4f8a8a9c4e5381c04/altair/vegalite/v5/schema/core.py#L454
	- this redeclares `_default_wrapper_classes` with `Iterator[type[Any]]`
- My assumption that the reason for the ignore, was to avoid displaying `_wrapper_classes` as a parameter for `Chart.from_dict`
- This makes the signatures of all derived classes more flexible
	- which I think is more important than, *what appears to me*, a niche feature

I'm doubting myself now, but does that additional context help explain my choice better?

altair/vegalite/v5/api.py
```diff
@@ -1998,13 +2030,13 @@ def transform_filter(
             returns chart to allow for chaining
         """
         if isinstance(filter, Parameter):
-            new_filter: dict[str, bool | str] = {"param": filter.name}
+            new_filter: dict[str, Any] = {"param": filter.name}
             if "empty" in kwargs:
                 new_filter["empty"] = kwargs.pop("empty")
             elif isinstance(filter.empty, bool):
                 new_filter["empty"] = filter.empty
-            filter = new_filter  # type: ignore[assignment]
```
dangotbanned, line 2006: https://github.com/vega/altair/pull/3480/files/419a4e944f231026322e2c4f137e7a3eb94735e8#r1679700744 

@binste 
> Why does this have to be an Any type? Can FilterTransform actually accept anything other than `bool` or `str`?

That one was a dual fix for the type ignores above and below this comment

altair/vegalite/v5/api.py
dangotbanned: See https://github.com/vega/altair/pull/3480/files/419a4e944f231026322e2c4f137e7a3eb94735e8#r1679846774

I think those errors were self-imposed?

altair/vegalite/v5/api.py
dangotbanned: > Why do we have to remove this? Do we ever pass anything other than a `NonNormalizedSpec`?

I was really confused about this annotation.

All `ChartType` classes would fail this, since they derive from `TopLevelSpec`

https://github.com/vega/altair/blob/9afd374a83f954e79a5fd6e4f8a8a9c4e5381c04/altair/vegalite/v5/schema/core.py#L23616

dangotbanned: @binste to demonstrate the effects of using `NonNormalizedSpec` here, vs no annotation in `TopLevelMixin` 

image

Whereas simply removing this annotation

image

---

So the current annotation results in `pylance` users getting a return type of `Any` and requires them to add a type ignore

dangotbanned:
> Great improvements! I especially like the type intersection :) Added some questions.

Thanks for the review @binste, I've just finished going through your comments.

No rush, but I'll wait for your response before reverting/making those changes

altair/vegalite/v5/api.py
binste: Hmm I can't replicate why I've added this in the first place. Thanks for the explanation and fixing it! Let's remove it as you've already done.

altair/utils/schemapi.py
binste: Interesting. Given that it's not used anywhere in the `altair` code base and it's prefixed with a `_`, what do you think about removing this completely and not accepting any arbitrary `kwds` at all? I'd be very surprised if anyone else has used this.

altair/vegalite/v5/api.py
binste: Not sure I can fully follow but if it gets rid of some type ignores I trust that it makes sense :)

altair/utils/schemapi.py
dangotbanned: > Interesting. Given that it's not used anywhere in the `altair` code base and it's prefixed with a `_`, what do you think about removing this completely and not accepting any arbitrary `kwds` at all? I'd be very surprised if anyone else has used this.

I wouldn't be opposed. 

I'll have a look to see if maybe this was used somewhere historically first. 
I find it quite strange that `_wrapper_classes` appears in the docstring - as well as appearing to be private.

dangotbanned: It was added in `v2.0.0` here, at the same time as the `_default_wrapper_classes` method

- https://github.com/vega/altair/commit/803f42e4d322f1ddcb6dc44621ecdb56a742a8d1

dangotbanned: Based on this PR description, I think we should remove it @binste

> This is a maintenance fix for a subtle bug I came across. The issue was that when running `from_dict`, altair would consider any subclass of `SchemaBase` as a potential wrapper for the generated class hierarchy.
> 
> What this meant is that if you, say, imported both **vega-lite **and **vega** wrappers, you could get cross-talk between them when using from_dict if they had common schemas.
> 
> This PR changes the auto-generated code so that each set of wrappers inherits from its own special base class, which ensures that this kind of cross-talk does not happen.

There are only `vega-lite` wrappers in `v5.0.0`, so this is no longer a concern

altair/vegalite/v5/api.py
dangotbanned: Resolved in https://github.com/vega/altair/pull/3480#discussion_r1681450713

dangotbanned:
@binste as https://github.com/vega/altair/pull/3480#discussion_r1681509490 was the last part to resolve, I thought I'd go ahead with it in [`21a3e84` (#3480)](https://github.com/vega/altair/pull/3480/commits/21a3e84400630f89d916b0d7adcdb71b3803f5a4)

Hopefully that will make this easier to close, as you can just revert that commit if needed

altair/utils/schemapi.py
binste: Nice git-archeology! ;) Thanks for digging it out and I agree, as we've removed the Vega-related code parts in v5, we can get rid of this.

binste (approved):
Thanks again! :)

PR #3466, opened 2024-07-09
feat: Reimplement `expr` as a class that is understood by IDEs

dangotbanned:
Related https://github.com/vega/altair/issues/2904

# `expr`
- all `ConstExpression`s are now read-only properties stored in a metaclass
- `expr` subclasses `ExprRef`, replacing the `__call__` w/ `__new__`
  - No instances of `expr` are ever created
- All `FunctionExpression`s are now classmethods
- Each `Expression` can now be referenced in autocomplete, without needing `IPython`/`Jupyter`

# Docs
- Docstrings have been reformatted to be compatible with `sphinx`
- Broken links in docstrings fixed (e.g. for d3)
- class-level doc for `expr`, with references & doctest
- Added to API Reference
  - uses the section heading proposed in https://github.com/vega/altair/pull/3427#discussion_r1659965947

# Demo
image

## Current (IPython)

https://github.com/vega/altair/assets/125183946/8ac49a65-c6cd-4e25-9d65-632253fa0b15

## Current (Static)

https://github.com/vega/altair/assets/125183946/ad30bba5-e126-41b8-910b-dc8b922ff896

## Proposed (Static)

https://github.com/vega/altair/assets/125183946/3f65d5d5-a006-421e-8037-a3dbfa7d796e

## Proposed (API Reference)

image

image

PR #3482, opened 2024-07-16
fix: Restrict static & runtime top-level imports

dangotbanned:
Fixes: https://github.com/vega/altair/discussions/3478#discussioncomment-10062733
Supersedes: https://github.com/vega/altair/pull/2927

# Description

This PR is primarily fixing `pylance` autocompletion that has regressed since https://github.com/vega/altair/pull/3431

However, this also required some runtime changes, which have been spread across commits to more clearly document what is/isn't accessible and where.
- The main fix comes from 2864c0789dbb632edb392d5617d6415be586c18e
- I'd recommend reviewing the commits sequentially, as a lot of what comes after is organizing/re-writing imports.

---

# Autocomplete

## Before
image

## After
- No `_T` aliases
- No* stdlib (e.g. `alt.annotations`)
	- `Any` will be removed following https://github.com/vega/altair/pull/3466
	- All others no longer appear for me in `pylance`

image

# `__all__` changes

Every instance of `from module_name import *` has either:
 1. `module_name` declares an `__all__`, which is what gets imported
 2. Every `import *` can be followed until reaching a defined `__all__`

https://github.com/user-attachments/assets/f2ebe3eb-1e58-4c76-95b0-9fb5baa15bc5

---

# Extra

An added bonus seems to be `@deprecated` is being recognised in `tests/**`
*Note:* deprecated functions *still appear*, but the visual cue & warning will be displayed

image


tests/vegalite/v5/test_api.py
```diff
@@ -23,6 +23,7 @@
 import polars as pl
 
 import altair as alt
+from altair.utils.schemapi import Undefined
```
binste, line 26: What's the reasoning behind importing `Undefined` from the submodule?

binste (commented):
Thanks for taking care of this! I've added one question as a comment. Also, in https://github.com/vega/altair/issues/3454#issuecomment-2232798877 you say that this PR is an opportunity to discuss the boundaries of what we want to consider as breaking. Are you aware of anything in this PR that could be considered a breaking change? Probably the removal of some entries from the top-level Altair namespace? I'm wondering if we want to re-add them and only fix the pylance related autocompletion in this PR. Cleaning up the top-level namespace could be a bigger initiative to tackle for v6. thoughts?

tests/vegalite/v5/test_api.py
dangotbanned: This was a similar change to that of [v5/test_params](https://github.com/vega/altair/pull/3482/files/1e0c12d2e10ddf23636233d1df6f432cccbe861d#diff-5bc52267a53307a825084d94af60cbf84eee6b53fa642e206a469a641802bc03).

While currently `Undefined` is available at the top-level, this may not be the case in a future version (hypothetically).

Using a *more* explicit import isolates these tests to the core functionality they are testing (`v5/api.py`) without imposing the dependencies of high level packages.

---

### Some scenarios
- `Undefined` is removed from top-level
	- This version would not need to be modified
	- Existing version would need updating
		- Then requires trying to follow where the new import location would be
- `Undefined` is removed from `altair.utils.__all__`
	- This version leaves a clear trace of where it was before, hinting where you should look for changes
		- May not even require a change, if it remained in `schemapi.py`, but was exported to `utils.__all__`
	- Existing version may/may not need updating
		- But since we relied on a higher level export, the origin of `Undefined` isn't clear

---

When refactoring code, things like [rename module](https://code.visualstudio.com/docs/python/editing#_rename-module) and [move symbol](https://code.visualstudio.com/docs/python/editing#_move-symbol) will be able to use these absolute imports and keep them in sync - since they are clearly traceable to the origin of the symbol.

dangotbanned:
> Thanks for taking care of this! I've added one question as a comment. 

Thank you @binste for reviewing!

> Also, in [#3454 (comment)](https://github.com/vega/altair/issues/3454#issuecomment-2232798877) you say that this PR is an opportunity to discuss the boundaries of what we want to consider as breaking. Are you aware of anything in this PR that could be considered a breaking change? Probably the removal of some entries from the top-level Altair namespace? I'm wondering if we want to re-add them and only fix the pylance related autocompletion in this PR. Cleaning up the top-level namespace could be a bigger initiative to tackle for v6. thoughts?

*I will update this with a itemized list of the `__all__` modifications.*

---

To elaborate on [#3454 (comment)](https://github.com/vega/altair/issues/3454#issuecomment-2232798877), this is an effort to define what we consider [**public API**](https://github.com/vega/altair/discussions/3478#discussioncomment-10104230).

I was hoping that by making some opinionated changes, that it would be easier for any maintainer to weigh in with either:
- a correction (identification of public API)
- an approval (identification of private API)

So for something like [`973dfc7` (#3482)](https://github.com/vega/altair/pull/3482/commits/973dfc7613b50b21b083b703dc46c0228f9a1c7b) I hoped to open a discussion on
- Do we expect users to rely on these directly?
- What is the benefit of exposing such symbols at the top-level?
- Were these exposed intentionally?

---

Does that help paint a clearer picture of what I was going for?

tests/vegalite/v5/test_api.py
binste: While I'd consider it unlikely right now that we remove `Undefined` from the top-level as it might break a lot of user code (personally, I'm working with a codebase which contains a lot of `alt.Undefined` references), I also don't see a downside so fine me. Thanks for the thorough explanation!

binste:
This helps, thanks! May I propose that this PR is changed to only fix the regression in the pylance autocompletion without removing elements from the top-level namespace. I think it's great that you push the discussion forward of what we want to consider the "public" part of the API and what guarantees we offer users. I see a lot of value in defining this.

I'd prefer if we look at defining the public API in one go holistically by going through all elements in a systematic way. By doing it in one go, we can hopefully reduce frictions for end users and provide clear indication of the changes, migration guides, and a guarantee of what is public going forward. It will involve tradeoffs between making Altair better for future users while not breaking too many things for existing codebases and users. Discussing this topic with an overview of all the parts of the package hopefully allows us to get this right.

Thanks again for continue to push Altair in a better direction! Much appreciated :)

dangotbanned:
> Thanks again for continue to push Altair in a better direction! Much appreciated :)

Thank you @binste 

> This helps, thanks! May I propose that this PR is changed to only fix the regression in the pylance autocompletion without removing elements from the top-level namespace.

That is absolutely fine with me!
I'll work on reverting those changes and simply link to my [suggestions](https://github.com/vega/altair/pull/3482#issuecomment-2241551416) in the *Motivation* section of https://github.com/vega/altair/issues/3454 for future reference

altair/__init__.py
dangotbanned, line 1: @binste `altair.__all__` is now identical to main 👍

binste:
> That is absolutely fine with me! I'll work on reverting those changes and simply link to my [suggestions](https://github.com/vega/altair/pull/3482#issuecomment-2241551416) in the _Motivation_ section of #3454 for future reference

That's perfect, thank you! :)

PR #3485, opened 2024-07-19
docs: Add example for *Bar Chart with Highlighting on Hover and Selection on Click*

dangotbanned:
Fixes https://github.com/vega/altair/issues/3301

Utilises `alt.when`, introduced in https://github.com/vega/altair/pull/3427

image

---

@mattijn I thought adding this first would be something to refer to in the follow-up PR discussed in https://github.com/vega/altair/pull/3427#issuecomment-2237588984

Also revisiting since https://github.com/vega/altair/issues/2759#issuecomment-2138158202 I have a better explanation for the `empty=False` placement.

For `select`, the `Parameter` is used twice - once with `empty=False` and another with the default.
Whereas `highlight` is used only once.

I think this could be helpful to add to the `alt.condition(empty)` and each `.when(empty)` doc, as this is a good justification for why you can pass `empty` there.

https://github.com/dangotbanned/altair/blob/ef81573b9c709989d3965f778a5b0c5d77f6d4c3/tests/examples_methods_syntax/interactive_bar_select_highlight.py#L26-L45

### DIFF
--- a/tests/examples_arguments_syntax/interactive_bar_select_highlight.py
+++ b/tests/examples_arguments_syntax/interactive_bar_select_highlight.py
@@ -27,10 +27,8 @@ select = alt.selection_point(name="select", on="click")
 highlight = alt.selection_point(name="highlight", on="pointerover", empty=False)
 
 stroke_width = (
-    alt.when(select)
-    .then(alt.value(2, empty=False))
-    .when(highlight)
-    .then(alt.value(1))
+    alt.when(select).then(alt.value(2, empty=False))
+    .when(highlight).then(alt.value(1))
     .otherwise(alt.value(0))
 )
 
--- a/tests/examples_arguments_syntax/interactive_bar_select_highlight.py
+++ b/tests/examples_arguments_syntax/interactive_bar_select_highlight.py
@@ -1,0 +1,47 @@
+"""
+Bar Chart with Highlighting on Hover and Selection on Click
+-----------------------------------------------------------
+This example shows a bar chart with highlighting on hover and selecting on click. (Inspired by Tableau's interaction style.)
+
+Based on https://vega.github.io/vega-lite/examples/interactive_bar_select_highlight.html
+"""
+
+# category: interactive charts
+import altair as alt
+
+source = {
+    "values": [
+        {"a": "A", "b": 28},
+        {"a": "B", "b": 55},
+        {"a": "C", "b": 43},
+        {"a": "D", "b": 91},
+        {"a": "E", "b": 81},
+        {"a": "F", "b": 53},
+        {"a": "G", "b": 19},
+        {"a": "H", "b": 87},
+        {"a": "I", "b": 52},
+    ]
+}
+
+select = alt.selection_point(name="select", on="click")
+highlight = alt.selection_point(name="highlight", on="pointerover", empty=False)
+
+stroke_width = (
+    alt.when(select)
+    .then(alt.value(2, empty=False))
+    .when(highlight)
+    .then(alt.value(1))
+    .otherwise(alt.value(0))
+)
+
+
+alt.Chart(
+    source,
+    height=200,
+    config=alt.Config(scale=alt.ScaleConfig(bandPaddingInner=0.2)),
+).mark_bar(fill="#4C78A8", stroke="black", cursor="pointer").encode(
+    x="a:O",
+    y="b:Q",
+    fillOpacity=alt.when(select).then(alt.value(1)).otherwise(alt.value(0.3)),
+    strokeWidth=stroke_width,
+).add_params(select, highlight)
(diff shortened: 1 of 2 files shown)


tests/examples_arguments_syntax/interactive_bar_select_highlight.py
```diff
@@ -0,0 +1,47 @@
...
+select = alt.selection_point(name="select", on="click")
+highlight = alt.selection_point(name="highlight", on="pointerover", empty=False)
+
+stroke_width = (
+    alt.when(select)
+    .then(alt.value(2, empty=False))
+    .when(highlight)
+    .then(alt.value(1))
+    .otherwise(alt.value(0))
+)
```
mattijn, line 35:
```suggestion
stroke_width = (
    alt.when(select).then(alt.value(2, empty=False))
    .when(highlight).then(alt.value(1))
    .otherwise(alt.value(0))
)
```

tests/examples_methods_syntax/interactive_bar_select_highlight.py
```diff
@@ -0,0 +1,45 @@
...
+select = alt.selection_point(name="select", on="click")
+highlight = alt.selection_point(name="highlight", on="pointerover", empty=False)
+
+stroke_width = (
+    alt.when(select)
+    .then(alt.value(2, empty=False))
+    .when(highlight)
+    .then(alt.value(1))
+    .otherwise(alt.value(0))
+)
```
mattijn, line 35:
```suggestion
stroke_width = (
    alt.when(select).then(alt.value(2, empty=False))
    .when(highlight).then(alt.value(1))
    .otherwise(alt.value(0))
)
```

mattijn (approved):
Thanks! If you have more insight where to place `empty=False`, that would be great. I thought we could control everything within the `alt.selection_point()`, but that seems not to be the case.

I added two suggestions that shows on a single line the `when().then()`'s

dangotbanned:
> Thanks! If you have more insight where to place `empty=False`, that would be great. I thought we could control everything within the `alt.selection_point()`, but that seems not to be the case.
> 
> I added two suggestions that shows on a single line the `when().then()`'s

Thanks for the review @mattijn 

> ```py
> 
> stroke_width = (
>     alt.when(select).then(alt.value(2, empty=False))
>     .when(highlight).then(alt.value(1))
>     .otherwise(alt.value(0))
> )
> 
> ```

I've no strong feelings on the formatting (happy to change), but two points I wanted to mention:

1. I simply ran:
```cmd
>>> ruff format ./tests/*/interactive_bar_select_highlight.py
```

I did notice `ruff` seems to be disabled for the examples, so if this changed in the future - we'd end up back at the original.

2. If a user copies the example, and is using `ruff | black` - they may find the result surprising/unfamiliar.

---

I would say generally though I'd be in favor of running all the examples through `ruff`, for greater consistency:
- https://github.com/vega/altair/blob/80939e74ef4b5b2fc2c5dbb9ea30d2478d523b9c/tests/examples_arguments_syntax/hexbins.py
- https://github.com/vega/altair/blob/80939e74ef4b5b2fc2c5dbb9ea30d2478d523b9c/tests/examples_arguments_syntax/distributions_and_medians_of_likert_scale_ratings.py
- https://github.com/vega/altair/blob/80939e74ef4b5b2fc2c5dbb9ea30d2478d523b9c/tests/examples_arguments_syntax/bar_chart_with_range.py
- https://github.com/vega/altair/blob/80939e74ef4b5b2fc2c5dbb9ea30d2478d523b9c/tests/examples_arguments_syntax/area_chart_gradient.py
- https://github.com/vega/altair/blob/80939e74ef4b5b2fc2c5dbb9ea30d2478d523b9c/tests/examples_arguments_syntax/candlestick_chart.py
- https://github.com/vega/altair/blob/80939e74ef4b5b2fc2c5dbb9ea30d2478d523b9c/tests/examples_arguments_syntax/comet_chart.py
- https://github.com/vega/altair/blob/80939e74ef4b5b2fc2c5dbb9ea30d2478d523b9c/tests/examples_arguments_syntax/dendrogram.py

mattijn:
From a programming perspective I can understand that you just want to `ruff` everything. 
But the reason why autoformatting is disabled on the examples is because the autoformatted syntax is often producing results of altair syntax that I would not advocate in writing as a user. If the user decide to autoformat the code once finished, that is fine by me. What do you think?

dangotbanned:
> What do you think?

@mattijn all good with me, I'll accept the suggestions and we can get this merged

dangotbanned:
@mattijn https://mypy-lang.blogspot.com/2024/07/mypy-111-released.html?m=1

🤦

dangotbanned:
#3487 will fix the unrelated `mypy` issue

mattijn:
Thanks! Tests pas now🥳

PR #3488, opened 2024-07-20
docs: Versioning policy

dangotbanned:
Will close: https://github.com/vega/altair/issues/3454

Based on discussions in https://github.com/vega/altair/issues/3454#issuecomment-2201979525 and https://github.com/vega/altair/issues/3454#issuecomment-2232623054

Public API definition ~~is~~ was blocking, see discussion in https://github.com/vega/altair/discussions/3478#discussioncomment-10171851.

# Tasks
- [x] Background on what versioning has been used so far
- [x] Define the [**Public API**](https://semver.org/#spec-item-1)
- [x] Define deprecation process 
	- [x] Period(s)
		- https://github.com/vega/altair/issues/3454#issuecomment-2203434526
		- Mostly *when* can changes be introduced, how *long* can users expect deprecated support
	- [x] Warnings
		- https://github.com/vega/altair/pull/3455
		- https://docs.pola.rs/development/versioning/#deprecation-warnings


dangotbanned:
# Special handling of upstream `vega-lite` schema changes
I've moved this from the main task list, since I'm not sure of a good solution but maybe someone else has an idea.

- Should `altair` support (w/ deprecation) input that is no longer compliant with the `vega-lite` schema?
- See  `alt.param` as of [`v5.0.0`](https://github.com/vega/altair/releases/tag/v5.0.0)

https://github.com/vega/altair/blob/7f05ecbf999766ae1398dcbeb6144c128413026c/altair/vegalite/v5/api.py#L448-L482

## Issues
The example I've given highlights the conflict between how [SemVer](https://semver.org/#how-should-i-handle-deprecating-functionality) specifies deprecation and the most recent major version of `altair`.

`v5.0.0` *should* be where a **breaking** change is made, but was instead a **backwards-compatible** change.

This is complicated by (AFAIK) the change occuring - without deprecation - in `vega-lite`.

dangotbanned:
@mattijn Do you think there is a need to state here about the level of docs for deprecated functionality?
E.g. removal from API Reference, reduced docstrings
- #3509

mattijn:
I trust you here👍

binste:
> # Special handling of upstream `vega-lite` schema changes
> I've moved this from the main task list, since I'm not sure of a good solution but maybe someone else has an idea.
> 
> * Should `altair` support (w/ deprecation) input that is no longer compliant with the `vega-lite` schema?
> * See  `alt.param` as of [`v5.0.0`](https://github.com/vega/altair/releases/tag/v5.0.0)
> 
> https://github.com/vega/altair/blob/7f05ecbf999766ae1398dcbeb6144c128413026c/altair/vegalite/v5/api.py#L448-L482
> 
> ## Issues
> The example I've given highlights the conflict between how [SemVer](https://semver.org/#how-should-i-handle-deprecating-functionality) specifies deprecation and the most recent major version of `altair`.
> 
> `v5.0.0` _should_ be where a **breaking** change is made, but was instead a **backwards-compatible** change.
> 
> This is complicated by (AFAIK) the change occuring - without deprecation - in `vega-lite`.

First thought is that if something is no longer compliant with the VL schema and it leads to a breaking change, we have to support it until the next major version following the policy you're drafting. If it's a feature which we know was heavily used by users and we can write some compatibility code around it, we can optionally consider deprecating it and still supporting it for a while. If we can make the lives of our users easier, I'm all for it :) What do you think?

dangotbanned:
> First thought is that if something is no longer compliant with the VL schema and it leads to a breaking change, we have to support it until the next major version following the policy you're drafting. If it's a feature which we know was heavily used by users and we can write some compatibility code around it, we can optionally consider deprecating it and still supporting it for a while. If we can make the lives of our users easier, I'm all for it :) What do you think?

Thanks @binste, reading your comment and then re-reading the policy I've drafted has made me think I've overcomplicated this. 
Will adapt that into the next draft

dangotbanned:
Ready for review @binste @mattijn 

Feeling confident all the ideas that came up are now documented.

However, I'm not a writer, so please feel free to *directly* edit to taste

mattijn (approved):

joelostblom (approved):
This looks good for me, thanks for putting it together.

dangotbanned:
Thanks both @joelostblom @mattijn

binste (approved):
Precise and succinct, great!

PR #3490, opened 2024-07-21
docs: Add `empty` as a explicit `condition` kwarg

dangotbanned:
Partial fix:
- https://github.com/vega/altair/pull/3485#pullrequestreview-2189012089
- https://github.com/vega/altair/pull/3485#issue-2418667202

This was already being [handled internally](https://github.com/vega/altair/blob/24e03154fb93c26eccf72d0d31fc6bc777cc0526/altair/vegalite/v5/api.py#L1516), but now aligns more closely with `when` and provides a documentable argument.

I was hoping to get this doc approved, [before moving onto](https://github.com/vega/altair/discussions/3478#discussioncomment-10104219) `when-then-otherwise` as this is a smaller change - but will be reused there multiple times

### DIFF
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -1480,15 +1480,6 @@ def condition(
 @overload
 def condition(
     predicate: _PredicateType,
-    if_true: str,
-    if_false: str,
-    *,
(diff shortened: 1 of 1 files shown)


mattijn:
Thanks!

PR #3501, opened 2024-07-24
feat: Support a wider range of iterables in `SchemaBase.to_dict`

dangotbanned:
Fixes: #2808 and #2877

Supersedes https://github.com/vega/altair/pull/3237

# CI Failures

See https://github.com/vega/altair/pull/3501#discussion_r1690163239

# Notes

Huge thanks to @MarcoGorelli as `narwhals` made this feature super simple to implement.

In addition to handling `Sequence` like types, an `Iterator` can safely be passed to `narwhals.from_native` without being consumed.

<details>
<summary>Code block</summary>

```py
import narwhals.stable.v1 as nw

elements = [1,2,3]
it = iter(elements)
some_obj = nw.from_native(it,strict=False)

>>> list(it) == elements
True
```

</details>

---

# #2808 Cases

<details>
<summary>Large screenshots</summary>

image
image

</details>

# #2877 Case

<details>
<summary>Large screenshot</summary>

image

</details>

# Tasks

- [x] Update User Guide


altair/utils/schemapi.py
```diff
@@ -32,6 +32,7 @@
 import jsonschema.exceptions
 import jsonschema.validators
 from packaging.version import Version
+import narwhals as nw
```
MarcoGorelli, line 35: perhaps use `narwhals.stable.v1`, as that's what's used in the rest of the codebase?

dangotbanned: > perhaps use `narwhals.stable.v1`, as that's what's used in the rest of the codebase?

Thanks @MarcoGorelli, will give this a try later.

Should I be expected any behavioural changes from `narwhals.from_native`?

MarcoGorelli: at the moment they should be identical, just trying to future-proof things 😎

dangotbanned: Makes sense, cheers

tests/utils/test_schemapi.py
```diff
@@ -524,23 +528,23 @@ def chart_error_example__wrong_tooltip_type_in_faceted_chart():
...
+        .encode(tooltip=[{5000}])
         .facet()
     )
 
 
 def chart_error_example__wrong_tooltip_type_in_layered_chart():
     # Error: Wrong data type to pass to tooltip
     return alt.layer(
-        alt.Chart().mark_point().encode(tooltip=[{"wrong"}]),
+        alt.Chart().mark_point().encode(tooltip=[{5000}]),
```
dangotbanned, line 539: These all needed updating since `set` is an `Iterable`.

tools/schemapi/schemapi.py
```diff
@@ -1226,6 +1241,12 @@ def _is_list(obj: Any | list[Any]) -> TypeIs[list[Any]]:
     return isinstance(obj, list)
 
 
+def _is_iterable(
+    obj: Any, *, exclude: type | tuple[type, ...] = (str, bytes)
```
dangotbanned, line 1245: `(str, bytes)` chosen as they are the remaining `Iterable` types rom https://github.com/vega/altair/pull/3237#issuecomment-1806868507

`dict` is handled in an earlier case and `type` isn't `Iterable`

altair/utils/schemapi.py
```diff
@@ -486,6 +487,14 @@ def _subclasses(cls: type[Any]) -> Iterator[type[Any]]:
             yield cls
 
 
+def _from_array_like(obj: Iterable[Any], /) -> list[Any]:
+    try:
+        ser = nw.from_native(obj, strict=True, series_only=True)
+        return ser.to_list()
+    except TypeError:
+        return list(obj)
```
dangotbanned, line 495: Since anything that got here **must** be an `Iterable`, it will always be accepted by `list` https://docs.python.org/3/library/stdtypes.html#list

altair/utils/schemapi.py
```diff
@@ -510,10 +519,16 @@ def _todict(obj: Any, context: dict[str, Any] | None, np_opt: Any, pd_opt: Any)
             for k, v in obj.items()
             if v is not Undefined
         }
-    elif hasattr(obj, "to_dict"):
+    elif (
+        hasattr(obj, "to_dict")
+        and (module_name := obj.__module__)
+        and module_name.startswith("altair")
```
joelostblom, line 525: What are these additional checks for and why weren't they needed previously?

tests/utils/test_schemapi.py
joelostblom: I'm not understanding the reason for updating these. I'm getting the same error message with both `"wrong"` and `5000`. What difference do you see on your end?

```
'['wrong']' is an invalid value for `field`. Valid values are of type 'string' or 'object
```

tests/utils/test_schemapi.py
```diff
@@ -948,3 +952,37 @@ def test_to_dict_expand_mark_spec():
...
+        list,
+        deque,
+        pl.Series,
+        pd.Series,
+        pd.Index,
+        pd.Categorical,
+        pd.CategoricalIndex,
+        np.array,
+    ],
+)
```
joelostblom, line 974: Should this include `set`?

joelostblom (commented):
Looking great overall @dangotbanned , thank you again! A few comments / questions.

altair/utils/schemapi.py
dangotbanned: Calling `list(iterable)` will defer to `iterable.__iter__`.

At least in the case of `pandas` and `polars`, calling `__iter__` vs `to_list` will perform a different operation.

`pandas`:
- [`__iter__`](https://github.com/pandas-dev/pandas/blob/63dc1bb4f99d24b46bacb113d740d54459fdbe5e/pandas/core/base.py#L873)
- [`to_list`](https://github.com/pandas-dev/pandas/blob/63dc1bb4f99d24b46bacb113d740d54459fdbe5e/pandas/core/base.py#L871)

`polars`:
- [`__iter__`](https://github.com/pola-rs/polars/blob/9978d88b15f3ea85195c36b1060d1d9f77369c93/py-polars/polars/series/series.py#L1224)
- [`to_list`](https://github.com/pola-rs/polars/blob/9978d88b15f3ea85195c36b1060d1d9f77369c93/py-polars/polars/series/series.py#L4036)

With the `polars` case, the simplest explanation is that explicitly asking for a `list` allows more of the operation to be performed in `rust`.

---

> which would make the code cleaner?

I'm not sure I agree on this part. 
I thought using `nw.from_native(obj, strict=True, series_only=True)` was:
- Very explicit in what is happening
- Defers a huge amount of type checking to an external library
- Ensures the most performant path to a `list` is taken
- Can potentially support even more types in the future, by simply changing the import to `import narwhals.stable.v2`

---

@joelostblom if you think this info would be helpful in the docstring for `_from_array_like` - I'd be happy to add it

altair/utils/schemapi.py
dangotbanned: That was added in [`3275c5e` (#3501)](https://github.com/vega/altair/pull/3501/commits/3275c5efa837e52f8d7f40f297cd298879d5b6a7)

> fix: Prevent `to_dict` method being called on `pd.Series`
> AFAIK, this is the intended case for `Parameter | Expression` not for converting arbitrary objects

To add to that, it is because that branch is *before*:

```py
elif _is_iterable(obj, exclude=(str, bytes)):
	return _todict(_from_array_like(obj), context, np_opt, pd_opt)
```

So that code would never be reached for any object like:

```py
from typing import Any, Iterable, Iterator

class SomeObj(Iterable):
    def to_dict(self) -> Iterator[Any]: ...
```

`pd.Series` met these criteria during testing

tests/utils/test_schemapi.py
dangotbanned: > I'm not understanding the reason for updating these. I'm getting the same error message with both `"wrong"` and `5000`. What difference do you see on your end?
> 
> ```
> '['wrong']' is an invalid value for `field`. Valid values are of type 'string' or 'object
> ```

This one was subtle enough that it took me a while when looking at to notice the problem.

> ```
> '{'wrong'}' is an invalid value for `field`. Valid values are of type 'string' or 'object'.
> ```

```py

original: list[set[str]] = [{"wrong"}]
replaced: list[set[int]] = [{5000}]

```

The original tests relied of `{"wrong"}` being passed through the bottom case of `_to_dict`:
https://github.com/dangotbanned/altair/blob/438ddf994230a547d01349a572404e21e65ae54f/altair/utils/schemapi.py#L533

However, the set elements are now iterated over in:
https://github.com/dangotbanned/altair/blob/438ddf994230a547d01349a572404e21e65ae54f/altair/utils/schemapi.py#L530

---

So this is why the error now reads as a `list`:

> \\[5000\\]

tests/utils/test_schemapi.py
dangotbanned: > Should this include `set`?

It could, but would require adapting the test as bit more to account for `set` not being ordered:

```py
    def test_to_dict_iterables(tp, expected) -> None:
        x_dict = alt.X("x:N", sort=tp(expected)).to_dict()
        actual = x_dict["sort"]  # type: ignore
>       assert actual == expected
E       AssertionError: assert ['e', 'b', 'c', 'a', 'f', 'd'] == ['c', 'd', 'f', 'a', 'b', 'e']
E
E         At index 0 diff: 'e' != 'c'
E         Use -v to get more diff

tests\utils\test_schemapi.py:979: AssertionError
================================================================================================ short test summary info ================================================================================================ 
FAILED tests/utils/test_schemapi.py::test_to_dict_iterables[set-expected0] - AssertionError: assert ['e', 'b', 'c', 'a', 'f', 'd'] == ['c', 'd', 'f', 'a', 'b', 'e']
============================================================================================= 1 failed, 19 passed in 18.88s ============================================================================================= 

```

Related to this, all the types I chose are `Sequence` except the `numpy.array` as mentioned here https://github.com/vega/altair/issues/2877

So while internally we are using `Iterable`, currently all of the public API uses the annotation `Sequence` and `set` is the former but not the latter.

dangotbanned:
> Looking great overall @dangotbanned , thank you again! A few comments / questions.

Thanks for taking the time to review @joelostblom 

Will wait for your response before making any changes

joelostblom:
Thanks for responding to the comment @dangotbanned ! At a cursory glance it all looks good but I will review more in detail and try it out when I'm back from vacation in about a week

altair/utils/schemapi.py
dangotbanned: @MarcoGorelli I was wondering if there was a relation to `narwhals.Series.__iter__` not being defined yet? 
https://github.com/narwhals-dev/narwhals/blob/94eb1e229915bf0613ff92588332406c79959c9f/narwhals/series.py#L18

Also please correct me on any of my points in https://github.com/vega/altair/pull/3501#discussion_r1691248267 if you have more insight

MarcoGorelli: hey @dangotbanned - I think it's not necessary? At least, there's `__getitem_` and `__len__`

The rest of your points look good to me, esp. using the native `to_list`

> Can potentially support even more types in the future, by simply changing the import to import narwhals.stable.v2

Unless there's some breaking API change we just can't work around, any new library should still be available for `narwhals.stable.v1`

dangotbanned: Thanks @MarcoGorelli 

> I think it's not necessary? At least, there's `__getitem_` and `__len__`

To be clear I agree and I'm not asking for `__iter__` to be included.
We're not even using `narwhals.Series` directly here 😅

tests/utils/test_schemapi.py
joelostblom: Great point, maybe we should strictly adhere to passing only ordered sequences/arrays and disallow passing unordered sets? To me that seems closer to the intended use case in both the examples motivating this PR: the user wants to specify both the elements and in which order they occur. It might be confusing if they pass a set without realizing that it is unordered and altair does not show an error but instead the elements show up in the wrong order. What do you think?

tools/schemapi/schemapi.py
joelostblom: I see, since the solution in this PR prevents the conversion to a list of byte strings it is preserving the current behavior which seems like a good move instead of introducing a change here. Passing a byte string will result in an error due to the schema validation like you mentioned and I think that is suitable.

joelostblom (commented):
Looking great! I found a day to review this more in depth and left a few comments. I should be able to take another look in ~4 days.

joelostblom:
As part of this PR, could you also update the docs here https://altair-viz.github.io/user_guide/encodings/index.html#sort-option to not only mention that lists can be passed but any ordered sequence would work?

dangotbanned:
> Looking great! I found a day to review this more in depth and left a few comments. I should be able to take another look in ~4 days.

Thanks for the second look @joelostblom 

I'm going to work through your comments after this, but thought I'd give a heads-up on https://github.com/vega/altair/pull/3501#commits-pushed-e15bbed

After reading your comments this morning, I got the feeling that the focus on ordering/sets had overlooked the much broader scope of this change. 

I think this was my mistake by writing tests that only covered `alt.X(sort=)`.
I've updated these to show that this applies **anywhere** that previously could only accept a `(list|tuple)` - provided the value type matches the schema.

The ordering requirement makes a lot of sense for `sort`, but wouldn't be applicable in every case - especially when someone may be using a `set` to get unique values.

tests/utils/test_schemapi.py
dangotbanned: @joelostblom I'm not sure how to explain this since my attempts in https://github.com/vega/altair/pull/3501#discussion_r1690163239 and https://github.com/vega/altair/pull/3501#discussion_r1691287871 were not helpful.

I have reverted back to the original tests, to show exactly the failures this resolves - which are in [`bc4f4aa` (#3501)](https://github.com/vega/altair/pull/3501/commits/bc4f4aae84f3b202fb48807397f3b04ee3290d43)

My original commit https://github.com/dangotbanned/altair/commit/1f3dfd33ee273e71c74b67e64ae4410ae0b7c058 fixes the tests that are currently failing

dangotbanned:
> As part of this PR, could you also update the docs here [altair-viz.github.io/user_guide/encodings/index.html#sort-option](https://altair-viz.github.io/user_guide/encodings/index.html#sort-option) to not only mention that lists can be passed but any ordered sequence would work?

Yeah that sounds like a good idea, thanks @joelostblom

tests/utils/test_schemapi.py
dangotbanned: I think I answered more of this than I planned to in https://github.com/vega/altair/pull/3501#issuecomment-2269295210 so I'll just respond to the bits I don't think were covered there

> It might be confusing if they pass a set without realizing that it is unordered and altair does not show an error but instead the elements show up in the wrong order. What do you think?

I totally get the intention, but if a user wants to use the standard library incorrectly - who are we to stop them?
The **second sentence** in the [python tutorial on sets](https://docs.python.org/3/tutorial/datastructures.html#sets) states they are unordered.

---

Additionally, passing a `set` will *fail type checking* for `Sequence`, which is what I meant when I said:

> So while internally we are using `Iterable`, currently all of the public API uses the annotation `Sequence` and `set` is the former but not the latter.

```py
>>> mypy tests
error: Argument "sort" to "X" has incompatible type "set[str]"; expected
"dict[Any, Any] | SchemaBase | Sequence[str] | Sequence[bool] | Sequence[float] | Sequence[dict[Any, Any] | SchemaBase] | Literal['ascending', 'descending', 'x', 'y', 'color', 'fill', 'stroke', 'strokeWidth', 'size', 'shape', 'fillOpacity', 'strokeOpacity', 'opacity', 'text', '-x', '-y', '-color', '-fill', '-stroke', '-strokeWidth', '-size', '-shape', '-fillOpacity', '-strokeOpacity', '-opacity', '-text'] | UndefinedType | None"
 [arg-type]
        attempt_set = alt.X("fieldname", sort={"1", "2", "3"})
                                              ^~~~~~~~~~~~~~~ 
Found 1 error in 1 file (checked 325 source files)
```

tests/utils/test_schemapi.py
dangotbanned: I've added some markup to the [CI log](https://github.com/vega/altair/actions/runs/10251914588/job/28361056799?pr=3501) to make it easier to read:


## short test summary info


<details>
<summary>Failed 1</summary>


**FAILED** tests/utils/test_schemapi.py::test_chart_validation_errors[chart_error_example__wrong_tooltip_type_in_faceted_chart-'{'wrong'}' is an invalid value for `field`. Valid values are of type 'string' or 'object'.$] 

```py
AssertionError: Regex pattern did not match.

Regex: "'{'wrong'}' is an invalid value for `field`. Valid values are of type 'string' or 'object'.$"
Input: "'['wrong']' is an invalid value for `field`. Valid values are of type 'string' or 'object'."
```

</details>


<details>
<summary>Failed 2</summary>

**FAILED** tests/utils/test_schemapi.py::test_chart_validation_errors[chart_error_example__wrong_tooltip_type_in_layered_chart-'{'wrong'}' is an invalid value for `field`. Valid values are of type 'string' or 'object'.$]

```py
AssertionError: Regex pattern did not match.

Regex: "'{'wrong'}' is an invalid value for `field`. Valid values are of type 'string' or 'object'.$"
Input: "'['wrong']' is an invalid value for `field`. Valid values are of type 'string' or 'object'."
```

</details>


<details>
<summary>Failed 3</summary>

**FAILED** tests/utils/test_schemapi.py::test_chart_validation_errors[chart_error_example__two_errors_in_layered_chart-Multiple errors were found.\n\nError 1: '{'wrong'}' is an invalid value for `field`. Valid values are of type 'string' or 'object'.\n\nError 2: `Color` has no parameter named 'invalidArgument'\n\n    Existing parameter names are:\n    shorthand      bin         legend   timeUnit   \n    aggregate      condition   scale    title      \n    bandPosition   field       sort     type       \n\n    See the help for `Color` to read the full description of these parameters$]

```py
AssertionError: Regex pattern did not match.

Regex: "Multiple errors were found.\n\nError 1: '{'wrong'}' is an invalid value for `field`. Valid values are of type 'string' or 'object'.\n\nError 2: `Color` has no parameter named 'invalidArgument'\n\n    Existing parameter names are:\n    shorthand      bin         legend   timeUnit   \n    aggregate      condition   scale    title      \n    bandPosition   field       sort     type       \n\n    See the help for `Color` to read the full description of these parameters$"
Input: "Multiple errors were found.\n\nError 1: '['wrong']' is an invalid value for `field`. Valid values are of type 'string' or 'object'.\n\nError 2: `Color` has no parameter named 'invalidArgument'\n\n    Existing parameter names are:\n    shorthand      bin         legend   timeUnit   \n    aggregate      condition   scale    title      \n    bandPosition   field       sort     type       \n\n    See the help for `Color` to read the full description of these parameters"
```

</details>


<details>
<summary>Failed 4</summary>


**FAILED** tests/utils/test_schemapi.py::test_chart_validation_errors[chart_error_example__two_errors_in_complex_concat_layered_chart-Multiple errors were found.\n\nError 1: '{'wrong'}' is an invalid value for `field`. Valid values are of type 'string' or 'object'.\n\nError 2: '4' is an invalid value for `bandPosition`. Valid values are of type 'number'.$]

```py
AssertionError: Regex pattern did not match.

Regex: "Multiple errors were found.\n\nError 1: '{'wrong'}' is an invalid value for `field`. Valid values are of type 'string' or 'object'.\n\nError 2: '4' is an invalid value for `bandPosition`. Valid values are of type 'number'.$"
Input: "Multiple errors were found.\n\nError 1: '['wrong']' is an invalid value for `field`. Valid values are of type 'string' or 'object'.\n\nError 2: '4' is an invalid value for `bandPosition`. Valid values are of type 'number'."
```

</details>


<details>
<summary>Failed 5</summary>


**FAILED** tests/utils/test_schemapi.py::test_chart_validation_errors[chart_error_example__three_errors_in_complex_concat_layered_chart-Multiple errors were found.\n\nError 1: '{'wrong'}' is an invalid value for `field`. Valid values are of type 'string' or 'object'.\n\nError 2: `Color` has no parameter named 'invalidArgument'\n\n    Existing parameter names are:\n    shorthand      bin         legend   timeUnit   \n    aggregate      condition   scale    title      \n    bandPosition   field       sort     type       \n\n    See the help for `Color` to read the full description of these parameters\n\nError 3: '4' is an invalid value for `bandPosition`. Valid values are of type 'number'.$]

```py
AssertionError: Regex pattern did not match.

Regex: "Multiple errors were found.\n\nError 1: '{'wrong'}' is an invalid value for `field`. Valid values are of type 'string' or 'object'.\n\nError 2: `Color` has no parameter named 'invalidArgument'\n\n    Existing parameter names are:\n    shorthand      bin         legend   timeUnit   \n    aggregate      condition   scale    title      \n    bandPosition   field       sort     type       \n\n    See the help for `Color` to read the full description of these parameters\n\nError 3: '4' is an invalid value for `bandPosition`. Valid values are of type 'number'.$"
Input: "Multiple errors were found.\n\nError 1: '['wrong']' is an invalid value for `field`. Valid values are of type 'string' or 'object'.\n\nError 2: `Color` has no parameter named 'invalidArgument'\n\n    Existing parameter names are:\n    shorthand      bin         legend   timeUnit   \n    aggregate      condition   scale    title      \n    bandPosition   field       sort     type       \n\n    See the help for `Color` to read the full description of these parameters\n\nError 3: '4' is an invalid value for `bandPosition`. Valid values are of type 'number'."
```
</details>

### *5 failed*, 1145 passed, 462 skipped, 4 warnings in 42.52s
...

joelostblom: Thanks for elaborating further on this and providing the annotated CI logs. In my testing, `list[set[str]]` and `list[set[int]]` both seem to produce the same error message, so what I am confused about is why the fix is to switch from `"wrong"` to `5000` as the set element; it seems like the important part is to change from expecting a set to expecting a list in the error message.

I just pushed commit https://github.com/vega/altair/commit/fb84361ea4142e2a911c416f930c17a240f9c687 to illustrate what I would expect the solution to look like. The tests are passing, so I wonder if there is something I'm missing here that your solution with using an int as the set element accounts for?

tests/utils/test_schemapi.py
joelostblom: Thanks for adding the additional tests accounting for more cases than `sort`. I agree that it is not Altair's responsibility to make sure users are using other libraries correctly. At the same time I think hints of correct usage with Altair can be helpful but as you said we already have that via the `Sequence` typing hint, which should be sufficient so let's leave this as is.

tests/utils/test_schemapi.py
```diff
@@ -629,13 +638,25 @@ def chart_error_example__four_errors():
...
+    """
+    if isinstance(val, types.FunctionType):
+        return val.__name__
+    else:
+        return ""
+
+
+# NOTE: Avoids all cases appearing in a failure traceback
+# At the time of writing, this is over 300 lines
+chart_funcs_error_message: list[tuple[Callable[..., Any], str]] = [
```
joelostblom, line 655: :+1:

tests/utils/test_schemapi.py
```diff
@@ -629,13 +638,25 @@ def chart_error_example__four_errors():
...
+    (
+        chart_error_example__invalid_y_option_value_unknown_x_option,
+        inspect.cleandoc(
```
joelostblom, line 658: Can we remove this `inspect.cleandoc` as well as you have done for the others? It seems to work for me locally.

tests/utils/test_schemapi.py
```diff
@@ -876,6 +867,7 @@ def test_chart_validation_errors(chart_func, expected_error_message):
     with warnings.catch_warnings():
         warnings.filterwarnings("ignore", category=UserWarning)
         chart = chart_func()
+    expected_error_message = inspect.cleandoc(expected_error_message)
```
joelostblom, line 870: :+1:

joelostblom (commented):
Thanks again for this contribution and your patience with the review process @dangotbanned. I only have a few minor comments/clarifications left and then we should be ready to merge.

tests/utils/test_schemapi.py
dangotbanned: > I just pushed commit [fb84361](https://github.com/vega/altair/commit/fb84361ea4142e2a911c416f930c17a240f9c687) to illustrate what I would expect the solution to look like. The tests are passing, so I wonder if there is something I'm missing here that your solution with using an int as the set element accounts for?

Thanks @joelostblom that works for me as well and is now in [`8739239` (#3501)](https://github.com/vega/altair/pull/3501/commits/87392396a05faf1039134e47f969d1e47ec19b74)

Apologies I misunderstood you before. 

In https://github.com/vega/altair/pull/3501#discussion_r1691287871 I mentioned that the change was *subtle* - but didn't elaborate that I switched to `{5000}` to draw more attention to it.

Since then - forgetting my own reasoning - I assumed it was a **necessary** change, but it evidently was not 😅 

Appreciate your patience with me also

joelostblom (approved):
Super, thanks for the quick turn around! I'm excited to merge this PR and no longer have to convert series etc to lists explicitly =) Thanks again!

Issue #3239, opened 2023-10-26
More convenient syntax instead of FieldOneOfPredicate


joelostblom:
Writing a condition that matches a single value in relatively straightforward:

```py
import altair as alt
from vega_datasets import data


source = data.cars()
alt.Chart(source).mark_point().encode(
    x='Horsepower',
    y='Miles_per_Gallon',
    color=alt.condition(
        alt.datum.Origin == 'Japan',
        alt.value('red'),
        alt.value('grey')
    )
)
```

But when we want to match multiple values, the `alt.datum.Origin == 'Japan',` needs to be replaced with the notably more compllicated:

```py
        alt.FieldOneOfPredicate(
            'Origin',
            ['Japan', 'Europe']
        ),
```

It would be nice to support the `in` operator here and allow the syntax to just be `alt.datum.Origin in ['Japan', 'Europe']`. I haven't looked into the change necessary to make this happen but noting it down as something to revisit in the future.

mattijn:
I think that you would need a custom Altair list instead of using a python `list` to control this behaviour in the predicate.
You could add a custom magic method for the membership operator (eg. `__contains__`) to the [`OperatorMixin`](https://github.com/altair-viz/altair/blob/main/altair/expr/core.py#L41), but the equivalent `in` and `not in` membership operators will only be applied to the right-hand-side operand and not on the left-hand-side operand. Basically the list decide what it contains and since this is a python list it will return a boolean.

joelostblom:
> ...the equivalent in and not in membership operators will only be applied to the right-hand-side operand and not on the left-hand-side operand. Basically the list decide what it contains and since this is a python list it will return a boolean.

I was wondering about this. What made me think that it might be possible was that we do use the standard Python  operators for the other comparisons, but I hear that this might not work the same for the `in` operator. I don't fully understand why it would only be applied to the right-hand side from just skimming the code, but I will try to have a closer look at that section of the code base in the future. Thanks for pointing this out and linking to where the logic is!

dangotbanned:
I have a couple of alternative ideas for this one.

# 1 - A wrapper function (`one_of`)
Note: the `...TimeUnit_T` aliases are from https://github.com/dangotbanned/altair/blob/ac23fd51460e69b1a7189da28b9a10e28eda2ac7/altair/vegalite/v5/schema/core.py#L16331-L16333

<details>
<summary>Code block</summary>

```py
from __future__ import annotations

from typing import TYPE_CHECKING

[...]

PR #3505, opened 2024-07-26
feat(RFC): Adds `agg`, `field` utility classes

dangotbanned:
- `field` proposed in https://github.com/vega/altair/issues/3239#issuecomment-2233593364
- `agg` was developed during in https://github.com/vega/altair/pull/3427#issuecomment-2228454241 as a solution to part of https://github.com/vega/altair/discussions/3476

I've isolated these into a new module. 
Currently, this is only to support discussions like https://github.com/vega/altair/issues/3239#issuecomment-2251433303 to test out the syntax/design. 

I'm not looking to add these into `api.py` until https://github.com/vega/altair/issues/3454 is resolved. 
They will be too difficult to find currently, but having this in a public branch will make refining the design easier.

### DIFF
--- a/altair/vegalite/v5/_api_rfc.py
+++ b/altair/vegalite/v5/_api_rfc.py
@@ -281,7 +281,7 @@ class field:
         cls,
         field: str,
         /,
-        *values: bool | float | dict[str, Any] | SchemaBase,
+        *values: str | bool | float | dict[str, Any] | SchemaBase,
         timeUnit: TimeUnitType = Undefined,
     ) -> SelectionPredicateComposition:
         tp: type[Any] = type(values[0])
--- a/altair/vegalite/v5/_api_rfc.py
+++ b/altair/vegalite/v5/_api_rfc.py
@@ -54,6 +54,7 @@ RangeType: TypeAlias = Union[
     Sequence[Union[Dict[str, Any], None, float, Parameter, SchemaBase]],
 ]
 ValueType: TypeAlias = Union[str, bool, float, Dict[str, Any], Parameter, SchemaBase]
+OneOfType: TypeAlias = Union[str, bool, float, Dict[str, Any], SchemaBase]
 
 
 _ENCODINGS = frozenset(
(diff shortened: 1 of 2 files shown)


altair/vegalite/v5/_api_rfc.py
```diff
@@ -0,0 +1,372 @@
...
+    {'field': 'Origin'}
+
+    >>> field("Origin:N")
+    {'field': 'Origin', 'type': 'nominal'}
+
+    >>> field.one_of("Origin", "Japan", "Europe")
+    SelectionPredicateComposition({'field': 'Origin', 'oneOf': ['Japan', 'Europe']})
+    """
+
+    def __new__(  # type: ignore[misc]
```
dangotbanned, line 307: Wanted to highlight this current design quirk.

```py
import altair as alt

>>> alt.field("Origin:N")
{'field': 'Origin', 'type': 'nominal'}

>>> alt.agg("Origin:N")
{'field': 'Origin', 'type': 'nominal'}
```

1. I'm anticipating users would expect `alt.field()` to work this way, despite shorthand being unrelated.
2. I also think using `alt.agg()` without specifying an aggregation wouldn't seem an intuitive way to define a field.
3. Each wrapped [`Field...Predicate`](https://github.com/vega/altair/blob/679a7ce30cb3366319bce2f4211dd62a5a80c87c/altair/vegalite/v5/schema/core.py#L16401) already has a constructor; the faux-parent [`Predicate`](https://github.com/vega/altair/blob/679a7ce30cb3366319bce2f4211dd62a5a80c87c/altair/vegalite/v5/schema/core.py#L16392) class is just a `Union` and the unrelated [`Field`](https://github.com/vega/altair/blob/679a7ce30cb3366319bce2f4211dd62a5a80c87c/altair/vegalite/v5/schema/core.py#L7353) class

altair/vegalite/v5/_api_rfc.py
dangotbanned, line 1: Pushed by accident - ignore

PR #3507, opened 2024-07-28
build: Add `ipykernel` optional dependency to `dev` group

dangotbanned:
This PR adds the optional `ipython` dependency `ipykernel`, to avoid the following issue.

# Bug

## Attempting to run a notebook
image

## Install failing due to `uv` instead of `pip`
image

## Error command

image

This has to be modified to fix:

```bash
# >>> python -m pip install ipykernel -U --force-reinstall
>>> python -m uv pip install ipykernel -U --force-reinstall
```

---

I have been manually working around this since we moved from `pip` -> `uv` in https://github.com/vega/altair/pull/3431#issuecomment-2168654915

But this change is much simpler and doesn't need to be repeated for each new environment.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,7 +68,7 @@ dev = [
     "hatch",
     "ruff>=0.5.3",
     "ibis-framework[polars]",
-    "ipython",
+    "ipython[kernel]",
     "pandas>=0.25.3",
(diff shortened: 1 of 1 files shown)


binste (approved):
Useful, thanks!

PR #3063, opened 2023-05-19
Deprecate `.ref()` instead of removing it

mattijn:
Fix #3059.

This PR reintroduce the `.ref()` function which was removed in Altair 5.0.0. 

While the function is not needed anymore it is better to deprecate first to ease transition.

Now the following from the [Altair Tutorial on selections](https://altair-viz.github.io/altair-tutorial/notebooks/06-Selections.html#binding-scales-to-other-domains):
```python
import altair as alt
from vega_datasets import data

weather = data.seattle_weather()

interval = alt.selection_interval(encodings=['x'])

base = alt.Chart(weather).mark_rule(size=2).encode(
    x='date:T',
    y='temp_min:Q',
    y2='temp_max:Q',
    color='weather:N'
)

chart = base.encode(
    x=alt.X('date:T', scale=alt.Scale(domain=interval.ref()))
).properties(
    width=800,
    height=300
)

view = base.add_selection(
    interval
).properties(
    width=800,
    height=50,
)

chart & view
```
```cmd
d:\Software\mattijn\altair\altair\utils\deprecation.py:65: AltairDeprecationWarning: 'ref' is deprecated. No need to call '.ref()' anymore.
  warnings.warn(message, AltairDeprecationWarning, stacklevel=1)
d:\Software\mattijn\altair\altair\utils\deprecation.py:65: AltairDeprecationWarning: 'add_selection' is deprecated. Use 'add_params' instead.
  warnings.warn(message, AltairDeprecationWarning, stacklevel=1)
```
And a working plot.
image

Issue #3059, opened 2023-05-17
`TypeError: 'GetAttrExpression' object is not callable` with `interval.ref()`


tkyoma:
Thanks for the wonderful library!

Using the previous version `4.2.2`, everything works fine, but with `5.0.0`, the following code produces the error shown:

```python
interval = alt.selection_interval(encodings=["x"])
interval.ref()

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[10], line 2
      1 interval = alt.selection_interval(encodings=["x"])
----> 2 interval.ref()

TypeError: 'GetAttrExpression' object is not callable
```

mattijn:
Thanks for the report @tkyoma! 

In Altair version 4.2.2 the [`.ref()`](https://github.com/altair-viz/altair/blob/v4.2.2/altair/vegalite/v4/api.py#L176) did return a dictionary, but from Altair version 5 the `alt.selection_interval()` returns a `Parameter` object and not a `Selection` object. 

Can you elaborate a little bit more in what way you are using this syntax? I'm not yet sure if this a bug or a breaking change. Would it for example be possible to use [`interval.to_dict()`](https://github.com/altair-viz/altair/blob/master/altair/vegalite/v5/api.py#L194) instead?

EtienneT:
I got the exact same problem today.

I installed altair 5.0.0, coded some graphs, then installed streamlit which requires altair 4.  I later realized that the code was not working anymore and uninstalled streamlit and reinstalled altair 5, but the .ref() error was still occurring even though I restarted my notebook kernel, restarted jupyter to make sure I was using the latest version.  alt.__version__ returns 5.0.0 too.

I can reproduce the problem by using this example from the docs:

```
import altair as alt
from vega_datasets import data

weather = data.seattle_weather()

interval = alt.selection_interval(encodings=['x'])

base = alt.Chart(weather).mark_rule(size=2).encode(
    x='date:T',
    y='temp_min:Q',
    y2='temp_max:Q',
    color='weather:N'
)

chart = base.encode(
    x=alt.X('date:T', scale=alt.Scale(domain=interval.ref()))
).properties(
    width=800,
    height=300
)

view = base.add_selection(
    interval
).properties(
    width=800,
    height=50,
)

chart & view
```

I am getting the following error:

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [1], in <cell line: 15>()
[...]

PR #3509, opened 2024-07-30
docs: Undoc deprecated functionality

dangotbanned:
Partially addresses #3454 

> - Should deprecated features be documented to the same level as their replacements?
>	- Deprecated: https://altair-viz.github.io/user_guide/generated/api/altair.selection_single.html#altair-selection-single 
>	- Replacement: https://altair-viz.github.io/user_guide/generated/api/altair.selection_point.html#altair.selection_point

---

Mainly impacts [API Functions](https://altair-viz.github.io/user_guide/api.html#api-functions) where it is much clearer which `selection_` functions should be used, since 3/5 are no longer there.

image

### DIFF
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -1980,7 +1980,6 @@ class TopLevelMixin(mixins.ConfigMethodMixin):
         vegaembed_version: str = VEGAEMBED_VERSION,
         embed_options: dict | None = None,
         json_kwds: dict | None = None,
-        webdriver: str | None = None,
         engine: str | None = None,
         inline=False,
         **kwargs,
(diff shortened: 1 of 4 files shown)


altair/vegalite/v5/api.py
```diff
@@ -350,7 +350,7 @@ def __init__(
 
     @utils.deprecated(version="5.0.0", alternative="to_dict")
     def ref(self) -> dict:
-        """'ref' is deprecated. No need to call '.ref()' anymore."""
+        """'ref' is deprecated. Use '.to_dict()' instead."""
```
mattijn, line 353: Hm, not sure about this. See related issue https://github.com/vega/altair/issues/3059 and PR https://github.com/vega/altair/pull/3063 why it was stating that there is no need to call `.ref()` anymore.

dangotbanned: > Hm, not sure about this. See related issue #3059 and PR #3063 why it was stating that there is no need to call `.ref()` anymore.

Thanks for linking those @mattijn 

I'm thinking both the docstring and the deprecation should be updated.

dangotbanned: I thought I'd push with the original message in [`23a8eb5` (#3509)](https://github.com/vega/altair/pull/3509/commits/23a8eb5babe9311d093134e9642a746efe78629b) and if you'd like to add context I'll accept the suggestion(s)

mattijn:
Thanks!

Issue #3337, opened 2024-02-24
Simplify package structure


binste:
Now that Altair no longer supports different Vega-Lite and even Vega versions, I'd be in favour of simplifying the folder structure for the next major release. This would help with

* version control: Now, all files in new `v5`, `v6`, ... folders show up as "new" which makes it difficult to compare against older versions
* type hints: See #3335 

I've added a new `v6` label we can use to tag issues which we can't tackle just yet but which require a major release.

mattijn:
cross referencing this comment: https://github.com/vega/altair/issues/3416#issuecomment-2334258014

PR #3513, opened 2024-08-02
refactor(ruff): Organize imports w/ (`I001`, `TID252`) rules

dangotbanned:
This PR defines and applies import rules.

Something like this had been on my radar for a while, adding this now seems to be the least disruptive time.

Autogenerated code already had `I001` applied for import sorting:
https://github.com/vega/altair/blob/a84783359b933b371dad1d1f1233304c5daf84e5/tools/schemapi/utils.py#L830-L847

In addition to the visual changes, this makes navigating via imports much easier.

# Ruff links
- https://docs.astral.sh/ruff/rules/#isort-i
- https://docs.astral.sh/ruff/settings/#lintisort

# Related Issues
- https://github.com/vega/altair/issues/3337
  - Will be much easier now that all references are absolute
  - https://code.visualstudio.com/docs/python/editing#_rename-module
  - https://code.visualstudio.com/docs/python/editing#_move-symbol

### DIFF
--- a/altair/expr/consts.py
+++ b/altair/expr/consts.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-
 CONST_LISTING = {
     "NaN": "not a number (same as JavaScript literal NaN)",
     "LN10": "the natural log of 10 (alias to Math.LN10)",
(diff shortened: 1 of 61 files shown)


mattijn:
Looks good, thanks.

Issue #3514, opened 2024-08-03
`polars.DataFrame.plot` support tracking


dangotbanned:
### What is your suggestion?

Related `polars` PR w/ @binste [comment](https://github.com/pola-rs/polars/pull/17995#issuecomment-2264929315)
- https://github.com/pola-rs/polars/pull/17995

Thought it would make sense to track in an issue here, not only in a `polars` PR.

If you haven't read though that thread, please check it first as there are many related issues/repos that have been linked.
Including `pandas`, `pd_vega`, `altair_pandas`

# Example end-goal
Providing a **limited** API via a `plot` accessor, that shortens paths like below:

```py
from datetime import date

from packaging.version import Version
import polars as pl
import altair as alt

df = pl.DataFrame(
        {
            "date": [date(2020, 1, 2), date(2020, 1, 3), date(2020, 1, 4)] * 2,
            "price": [1, 4, 6, 1, 5, 2],
            "stock": ["a", "a", "a", "b", "b", "b"]
        }
    )

if Version(pl.__version__) >= Version("1.5.0"): # Currently 1.4.0
    df.plot.line(x="date", y="price", color="stock")
else:
    alt.Chart(df).mark_line().encode(x="date", y="price", color="stock")
```

This should provide:
- Access to *common* marks
- Arguments passed to `.encode()`
- Some level of typing on those parameters

@MarcoGorelli has already done some great work in the original PR, and the simplest approach would be exposing some types for them to import

As mentioned [here](https://github.com/pola-rs/polars/pull/17995#issuecomment-2265195835) and [here](https://github.com/pola-rs/polars/pull/17995#issuecomment-2266650913) an `alt.Chart` is returned - which can be used for further customisation using the full `altair` API.

PR #3506, opened 2024-07-27
refactor: replace archived `m2r` with updated `mistune`

dangotbanned:
The `dev` dependency [m2r](https://github.com/miyakogi/m2r) was archived on 2022-11-17.
The most popular fork [m2r2](https://github.com/CrossNox/m2r2) was last updated on 2023-01-30.

`m2r` was used to convert markdown to rst, extending https://github.com/lepture/mistune/releases/tag/v0.8.4

This [functionality](https://mistune.lepture.com/en/latest/api.html#mistune.renderers.rst.RSTRenderer) is part of the current version of `mistune` 

`mistune==0.8.4` currently raises the following error during [doc:built-html](https://github.com/vega/altair/actions/runs/10125141709/job/28000305726):

```cmd
Run hatch run doc:build-html
Creating environment: doc
Installing project in development mode
Checking dependencies
cmd [1] | mkdir -p doc/_images
cmd [2] | sphinx-build -b html -d doc/_build/doctrees doc doc/_build/html
Running Sphinx v7.4.7
loading translations [en]... done
/home/runner/.local/share/hatch/env/virtual/altair/801ceKDt/doc/lib/python3.12/site-packages/mistune.py:435: SyntaxWarning: invalid escape sequence '\|'
  cells[i][c] = re.sub('\\\\\|', '|', cell)
making output directory... done
```

---

# Changes
Most of the diff is due to `m2r` inserting whitespace where it wasn't needed. 

As a bonus, some bulleted lists that appear within a parameter description are now formatted correctly.

[Before](https://github.com/vega/altair/blob/679a7ce30cb3366319bce2f4211dd62a5a80c87c/altair/vegalite/v5/schema/core.py#L754-L784) -> [After](https://github.com/dangotbanned/altair/blob/a72ab7ff70f73ee4243189c9d3a74ce532936f1a/altair/vegalite/v5/schema/core.py#L752-L782)

PR #3515, opened 2024-08-03
feat(typing): Adds public `altair.typing` module

dangotbanned:
Will close https://github.com/vega/altair/issues/3514

# Related
- https://github.com/vega/altair/pull/3515#discussion_r1703163151
- https://github.com/vega/altair/discussions/3478#discussioncomment-10190173
- https://github.com/vega/altair/pull/3420
- https://github.com/pola-rs/polars/pull/17995


tools/generate_schema_wrapper.py
```diff
@@ -888,6 +888,9 @@ def _create_encode_signature(
             # We could be more specific about what types are accepted in the list
             # but then the signatures would get rather long and less useful
             # to a user when it shows up in their IDE.
+
+            # NOTE: Currently triggered only for `detail`, `order`, `tooltip`
+            # I think the `tooltip` one especially we should be more specific
```
dangotbanned, line 893: **NOTE**: Currently triggered only for `detail`, `order`, `tooltip`
I think the `tooltip` one especially we should be more specific.

altair/typing.py
```diff
@@ -0,0 +1,63 @@
...
+    shape: ChannelType
+    size: ChannelType
+    stroke: ChannelType
+    strokeDash: ChannelType
+    strokeOpacity: ChannelType
+    strokeWidth: ChannelType
+    text: ChannelType
+    theta: ChannelType
+    theta2: ChannelType
+    tooltip: ChannelType | list[Any]
```
dangotbanned, line 52: Currently we'd need to annotate as `list[Any]`, but think it could be improved in https://github.com/vega/altair/pull/3515/files#r1702733945

mattijn:
Again, just FYI @dangotbanned: https://github.com/vega/.github?tab=readme-ov-file#adding-maintainers 😉

dangotbanned:
> Again, just FYI @dangotbanned: https://github.com/vega/.github?tab=readme-ov-file#adding-maintainers 😉

Alright I'm interested @mattijn, if you wanna hold a vote. 

No hard feelings if I remain a contributor - I'm still very keen to continue working on `altair` regardless

binste:
Good idea, thanks @dangotbanned for opening it! Yeah the reverse works. I'll commit to this branch tomorrow. Of course if you want to take a stab at it today, feel free to go for it! As long as we're not working on it at the same time :)





> > Again, just FYI @dangotbanned: https://github.com/vega/.github?tab=readme-ov-file#adding-maintainers 😉
> 
> Alright I'm interested @mattijn, if you wanna hold a vote.
> 
> No hard feelings if I remain a contributor - I'm still very keen to continue working on `altair` regardless

Awesome! Fully agree with Mattijn. It would make it easier for you to collaborate with us and also as a recognition of the significant contributions you made to Altair recently. I've created a vote in the maintainers Slack channel. It's up for 7 days, I'll let you know afterwards.

altair/vegalite/v5/schema/channels.py
```diff
@@ -30945,7 +30945,7 @@ def encode(
         text: Optional[str | Text | Map | TextDatum | TextValue] = Undefined,
         theta: Optional[str | Theta | Map | ThetaDatum | ThetaValue] = Undefined,
         theta2: Optional[str | Theta2 | Map | Theta2Datum | Theta2Value] = Undefined,
-        tooltip: Optional[str | Tooltip | Map | list | TooltipValue] = Undefined,
+        tooltip: Optional[OneOrSeq[str | Tooltip | Map | TooltipValue]] = Undefined,
```
dangotbanned, line 30948: - 3 extra characters
- doesn't break into a new line
- generic over `Sequence` instead of `list`

Personally think this reads nicely:

> tooltip is optional, accepting one or a `Sequence` of `str`, `Tooltip`, `Map`, `TooltipValue`

dangotbanned: What I have in mind for the `TypedDict`, is that since it doesn't need a default (`Undefined`) we can reduce this further there:

```py
tooltip_chart: Optional[OneOrSeq[str | Tooltip | Map | TooltipValue]] = Undefined
tooltip_tdict: OneOrSeq[str | Tooltip | Map | TooltipValue]
```

Passing `Undefined` doesn't make any sense from the context of `pl.DataFrame.plot`

dangotbanned:
> Awesome! Fully agree with Mattijn. It would make it easier for you to collaborate with us and also as a recognition of the significant contributions you made to Altair recently. I've created a vote in the maintainers Slack channel. It's up for 7 days, I'll let you know afterwards.

Thanks @binste @mattijn really appreciate it

> Good idea, thanks @dangotbanned for opening it! Yeah the reverse works. I'll commit to this branch tomorrow. Of course if you want to take a stab at it today, feel free to go for it! As long as we're not working on it at the same time :)

All yours as of [`51a84a5` (#3515)](https://github.com/vega/altair/pull/3515/commits/51a84a5db3e991bbe41ff3673f4a1b8da0ca868d)

I got more of it done than I was expecting, not fully there yet.

Will add some notes tomorrow, but feel free to shoot over any questions if I havent got to it yet

altair/vegalite/v5/schema/channels.py
```diff
@@ -31189,3 +31189,48 @@ def encode(
         encoding.update(kwargs)
         copy.encoding = core.FacetedEncoding(**encoding)
         return copy
+
+
+class _EncodeKwds(TypedDict, total=False):
+    """Placeholder (FIXME)."""
+
+    angle: str | Angle | Map | AngleDatum | AngleValue
```
dangotbanned, line 31197: I think the docstring here should have the parameter descriptions, but **not** the types.

This won't be going through `sphinx` so there is no need for the duplication IMO.

Also it will fit in beter with https://github.com/MarcoGorelli/polars/blob/db6d8f76782aa02bb72b923cf46e2f638ede6bef/py-polars/polars/dataframe/plotting.py#L54-L65

altair/vegalite/v5/schema/channels.py
```diff
@@ -31189,3 +31189,48 @@ def encode(
...
+    order: OneOrSeq[str | Order | Map | OrderValue]
+    radius: str | Radius | Map | RadiusDatum | RadiusValue
+    radius2: str | Radius2 | Map | Radius2Datum | Radius2Value
+    row: str | Row | Map
+    shape: str | Shape | Map | ShapeDatum | ShapeValue
+    size: str | Size | Map | SizeDatum | SizeValue
+    stroke: str | Stroke | Map | StrokeDatum | StrokeValue
+    strokeDash: str | StrokeDash | Map | StrokeDashDatum | StrokeDashValue
+    strokeOpacity: str | StrokeOpacity | Map | StrokeOpacityDatum | StrokeOpacityValue
+    strokeWidth: str | StrokeWidth | Map | StrokeWidthDatum | StrokeWidthValue
```
dangotbanned, line 31221: Pure luck that these all ended up being 1 line 😅

altair/typing.py
```diff
@@ -1,63 +1,5 @@
...
-from typing import Any, Mapping, Union
-from typing_extensions import TypedDict
+__all__ = ["EncodeKwds"]
 
-if sys.version_info >= (3, 10):
-    from typing import TypeAlias
-else:
-    from typing_extensions import TypeAlias
-
-ChannelType: TypeAlias = Union[str, Mapping[str, Any], Any]
```
dangotbanned: @binste I know I removed `ChannelType` here, but `altair.typing` could define the precise types for each of the positional-args:

https://github.com/pola-rs/polars/pull/17995#issuecomment-2263609625

image

These did vary between each method, but I think this would be 6 unique channel type aliases total:

- https://github.com/MarcoGorelli/polars/blob/db6d8f76782aa02bb72b923cf46e2f638ede6bef/py-polars/polars/dataframe/plotting.py#L30
- https://github.com/MarcoGorelli/polars/blob/db6d8f76782aa02bb72b923cf46e2f638ede6bef/py-polars/polars/dataframe/plotting.py#L92
- https://github.com/MarcoGorelli/polars/blob/db6d8f76782aa02bb72b923cf46e2f638ede6bef/py-polars/polars/dataframe/plotting.py#L161


Unsure if there would be more needed for the yet-unwritten methods

binste: You already got very far! Great to see that we're almost there 😄 

I'm going through it now but I'm also wondering if we maybe just want to specify, for each channel, the types separately. `ChannelX`, `ChannelY`, etc. works for me regarding naming. So basically what's there now for `EncodingKwgs` but as individual type aliases.

I don't know if we then even need a `TypedDict`, what do you think? `Unpack` supports the usage suggested in https://github.com/pola-rs/polars/pull/17995#issuecomment-2263609625 only for Python >= 3.12. And I think the plot methods should then only accept `**kwargs: Unpack[EncodeKwgs]` as else the arguments which are explicitly defined are defined twice. See [typing docs](https://typing.readthedocs.io/en/latest/spec/callables.html#keyword-collisions).

Downside would be that in Polars, users only get autocomplete and type checking for the channels which are explicitly defined but for simple exploratory charts, this could be enough?

altair/vegalite/v5/schema/channels.py
binste: I like the usage of `OneOrSeq` and agree that `Undefined` is not relevant for these type hints.

altair/__init__.py
```diff
@@ -654,6 +655,7 @@ def __dir__():
 from altair.jupyter import JupyterChart
 from altair.expr import expr
 from altair.utils import AltairDeprecationWarning, parse_shorthand, Optional, Undefined
+from altair import typing
```
dangotbanned, line 658: @mattijn just want to be super clear on the scope of this, in reference to [comment](https://github.com/pola-rs/polars/pull/17995#issuecomment-2265137793).
It may even be worth adapting my notes here into the `altair.typing` module docstring.

- `altair.typing` will contain *some* public-facing types, that may be useful for extending/integrating with `altair`
	- Doesn't need to include everything
	- Some things may be useful both int/externally (e.g. `ChartType`)
- The module itself **is** accessible at the top-level, but **not** it's members
	- ✔️ `from altair.typing import EncodeKwds`
	- ✔️ `import altair.typing as alt_t; alt_t.EncodeKwds`
	- ❌ `alt.EncodeKwds` 
	- ❌ `alt.typing.EncodeKwds` 
- This is a common practice in other packages
	- https://github.com/numpy/numpy/blob/main/numpy/typing/__init__.py
	- https://github.com/numpy/numpy/blob/main/numpy/dtypes.pyi
	- https://github.com/pola-rs/polars/blob/0c2b5d82fd2726df7d7b318e5c21581a27fffdba/py-polars/polars/type_aliases.py
	- https://github.com/pandas-dev/pandas/blob/main/pandas/_typing.py

mattijn: Sounds useful to have public-facing types within `altair.typing`!

tools/schemapi/utils.py
```diff
@@ -363,53 +372,47 @@ def title(self) -> str:
...
+    def get_python_type_representation(
+        self,
+        for_type_hints: bool = ...,
+        return_as_str: Literal[False] = ...,
+        additional_type_hints: list[str] | None = ...,
+    ) -> list[str]: ...
     def get_python_type_representation(
         self,
         for_type_hints: bool = False,
-        altair_classes_prefix: str | None = None,
```
binste, line 369: Tried to figure it out myself but I'm a bit lost. Where does the code generation now add the `core` prefix?

tools/generate_schema_wrapper.py
```diff
@@ -859,48 +863,64 @@ def vegalite_main(skip_download: bool = False) -> None:
...
+    Wrapper for what was previously `_create_encode_signature()`.
+
+    Now also creates a `TypedDict` as part of https://github.com/pola-rs/polars/pull/17995
+    """
+
+    channel_infos: dict[str, ChannelInfo]
+    fmt_method: str
+    fmt_typed_dict: str
+
+    def __iter__(self) -> Iterator[str]:
```
binste, line 878: I enjoyed reading an iterable data class but I'm not sure if there is any advantage of writing it as such? It's used as something which takes 3 arguments and outputs a sequence -> Could be a simple function `generate_encoding_artifacts(channel_infos, fmt_method, fmt_typed_dict)` which would be more in line with the rest of the codebase. Just to keep the cognitive load as low as possible for new joiners to the codebase :)

altair/typing.py
dangotbanned: > You already got very far! Great to see that we're almost there 😄

Thanks @binste for such a thorough response!

> And I think the plot methods should then only accept `**kwargs: Unpack[EncodeKwds]` as else the arguments which are explicitly defined are defined twice. See [typing docs](https://typing.readthedocs.io/en/latest/spec/callables.html#keyword-collisions).

I was trying to demonstrate the correct way to use `Unpack` there, to avoid that issue (the second one):

```py

def foo(name, **kwargs: Unpack[Movie]) -> None: ...     # WRONG! "name" will always bind to the first parameter.

def foo(name, /, **kwargs: Unpack[Movie]) -> None: ...  # OK! "name" is a positional-only parameter,
           # ^^^                                        # so **kwargs can contain a "name" keyword.
```

<details>
<summary>Long screenshot</summary>

image

</details>


> I don't know if we then even need a `TypedDict`, what do you think? `Unpack` supports the usage suggested in [pola-rs/polars#17995 (comment)](https://github.com/pola-rs/polars/pull/17995#issuecomment-2263609625) only for Python >= 3.12. 

I'm not sure there's an issue with [`typing_extensions.Unpack`](https://typing-extensions.readthedocs.io/en/latest/#typing_extensions.Unpack) which I used unconditionally there?
Usually `ruff` or `pylance` will warn me when I do something that isn't allowed in `py3.8`

image

---
@binste 
> Downside would be that in Polars, users only get autocomplete and type checking for the channels which are explicitly defined but for simple exploratory charts, this could be enough?

https://github.com/pola-rs/polars/pull/17995#issuecomment-2263609625
@dangotbanned 
> You could then repeat the `/, **kwargs: Unpack[EncodeKwds]` for the other methods - maybe changing the positional-only ones if needed

This was essentially the compromise of that, if that makes sense?
You have your common ones that can be called *without* kwargs, but more options *with* kwargs

tools/schemapi/utils.py
dangotbanned: > Tried to figure it out myself but I'm a bit lost. Where does the code generation now add the `core` prefix?

It doesn't have to 😎 

Since we're using `from __future__ import annotations`, a lot of that code wasn't needed anymore.
`core` specifically was removed in the only 2 cases it appeared:

https://github.com/vega/altair/blob/15f115134e6a6d3234ccc7230f65b1fb1d10f8c8/tools/schemapi/utils.py#L406-L412

Which was the major source of reducing code in https://github.com/vega/altair/pull/3431

---

So I removed a lot of code that hadn't had an effect since then.
I tried to demonstrate that by pushing this commit [`0014cc8` (#3515)](https://github.com/vega/altair/pull/3515/commits/0014cc8e5527f8115c4cd2f4cef9e3e382cd4379) which was the one prior to this

binste: Nice! :D

altair/typing.py
binste: Yes, you're absolutely right! Thanks for the detailed explanation. Lost a bit the overview there but this brought me back on track :)

Just tested `/, **kwargs: Unpack[EncodeKwds]` with `typing_extensions.Unpack` on Python 3.11 and works as expected. 

Reviewing the rest now.

altair/vegalite/v5/schema/channels.py
binste: Agree. Working on this now.

tools/generate_schema_wrapper.py
dangotbanned: > I enjoyed reading an iterable data class but I'm not sure if there is any advantage of writing it as such?

I will admit it is the ugly duckling at the moment, no major objections if changed to a function (especially for something outside of `altair`).

>  Just to keep the cognitive load as low as possible for new joiners to the codebase :)

Following this up with another comment explaining the choice, since I think we're both seeking the same goal - but approaching it differently.
In case that takes a while to write up nicely, this would be one step *among others* to gradually break down a lot of the code into smaller pieces

altair/vegalite/v5/schema/channels.py
binste: [2b9ad2c](https://github.com/vega/altair/pull/3515/commits/2b9ad2cd74cf7de66655304be57214acd9e83758)

tools/generate_schema_wrapper.py
```diff
@@ -905,21 +907,31 @@ def _gen_artifacts(self) -> None:
...
+                    f"    {process_description(info.deep_description)}",
+                )
+            )
+        signature_docstring_parameters += [""]
+        typed_dict_docstring_parameters += [""]
+        signature_doc = indent_docstring(
+            signature_docstring_parameters, indent_level=8, width=100, lstrip=False
+        )
+        typed_dict_doc = indent_docstring(
+            typed_dict_docstring_parameters, indent_level=8, width=100, lstrip=False
```
dangotbanned, line 928: So I actually prefer the look of the shorter wrapping in `_EncodeKwds`, but I think you want the `indent_level=4`, since it is on the class and not the method

I thought I made the change to `8` in https://github.com/vega/altair/pull/3506 but I can't find it

tools/generate_schema_wrapper.py
binste: I took the liberty to rewrite it in [Rewrite EncodeArtifacts dataclass as a function](https://github.com/vega/altair/pull/3515/commits/79f317dcaa8b74fff6891530dab7b185fec489b5) but we can also revert it back.

altair/vegalite/v5/schema/channels.py
dangotbanned: > [2b9ad2c](https://github.com/vega/altair/pull/3515/commits/2b9ad2cd74cf7de66655304be57214acd9e83758) @dangotbanned thoughts?

I was writing my [comment](https://github.com/vega/altair/pull/3515/commits/2b9ad2cd74cf7de66655304be57214acd9e83758#r1703209502) at the same time as this apologies

tools/generate_schema_wrapper.py
binste: Just for context, your comment didn't load for me before I wrote mine :)

tools/generate_schema_wrapper.py
binste: 8 instead of 4 gives the additional indent marked in red:
image

Same as the `function_with_pep484_type_annotations ` example in https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html which also does not have type hints in the docstring. Or maybe I'm misreading your message?

binste:
@dangotbanned I have nothing further to add apart from the open discussions. This looks great to me, thanks for putting it together :)

@MarcoGorelli Would this work for you? The typeddict is available as `from altair.typing import EncodeKwds`. See [this comment](https://github.com/vega/altair/pull/3515#discussion_r1703175536) by @dangotbanned  which explains the proposed usage (same as proposed over in https://github.com/pola-rs/polars/pull/17995)

tools/generate_schema_wrapper.py
dangotbanned: > I took the liberty to rewrite it in [Rewrite EncodeArtifacts dataclass as a function](https://github.com/vega/altair/pull/3515/commits/79f317dcaa8b74fff6891530dab7b185fec489b5) but we can also revert it back.

As a compromise, could we do:

```py
def generate_encoding_artifacts(
    channel_infos: dict[str, ChannelInfo], fmt_method: str, fmt_typed_dict: str
) -> Iterator[str]:
    # do stuff
    # ...
	artifacts = ["1", "2"]
    yield from artifacts
```

Essentially then you can update `generate_encoding_artifacts` to do whatever in the future, but keep the same annotation and usage here:

```py
*generate_encoding_artifacts(channel_infos, ENCODE_METHOD, ENCODE_TYPED_DICT)
```

So this always means the same thing, regardless of how many `artifacts`

binste: Changed it in https://github.com/vega/altair/pull/3515/commits/0287eba43a18b8a4b4a4c2963e50c92f23726bc5 to an iterator as suggested. Feels like I can learn something here :) To put it into different words to see if I understood it: Is it correct that you proposed this so that there is no promise that `generate_encoding_artifacts` always returns just 2 strings but could be expanded in the future?

dangotbanned:
Also sorry @binste if my somewhat live-reviewing wasn't helpful.
Every time I finished a comment you had either replied or pushed more 🚤

tools/generate_schema_wrapper.py
dangotbanned: No that part is good - its the right hand side wraps early, thinking its hit 100

dangotbanned: > So I actually prefer the look of the shorter wrapping in `_EncodeKwds`, but I think you want the `indent_level=4`, since it is on the class and not the method @binste
> 
> I thought I made the change to `8` in #3506 but I can't find it

So the bottom one should be able to extend further.
But I was saying I prefer wrapping earlier, although it doesn't match the style elsewhere in the docs.

I think 100 is a jupyter thing

image

binste:
It was very helpful! And you basically did the whole PR yourself ;) I'm sitting outside with a mediocre internet connection so sometimes comments did not load right away.

We try to keep discussions as open as possible and hence on GitHub but for circumstances such as this one, we also have Slack to enable better live communication. If you want, feel free to join with this link: https://join.slack.com/t/vega-js/shared_invite/zt-2ob9c7x5s-mnfq~zxwowpJKmtYTHPh0g

tools/generate_schema_wrapper.py
dangotbanned: > Changed it in [0287eba](https://github.com/vega/altair/commit/0287eba43a18b8a4b4a4c2963e50c92f23726bc5) to an iterator as suggested. Feels like I can learn something here :) To put it into different words to see if I understood it: Is it correct that you proposed this so that there is no promise that `generate_encoding_artifacts` always returns just 2 strings but could be expanded in the future?

Yeah exactly you've got it!


I'm not sure if you've run into this before, but for me when a function returns multiple things - I then need to think *"what is each one?"*, *"does the order matter?"*, etc.

Now it is just that this returns strings and it does that exactly once

altair/typing.py
dangotbanned: @MarcoGorelli when you see this

The need for this solution is that `alt.Chart.encode` has alphabetically ordered parameters, whereas `pl.DataFrame.plot` has:
- pos-only (in a different order)
- *args
- **kwargs

tools/generate_schema_wrapper.py
binste: Definitely! Great if code can make these things clearer. Will also use it more going forward, thanks.

dangotbanned: Some more useful examples `alt.when` following [api._parse_when_compose](https://github.com/vega/altair/blob/95f654fcd036069e1101883d176ec67c6f13fa25/altair/vegalite/v5/api.py#L711)

tools/generate_schema_wrapper.py
dangotbanned: This had less of an impact than I expected but we have 4 extra chars of width in [`bac1f67` (#3515)](https://github.com/vega/altair/pull/3515/commits/bac1f67c97760c00d13d2b3100a468c484a36946)

dangotbanned:
> **Edit:** If you Marco want, we can also add the channel-specific type hints as separate individual type hints in addition to the TypedDict, e.g. `altair.typing.ChannelX`, `altair.typing.ChannelY` ... so you can use it in the individual annotations of the explicitly specified channels.

@binste IMO I think @MarcoGorelli would want this, since it would be easier to keep an alias in sync - rather than creating new ones

If it seems simple enough for you to generate them, you could give that a try when you're free?

Otherwise, I'll loop back to this once I've caught up on some other PR/issues and check in with you

MarcoGorelli:
Yup, exactly right, thanks! Creating and maintaining new type aliases in Polars might not go down so well, but just importing `EncodeKwds` from `altair.typing`, would be very easy

And if it's just as easy to import `ChannelX` / `ChannelY` etc., then even better! But even just having `EncodeKwds` would already be a nice improvement 📈 

I presume Altair-ally might also want to make use of these, so `altair.typing` might be more broadly useful beyond just Polars' usage

altair/typing.py
dangotbanned: **Marking as unresolved purely so this is easier to find during review**

It is actually resolved unless https://github.com/vega/altair/pull/3515#discussion_r1703238386 turns up anything new

tools/generate_schema_wrapper.py
```diff
@@ -894,8 +896,13 @@ def generate_encoding_artifacts(
         if info.supports_arrays:
             docstring_union_types.append("List")
             tp_inner = f"OneOrSeq[{tp_inner}]"
-        signature_args.append(f"{channel}: Optional[{tp_inner}] = Undefined")
-        typed_dict_args.append(f"{channel}: {tp_inner}")
+
+        alias_name = f"Channel{channel[0].upper()}{channel[1:]}"
+        channel_type_aliases.append(f"{alias_name}: TypeAlias = {tp_inner}")
```
dangotbanned, line 901: @binste you can't reuse this outside of annotation scope until `py3.10` IIRC, due to ` | `

dangotbanned: Ignore me if `ruff` has already given you the warning

binste: Wow, live error feedback, thanks! :) I'll need to get used to the fact that we can now taste some of the improved typing syntax of newer versions thanks to the futures import but not in all scopes.

binste: Actually, ruff did not give me any warning although I have the ruff vscode extension installed. Any setting you can recommend for this?

dangotbanned: > Actually, ruff did not give me any warning although I have the ruff vscode extension installed. Any setting you can recommend for this?

You can run `hatch` commands against an earlier python version for this

https://github.com/vega/altair/blob/main/CONTRIBUTING.md#failures-on-specific-python-versions

I added it to the guide, but maybe only @mattijn saw

dangotbanned: There are some other changes on there as well, you should be able to run everything from `hatch` now

binste: Great, thanks!

tools/generate_schema_wrapper.py
```diff
@@ -861,59 +865,72 @@ def vegalite_main(skip_download: bool = False) -> None:
...
-            # to a user when it shows up in their IDE.
-            union_types.append("list")
             docstring_union_types.append("List")
+            tp_inner = f"OneOrSeq[{tp_inner}]"
 
-        union_types = union_types + datum_and_value_class_names
-        docstring_union_types = docstring_union_types + [
-            rst_syntax_for_class(c) for c in datum_and_value_class_names
-        ]
+        alias_name = f"Channel{channel[0].upper()}{channel[1:]}"
```
dangotbanned, line 900: How're you planning to get each `alias_name` into `altair.typing` without including it in `channels.__all__`?

binste: I'm now trying to rename them to `_ChannelX` and then import directly in `altair.typing` and expose it there via `__all__`. Same as for `EncodeKwds`. Don't think it's worth it right now to autogenerate `altair.typing` as there are very rarely new additions of channels but would of course be a nice-to-have. What do you think?

binste: I'll also add it to the API docs so that it's declared as public. I will finish and push it tomorrow.

dangotbanned: > I'm now trying to rename them to `_ChannelX` and then import directly in `altair.typing` and expose it there via `__all__`. Same as for `EncodeKwds`. Don't think it's worth it right now to autogenerate `altair.typing` as there are very rarely new additions of channels but would of course be a nice-to-have. What do you think?

You can keep the name the same btw. They won't be added to `channels.__all__` unless you write code to do that. 

Just thinking writing out 40ish renames in `typing` might be a lot

binste: 👍 I've now also renamed `_EncodeKwds` to `EncodeKwds` as else it shows up like this in Sphinx (could probably also be fixed but now it's also consistent with the `Channel...` type aliases): 
image

In the latest commit, I've also:
* Added channel type aliases to typing module
* Added 'Type Hints' section to API references in docs.
* Removed is_chart_type from 'API' docs section as it's now in typing. Rename '_EncodeKwds' to 'EncodeKwds'

Would you mind having another look?

dangotbanned: > Would you mind having another look?

@binste Wow excellent!

Yeah I will go through it holistically later today, but good work on those points 👍

dangotbanned: @binste I do have some things I wanna discuss before making any changes other than https://github.com/vega/altair/pull/3515#commits-pushed-b3fbe9c

But overall I think we're very close to finished on this 😄

tools/generate_schema_wrapper.py
```diff
@@ -861,59 +865,65 @@ def vegalite_main(skip_download: bool = False) -> None:
...
-    docstring = indent_docstring(
-        docstring_parameters, indent_level=8, width=100, lstrip=False
+        type_aliases.append(f"{alias_name}: TypeAlias = {tp_inner}")
+        typed_dict_args.append(f"{channel}: {alias_name}")
+        signature_args.append(f"{channel}: Optional[{alias_name}] = Undefined")
+
+        description: str = f"    {process_description(info.deep_description)}"
+
+        signature_doc_params.extend((f"{channel} : {doc_types_flat}", description))
+        typed_dict_doc_params.extend((f"{channel}", description))
```
dangotbanned, line 915: @binste I had to lower the resolution to get below 10MB, but hopefully you can still see what the issue is.


https://github.com/user-attachments/assets/9f72ba05-e376-48cf-8182-672122483927

Essentially, the `TypeAlias`'s make it more difficult to find out what the correct type is.
I still think we need them for the individual args in `polars`, but I think they only worsen the UX for the `TypedDict` case - since you need to dig through refs to find the right types.

Simply using the expanded types still keeps everything on one row, but lets you see the types in more places.

Do you have the same issue in your IDE?

dangotbanned (commented):
Delivering as promised in https://github.com/vega/altair/pull/3515#discussion_r1705648767

This is a mostly visual review @binste but I think it is really important to make sure the result of this is as helpful as possible.

tools/generate_schema_wrapper.py
binste: Oh yeah, thanks for checking! Same for me (also vs code) and the expanded list is definitely better. I've changed it in https://github.com/vega/altair/pull/3515/commits/6fefd1270d59344bc2b44d7ee4491b8cb84b56ee

tools/generate_api_docs.py
```diff
@@ -72,6 +72,17 @@
    :nosignatures:
 
    {api_classes}
+
+Type Hints
```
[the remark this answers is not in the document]
binste: As we now define our public API via the API section (#3488) I think we should offer this guarantee to polars that there are no breaking changes in `altair.typing` within minor versions. As 5.4.0 will likely be the minimum required version, it would be great to already have it.

I fully agree with you that the docs can be made more user-friendly for these objects. However, I do not (yet) see it as a blocking issue. When I generate the docs, all type hint aliases are correctly identified by sphinx as such, nothing is identified as `TypeAliasType` as in your screenshot:

image

image

Did you modify anything to produce this on purpose or do we have a difference? Could you maybe try again with `hatch run doc:clean-generated && hatch run doc:build-html`?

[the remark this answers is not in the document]
dangotbanned: > As we now define our public API via the API section (#3488) I think we should offer this guarantee to polars that there are no breaking changes in `altair.typing` within minor versions. As 5.4.0 will likely be the minimum required version, it would be great to already have it.

Yeah you do make a good point, great to see (#3488) mentioned!

> I fully agree with you that the docs can be made more user-friendly for these objects. However, I do not (yet) see it as a blocking issue. When I generate the docs, all type hint aliases are correctly identified by sphinx as such, nothing is identified as `TypeAliasType` as in your screenshot:

So I don't think any of them were identified with the labels (TypeAlias, TypeAliasType, ..., TypedDict) I gave in https://github.com/vega/altair/pull/3515#discussion_r1705732684. 
I used those in the comment to explain why `sphinx` had treated them differently.

> Did you modify anything to produce this on purpose or do we have a difference? Could you maybe try again with `hatch run doc:clean-generated && hatch run doc:build-html`?

Didn't modify anything to produce that intentionally.
I run on a clean build, here is another screenshot w/ another build - in case this is `sphinx` version related 

<details>
<summary>Screenshot take 2</summary>

image

</details>


Maybe the issue is `hatch` isn't updating your `sphinx`, since the dependencies don't state a minimum version? 
https://github.com/vega/altair/blob/95f654fcd036069e1101883d176ec67c6f13fa25/pyproject.toml#L84-L85

Otherwise, maybe it is the `python` version we're using. I'm using `3.12`, which would be the first one to introduce [TypeAliasType](https://docs.python.org/3/library/typing.html#typing.TypeAliasType). 
Is it possible that `sphinx` would only use the symbol if built using `sys.version_info >= (3, 12)`?
That seems odd to me, as this was using `typing_extensions`.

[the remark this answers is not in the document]
dangotbanned: @binste hopefully we'll get the same `sphinx` output now I've merged this https://github.com/vega/altair/pull/3527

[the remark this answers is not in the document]
binste: Weird, even with #3527, I don't get this. I'm using Python 3.11 so maybe it's something like this...

I can build the docs with this setup when making the release. I'd suggest that if I merge this on the weekend and then make the release as-is. This makes it part of the public API and we can figure out doc issues at a later stage. Let me know if I'm missing anything, thanks! :)

binste:
@mattijn Just pinging you in case you want to have a look at this. Good to go from my side. I'll wait for @dangotbanned's final confirmation as well. I'd like to merge this on the weekend and cut the release so that they can continue with the work in Polars.

mattijn:
I'm on holiday coming period, trust you here👍

tools/generate_api_docs.py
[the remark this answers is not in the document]
dangotbanned: > Weird, even with #3527, I don't get this. I'm using Python 3.11 so maybe it's something like this...
> 
> I can build the docs with this setup when making the release. I'd suggest that if I merge this on the weekend and then make the release as-is. This makes it part of the public API and we can figure out doc issues at a later stage. Let me know if I'm missing anything, thanks! :)

I agree @binste, if you have consistency with `3.11` then go with that for now 👍 

Also, and this is non-blocking for me, but I'd suggest renaming the section from *Type Hints* -> *Typing*

Has been fun working with you on this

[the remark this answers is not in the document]
binste: Done 👍  And same, was fun :)

PR #3516, opened 2024-08-03
feat: Adds 4 missing `carbon` themes, provide autocomplete

dangotbanned:
# Related 
- https://github.com/vega/vega-themes/pull/587
- https://github.com/pola-rs/polars/pull/17995#issuecomment-2266683585

# Examples 
- https://vega.github.io/vega-themes/?theme=carbonwhite
- https://vega.github.io/vega-themes/?theme=carbong10
- https://vega.github.io/vega-themes/?theme=carbong90
- https://vega.github.io/vega-themes/?theme=carbong100

image

image

image

image

# Bonus

Support autocomplete for `themes.enable(name)`.

image

### DIFF
--- a/altair/vegalite/v5/theme.py
+++ b/altair/vegalite/v5/theme.py
@@ -7,6 +7,10 @@ from typing import Final
 from altair.utils.theme import ThemeRegistry
 
 VEGA_THEMES = [
+    "carbonwhite",
+    "carbong10",
+    "carbong90",
+    "carbong100",
     "dark",
     "excel",
     "fivethirtyeight",
--- a/altair/vegalite/v5/theme.py
+++ b/altair/vegalite/v5/theme.py
@@ -14,6 +14,7 @@ if TYPE_CHECKING:
     else:
         from typing_extensions import TypeAlias
 
+    # If you add a theme here, also add it in `VEGA_THEMES` below.
     _ThemeName: TypeAlias = Literal[
         "default",
         "carbonwhite",
@@ -33,6 +34,7 @@ if TYPE_CHECKING:
         "vox",
     ]
 
+# If you add a theme here, also add it in `_ThemeName` above.
 VEGA_THEMES = [
     "carbonwhite",
     "carbong10",
--- a/altair/vegalite/v5/theme.py
+++ b/altair/vegalite/v5/theme.py
@@ -7,16 +7,16 @@ from typing import Final
 from altair.utils.theme import ThemeRegistry
 
 VEGA_THEMES = [
-    "ggplot2",
-    "quartz",
-    "vox",
-    "fivethirtyeight",
     "dark",
-    "latimes",
-    "urbaninstitute",
     "excel",
+    "fivethirtyeight",
+    "ggplot2",
     "googlecharts",
+    "latimes",
     "powerbi",
+    "quartz",
+    "urbaninstitute",
+    "vox",
 ]
 
 
--- a/altair/vegalite/v5/theme.py
+++ b/altair/vegalite/v5/theme.py
@@ -2,9 +2,36 @@
 
 from __future__ import annotations
 
-from typing import Final
+from typing import TYPE_CHECKING, Final, Literal
 
 from altair.utils.theme import ThemeRegistry
+
+if TYPE_CHECKING:
+    import sys
+
+    if sys.version_info >= (3, 10):
+        from typing import TypeAlias
+    else:
+        from typing_extensions import TypeAlias
+
+    _ThemeName: TypeAlias = Literal[
+        "default",
+        "carbonwhite",
+        "carbong10",
+        "carbong90",
+        "carbong100",
+        "dark",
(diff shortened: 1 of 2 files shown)


altair/vegalite/v5/theme.py
```diff
@@ -2,21 +2,52 @@
...
+        "googlecharts",
+        "latimes",
+        "opaque",
+        "powerbi",
+        "quartz",
+        "urbaninstitute",
+        "vox",
+    ]
+
 VEGA_THEMES = [
```
binste, line 36: Instead of defining the theme names twice, I think you could specify `_ThemeName` as you've done here and then `VEGA_THEMES` could become `typing.get_args(_ThemeName`). See https://docs.python.org/3/library/typing.html#typing.get_args

binste (changes requested):
Thanks! And great to have autocomplete on this :) Added one suggestion

altair/vegalite/v5/theme.py
dangotbanned: Thanks @binste for the review!

I think I could rework this to use `typing.get_args`, but wanted to note some things first that led to my duplication here.

---

I'm viewing `_ThemeName` as a *short-term* solution. 
Explicitly marking it *"private"*, and **defining within** a `TYPE_CHECKING` block to prevent it existing at runtime.

I don't think I've used this pattern anywhere in `altair` before, but some examples in other projects (for their own reasons):
<details>
<summary>Examples</summary>

- https://github.com/pytest-dev/pytest/blob/dc756f4117f7cbabbe0866bab1fe562ac287cc03/src/_pytest/mark/structures.py#L428-L482
- https://github.com/pytest-dev/pytest/blob/dc756f4117f7cbabbe0866bab1fe562ac287cc03/src/_pytest/mark/structures.py#L501-508
- https://github.com/narwhals-dev/narwhals/blob/6fcc6a6ab3b3a5831db14d0e7e18fc45c43a0594/narwhals/schema.py#L13-L19
- https://github.com/narwhals-dev/narwhals/blob/6fcc6a6ab3b3a5831db14d0e7e18fc45c43a0594/narwhals/typing.py#L9-L32

</details>


IIRC, I'd need to move `_ThemeName` out of that block or use [`typing.get_type_hints`](https://github.com/python/cpython/blob/e319f774f9e766a2b92949444a2d46081df3363a/Lib/typing.py#L1185)

I'd also need to remove `"default", "opaque"` in the def of `_ThemeName` *or* later during iteration:

<details>
<summary>Registration source</summary>


https://github.com/vega/altair/blob/c12ca27362da791a57b89e0b58d3352b75cbc8ca/altair/vegalite/v5/theme.py#L45-L61

</details>


---

For a *long-term* solution, I'd want to:
- Extract the names directly from [source](https://github.com/vega/vega-themes/blob/main/src/index.ts) as (`VegaThemes: Literal[...]`)
- Define `altair` themes as (`AltairThemes: Literal["opaque", ...`)
	- [Your comment](https://github.com/pola-rs/polars/pull/17995#issuecomment-2266683585) which was what sparked this PR 😄 
	- [changing-the-theme](https://altair-viz.github.io/user_guide/customization.html#changing-the-theme) which states a plan to add more in the future
- ~~`_ThemeName`~~ -> `DefaultThemes: TypeAlias = AltairThemes | VegaThemes`
- Annotate `themes.enable(name: DefaultThemes)`
- Revisit the need for defining this across two `theme.py` modules, as part of https://github.com/vega/altair/issues/3337

---

Essentially I want to keep this as *flexible* as possible *for now*.

Planning to open another issue to discuss **themes** more generally.
There are other ideas I've had and I got the impression you *might* also have some of your own 🤝

binste: Steps re long-term solution sound great! We could probably keep it simple and just fetch the themes from the latest source file in vega-themes which you linked. Or the file which was part of the latest release.

Let's continue the discussion in a new issue as you suggested, thanks for bringing it up! :) I have some notes on providing a better default theme and additional documentation around customization (advanced legend manipulation, vertical y-axis titles, y-axis on the right side with tick labels on top of the ticks, ...)

binste: I've added 2 comments to mention that these two variables need to be kept in sync for now. Just in case we don't get to an improved solution for a while.

binste (approved):

altair/vegalite/v5/theme.py
dangotbanned: > Steps re long-term solution sound great! We could probably keep it simple and just fetch the themes from the latest source file in vega-themes which you linked. Or the file which was part of the latest release.
> 
> Let's continue the discussion in a new issue as you suggested, thanks for bringing it up! :) I have some notes on providing a better default theme and additional documentation around customization (advanced legend manipulation, vertical y-axis titles, y-axis on the right side with tick labels on top of the ticks, ...)

Perfect, thanks again @binste

PR #3522, opened 2024-08-05
chore: Remove filterwarnings from tests for cross-version pandas compatibility

MarcoGorelli:
pandas has two month-related offsets:
- MonthBegin, which is aliased to 'MS'
- MonthEnd, which is aliased to 'M' and is being renamed to 'ME' (because it was a common source of confusion, people would see 'M' and expect it to mean 'month', not 'month end'... https://github.com/pandas-dev/pandas/issues/9586)

When doing a monthly group-by, I think the first one gives the results that people are usually actually looking for, and is the same across versions, hence removing the need for filterwarnings or version checks

In the `bump_chart.py` example:
- grouping by `'6MS'` means that the windows are:
  - [2000-01-01, 2000-07-01)
  - [2000-07-01, 2001-01-01)
  - ...
- grouping by '6M' (or, now, '6ME') means that the windows are:
  - (1999-07-31 23:59:59.999999999, 2000-01-31 23:59:59.999999999]
  - (2000-01-31 23:59:59.999999999, 2000-07-31 23:59:59.999999999]
  - (2000-07-31 23:59:59.999999999, 2001-01-31 23:59:59.999999999]

I think the former is a much more natural way of thinking about time series, and so is better suited to the examples anyway (in addition to helping avoid `filterwarnings`

---

Output:

**before**

image

**after**

image

Due to the explanation above, I think the second one would align more closely without how people would read the chart

### DIFF
--- a/tests/examples_arguments_syntax/bump_chart.py
+++ b/tests/examples_arguments_syntax/bump_chart.py
@@ -12,7 +12,7 @@ from vega_datasets import data
 import pandas as pd
 
 stocks = data.stocks()
-source = stocks.groupby([pd.Grouper(key="date", freq="6M"),"symbol"]).mean().reset_index()
+source = stocks.groupby([pd.Grouper(key="date", freq="6MS"),"symbol"]).mean().reset_index()
 
 alt.Chart(source).mark_line(point = True).encode(
     x = alt.X("date:O", timeUnit="yearmonth", title="date"),
--- a/tests/examples_methods_syntax/bump_chart.py
+++ b/tests/examples_methods_syntax/bump_chart.py
@@ -12,7 +12,7 @@ from vega_datasets import data
 import pandas as pd
 
 stocks = data.stocks()
-source = stocks.groupby([pd.Grouper(key="date", freq="6M"),"symbol"]).mean().reset_index()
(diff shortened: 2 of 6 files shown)


dangotbanned:
@MarcoGorelli I tried fixing this in [`3e07902` (#3431)](https://github.com/vega/altair/pull/3431/commits/3e07902f2db1dffb6868082619370813c0f93b4a) which might be what led you here. IIRC the warning suggested `'ME'`.

Is this because `'MS'` was always an alias, but `'ME'` was a newer introduction?

MarcoGorelli:
Yup, that's right

The warning suggests 'ME' as that preserves the 'M' behaviour - but my claim is that people almost certainly didn't really want `'M'` in the first places and should've been using `'MS'`. I'll write a blog post about this soon ✍️

dangotbanned (approved):
This removes the need for `filterwarnings` I introduced back in [`3e07902` (#3431)](https://github.com/vega/altair/pull/3431/commits/3e07902f2db1dffb6868082619370813c0f93b4a)

dangotbanned:
Thanks @MarcoGorelli 

I already reviewed earlier this week, but appreciate the extra visual 👍 

If you come across any other `pandas`-related quirks in the examples, another thing to check is if it comes from https://vega.github.io/vega-lite/examples/

This one seems to be an `altair` original, so this change works

PR #3525, opened 2024-08-07
feat: Update Vega-Lite to 5.20.1

binste:
Compared to 5.19, it includes the new parameters `react` and `tickBandPaddingInner`.

Edit: I'll need to update the tests next. Note to myself to also update `NOTES_FOR_MAINTAINERSE.md` to mention this.


jonmmease:
@binste, feel free to update the `tests/vegalite/v5/test_api.py::test_to_url` test to to only check the start of the URL (the way we do when checking PNG/SVG images). Testing that the full URL is accurate is really vl-convert's responsibility, so I don't think we need to check the whole thing here, especially since it breaks on every Vega-Lite update.

binste:
Good point! Changed it

Issue #3519, opened 2024-08-04
Improvements to `alt.themes`


dangotbanned:
### What is your suggestion?

Following discussion w/ @binste in
- https://github.com/vega/altair/pull/3516#discussion_r1703175162

<details>
<summary>Original comment relating to autocomplete</summary>

For a *long-term* solution, I'd want to:
- Extract the names directly from [source](https://github.com/vega/vega-themes/blob/main/src/index.ts) as (`VegaThemes: Literal[...]`)
- Define `altair` themes as (`AltairThemes: Literal["opaque", ...`)
	- [Your comment](https://github.com/pola-rs/polars/pull/17995#issuecomment-2266683585) which was what sparked this PR 😄 
	- [changing-the-theme](https://altair-viz.github.io/user_guide/customization.html#changing-the-theme) which states a plan to add more in the future
- ~~`_ThemeName`~~ -> `DefaultThemes: TypeAlias = AltairThemes | VegaThemes`
- Annotate `themes.enable(name: DefaultThemes)`
- Revisit the need for defining this across two `theme.py` modules, as part of https://github.com/vega/altair/issues/3337

</details>

The larger goal would be making the process of creating/choosing a theme as easy as possible.

After linking to the existing docs in https://github.com/pola-rs/polars/pull/17995#issuecomment-2266657656, I thought we could improve both the docs and the UX itself.

---

# Ideas
I'll fill this out some more, but quick run-down:

## Feat
- More default themes, as stated in [User Guide](https://altair-viz.github.io/user_guide/customization.html#changing-the-theme)
- #3536 
- #3526 
- #3523 

## Docs
- Extend [User Guide chart-themes](https://altair-viz.github.io/user_guide/customization.html#chart-themes) to demo more
- #3528 
- Recreating the interactive demo in `altair`?

I know that @binste had some ideas as well, maybe others do too

mattijn:
Maybe useful, maybe not, but themes are also known within [vl-convert](https://github.com/vega/vl-convert):
```python
import vl_convert as vlc
list(vlc.get_themes().keys())
```

> ```python
> ['carbong10',
>  'carbong100',
>  'carbong90',
>  'carbonwhite',
>  'dark',
>  'excel',
>  'fivethirtyeight',
>  'ggplot2',
>  'googlecharts',
>  'latimes',
>  'powerbi',
>  'quartz',
>  'urbaninstitute',
>  'vox']
> ```

dangotbanned:
[...]

PR #3523, opened 2024-08-05
feat: Adds `vega-themes.json` using `vl_convert`

dangotbanned:
https://github.com/vega/altair/issues/3519#issuecomment-2267875356

Provides inline definitions from `vega-themes`.

Thought this made more sense as `.json`, rather than a `dict` within a module.

Eventually these could each be read into a Theme/Config `TypedDict` as mentioned in https://github.com/vega/altair/issues/3519

PR #3526, opened 2024-08-07
feat: Adds `@register_theme` decorator

dangotbanned:
Resolves one item in #3519

Fairly simple addition, but it is a nice convenience when you only use a single theme.

I also added an example that might be relevant to @joelostblom https://github.com/vega/altair/issues/3519#issuecomment-2267663552

The decorator also works well with type checkers: 

image


dangotbanned:
Will probably need to resolve some conflicts after #3523 is merged

altair/vegalite/v5/theme.py
```diff
@@ -2,21 +2,33 @@
...
+        from typing import LiteralString
+    else:
+        from typing_extensions import LiteralString
     if sys.version_info >= (3, 10):
         from typing import TypeAlias
     else:
         from typing_extensions import TypeAlias
 
+P = ParamSpec("P")
+R = TypeVar("R", bound=Dict[str, Any])
```
dangotbanned, line 29: The bounded `TypeVar` here can later be replaced with the `alt.Config` based `TypedDict` mentioned in https://github.com/vega/altair/issues/3519#issue-2447214476

While this currently provides *some* degree of type safety, the `TypedDict` solution will give detailed feedback on issues with the returned `dict`.

altair/vegalite/v5/theme.py
```diff
@@ -60,3 +72,75 @@ def __repr__(self) -> str:
     themes.register(theme, VegaTheme(theme))
 
 themes.enable("default")
+
+
+def register_theme(
+    name: LiteralString, *, enable: bool
```
binste, line 78: Any specific reason to restrict to `LiteralString` here? The functionality works for any `str` and although it might be niche, users could autogenerate theme names. Maybe they would like to register multiple themes in various combinations and then create theme names with f string syntax.

binste (changes requested):
Very useful! Thanks for adding this and sorry that you had to wait a month for a review and keep this branch in sync ;)

Only one small question regarding `LiteralString`, otherwise this looks good to me.

altair/vegalite/v5/theme.py
dangotbanned: So it is related to needing the string to be literal here, as having `str` instead would "swallow" the other `Literal` aliases - preventing them from showing on hover:

https://github.com/vega/altair/blob/5207768b6e533c0509218376942309d1c7bac22f/altair/utils/theme.py#L22-L25

For the same reason I had to update the type here https://github.com/dangotbanned/altair/blob/027c82cbc469c811a18e6f914e3a3c1d8cf12a00/altair/vegalite/v5/theme.py#L31

---

FWIW I did try `str` first. 
Settling on `LiteralString` was more of a pragmatic decision on my part, trying to prioritise the UX of the most general case. 
That is, having autocomplete with the *occasional* `#type: ignore` is preferable to no autocomplete at all.

Does that additional context help at all @binste?

dangotbanned:
> Very useful! Thanks for adding this and sorry that you had to wait a month for a review and keep this branch in sync ;)

Thanks for getting to this @binste and no worries on the timing; this is a mini-feature after all 😄

binste (approved):
Good point, having autocompletion trumps type hint flexibility in this case, wasn't aware that we can't have both. Thanks for the explanation!

Would you mind just adding the link to your GitHub comment as a Python comment next to the `LiteralString` type hints in `register_theme` and in the `ThemeRegistry`? Else, I'll probably change the type hint in a few months when I've forgotten why `LiteralString` is better :)

altair/vegalite/v5/theme.py
dangotbanned: Added [requested comments](https://github.com/vega/altair/pull/3526#pullrequestreview-2281646031) in [`9ce798c` (#3526)](https://github.com/vega/altair/pull/3526/commits/9ce798c15ec01c57d2d810d9215815a3a75bda92)

PR #3520, opened 2024-08-04
refactor: Simplify unreachable compound chart cases

dangotbanned:
Fixes https://github.com/vega/altair/issues/3518

# Related
- https://github.com/vega/altair/pull/3508

# Note
See links in [`1fdfe90` (#3520)](https://github.com/vega/altair/pull/3520/commits/1fdfe901405111dfb0f7919a6bc2fb861b70afe0) for more info on the `ValueError` -> `TypeError`.
It is a subclass so this doesn't affect anything downstream, and is more consistent with `stdlib` definition.

PR #3531, opened 2024-08-09
ci(ruff): Enforce the default `C901` complexity

dangotbanned:
# References
- https://docs.astral.sh/ruff/rules/complex-structure/#why-is-this-bad
- https://docs.astral.sh/ruff/settings/#lint_mccabe_max-complexity

Previously was set at **18**, the default is **10**.
As the rule was not in `[tool.ruff.lint.select]`, the setting was never used.
Setting such a high limit hides opportunities to simplify large, complex functions.

Many cases were resolved/improved in vega#3431.
There were 2 I fixed in this PR as they were quite simple changes.
The remainder use the *# noqa: C901* comment to serve as a TODO.

### DIFF
--- a/altair/jupyter/jupyter_chart.py
+++ b/altair/jupyter/jupyter_chart.py
@@ -218,7 +218,7 @@ class JupyterChart(anywidget.AnyWidget):
         )
 
     @traitlets.observe("chart")
-    def _on_change_chart(self, change):
+    def _on_change_chart(self, change):  # noqa: C901
         """Updates the JupyterChart's internal state when the wrapped Chart instance changes."""
         new_chart = change.new
         selection_watches = []
--- a/altair/vegalite/v5/api.py
+++ b/altair/vegalite/v5/api.py
@@ -1725,7 +1725,7 @@ class TopLevelMixin(mixins.ConfigMethodMixin):
     _class_is_valid_at_instantiation: bool = False
     data: Any
 
-    def to_dict(
+    def to_dict(  # noqa: C901
         self,
         validate: bool = True,
         *,
@@ -3876,7 +3876,8 @@ def _check_if_valid_subspec(
             raise ValueError(err.format(attr, classname))
 
 
-def _check_if_can_be_layered(spec: LayerType | dict[str, Any]) -> None:
+# C901 fixed in https://github.com/vega/altair/pull/3520
+def _check_if_can_be_layered(spec: LayerType | dict[str, Any]) -> None:  # noqa: C901
     """Check if the spec can be layered."""
 
     def _get(spec: LayerType | dict[str, Any], attr: str) -> Any:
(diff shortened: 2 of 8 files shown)


altair/jupyter/jupyter_chart.py
```diff
@@ -218,7 +218,7 @@ def __init__(
         )
 
     @traitlets.observe("chart")
-    def _on_change_chart(self, change):
+    def _on_change_chart(self, change):  # noqa: C901
```
dangotbanned, line 221: I think this was the highest, at around 20

altair/vegalite/v5/api.py
```diff
@@ -3876,7 +3876,8 @@ def _check_if_valid_subspec(
             raise ValueError(err.format(attr, classname))
 
 
-def _check_if_can_be_layered(spec: LayerType | dict[str, Any]) -> None:
+# C901 fixed in https://github.com/vega/altair/pull/3520
+def _check_if_can_be_layered(spec: LayerType | dict[str, Any]) -> None:  # noqa: C901
```
dangotbanned, line 3880: See the resolution for this in #3520

jonmmease:
Thanks for working through these rules @dangotbanned!

I'll defer to others, but I don't personally find C901 all that helpful. We should of course use functions and classes to minimize duplication of logic, but for an inherently complex function that doesn't contain logic that's useful elsewhere, I don't personally get much benefit from it being broken down into smaller functions.  And I suppose this is reflected in my coding style 😄

dangotbanned:
> Thanks for working through these rules @dangotbanned!
> 
> I'll defer to others, but I don't personally find C901 all that helpful. We should of course use functions and classes to minimize duplication of logic, but for an inherently complex function that doesn't contain logic that's useful elsewhere, I don't personally get much benefit from it being broken down into smaller functions. And I suppose this is reflected in my coding style 😄

Totally valid point @jonmmease! Appreciate your thoughts.

Choosing to use `C901` (or any rule really) is ultimately opt-in per *"violation"*.
I shouldn't have phrased it in the description as *every* case needing to be *"fixed"*.

Personally, the benefit to me when writing new code that the warning prompts me to consider *if* there is an alternative way to write the function.

> `function` is too complex (15 > 10)

If there isn't, or if it is time sensitive - add a *# noqa: C901* and move on.
The important part is the **moment of pause** the rule introduces, which *for me* usually leads to thinking more critically about my own code

jonmmease:
Yeah, it's easy to disable when needed. So no objection, and the small changes you needed to make to pass the check all look positive to me.

jonmmease (approved):
Looks good. Feel free to merge when you're ready!

dangotbanned:
> Looks good. Feel free to merge when you're ready!

Thanks @jonmmease!

FYI if at any time you find it frustrating to add a lot of ignore comments (for a single file), you can always add to our [tool.ruff.lint.per-file-ignores](https://github.com/dangotbanned/altair/blob/13c06a7ff6585cad278af540ca6b512544d7ae17/pyproject.toml#L388-L392) table instead

https://docs.astral.sh/ruff/settings/#lint_per-file-ignores

PR #3533, opened 2024-08-12
fix: Raise informative error message if a non-existent column name is passed

### LINKED ISSUE #3532: regression: unclear error message when plotting non-existent column name
### What happened?

I ran
```python
import pandas as pd
import altair as alt

df = pd.DataFrame({'a':[1,2], 'b':[4,5]})
alt.Chart(df).mark_line().encode(x='a', y='c').to_json()
```

✅ On Altair 5.3.0, I get
```
ValueError: Unable to determine data type for the field "c"; verify that the field name is not misspelled. If you are referencing a field from a transform, also confirm that the data type is specified correctly.
```

🔴 On Altair 5.4.0, I get
```
ValueError: c encoding field is specified without a type; the type cannot be automatically inferred because the data is not specified as a pandas.DataFrame.
```

Keep the 5.3.0 error

This one's my fault from the Narwhalification, sorry about that - submitting a fix now

5.4.0

MarcoGorelli:
closes #3532

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -504,7 +504,7 @@ def to_eager_narwhals_dataframe(data: IntoDataFrame) -> nw.DataFrame[Any]:
 
 def parse_shorthand(  # noqa: C901
     shorthand: dict[str, Any] | str,
-    data: pd.DataFrame | DataFrameLike | None = None,
+    data: IntoDataFrame | None = None,
     parse_aggregates: bool = True,
     parse_window_ops: bool = False,
     parse_timeunits: bool = True,
--- a/altair/vegalite/v5/schema/channels.py
+++ b/altair/vegalite/v5/schema/channels.py
@@ -14,7 +14,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Literal, Sequence, TypedDict, Union, overload
 from typing_extensions import TypeAlias
 
-from narwhals.dependencies import is_pandas_dataframe as _is_pandas_dataframe
+import narwhals.stable.v1 as nw
 
 from altair.utils import infer_encoding_types as _infer_encoding_types
 from altair.utils import parse_shorthand
@@ -179,7 +179,7 @@ class FieldChannelMixin:
                 # We still parse it out of the shorthand, but drop it here.
                 parsed.pop("type", None)
             elif not (type_in_shorthand or type_defined_explicitly):
-                if _is_pandas_dataframe(context.get("data", None)):
+                if isinstance(context.get("data", None), nw.DataFrame):
                     msg = (
                         f'Unable to determine data type for the field "{shorthand}";'
                         " verify that the field name is not misspelled."
--- a/altair/vegalite/v5/schema/channels.py
+++ b/altair/vegalite/v5/schema/channels.py
@@ -170,7 +170,7 @@ class FieldChannelMixin:
         if shorthand is Undefined:
             parsed = {}
         elif isinstance(shorthand, str):
-            data = context.get("data", None)
+            data: nw.DataFrame | Any = context.get("data", None)
             parsed = parse_shorthand(shorthand, data=data)
             type_required = "type" in self._kwds  # type: ignore[attr-defined]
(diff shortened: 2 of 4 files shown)


altair/vegalite/v5/schema/channels.py
```diff
@@ -179,7 +179,7 @@ def to_dict(
                 # We still parse it out of the shorthand, but drop it here.
                 parsed.pop("type", None)
             elif not (type_in_shorthand or type_defined_explicitly):
-                if _is_pandas_dataframe(context.get("data", None)):
+                if isinstance(context.get("data", None), nw.DataFrame):
```
MarcoGorelli, line 182: with the changes since #3452, by the time the code gets here, `context.get('data')` would have been "Narwhalified", and not a raw pandas dataframe anymore

dangotbanned (changes requested):
Thanks for the quick fix @MarcoGorelli!

I wasn't expecting you'd object to [`d9e219b` (#3533)](https://github.com/vega/altair/pull/3533/commits/d9e219bc15d4155e3fb1a1d4dd8b27544a267f8f) but let me know if you do.

Other than that, I only had one minor typing-related suggestion

altair/vegalite/v5/schema/channels.py
MarcoGorelli: thanks @dangotbanned !

agree on the commit you added

I've factored out `context.get('data', None)`, I'm just not sure about the annotation, as this can be a lot more than `SomeFrame` - check the return type of `_prepare_data`

dangotbanned: Thanks @MarcoGorelli and for pointing to `_prepare_data`, that really helped.

So if I've understood this correctly, starting in `Chart`:

https://github.com/vega/altair/blob/0062e6228134f176bb9fbaa944724039bb6b043a/altair/vegalite/v5/api.py#L3661-L3663

`context["data"]` is set in `TopLevelMixin.to_dict`:

https://github.com/vega/altair/blob/0062e6228134f176bb9fbaa944724039bb6b043a/altair/vegalite/v5/api.py#L1792-L1799

Which tries with `narwhals`:
https://github.com/vega/altair/blob/0062e6228134f176bb9fbaa944724039bb6b043a/altair/utils/core.py#L486

And prepares in:

https://github.com/vega/altair/blob/0062e6228134f176bb9fbaa944724039bb6b043a/altair/vegalite/v5/api.py#L253-L255

@MarcoGorelli 
> I'm just not sure about the annotation, as this can be a lot more than `SomeFrame` - check the return type of `_prepare_data`

So with this in mind - I guess this is more of a question now - can we assume the type of `context["data"]` has been **narrowed at all** by this point?

I would have thought for the cases `narwhals` can handle, they'd at least be wrapped by now.
If so, that would be information I'd find useful if it can be conveyed through typing.

Please let me know if I'm missing something here

MarcoGorelli: > I would have thought for the cases narwhals can handle, they'd at least be wrapped by now

Yup! 👍 

> If so, that would be information I'd find useful if it can be conveyed through typing.

Would `: nw.DataFrame | Any` be acceptable to you?

dangotbanned: > Would `: nw.DataFrame | Any` be acceptable to you? For someone reading this particular function, that might be all they need to know?

Yeah @MarcoGorelli that seems the simplest

altair/utils/core.py
```diff
@@ -504,7 +504,7 @@ def to_eager_narwhals_dataframe(data: IntoDataFrame) -> nw.DataFrame[Any]:
 
 def parse_shorthand(  # noqa: C901
     shorthand: dict[str, Any] | str,
-    data: pd.DataFrame | DataFrameLike | None = None,
+    data: IntoDataFrame | None = None,
```
MarcoGorelli, line 507: I think this wasn't quite right to begin with, looking into it

dangotbanned (approved):
Thanks @MarcoGorelli

PR #3535, opened 2024-08-12
feat(typing): Generate `Literal` aliases in `channels` overload(s)

dangotbanned:
# Related
- #3469
- #3464
- #3431

# Description
Discovered this while working on a larger refactoring to support https://github.com/vega/altair/issues/3519#issuecomment-2281012908 in #3536 

This change alone though is very minor, but has a nice **-6k** to lines of code


binste (approved):
I really like that you keep reducing the file sizes significantly! Apart from overall readability, it also reduces the size of the wheel file which is nice for pyodide. Altair 5.3.0 was 850KB and 5.4.0 only 650KB :)

PR #3536, opened 2024-08-12
feat: Adds `ThemeConfig` (`TypedDict`)

### LINKED ISSUE #3519: Improvements to `alt.themes`

dangotbanned:
Resolves one item in #3519

# Description
This PR aims to improve the UX when authoring a theme.
By utilising [`TypedDict`(s)](https://typing.readthedocs.io/en/latest/spec/typeddict.html) we can provide a very similar experience to that of writing in [Vega Editor](https://vega.github.io/editor/#/), with rich **autocompletion** and **static** validation feedback.

## Main Additions
- A hierarchy of `TypedDict` classes, mirroring `SchemaBase` classes - for the subset of objects that are used within a theme.
- A new module [`_.config.py`](https://github.com/dangotbanned/altair/blob/6e4af95becc8ff9be033535d45ce3db9b2c749a3/altair/vegalite/v5/schema/_config.py) where these classes reside.
- Refactoring `tools/` (w/ improved documentation)  to support *both* existing schema generation **and** these new `TypedDict`(s):
	- https://github.com/vega/altair/pull/3536/files#diff-ed138919703443179382e2761f4e08ae579ea9f51d88f29ee59745de1e7a96be
	- https://github.com/vega/altair/pull/3536/files#diff-4ed6d13a6ff1bb029ccb8835ed2aed8e3d790f817e42629850b42a59e97df468
	- https://github.com/vega/altair/pull/3536/files#diff-26bb768d345282c152ff21c1581804be0e84fdfaba9e8f530600d9ff068bd7d8

## Interactions with public API
These classes are entirely isolated from the main `alt.___` namespace, but can still be used anywhere a `dict|Map` annotation is present in existing code.
Currently:
- The top-level object `ThemeConfig` is exported to `altair.typing`
- All others are accessible via `altair.typing.theme`.

Using the classes themselves is entirely *optional*, as can be seen in [`test_theme.py`](https://github.com/vega/altair/pull/3536/files#diff-e52efec94b8b7f0f1b27eac85385173dacfaca37d65afd0df9839e09e58bf377).
Simply annotating `ThemeConfig` (and using a static type checker) allows the user to specify a theme using native `python` types.

# Examples
## Autocomplete w/ native `python` types
<p>

```py
import altair as alt
from altair.typing import ThemeConfig # noqa: TCH001


@alt.register_theme("custom theme", enable=True)
def my_theme() -> ThemeConfig:
    return {""}
```

</p>
</details> 

image

## Type checking w/ mixed `python|altair` types
<p>

```py
import altair as alt
from altair.typing import ThemeConfig # noqa: TCH001
from altair.typing.theme import ConfigKwds

@alt.register_theme("custom theme", enable=True)
def my_theme() -> ThemeConfig:
    return {
        "autosize": {"contains": "padding"},
        "config": ConfigKwds(font="Segoe UI", axis=9999999),
        "width": "container",
        "height": [500],
    }
```

</p>
</details> 

image

## Options for imports & runtime typing

<p>

```py
import altair as alt
from altair.typing import ThemeConfig, theme  # noqa: F811
from altair.typing.theme import AxisConfigKwds

custom_theme = ThemeConfig(
    config=theme.ConfigKwds(
        axis=AxisConfigKwds(grid=True, labelFont="system-ui"),
        axisX=AxisConfigKwds(labelOpacity=0.5),
        circle=theme.MarkConfigKwds(fill="purple"),
    )
)

print(
    f"{custom_theme!r}\n\n"
    f"The above theme was defined using {ThemeConfig.__name__!r}, "
    f"but at runtime the type is {type(custom_theme).__name__!r}"
)

# Manual registration
alt.themes.register("demo", lambda: custom_theme)
alt.themes.enable("demo")

alt.Chart(data.cars()).mark_circle().encode(x="Horsepower", y="Miles_per_Gallon")
```

</p>
</details> 

image

### Default

<p>

```py
import altair as alt
from altair.typing import ThemeConfig
from altair.typing.theme import ConfigKwds


source = data.stocks()
lines = (
    alt.Chart(source, title=alt.Title("Stocks"))
    .mark_line()
    .encode(x="date:T", y="price:Q", color="symbol:N")
    .interactive(bind_y=False)
)
lines
```

</p>
</details> 

image

### With a registered theme

<p>

```py
@alt.register_theme("param_font_size", enable=True)
def custom_theme_2() -> ThemeConfig:
    sizes = 12, 14, 16, 18, 20
    return {
        "autosize": {"contains": "content", "resize": True},
        "background": "#F3F2F1",
        "config": ConfigKwds(
            axisX={"labelFontSize": sizes[1], "titleFontSize": sizes[1]},
            axisY={"labelFontSize": sizes[1], "titleFontSize": sizes[1]},
            font="'Lato', 'Segoe UI', Tahoma, Verdana, sans-serif",
            headerColumn={"labelFontSize": sizes[1]},
            headerFacet={"labelFontSize": sizes[1]},
            headerRow={"labelFontSize": sizes[1]},
            legend={"labelFontSize": sizes[0], "titleFontSize": sizes[1]},
            text={"fontSize": sizes[0]},
            title={"fontSize": sizes[-1]},
        ),
        "height": {"step": 28},
        "width": 350,
    }


lines
```

</p>
</details> 

image

## Early POC Demo
**These screenshots were taken mid-development**

<details>
<summary>Providing fully nested auto-complete </summary>

image

</details>

<details><summary>Type checking feedback</summary>
<p>

image

</p>
</details> 

# Tasks
- [x] Resolve `ThemeConfig` "parent`
	- See https://github.com/vega/altair/pull/3536#issuecomment-2343859781
- [x] "Final" refactor/tidy up before review
	- See https://github.com/vega/altair/pull/3536#commits-pushed-44624ef
- [x] Address schema/default themes inconsistencies
	- See [failures](https://github.com/vega/altair/actions/runs/10779550963/job/29893320619?pr=3536)
- [x] Add pure type checking tests
- [x] Add docs, using the style seen in [`EncodeKwds`](https://github.com/vega/altair/blob/b996fa45d66c94266b97475eab0723b1841436cf/altair/vegalite/v5/schema/channels.py#L31247), where docs do not contain typing information
	- [x] Draft 1
	- [x] Add a useful summary to `ThemeConfig`
- [x] Reducing all annotations to `TypedDict`, `Literal`, or a composition that contains to **0** `SchemaBase` types
	- [`bb99389` (#3536)](https://github.com/vega/altair/pull/3536/commits/bb99389366011caecc887c1afb19e787aeee5b36)
- [x] Replace [`MANUAL_DEFS`](https://github.com/dangotbanned/altair/blob/bb99389366011caecc887c1afb19e787aeee5b36/tools/generate_schema_wrapper.py#L843-L882) w/ a recursive/graph-based solution
	- Working through changes like [`3c9aab5` (#3536)](https://github.com/vega/altair/pull/3536/commits/3c9aab5a094c1fbd18ebc8edcd70734bcddae545) to better understand the surrounding code
	- Huge thanks to @binste for unblocking this with [suggestion](https://github.com/vega/altair/pull/3536#discussion_r1743942369)
- [x] Transfer all *FIXME*, *HACK*, *TODO* comments into tasks/threads
	- Some of these have since been moved back to comments
	- where I see they may not get resolved during this PR


altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    align: Align_T
+    angle: float
+    aria: bool
+    ariaRole: str
+    ariaRoleDescription: str
+    aspect: bool
+    baseline: str | Baseline_T
+    binSpacing: float
+    blend: Blend_T
+    color: str | LinearGradient | RadialGradient | ColorName_T
```
dangotbanned, line 27: TODO: Reduce `LinearGradient | RadialGradient`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    continuousBandSize: float
+    cornerRadius: float
+    cornerRadiusBottomLeft: float
+    cornerRadiusBottomRight: float
+    cornerRadiusTopLeft: float
+    cornerRadiusTopRight: float
+    cursor: Cursor_T
+    description: str
+    dir: TextDirection_T
+    discreteBandSize: float | RelativeBandSize
```
dangotbanned, line 37: TODO: `RelativeBandSize`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    dx: float
+    dy: float
+    ellipsis: str
+    endAngle: float
+    fill: str | None | LinearGradient | RadialGradient | ColorName_T
+    fillOpacity: float
+    filled: bool
+    font: str
+    fontSize: float
+    fontStyle: str | FontStyle
```
dangotbanned, line 47: TODO: `FontStyle`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    lineHeight: float
+    minBandSize: float
+    opacity: float
+    order: bool | None
+    orient: Orientation_T
+    outerRadius: float
+    padAngle: float
+    radius: float
+    radius2: float
+    shape: str | SymbolShape
```
dangotbanned, line 65: TODO: `SymbolShape`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    strokeOffset: float
+    strokeOpacity: float
+    strokeWidth: float
+    tension: float
+    text: str | Sequence[str]
+    theta: float
+    theta2: float
+    timeUnitBandPosition: float
+    timeUnitBandSize: float
+    tooltip: str | bool | None | float | TooltipContent
```
dangotbanned, line 84: TODO: `TooltipContent`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    limit: float
+    line: bool | OverlayMarkDef
+    lineBreak: str
+    lineHeight: float
+    opacity: float
+    order: bool | None
+    orient: Orientation_T
+    outerRadius: float
+    padAngle: float
+    point: str | bool | OverlayMarkDef
```
dangotbanned, line 138: TODO: `OverlayMarkDef`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    radius: float
+    radius2: float
+    shape: str | SymbolShape
```
dangotbanned, line 141: TODO: `SymbolShape`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    description: str
+    disable: bool
+    domain: bool
+    domainCap: StrokeCap_T
+    domainColor: str | None | ColorName_T
+    domainDash: Sequence[float]
+    domainDashOffset: float
+    domainOpacity: float
+    domainWidth: float
+    format: str | Dict
```
dangotbanned, line 183: TODO: `Dict`

This is `core.Dict`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    maxExtent: float
+    minExtent: float
+    offset: float
+    orient: AxisOrient_T
+    position: float
+    style: str | Sequence[str]
+    tickBand: Literal["center", "extent"]
+    tickCap: StrokeCap_T
+    tickColor: str | None | ColorName_T
+    tickCount: float | TimeIntervalStep | TimeInterval_T
```
dangotbanned, line 221: TODO: `TimeIntervalStep`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    titleFontStyle: str | FontStyle
+    titleFontWeight: FontWeight_T
+    titleLimit: float
+    titleLineHeight: float
+    titleOpacity: float
+    titlePadding: float
+    titleX: float
+    titleY: float
+    translate: float
+    values: Sequence[str] | Sequence[bool] | Sequence[float] | Sequence[DateTime]
```
dangotbanned, line 249: TODO: `DateTime`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    labelFontSize: float
+    labelFontStyle: str | FontStyle
+    labelFontWeight: FontWeight_T
+    labelLimit: float
+    labelOffset: float
+    labelOpacity: float
+    labelOverlap: str | bool
+    labelPadding: float
+    labelSeparation: float
+    layout: Map
```
dangotbanned, line 592: HACK: `Map` is being used since it would be `dict | ExprRef`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    clipExtent: Sequence[Sequence[float]]
+    coefficient: float
+    distance: float
+    extent: Sequence[Sequence[float]]
+    fit: (
+        GeoJsonFeature
+        | GeoJsonFeatureCollection
+        | Sequence[GeoJsonFeature]
+        | Sequence[GeoJsonFeature | GeoJsonFeatureCollection | Sequence[GeoJsonFeature]]
+    )
```
dangotbanned, line 723: TODO: Multiple `GeoJSON...`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    useUnaggregatedDomain: bool
+    xReverse: bool
+    zero: bool
+
+
+class SelectionConfigKwds(TypedDict, total=False):
+    """Placeholder doc."""
+
+    interval: IntervalSelectionConfigWithoutType
+    point: PointSelectionConfigWithoutType
```
dangotbanned, line 812: Think both of these need `TypedDict`, but were missed since they don't end with `Config`

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+
+
+class ThemeConfig(TypedDict, total=False):
+    """Placeholder doc."""
+
+    arc: RectConfigKwds
+    area: AreaConfigKwds
+    aria: Any  # TODO
+    autosize: Any  # TODO
+    axis: AxisConfigKwds
```
dangotbanned, line 988: Top-level `TypedDict`.

Anything labelled like `    aria: Any  # TODO` means that it is a simple annotation.

Leaving those until all other issues are resolved

altair/vegalite/v5/schema/_config.py
dangotbanned: Both occurrences are for `format: str | Dict`.

- https://vega.github.io/vega-lite/docs/header.html#labels
- https://vega.github.io/vega-lite/docs/format.html

Reading through the `vega-lite` docs, I can't see how this would be used.

---

Related, didn't realise `Dict` was just an aliased js `Object`. 
I suppose we could use `Map`, but an example of how this is used would be helpful

dangotbanned: Fixed in [`b4f7ede` (#3536)](https://github.com/vega/altair/pull/3536/commits/b4f7ededff4c5f4f6371f6d0f049b4766e99c74b)

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1055 @@
...
+    size: str | ScaleInvalidDataShowAsValuesize
+    stroke: str | ScaleInvalidDataShowAsValuestroke
+    strokeDash: str | ScaleInvalidDataShowAsValuestrokeDash
+    strokeOpacity: str | ScaleInvalidDataShowAsValuestrokeOpacity
+    strokeWidth: str | ScaleInvalidDataShowAsValuestrokeWidth
+    theta: str | ScaleInvalidDataShowAsValuetheta
+    x: str | ScaleInvalidDataShowAsValuex
+    xOffset: str | ScaleInvalidDataShowAsValuexOffset
+    y: str | ScaleInvalidDataShowAsValuey
+    yOffset: str | ScaleInvalidDataShowAsValueyOffset
```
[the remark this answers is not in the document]
dangotbanned: All of these will be using a `TypedDict`, related [`a8ac7ec` (#3536)](https://github.com/vega/altair/pull/3536/commits/a8ac7ec647f1c27ffbffe9c3d1d7f5f866659835)

Will be roughly:

```py
from typing import Generic, TypeVar
from typing_extensions import TypedDict

T = TypeVar("T")

class Value(TypedDict, Generic[T]):
    value: T
```

Then instead of creating a `TypedDict` per `ScaleInvalidDataShowAsValue...`, they become:

```py
class ScaleInvalidDataConfigKwds(TypedDict, total=False):
    """Placeholder doc."""

    angle: str | Value[float]
    color: str | Value[ColorName_T]
    fill: str | Value[ColorName_T]
    fillOpacity: str | Value[float]
    opacity: str | Value[float]
    radius: str | Value[float]
    shape: str | Value[str]
    size: str | Value[float]
    stroke: str | Value[ColorName_T]
    strokeDash: str | Value[Sequence[float]]
    strokeOpacity: str | Value[float]
    strokeWidth: str | Value[float]
    theta: str | Value[float]
    x: str | Value[str | float]
    xOffset: str | Value[float]
    y: str | Value[str | float]
    yOffset: str | Value[float]

```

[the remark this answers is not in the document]
dangotbanned: Most of this is finished now in [`a9c8b99` (#3536)](https://github.com/vega/altair/pull/3536/commits/a9c8b99cb09e64a81b18a833019d7b1fdde5465e)

pyproject.toml
```diff
@@ -195,6 +195,7 @@ exclude = [
     "tests/examples_methods_syntax",
     "tests/test_transformed_data.py",
 ]
+extend-include = ["*.ipynb"]
```
dangotbanned, line 198: Will probably raise this in another PR, not intending to keep this here once out of draft

altair/vegalite/v5/schema/_config.py
dangotbanned: Also noting here anywhere there is:

```py
color: str | ColorName_T
```

There is potentially a use-case for `typing.Annotated` (mentioned in https://github.com/vega/altair/discussions/3476#discussioncomment-10162300).
The `str` case is for `HexColor`, which should be a simple enough regex pattern to reproduce.

dangotbanned: > There is potentially a use-case for `typing.Annotated` (mentioned in [#3476 (comment)](https://github.com/vega/altair/discussions/3476#discussioncomment-10162300)). The `str` case is for `HexColor`, which should be a simple enough regex pattern to reproduce.

Looked into further and a possible issue will be using `Annotated` within a `TypedDict`, since there won't be an object available at runtime to introspect.

So we could annotate it here, with a useful name, but utilising the metadata provided (the regex pattern to be matched) would have to happen in `ThemeRegistry` on `.register(theme)`

altair/vegalite/v5/schema/_config.py
dangotbanned: Turns out this is a direct alias for `str`, with no subclasses or properties.
Seems like a good rule for `use_concrete=True` to skip `SchemaInfo.title` if `SchemaInfo` meets criteria like this.

https://github.com/vega/altair/blob/cfab6601aa649227c20880d05ae27799297694c3/altair/vegalite/v5/schema/vega-lite-schema.json#L10983-L10985

dangotbanned: Fixed in: 
- [`ecf7fd5` (#3536)](https://github.com/vega/altair/pull/3536/commits/ecf7fd5f1dd45a68083c62b731f893c95e073519)
- [`f4cb0d4` (#3536)](https://github.com/vega/altair/pull/3536/commits/f4cb0d44aa7bfa7f3738472782a41edaaab6924b)

altair/vegalite/v5/schema/_config.py
dangotbanned: Fixed in: 
- [`ecf7fd5` (#3536)](https://github.com/vega/altair/pull/3536/commits/ecf7fd5f1dd45a68083c62b731f893c95e073519)
- [`f4cb0d4` (#3536)](https://github.com/vega/altair/pull/3536/commits/f4cb0d44aa7bfa7f3738472782a41edaaab6924b)

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,1054 @@
...
+    cornerRadiusTopRight: float
+    cursor: Cursor_T
+    description: str
+    dir: TextDirection_T
+    discreteBandSize: float
+    dx: float
+    dy: float
+    ellipsis: str
+    endAngle: float
+    fill: str | None | LinearGradient | RadialGradient | ColorName_T
```
dangotbanned, line 41: TODO: `LinearGradient | RadialGradient`

dangotbanned: Technically representing `Gradient`, which is a `Union` of these types.

Currently `altair` defines these using an empty parent class:

https://github.com/vega/altair/blob/cfab6601aa649227c20880d05ae27799297694c3/altair/vegalite/v5/schema/core.py#L8039-L8045

altair/vegalite/v5/schema/_config.py
dangotbanned: Fixed in [`da516c1` (#3536)](https://github.com/vega/altair/pull/3536/commits/da516c1a6753b1fc3520d6d9b0dc689a2b22de00)

altair/vegalite/v5/schema/_config.py
dangotbanned: Fixed in [`e5c4b7f` (#3536)](https://github.com/vega/altair/pull/3536/commits/e5c4b7fb4e9bb55d51dd047229cd0f098f744a54)

pyproject.toml
dangotbanned: No longer needed since #3539, now a default setting for `ruff`

dangotbanned:
@binste FYI I spent a good chunk of yesterday trying (and failing) to make progress on this.
If you have any ideas I'm all ears - will loop back to this eventually but it really stumped me

> # Tasks
> 
> - [ ]  Replace [`MANUAL_DEFS`](https://github.com/dangotbanned/altair/blob/bb99389366011caecc887c1afb19e787aeee5b36/tools/generate_schema_wrapper.py#L843-L882) w/ a recursive/graph-based solution
> 	- Need a better understanding of the existing code
>	- Working through changes like [`3c9aab5` (#3536)](https://github.com/vega/altair/pull/3536/commits/3c9aab5a094c1fbd18ebc8edcd70734bcddae545)

tools/generate_schema_wrapper.py
```diff
@@ -702,82 +761,146 @@ def generate_vegalite_channel_wrappers(
...
+        "IntervalSelectionConfig",
+        "BindInput",
+        "BindRange",
+        "BindDirect",
+        "BindCheckbox",
+        "BindRadioSelect",
+        "NumberLocale",
+        "TimeLocale",
+        "LegendStreamBinding",
+    )
```
dangotbanned, line 864: FIXME: Generate this selection, instead of manually defining

tools/schemapi/utils.py
```diff
@@ -373,140 +375,156 @@ def title(self) -> str:
...
-        else:
-            return type_representations
+        if use_concrete:
+            if tps >= {"ColorHex", "ColorName_T", "str"}:
+                # HACK: Remove regular `str` if HEX & CSS color codes are present as well
+                tps.discard("str")
+            elif len(tps) == 0 and as_str:
+                # HACK: There is a single case that ends up empty here
+                # (LegendConfig.layout)
+                tps = {"Map"}
```
dangotbanned, line 476: Related: https://github.com/vega/altair/pull/3536#discussion_r1714344162

tools/schemapi/utils.py
```diff
@@ -373,140 +375,156 @@ def title(self) -> str:
...
+                spell_nested_sequence(self, target=target, use_concrete=use_concrete)
             )
-            type_representations.append(f"Sequence[{s}]")
         elif self.type in jsonschema_to_python_types:
-            type_representations.append(jsonschema_to_python_types[self.type])
+            if self.is_object() and use_concrete:
+                ...  # HACK: Fall-through case to avoid `dict` added to `TypedDict`
+            else:
+                tps.add(jsonschema_to_python_types[self.type])
         else:
```
dangotbanned, line 465: HACK: Fall-through case to avoid `dict` added to `TypedDict`

tools/generate_schema_wrapper.py
binste: This is how far I got, maybe it helps.

```python
import itertools as it
import json

from pathlib import Path
from typing import Any

from tools.schemapi.utils import SchemaInfo, get_valid_identifier

def load_schema(fp: Path, /) -> dict[str, Any]:
    """Reads and returns the root schema from ``fp``."""
    with fp.open(encoding="utf8") as f:
        root_schema = json.load(f)
    return root_schema


schema = load_schema(Path("altair/vegalite/v5/schema/vega-lite-schema.json"))
config = SchemaInfo({"$ref": "#/definitions/Config"}, rootschema=schema)


def is_relevant_object(schema: SchemaInfo) -> bool:
    return (
        schema.is_object()
        and not schema.is_value()
        and schema.title.strip()
        and schema.properties
    )


def get_all_objects(
    schema: SchemaInfo, recursion_depth: int = 0
) -> dict[str, SchemaInfo]:
    objects: dict[str, SchemaInfo] = {}
    if is_relevant_object(schema):
        objects[schema.title] = schema

    if recursion_depth < 10:
        for prop_info in it.chain(
            schema.properties.values(),
            schema.anyOf,
            schema.allOf,
            schema.oneOf,
            [schema.child(schema.items)] if schema.items else [],
        ):
            objects.update(
                get_all_objects(prop_info, recursion_depth=recursion_depth + 1)
            )
    return objects
    
    
all_objects = get_all_objects(config)
def_names = list(all_objects.keys())
MANUAL_DEFS = (
        "ScaleInvalidDataConfig",
        "OverlayMarkDef",
        "LinearGradient",
        "RadialGradient",
        "GradientStop",
        "TooltipContent",
        "DateTime",
        "TimeIntervalStep",
        "IntervalSelectionConfigWithoutType",
        "PointSelectionConfigWithoutType",
        "Feature<Geometry,GeoJsonProperties>",
        "GeoJsonFeature",
        "GeoJsonFeatureCollection",
        "GeometryCollection",
        "Point",
        "Polygon",
        "LineString",
        "MultiPoint",
        "MultiPolygon",
        "MultiLineString",
        "BrushConfig",
        "MergedStream",
        "DerivedStream",
        "AutoSizeParams",
        "Locale",
        "VariableParameter",
        "TopLevelSelectionParameter",
        "PointSelectionConfig",
        "IntervalSelectionConfig",
        "BindInput",
        "BindRange",
        "BindDirect",
        "BindCheckbox",
        "BindRadioSelect",
        "NumberLocale",
        "TimeLocale",
        "LegendStreamBinding",
    )
MANUAL_DEFS_VALID = [get_valid_identifier(k) for k in MANUAL_DEFS]
assert all(m in def_names for m in MANUAL_DEFS_VALID)  # Works
set(MANUAL_DEFS_VALID).symmetric_difference(def_names)  # def_names also includes the `Config` entries such as `AxisConfig` and some additional ones.
```

Some observations:
* The rules for reading out the relevant schema definitions is a bit wonky. I just tweaked it until I had what I wanted.
  * The recursion depth limit is pretty random but I had to set something. Unclear if there is a better way to stop earlier?
* It also includes predicate classes (`FieldRangePredicate`, ...) which are not present in your current `_config.py`. I'm not sure if they should be excluded or if they are relevant.
* The typed dicts should probably also accept strings (or dicts?) wherever a `ExprRef` is accepted. For example `AxisConfig.aria`.

dangotbanned: > This is how far I got, maybe it helps.
> ...

Will need to look into [this](https://github.com/vega/altair/pull/3536#discussion_r1743942369) in a bit more depth soon, but at first glance this is great thank you for the help @binste!

> It also includes predicate classes (`FieldRangePredicate`, ...) which are not present in your current `_config.py`. I'm not sure if they should be excluded or if they are relevant.

IIRC, I think I wouldn't have wanted that in there - but looking at `MANUAL_DEFs` again now maybe there are others that make less sense to apply in a theme?

Essentially I was trying to create a `TypedDict` if:
- After resolving references starting at `Config`, it would appear in some definition
- **And** it would be resonable to set a default for a property

All of the `...Config` classes are obvious choices, but some others I'm less sure on.

---

Regardless - really appreciate your time on this!

dangotbanned: # Update
Just tested this out and I can work with this @binste - thank you!

Hopefully will make some more progress tomorrow

binste: Awesome! Looking forward to it :) I'll definitely rewrite my own theme code with `ThemeConfig` once we release it.

binste: I haven't fully tested this but it might be easier to just go through all definitions in the top-level to get a mapping of types to the ones which it uses directly, `graph` in the example below. Next step would be to start with Config and look up children and always draw the new descendants from `graph`. Maybe it's the same, thought I just share it in case it sparks any ideas for you.

```python
from pathlib import Path
from typing import Any
import json
from tools.schemapi.utils import SchemaInfo

def load_schema(fp: Path, /) -> dict[str, Any]:
    """Reads and returns the root schema from ``fp``."""
    with fp.open(encoding="utf8") as f:
        root_schema = json.load(f)
    return root_schema


rootschema = load_schema(Path("altair/vegalite/v5/schema/vega-lite-schema.json"))

definitions: dict[str, SchemaInfo] = {}

for name in rootschema["definitions"]:
    schema = SchemaInfo({"$ref": "#/definitions/" + name}, rootschema=rootschema)
    definitions[name] = schema

graph: dict[str, list[str]] = {}

import itertools as it
for name, schema in definitions.items():
    graph[name] = []
    
    for prop_info in it.chain(
            schema.properties.values(),
            schema.anyOf,
            schema.allOf,
            schema.oneOf,
            [schema.child(schema.items)] if schema.items else [],
        ):
        if prop_info.refname:
            graph[name].append(prop_info.refname)
```

binste: Maybe this is easier, it doesn't run into any recursion issue and gives you a dictionary `graph` which you can walk. You could start from `Config` and then go down from there:

```python
from pathlib import Path
from typing import Any
import json
from tools.schemapi.utils import SchemaInfo

def load_schema(fp: Path, /) -> dict[str, Any]:
    """Reads and returns the root schema from ``fp``."""
    with fp.open(encoding="utf8") as f:
        root_schema = json.load(f)
    return root_schema


rootschema = load_schema(Path("altair/vegalite/v5/schema/vega-lite-schema.json"))

definitions: dict[str, SchemaInfo] = {}

for name in rootschema["definitions"]:
    schema = SchemaInfo({"$ref": "#/definitions/" + name}, rootschema=rootschema)
    definitions[name] = schema

# Key is the name of a definition and the value is a set of names of definitions
# that are children of the key definition.
graph: dict[str, set[str]] = {}

import itertools as it

def get_children(schema_info: SchemaInfo) -> set[str]:
    children: set[str] = set()
    
    for prop_info in it.chain(
            schema_info.properties.values(),
            schema_info.anyOf,
            [schema_info.child(schema_info.items)] if schema_info.items else [],
        ):
        if prop_info.refname and prop_info.is_reference():
            children.add(prop_info.refname)
        else:
            children.update(get_children(prop_info))
            
    return children
    

for name, schema in definitions.items():
    graph[name] = get_children(schema)       
```

I'm still double-checking if this captures everything but so far it looks promising :) Unrelated to this PR, I'm trying to analyze how types are related in general and how we could simplify this in Vega-Lite.

dangotbanned: @binste 

Adapting your original code, I've managed to select the schemas I wanted with:

```py
import keyword
def is_relevant_object(schema: SchemaInfo) -> bool:
    EXCLUDE = {"ExprRef", "RelativeBandSize", "ParameterPredicate"}
    return (
        schema.ref
        and schema.refname not in EXCLUDE
        and schema.properties
        and schema.type == "object"
        and not schema.is_value()
        and "field" not in schema.required
        and not (keyword.iskeyword(next(iter(schema.required), "")))
    )
```

Can you try plugging that in as a filter somewhere in your non-recursive version? https://github.com/vega/altair/pull/3536#discussion_r1748059188

tools/schemapi/utils.py
```diff
@@ -373,140 +387,163 @@ def title(self) -> str:
...
-                    subschema.get_python_type_representation(
-                        for_type_hints=for_type_hints
-                    )
+                    subschema.to_type_repr(target=target, use_concrete=use_concrete)
                 )
-            type_representations.extend(options)
-        elif self.is_object() and not for_type_hints:
-            type_representations.append("dict")
+            tps.update(options)
         elif self.is_array():
```
dangotbanned, line 469: @binste

Related: https://github.com/vega/altair/pull/3536#discussion_r1748059188

If we can factor out this assignment, then there is a pathway to using `SchemaInfo` objects themselves in a `set`.

Provided we can find a good candidate for `SchemaInfo.__hash__()`

tools/generate_schema_wrapper.py
binste: If you want, you can paste your code here and I'll add it directly in there as I'm not sure how it's most useful to you/what the output is you need. Else, you can try to replace `if prop_info.refname and prop_info.is_reference()` in my example code with the latest `is_relevant_object`

dangotbanned: > If you want, you can paste your code here and I'll add it directly in there as I'm not sure how it's most useful to you/what the output is you need.

<details>
<summary>Full (old) code</summary>

```py
import keyword
from itertools import chain

from tools import generate_schema_wrapper as gsw
from tools.schemapi.utils import SchemaInfo

schema = gsw.load_schema(Path("altair/vegalite/v5/schema/vega-lite-schema.json"))
config = SchemaInfo({"$ref": "#/definitions/Config"}, rootschema=schema)


def is_relevant_object(schema: SchemaInfo):
    EXCLUDE = {"ExprRef", "RelativeBandSize", "ParameterPredicate"}
    return (
        schema.ref
        and schema.refname not in EXCLUDE
        and schema.properties
        and schema.type == "object"
        and not schema.is_value()
        and "field" not in schema.required
        and not (keyword.iskeyword(next(iter(schema.required), "")))
    )


def get_all_objects(
    schema: SchemaInfo, recursion_depth: int = 0
) -> dict[str, SchemaInfo]:
    objects: dict[str, SchemaInfo] = {}
    if is_relevant_object(schema):
        objects[schema.refname] = schema

    if recursion_depth < 10:
        it = chain(
            schema.properties.values(),
            schema.anyOf,
            [schema.child(schema.items)] if schema.items else (),
        )
        for prop_info in it:
            objects.update(
                get_all_objects(prop_info, recursion_depth=recursion_depth + 1)
            )
    return objects


all_objects = get_all_objects(config)
all_objects
```

</details>

I didn't get to the recursive part yet, but that filter works

tools/schemapi/utils.py
dangotbanned: I'm not sure if the benefit to this is immediately obvious, but I can explain if not?

binste: I think I see where you're coming from. Just creating `SchemaInfo` based on strings sounds like a simple way which works now instead of trying to rewrite that part of the code and think about hashing the `SchemaInfo` :) Except you see a very simple way, I'm not sure if it's worth the effort.

binste: Unrelated, but while we're chatting, I'm trying to organize a call among Altair maintainers to catch up, give an update on the NumFOCUS project summit Mattijn and I went too, and discuss priorities. Of course there is no obligation to participate and we can move important discussions to GitHub. I did not feel an email where I could invite you under so if this is of interest to you, feel free to reach out to me at binder_stefan@outlook.com so I have your contact info.

dangotbanned: > Unrelated, but while we're chatting, I'm trying to organize a call among Altair maintainers to catch up, give an update on the NumFOCUS project summit Mattijn and I went too, and discuss priorities. Of course there is no obligation to participate and we can move important discussions to GitHub. I did not feel an email where I could invite you under so if this is of interest to you, feel free to reach out to me at [binder_stefan@outlook.com](mailto:binder_stefan@outlook.com) so I have your contact info.

Definitely interested, just pinged you now. 

Will probably make me email public at some point, but you can share it with the other maintainers

pyproject.toml
```diff
@@ -448,3 +448,10 @@ extraPaths=["./tools"]
 pythonPlatform="All"
 pythonVersion="3.8"
 reportUnusedExpression="none"
+include=[
+    "./altair/**/*.py",
+    "./doc/*.py",
+	"./sphinxext/**/*.py",
+	"./tests/**/*.py",
+	"./tools/**/*.py",
```
dangotbanned, line 456: # TODO
Fix this on `main`

tools/generate_schema_wrapper.py
```diff
@@ -894,6 +895,44 @@ def generate_config_typed_dicts(fp: Path, /) -> Iterator[str]:
...
+    if depth < MAX_DEPTH:
+        for prop_info in iter_children(info):
+            seen.update(find_relevant_schemas(prop_info, depth=depth + 1))
+    return seen
+
+
+def wip_config_schemas(fp: Path, /) -> set[SchemaInfo]:
+    schema = load_schema(fp)
+    config = SchemaInfo({"$ref": "#/definitions/Config"}, rootschema=schema)
+    return find_relevant_schemas(config)
```
dangotbanned, line 933: @binste thought I'd leave this here in case you decided to dip in

tools/generate_schema_wrapper.py
```diff
@@ -894,6 +895,44 @@ def generate_config_typed_dicts(fp: Path, /) -> Iterator[str]:
...
+def find_relevant_schemas(info: SchemaInfo, depth: int = 0) -> set[SchemaInfo]:
+    """Equivalent to `get_all_objects`."""
+    MAX_DEPTH = 6
+    seen: set[SchemaInfo] = set()
+    if is_relevant_schema(info) and info not in seen:
+        seen.add(info)
+    if depth < MAX_DEPTH:
+        for prop_info in iter_children(info):
+            seen.update(find_relevant_schemas(prop_info, depth=depth + 1))
+    return seen
```
dangotbanned, line 927: Thinking the `if depth < MAX_DEPTH` might be avoidable. 

# TODO
Try

```py
def find_relevant_schemas(info: SchemaInfo, /) -> set[SchemaInfo]:
    """Equivalent to `get_all_objects`."""
    seen: set[SchemaInfo] = set()
    if is_relevant_schema(info) and info not in seen:
        seen.add(info)
    for prop_info in iter_children(info):
        if prop_info not in seen:
            seen.update(find_relevant_schemas(prop_info))
    return seen
```

tools/schemapi/utils.py
```diff
@@ -362,6 +365,16 @@ def __setattr__(self, name: str, value: Any) -> Never:
...
 
+    def __hash__(self) -> int:
+        return hash(_HASH_ENCODER.encode(self.schema))
+
+    def __eq__(self, value: object) -> bool:
+        if isinstance(value, SchemaInfo):
+            if self.ref:
+                return self.ref == value.ref
+            return hash(self) == hash(value)
+        return False
```
dangotbanned, line 376: # TODO
May be able to avoid the `hash` branch, with 

```py
class SchemaInfo:
	def __eq__(self, value: object) -> bool:
		if isinstance(value, SchemaInfo):
			if self.ref:
				return self.ref == value.ref
			return self.schema == value.schema
		return False
```

https://docs.python.org/3.8/library/stdtypes.html#dict.values
> Dictionaries compare equal if and only if they have the same `(key, value)` pairs (regardless of ordering). 
> Order comparisons (‘<’, ‘<=’, ‘>=’, ‘>’) raise [TypeError](https://docs.python.org/3.8/library/exceptions.html#TypeError).

Not sure if `dict.__eq__` short-circuits on the first non-eq pair, but could write that in if needed

tools/generate_schema_wrapper.py
```diff
@@ -894,6 +895,44 @@ def generate_config_typed_dicts(fp: Path, /) -> Iterator[str]:
...
+        and "field" not in info.required
+        and not (iskeyword(next(iter(info.required), "")))
+    )
+
+
+def iter_children(info: SchemaInfo, /) -> Iterator[SchemaInfo]:
+    yield from info.properties.values()
+    yield from info.anyOf
+    if info.items:
+        yield info.child(info.items)
```
dangotbanned, line 915: # TODO

Find a more accurate name for what this is doing and add as `SchemaInfo.iter_...` (instance method).

I don't like the overlap with `SchemaInfo.child`, since they aren't really related

tools/schemapi/utils.py
dangotbanned: Fixed in [`a70f259` (#3536)](https://github.com/vega/altair/pull/3536/commits/a70f259f0e711fbaa69205976ed5a04e1633611c)

altair/vegalite/v5/schema/_config.py
dangotbanned: I think this might not be relevant anymore since `Map` is used everywhere now instead of `dict`
- https://github.com/vega/altair/pull/3573

dangotbanned: https://altair-viz.github.io/user_guide/generated/core/altair.LegendConfig.html#altair.LegendConfig

`LegendConfig.layout` seems to not have a description?

pyproject.toml
dangotbanned: Fixed in https://github.com/vega/altair/pull/3583

tools/generate_schema_wrapper.py
```diff
@@ -702,82 +763,127 @@ def generate_vegalite_channel_wrappers(
...
+    EXCLUDE = {"ExprRef", "RelativeBandSize", "ParameterPredicate"}
+    return bool(
+        info.ref
+        and info.refname not in EXCLUDE
+        and info.properties
+        and info.type == "object"
+        and not info.is_value()
+        and "field" not in info.required
+        and not (iskeyword(next(iter(info.required), "")))
+    )
```
dangotbanned, line 865: # TODO
Rename `is_relevant_schema` and add as a `SchemaInfo` method.

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,7078 @@
...
+        ``"normal"``) or a number (``100``, ``200``, ``300``, ..., ``900`` where
+        ``"normal"`` = ``400`` and ``"bold"`` = ``700``).
+    frame
+        The reference frame for the anchor position, one of ``"bounds"`` (to anchor relative
+        to the full bounding box) or ``"group"`` (to anchor relative to the group width or
+        height).
+    limit
+        The maximum allowed length in pixels of title and subtitle text.
+    lineHeight
+        Line height in pixels for multi-line tit
```
dangotbanned, line 7060: `padding` is an interesting edge case. 

The definition is a union that contains an inline `dict`.

Currently the `"googlecharts"` theme fails type checking, due usage of a `dict` here:

```json
"Padding": {
      "anyOf": [
        {
          "type": "number"
        },
        {
          "additionalProperties": false,
          "properties": {
            "bottom": {
              "type": "number"
            },
            "left": {
              "type": "number"
            },
            "right": {
              "type": "number"
            },
            "top": {
              "type": "number"
            }
          },
          "type": "object"
        }
      ],
      "minimum": 0
    }
```

tests/vegalite/v5/test_theme.py
```diff
@@ -52,3 +53,665 @@ def custom_theme() -> dict[str, int]:
...
+                "fill": "#525252",
+                "font": 'IBM Plex Sans,system-ui,-apple-system,BlinkMacSystemFont,".sfnstext-regular",sans-serif',
+                "fontWeight": 400,
+            },
+            "guide-title": {
+                "fill": "#525252",
+                "font": 'IBM Plex Sans,system-ui,-apple-system,BlinkMacSystemFont,".sfnstext-regular",sans-serif',
+                "fontWeight": 400,
+            },
+        },
```
dangotbanned, line 148: `"guide-(style|label)"` seem to have been skipped during schema gen, assuming the issue is the hyphen.

There are 4 hyphenated properties here:
https://github.com/vega/altair/blob/682b1695526d523b769f419b2dade1e32f5383de/altair/vegalite/v5/schema/vega-lite-schema.json#L28138-L28153

dangotbanned: @binste are you aware of this issue?

They appear in the docstring, but not in the signature for `core.StyleConfigIndex`

https://github.com/vega/altair/blob/682b1695526d523b769f419b2dade1e32f5383de/altair/vegalite/v5/schema/core.py#L22523-L22552

dangotbanned: Fixed in [`0bdfa72` (#3536)](https://github.com/vega/altair/pull/3536/commits/0bdfa72470551b2bc39e4adc76b7e989ccdf259b)

altair/vegalite/v5/schema/_config.py
dangotbanned: # FIXME

Temporary solution in [`e564ca5` (#3536)](https://github.com/vega/altair/pull/3536/commits/e564ca5a9a163d40669a0e9f1634fe93d16d9ea5)

This is enough for CI passing, but ideally don't want this merged

dangotbanned: Fixed in [`4933e54` (#3536)](https://github.com/vega/altair/pull/3536/commits/4933e547ae863eb247ea989f5fde37e2dce6019e)

altair/typing.py
```diff
@@ -46,11 +46,13 @@
     "ChartType",
     "EncodeKwds",
     "Optional",
+    "ThemeConfig",
     "is_chart_type",
 ]
 
 from altair.utils.schemapi import Optional
 from altair.vegalite.v5.api import ChartType, is_chart_type
+from altair.vegalite.v5.schema._config import ThemeConfig
```
dangotbanned, line 55: I'm starting off quite conservative here, since there are many new symbols added in `_config.py`.

You can utilise all the `TypedDict`(s) without actually using the classes, like in https://github.com/vega/altair/pull/3536/files#diff-e52efec94b8b7f0f1b27eac85385173dacfaca37d65afd0df9839e09e58bf377

Personally, I'd find it helpful to have a select few available in `altair.typing` and then the rest from one level down.

So this could look like:

```py

from altair.typing import ThemeConfig, ConfigKwds
from altair.typing.theme import LegendConfigKwds, DateTimeKwds, ...
```

The above is just an example, particularly trying to avoid the conflict with `themes` being a variable

# Related
- https://github.com/vega/altair/issues/2918
- https://github.com/vega/altair/issues/3337

altair/vegalite/v5/schema/_typing.py
```diff
@@ -86,6 +119,85 @@ def func(
...
+
+    column: T
+    row: T
+
+
+class PaddingKwds(TypedDict, total=False):
+    bottom: float
+    left: float
+    right: float
+    top: float
```
dangotbanned, line 198: I can define these in `_config.py` instead if there's interest?

Tried to wrap things up since this PR has gotten quite large

altair/vegalite/v5/schema/vega-themes.json
```diff
@@ -28,12 +28,6 @@
     "circle": {
       "fill": "#6929c4"
     },
-    "group": {
-      "fill": "#ffffff"
-    },
-    "path": {
-      "stroke": "#6929c4"
-    },
```
dangotbanned, line 36: See 79804ec6b1ee009c83d6993e390a040832e3bafc

altair/utils/theme.py
```diff
@@ -16,12 +17,12 @@
     from altair.utils.plugin_registry import PluginEnabler
     from altair.vegalite.v5.theme import AltairThemes, VegaThemes
 
-ThemeType = Plugin[Dict[str, Any]]
+ThemeType = Plugin[ThemeConfig]
```
dangotbanned, line 20: Keeping this `ThemeType` alias for backwards compat, but using `ThemeConfig` in annotations.

`ThemeConfig` has full docs and can be used to navgiate through the other `TypedDict`(s)

tests/vegalite/v5/test_theme.py
```diff
@@ -23,10 +36,952 @@ def test_vega_themes(chart) -> None:
...
+    """
+    Only including **1/4** of `carbon`_ , which gives sufficient structural coverage.
+
+    .. _carbon:
+        https://github.com/vega/vega-themes/blob/5f1a5c5b22cc462cf3d46894212152b71cfe964f/src/carbongen.ts
+    """
+    return ThemeConfig(
+        config={
+            "arc": {"fill": "#6929c4"},
+            "area": {"fill": "#6929c4"},
```
dangotbanned, line 77: Noting here that `mypy` runs are now **significantly** longer, due to all the inference required for `test_theme.py`.

I think all of this is worth it right now for completeness, but if this is a concern I can trim these tests down

tests/vegalite/v5/test_theme.py
```diff
@@ -23,10 +36,952 @@ def test_vega_themes(chart) -> None:
...
+                    "#666666",
+                    "#DCDCDC",
+                ]
+            },
+            "rect": {"fill": "#000"},
+        }
+    }
+
+
+# FIXME: Support key-completion for `ThemeConfig.padding` ``Padding``
```
dangotbanned, line 338:
```suggestion
```

tests/vegalite/v5/test_theme.py
binste: Really looking forward to when [`ruff` adds `mypy` compatible linting](https://github.com/astral-sh/ruff/issues/3893#issuecomment-2119208889) :)

altair/typing.py
binste: Having it in `typing.theme` sounds good to me! I would probably also want to use some of them in my own code, do see the value of having a select few available but totally fine if we tackle this in a follow-up PR, whatever works best for you!

altair/vegalite/v5/schema/_config.py
```diff
@@ -0,0 +1,7773 @@
+# The contents of this file are automatically written by
+# tools/generate_schema_wrapper.py. Do not modify directly.
+
+from __future__ import annotations
+
+import sys
+from typing import TYPE_CHECKING, Any, Literal, Sequence, TypedDict
+
+if sys.version_info >= (3, 14):
```
binste, line 9: Maybe already on your radar with #3587. This probably requires a change in `pyproject.toml` so that `typing_extensions` is already installed for Python 3.13? https://github.com/vega/altair/blob/bf29786819a4287de83adfe034d927d1cc208ac8/pyproject.toml#L18

altair/vegalite/v5/schema/_typing.py
```diff
@@ -86,6 +119,85 @@ def func(
...
+
+.. _hexadecimal:
+    https://www.w3schools.com/html/html_colors_hex.asp
+"""
+
+
+def is_color_hex(obj: Any) -> TypeIs[ColorHex]:
+    """Return ``True`` if the object is a hexadecimal color code."""
+    # NOTE: Extracts compiled pattern from metadata,
+    # to avoid defining  in multiple places.
```
binste, line 169:
```suggestion
    # to avoid defining in multiple places.
```

altair/vegalite/v5/schema/vega-themes.json
binste: Wow nice! Wasn't aware that such cases exist

binste (approved):
This PR blows my mind! 🤯 Really appreciate all the work you put into it, this obviously required going very deep in the VL schema and the code generation and I think the outcome is super helpful for further advancing the UX of themes. Now I really want to write some new themes 😄 

I only have some very minor comments. If you feel good with it, I'm ok if you merge it afterwards!

altair/vegalite/v5/schema/_config.py
dangotbanned: > Maybe already on your radar with #3587. This probably requires a change in `pyproject.toml` so that `typing_extensions` is already installed for Python 3.13?

Yeah definitely on my radar, I'd already made the change in https://github.com/vega/altair/pull/3591/files

That PR will be blocked until https://github.com/geopandas/pyogrio/issues/470 is resolved though

@binste what do you think about splitting that change into another PR that we can merge quickly?

I'm thinking it would be helpful to have a dependency change in our release notes; even though it is technically just "unbreaking" `3.13`

altair/vegalite/v5/schema/_typing.py
dangotbanned: In [`aaffebd` (#3536)](https://github.com/vega/altair/pull/3536/commits/aaffebd2b2d037024a2f4af1e6d942bdd9e24174) I settled on including them in `_config.__all__`

altair/typing.py
dangotbanned: > Having it in `typing.theme` sounds good to me! I would probably also want to use some of them in my own code, do see the value of having a select few available but totally fine if we tackle this in a follow-up PR, whatever works best for you!

Perfect!

We can always iron this out before `v5.5.0`, but I built the docs with [`12b6410` (#3536)](https://github.com/vega/altair/pull/3536/commits/12b6410b85d278a051b16cbf3054b6f0a261be83) and it came out fine

dangotbanned:
> This PR blows my mind! 🤯 Really appreciate all the work you put into it, this obviously required going very deep in the VL schema and the code generation and I think the outcome is super helpful for further advancing the UX of themes. Now I really want to write some new themes 😄
> 
> I only have some very minor comments. If you feel good with it, I'm ok if you merge it afterwards!

Thanks so much for the review @binste! (*and helping me get unstuck midway through*)

Hopefully this will make it easier to explore some of your points in https://github.com/vega/altair/issues/3519#issuecomment-2292010192 once we've merged

I'll leave this open for a few hours in case you had any thoughts on https://github.com/vega/altair/pull/3536#discussion_r1761463506 - otherwise I'll follow this up with another PR as mentioned

mattijn:
Thanks for all the work. A concise PR description with an explanation of the what/why/how would really help future readers coming to this PR.
This description is also helpful for making the release notes better. Thanks again!

dangotbanned:
> Thanks for all the work. A concise PR description with an explanation of the what/why/how would really help future readers coming to this PR. This description is also helpful for making the release notes better. Thanks again!

No worries, will work on that now thanks @mattijn

dangotbanned:
@mattijn just updated the description, hopefully you (and others) find it helpful 😄

mattijn:
Thanks! Can you add also how this feature is being used? Do we want users to import the following? 

```python
import altair.vegalite.v5.schema._config as theme
```
And
```python
from altair.vegalite.v5.schema._config import ThemeConfig
```
And then? I'd love to have a brief description that explains how a new user can start using this feature. It would be great if you could provide a simple example that highlights just one feature controlled by the `ThemeConfig`. Thanks again for all your hard work!

joelostblom:
Thanks for expanding on the PR description, that was helpful for me to appreciate just how much this PR improves the experience of defining custom themes, wow! In line with what Mattijn suggested, it would be great to [update the custom theme section of the docs](https://altair-viz.github.io/user_guide/customization.html#defining-a-custom-theme) to use this approach and mention the autocompletion behavior.

dangotbanned:
> Thanks! Can you add also how this feature is being used? Do we want users to import the following?
> 
> ```python
> import altair.vegalite.v5.schema._config as theme
> ```
> 
> And
> 
> ```python
> from altair.vegalite.v5.schema._config import ThemeConfig
> ```
> 
> And then? I'd love to have a brief description that explains how a new user can start using this feature. It would be great if you could provide a simple example that highlights just one feature controlled by the `ThemeConfig`. Thanks again for all your hard work!

Reading this made me spot a bug with the import behavior, just fixed in [`d4bd6db` (#3536)](https://github.com/vega/altair/pull/3536/commits/d4bd6db3721adbb05aa688289ba52f9a206b7fa9)

For reference, `@register_theme` has an example of the intended import path

https://github.com/dangotbanned/altair/blob/6e4af95becc8ff9be033535d45ce3db9b2c749a3/altair/vegalite/v5/theme.py#L94-L135

dangotbanned:
@joelostblom @mattijn thanks for your comments!

I'll be updating the description with some example tomorrow, but hopefully you get a rough idea with [`d4bd6db` (#3536)](https://github.com/vega/altair/pull/3536/commits/d4bd6db3721adbb05aa688289ba52f9a206b7fa9)

> it would be great to [update the custom theme section of the docs](https://altair-viz.github.io/user_guide/customization.html#defining-a-custom-theme) to use this approach and mention the autocompletion behavior.

Absolutely agree, I added that to the list in https://github.com/vega/altair/issues/3519#issue-2447214476 an hour ago - I think it will make the remaining items much easier also

mattijn:
Awesome! Looking forward to! Another question I have, why is this `ThemeConfig` not imported to the root level of Altair? It sounds like a top level object to me. Can you elaborate on this? With feature comes questions😀. Thanks!

dangotbanned:
> Awesome! Looking forward to! Another question I have, why is this `ThemeConfig` not imported to the root level of Altair? It sounds like a top level object to me. Can you elaborate on this? With feature comes questions😀. Thanks!

For `v6` I would agree, but right now I think this is much easier to find.

I can't seem to find the thread, but I suggested to @binste having a small number of `...Kwds` available in `altair.typing` - with the rest in `altair.typing.theme`.

With the current layout, I think this keeps the concept that these are all linked and that `ThemeConfig` is the entry point.

I'm not sure if all that made sense, if not apologies, I was supposed to call it day already 😅

altair/vegalite/v5/schema/_config.py
dangotbanned: Resolved in https://github.com/vega/altair/pull/3593

dangotbanned:
@mattijn @joelostblom 

FYI the [conversation](https://github.com/vega/altair/pull/3536/files#r1764650814) marked as *unresolved* is actually resolved. 

I haven't updated the status of it to block merging until you've both seen the [updated description](https://github.com/vega/altair/pull/3536#issue-2461819263)

Otherwise this should be ready to merge 👍

joelostblom:
I don't have much to comment on the changes to the import, so will refer to @mattijn for that. 

On first look, the class based syntax looks maybe a bit verbose:

image

So I would probably opt for the combination with dicts myself:

image

But maybe I will change my mind when I try them; it's great that both exist and support autocompletion!

dangotbanned:
@mattijn I'm planning to merge this tomorrow if you don't have any objections.

In #3600 I've been trying to work around not having the refactored `tools/...` and have made a lot of progress; but I need to start making changes in `tools.generate_schema_wrapper.py` and `tools.schemapi.utils.py` soon for generating `alt.expr.__init__.py`

mattijn:
Thanks for asking! I still cannot understand why this `ThemeConfig` feature is available within the `altair.types`. I thought we declared that `altair.types` as the public interface of types within altair. But now we are placing features such as `ThemeConfig` in there. Is it a type? Therefore it does not help me with discoverability either. 

Also, as a user, I really prefer a feature explanation that I can copy and paste. I don't see that in this PR. I see a few screenshots, but that is really not the advocated route for a [Minimal, Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example). For me it is still not clear how I can use this from the PR description.

I found the following in the code:

```python
import altair as alt
from altair.typing import ThemeConfig


@alt.register_theme("param_font_size", enable=True)
def custom_theme() -> ThemeConfig:
    sizes = 12, 14, 16, 18, 20
    return {
        "autosize": {"contains": "content", "resize": True},
        "background": "#F3F2F1",
        "config": {
            "axisX": {"labelFontSize": sizes[1], "titleFontSize": sizes[1]},
            "axisY": {"labelFontSize": sizes[1], "titleFontSize": sizes[1]},
            "font": "'Lato', 'Segoe UI', Tahoma, Verdana, sans-serif",
            "headerColumn": {"labelFontSize": sizes[1]},
            "headerFacet": {"labelFontSize": sizes[1]},
            "headerRow": {"labelFontSize": sizes[1]},
            "legend": {"labelFontSize": sizes[0], "titleFontSize": sizes[1]},
            "text": {"fontSize": sizes[0]},
            "title": {"fontSize": sizes[-1]},
        },
        "height": {"step": 28},
        "width": 350,
    }


source = data.stocks()
lines = (
    alt.Chart(source, title=alt.Title("Stocks"))
    .mark_line()
    .encode(x="date:T", y="price:Q", color="symbol:N")
)
lines.interactive(bind_y=False)
```

That works! But I also see another approach. Here you define a `custom_theme` from the `ThemeConfig`

```python
from altair.typing import ThemeConfig, theme

custom_theme = ThemeConfig(
    config=theme.ConfigKwds(
        axis=theme.AxisConfigKwds(grid=True, labelFont="system-ui"),
        circle=theme.MarkConfigKwds(fill="purple"),
    )
)

custom_theme
```
How do I register this as theme?

I searched for `register_theme` within the altair documentation page, but nothing was returned https://altair-viz.github.io/search.html?q=register_theme. A brief description to the documentation page would help the average user a lot here.

Last, after I enabled the custom theme, how can I de-register this custom theme and go back to the default settings?

Thank you for your patience!

dangotbanned:
Thanks for your [response](https://github.com/vega/altair/pull/3536#issuecomment-2365430472) @mattijn!

I'm going to try my best to address your concerns in a few follow-up comments.

Just finished writing up the first (of 2 or 3), so please know there is more to come

dangotbanned:
@mattijn 
> I still cannot understand why this `ThemeConfig` feature is available within the `altair.types`. I thought we declared that `altair.types` is the interface of public types within altair. But now we are placing features such as `ThemeConfig` in there. Is it a type? Therefore it does not help me with discoverability either.

So in the description I provided a link that can clear this up:
> This PR aims to improve the UX when authoring a theme.
> By utilising [TypedDict(s)](https://typing.readthedocs.io/en/latest/spec/typeddict.html) ...

I've picked out some key quotes, but I would really recommend reading the section [Using TypedDict Types](https://typing.readthedocs.io/en/latest/spec/typeddict.html#using-typeddict-types) as this illustrates how `TypedDict`(s) differ from regular classes:
> - The created `TypedDict` type object is not a real class object. 
> - When called, the `TypedDict` type object returns an ordinary dictionary object at runtime
> - In particular, `TypedDict` type objects cannot be used in `isinstance()` tests such as `isinstance(d, Movie)`.

Back to your question:
> Is it a type?

I'm afraid this is complex question to answer, but I will do my best to summarize.
- `TypedDict` is defined in [typing](https://docs.python.org/3/library/typing.html) - **not** as a class - but [as a function](https://github.com/python/cpython/blob/342e654b8eda24c68da64cc21bc9583e480d9e8e/Lib/typing.py#L3267)
	- <details><summary>Screenshot</summary></details>
- There is currently an open proposal [PEP 747](https://peps.python.org/pep-0747) which would define (among other symbols) that a `TypedDict` is a [TypeForm](https://peps.python.org/pep-0747/#motivation)
	- The section [Why not `type[C]`?](https://peps.python.org/pep-0747/#why-not-type-c) also further illustrates the distinction

The short (but seemingly a non)answer is that it is **not a real class**.

I believe that if we were to add these `TypedDict`(s) to the top-level, that it *could* cause confusion for users that (understandably) expect them to work the same as `SchemaBase` classes.

dangotbanned:
# Screenshots vs Code blocks

@mattijn
> Also, as a user, I really prefer a feature explanation that I can copy and paste. I don't see that in this PR. I see a few screenshots, but that is really not the advocated route for a [Minimal, Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example). For me it is still not clear how I can use this from the PR description.

I have now updated the PR description with collapsible code blocks to accompany each screenshot.

I would absolutely agree that providing a [Minimal, Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example) is helpful, although I thought that related more to bugs/issues?

There were a few reasons leaned towards screenshots:

1. (**Most importantly**) the key benefits to this PR are seen in an **IDE** - not at runtime:
    - Autocompletion for `dict` keys
    - Static type checking feedback
    - Documentation
2. I had already provided a comprehensive [test suite](https://github.com/dangotbanned/altair/blob/8519444f30c6421aae18403bff5c9aca1f366e63/tests/vegalite/v5/test_theme.py) containing code for 13 theme definitions [when asked](https://github.com/vega/altair/pull/3536#issuecomment-2353524664) to update the description
    - I seem to have misinterpreted your request as wanting something *more* than copy/pasting these tests
    - I did [mention them](https://github.com/vega/altair/pull/3536#issue-2461819263) and provide a link before presenting the screenshots
        > Using the classes themselves is entirely *optional*, as can be seen in [`test_theme.py`](https://github.com/vega/altair/pull/3536/files#diff-e52efec94b8b7f0f1b27eac85385173dacfaca37d65afd0df9839e09e58bf377).
3. Most other typing-focused PRs have also used screenshots to communicate the impact
    - https://github.com/vega/altair/pull/3208
    - https://github.com/vega/altair/pull/2949
    - https://github.com/vega/altair/pull/2927
    - https://github.com/vega/altair/pull/3455
    - https://github.com/vega/altair/pull/3526
    - https://github.com/vega/altair/pull/3516

dangotbanned:
# @alt.register_theme
@mattijn 
> How do I register this as theme?

I've updated the PR description, this is the example under the heading *Options for imports & runtime typing*

> I searched for `register_theme` within the altair documentation page, but nothing was returned [altair-viz.github.io/search.html?q=register_theme](https://altair-viz.github.io/search.html?q=register_theme). A brief description to the documentation page would help the average user a lot here.

Well spotted!
I wasn't aware this hadn't been included in the API reference yet, but will open a new issue/PR to address it - since that is orthogonal to this PR.

I was waiting until this PR was finished before introducing `@register_theme` to the **User Guide**, since they complement eachother well.

There are a few other changes I've noted for the **User Guide** in https://github.com/vega/altair/issues/3519, including `ThemeConfig` as mentioned in https://github.com/vega/altair/pull/3536#issuecomment-2353877227

> Last, after I enabled the custom theme, how can I de-register this custom theme and go back to the default settings?

Exactly the same way as documented in https://altair-viz.github.io/user_guide/customization.html#defining-a-custom-theme

In https://github.com/vega/altair/pull/3526 I've described this as a convenience for the common case where you use a single theme.
That isn't the only possible use-case, but it is the motivating one

dangotbanned:
I didn't mean to trigger a merge to `main` with https://github.com/vega/altair/commit/712680b2965eede4c40d93f52d3f963c5d181587

I was trying to fix the conflicts but now I see the comment I'd mentioned was unresolved in https://github.com/vega/altair/pull/3536#issuecomment-2357877156 has been resolved

@binste can we revert this so that @mattijn has time to respond please?

dangotbanned:
Thanks @mattijn for https://github.com/vega/altair/commit/3d97faeeb53141cffb1a5bacaa90018982414090

I'm not sure what would be the best way to proceed here, should I open a new PR?

mattijn:
Thanks for the added explanations! Appreciated. I like the suggestion you provide in https://github.com/vega/altair/pull/3536#issuecomment-2366736230 to consolidate everything under a single namespace, including the proposed `ThemeConfig`. I’m fine with doing that now, rather than releasing a version first and requiring deprecation immediately after.

However, instead of phasing out the plural `altair.themes` in favour of the singleton `altair.theme` can we consolidate this PR's work into the existing `altair.themes`? For example, we can have `altair.themes.ThemeConfig` (and `altair.themes.AxisConfigKwds`, etc.) aligning with the existing `altair.themes.register()` and `altair.themes.enable()`.

Aesthetically I like the singleton `altair.theme`, but I think sticking with `altair.themes` is more practical, if that is also an option. Can you reflect on this?

Thank you for the added explanations regarding the type of `ThemeConfig`! I currently I don't think it needs to be registered as part of interface of altair its public types. The current `_TypedDictMeta` type seems fine and even if it later becomes something like `TypeForm`, it's essentially a dictionary (great!) at runtime if I understand correctly.

Also, much appreciated for including both code blocks and screenshots! Great combination as such!

mattijn:
I don't think it is possible to re-open a merged PR, even when the changes are reverted with the latest commit, probably the best option is to open a new PR😔

dangotbanned:
> Thanks for the added explanations! Appreciated. I like the suggestion you provide in [#3536 (comment)](https://github.com/vega/altair/pull/3536#issuecomment-2366736230) to consolidate everything under a single namespace, including the proposed `ThemeConfig`. I’m fine with doing that now, rather than releasing a version first and requiring deprecation immediately after.

> Thank you for the added explanations regarding the type of `ThemeConfig`! I currently don't think it needs to be registered as part of interface of altair its public types. The current `_TypedDictMeta` type seems fine and even if it later becomes something like `TypeForm`, it's essentially a dictionary (great!) at runtime if I understand correctly.
> 
> Also, much appreciated for including both code blocks and screenshots! Great combination as such!

Thanks @mattijn 

---

> However, instead of phasing out the plural `altair.themes` in favour of the singleton `altair.theme` can we consolidate this PR's work into the existing `altair.themes`? For example, can we have `altair.themes.ThemeConfig` (and `altair.themes.AxisConfigKwds`, etc.) aligning with the existing `altair.themes.register()` and `altair.themes.enable()`.
> 
> Aesthetically I like the singleton `altair.theme`, but I think sticking with `altair.themes` is more practical, if that is also an option. Can you reflect on this?

Okay so rather than fully revert this PR, I think the changes we want are strictly concerning a small refactor of the public interface.
#3526 was merged *after* the last release, so renaming + moving `@register_theme` is all good

I propose:
- We continue this discussion in a new issue
	- On the condition that **resolving it is blocking for the next release**
- Whatever solution we end up on becomes a small `refactor` PR to implement this interface
- (If necessary) update the title(s), description(s) for #3526, #3536
	- Either noting the interface changed prior to release
	- Or simply replacing references to the new symbols/import locations

How does that sound to you?

mattijn:
The changes requested were only concerning a small refactor of the public interface. Proposal sounds good!

PR #3538, opened 2024-08-15
feat(typing): Generate `Literal`(s) using `"const"`

dangotbanned:
# Description
Like #3535, this was another discovery while working through issues in #3536.
Split into another PR as this alters existing generated code.

The core change here is using `Literal` for both (`JSON`) properties `"enum"` and `"const"`.

In `python`'s type system, there is no need to distinguish between the two as `Literal` can represent one or more values.


binste (approved):
Nice! :) Always fascinating to see what changes in the autogenerated code based on changes in `tools`.

PR #3539, opened 2024-08-15
ci: bump `ruff>=0.6.0`

dangotbanned:
https://github.com/astral-sh/ruff/releases/tag/0.6.0

Resolves the need for another PR mentioned in https://github.com/vega/altair/pull/3536#discussion_r1715709800

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ all = [
 ]
 dev = [
     "hatch",
-    "ruff>=0.5.7",
+    "ruff>=0.6.0",
     "ibis-framework[polars]",
     "ipython[kernel]",
     "pandas>=0.25.3",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -231,7 +231,9 @@ extend-safe-fixes=[
     # ends-in-period
     "D400",
     # missing-return-type-special-method 
-    "ANN204"
+    "ANN204",
+    # unnecessary-dict-comprehension-for-iterable
+    "C420",
 ]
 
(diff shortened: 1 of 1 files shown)


joelostblom:
Just a note if you wonder why I added an additional label to this PR @dangotbanned . We try to always have one of the labels listed here on each PR: https://github.com/vega/altair/blob/main/.github/release.yml. GitHub then organize each PR into its corresponding section of the release notes automatically when creating the release via GitHub's interface. I don't have a strong opinion about labels not listed there, so feel free to add more to that list if there are some additional labels/sections you think would be useful.

I looked into automating the addition of a label to each PR via a GitHub action based on the conventional commit `type` mentioned in the PR title, but there was no such action so I just check through the closed PRs every now and then and label the ones that are missing one of the four labels our release notes are based on.

dangotbanned:
No worries @joelostblom 

I won't be able to test right now, but https://github.com/vega/altair/blob/main/.github/release.yml looks like you can map multiple labels to each title?

E.g. this had the `dependencies` label, which could fall under `maintenance`.

If that can work, I think that would be worth doing as there are other labels that wouldnt be picked up currently

joelostblom:
Yup, [we can map additional labels](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#example-configurations) just like you said. Another option would be to remove the labels we haven't used much in the past like `dependencies` and `usability`. I don't really have a preference since I mostly use the four other labels, but if you believe that these additional labels are useful, then I'm happy for them to be added.

PR #3545, opened 2024-08-18
fix(typing): Resolve misc type ignores in `schemapi.py`

dangotbanned:
These are all very low-priority, but allows using `"python.analysis.typeCheckingMode": "basic"` in this module, without any warnings.

Planning to investigate improving the peformance of validation, which will be much easier with a live type checker

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -114,7 +114,7 @@ def validate_jsonschema(
     rootschema: dict[str, Any] | None = ...,
     *,
     raise_error: Literal[True] = ...,
-) -> None: ...
+) -> Never: ...
 
 
 @overload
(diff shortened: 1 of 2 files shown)


binste:
Thanks! I've also noticed that validation can take up a significant amount of time in contexts where milliseconds matter such as in web applications. In an application at work, I had to disable validation of charts when it's deployed to production. Maybe there are things in Python we could improve but I think part of it is also just how the `jsonschema` library works. Curious to see what you find out @dangotbanned! There are other libraries such as https://github.com/horejsek/python-fastjsonschema and https://github.com/Stranger6667/jsonschema-rs but we'd need something which is also available in pyodide and from which we can continue to produce user-friendly error messages.

PR #3547, opened 2024-08-19
perf(DRAFT): Lazy schema validation & error messages

dangotbanned:
First non-failing version.
Have left most of the original code in. 

Planning to migrate & adapt the comments before removing.

### DIFF
--- a/tools/schemapi/schemapi.py
+++ b/tools/schemapi/schemapi.py
@@ -41,7 +41,7 @@ from altair import vegalite
 if TYPE_CHECKING:
     from typing import ClassVar
 
-    from jsonschema.protocols import Validator
+    from jsonschema.protocols import Validator, _JsonParameter
     from referencing import Registry
 
     from altair.typing import ChartType
@@ -109,7 +109,7 @@ def debug_mode(arg: bool) -> Iterator[None]:
 
 @overload
 def validate_jsonschema(
-    spec: Any,
+    spec: _JsonParameter,
     schema: dict[str, Any],
     rootschema: dict[str, Any] | None = ...,
     *,
@@ -119,7 +119,7 @@ def validate_jsonschema(
 
 @overload
 def validate_jsonschema(
-    spec: Any,
+    spec: _JsonParameter,
     schema: dict[str, Any],
     rootschema: dict[str, Any] | None = ...,
     *,
@@ -128,7 +128,7 @@ def validate_jsonschema(
 
 
 def validate_jsonschema(
-    spec,
+    spec: _JsonParameter,
     schema: dict[str, Any],
     rootschema: dict[str, Any] | None = None,
     *,
@@ -169,7 +169,7 @@ def validate_jsonschema(
 # Everything else is skipped if this returns an empty `list`
 # TODO: Refactor to peek at possible error w/ `next(validator.iter_errors(spec))`
 def _get_errors_from_spec(
-    spec: dict[str, Any],
+    spec: _JsonParameter,
     schema: dict[str, Any],
     rootschema: dict[str, Any] | None = None,
 ) -> ValidationErrorList:
--- a/tools/schemapi/schemapi.py
+++ b/tools/schemapi/schemapi.py
@@ -205,13 +205,10 @@ def _get_errors_from_spec(
         validator_kwargs["registry"] = _get_referencing_registry(
             rootschema or schema, json_schema_draft_url
         )
-
     else:
         # No resolver is necessary if the schema is already the full schema
         validator_kwargs["resolver"] = (
-            jsonschema.RefResolver.from_schema(rootschema)
-            if rootschema is not None
-            else None
+            jsonschema.RefResolver.from_schema(rootschema) if rootschema else rootschema
         )
 
     validator = validator_cls(schema, **validator_kwargs)
--- a/tools/schemapi/schemapi.py
+++ b/tools/schemapi/schemapi.py
@@ -42,6 +42,7 @@ from altair import vegalite
 if TYPE_CHECKING:
     from typing import ClassVar
 
+    from jsonschema.protocols import Validator
     from referencing import Registry
 
     from altair.typing import ChartType
@@ -191,8 +192,9 @@ def _get_errors_from_spec(
     URI reference due to the characters such as '<'.
     """
     json_schema_draft_url = _get_json_schema_draft_url(rootschema or schema)
-    validator_cls = jsonschema.validators.validator_for(
-        {"$schema": json_schema_draft_url}
+    validator_cls: type[Validator] = cast(
+        "type[Validator]",
+        jsonschema.validators.validator_for({"$schema": json_schema_draft_url}),
     )
     validator_kwargs: dict[str, Any] = {}
     if hasattr(validator_cls, "FORMAT_CHECKER"):
--- a/tools/schemapi/schemapi.py
+++ b/tools/schemapi/schemapi.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import contextlib
 import inspect
 import json
+import operator
 import sys
 import textwrap
 from collections import defaultdict
@@ -13,6 +14,7 @@ from math import ceil
 from typing import (
     TYPE_CHECKING,
     Any,
+    Callable,
     Dict,
     Final,
     Iterable,
@@ -28,7 +30,6 @@ from typing import (
 from typing_extensions import TypeAlias
 
 import jsonschema
-import jsonschema.exceptions
 import jsonschema.validators
 import narwhals.stable.v1 as nw
 from packaging.version import Version
@@ -82,9 +83,31 @@ _DEFAULT_JSON_SCHEMA_DRAFT_URL: Final = "http://json-schema.org/draft-07/schema#
 # class-level _class_is_valid_at_instantiation attribute to False
 DEBUG_MODE: bool = True
 
+_JSONSCHEMA_VERSION = Version(importlib_version("jsonschema"))
+_USING_REFERENCING: Final[bool] = _JSONSCHEMA_VERSION >= Version("4.18")  # noqa: SIM300
+"""
+``jsonschema`` deprecated ``RefResolver`` in favor of ``referencing``.
 
-_USING_REFERENCING: Final[bool] = Version(importlib_version("jsonschema")) >= Version("4.18")  # fmt: off
-"""In version 4.18.0, the ``jsonschema`` package deprecated RefResolver in favor of the ``referencing`` library."""
+See https://github.com/python-jsonschema/jsonschema/releases/tag/v4.18.0a1
+"""
+
+if _JSONSCHEMA_VERSION >= Version("4.0.1"):  # noqa: SIM300
+    _json_path: Callable[[ValidationError], str] = operator.attrgetter("json_path")
+else:
+
+    def _json_path(err: ValidationError, /) -> str:
+        """
+        Vendored backport for ``jsonschema.ValidationError.json_path`` property.
+
+        See https://github.com/vega/altair/issues/3038.
+        """
+        path = "$"
+        for elem in err.absolute_path:
+            if isinstance(elem, int):
+                path += "[" + str(elem) + "]"
+            else:
+                path += "." + elem
+        return path
 
 
 def enable_debug_mode() -> None:
@@ -279,23 +302,7 @@ def _get_referencing_registry(
     )
 
 
-def _json_path(err: jsonschema.exceptions.ValidationError) -> str:
-    """
-    Drop in replacement for the .json_path property of the jsonschema ValidationError class.
-
-    This is not available as property for ValidationError with jsonschema<4.0.1.
-
-    More info, see https://github.com/vega/altair/issues/3038.
-    """
-    path = "$"
-    for elem in err.absolute_path:
-        if isinstance(elem, int):
-            path += "[" + str(elem) + "]"
-        else:
-            path += "." + elem
-    return path
-
-
+# NOTE: Review function (2)
 def _group_errors_by_json_path(
     errors: ValidationErrorList,
 ) -> GroupedValidationErrors:
@@ -308,8 +315,7 @@ def _group_errors_by_json_path(
     """
     errors_by_json_path = defaultdict(list)
     for err in errors:
-        err_key = getattr(err, "json_path", _json_path(err))
-        errors_by_json_path[err_key].append(err)
+        errors_by_json_path[_json_path(err)].append(err)
     return dict(errors_by_json_path)
 
 
@@ -563,7 +569,7 @@ class SchemaValidationError(jsonschema.ValidationError):
         super().__init__(**err._contents())
         self.obj = obj
         self._errors: GroupedValidationErrors = getattr(
-            err, "_all_errors", {getattr(err, "json_path", _json_path(err)): [err]}
+            err, "_all_errors", {_json_path(err): [err]}
         )
         # This is the message from err
         self._original_message = self.message
--- a/tools/schemapi/schemapi.py
+++ b/tools/schemapi/schemapi.py
@@ -165,6 +165,9 @@ def validate_jsonschema(
         return None
 
 
+# NOTE: Entry for creating a `list` of errors
+# Everything else is skipped if this returns an empty `list`
+# TODO: Refactor to peek at possible error w/ `next(validator.iter_errors(spec))`
 def _get_errors_from_spec(
     spec: dict[str, Any],
     schema: dict[str, Any],
--- a/tools/schemapi/schemapi.py
+++ b/tools/schemapi/schemapi.py
@@ -171,20 +171,25 @@ def _get_errors_from_spec(
     rootschema: dict[str, Any] | None = None,
 ) -> ValidationErrorList:
     """
-    Uses the relevant jsonschema validator to validate the passed in spec against the schema using the rootschema to resolve references.
+    Uses the relevant ``jsonschema`` validator to validate ``spec`` against ``schema`` using `` rootschema`` to resolve references.
 
-    The schema and rootschema themselves are not validated but instead considered as valid.
+    ``schema`` and ``rootschema`` are not validated but instead considered as valid.
+
+    We don't use ``jsonschema.validate`` as this would validate the ``schema`` itself.
+    Instead, we pass the ``schema`` directly to the validator class.
+
+    This is done for two reasons:
+
+    1. The schema comes from Vega-Lite and is not based on the user
+    input, therefore there is no need to validate it in the first place.
+    2. The "uri-reference" format checker fails for some of the
+    references as URIs in "$ref" are not encoded, e.g.:
+
+        '#/definitions/ValueDefWithCondition<MarkPropFieldOrDatumDef, (Gradient|string|null)>'
+
+    would be a valid $ref in a Vega-Lite schema but it is not a valid
+    URI reference due to the characters such as '<'.
     """
-    # We don't use jsonschema.validate as this would validate the schema itself.
-    # Instead, we pass the schema directly to the validator class. This is done for
-    # two reasons: The schema comes from Vega-Lite and is not based on the user
-    # input, therefore there is no need to validate it in the first place. Furthermore,
-    # the "uri-reference" format checker fails for some of the references as URIs in
-    # "$ref" are not encoded,
-    # e.g. '#/definitions/ValueDefWithCondition<MarkPropFieldOrDatumDef,
-    # (Gradient|string|null)>' would be a valid $ref in a Vega-Lite schema but
-    # it is not a valid URI reference due to the characters such as '<'.
-
     json_schema_draft_url = _get_json_schema_draft_url(rootschema or schema)
     validator_cls = jsonschema.validators.validator_for(
         {"$schema": json_schema_draft_url}
--- a/tools/schemapi/schemapi.py
+++ b/tools/schemapi/schemapi.py
@@ -41,6 +41,7 @@ from altair import vegalite
 if TYPE_CHECKING:
     from typing import ClassVar
 
+    from jsonschema import ValidationError
     from jsonschema.protocols import Validator, _JsonParameter
     from referencing import Registry
 
@@ -56,7 +57,7 @@ if TYPE_CHECKING:
     else:
         from typing_extensions import Never, Self
 
-ValidationErrorList: TypeAlias = List[jsonschema.exceptions.ValidationError]
+ValidationErrorList: TypeAlias = List[jsonschema.ValidationError]
 GroupedValidationErrors: TypeAlias = Dict[str, ValidationErrorList]
 
 # This URI is arbitrary and could be anything else. It just cannot be an empty
@@ -124,7 +125,7 @@ def validate_jsonschema(
     rootschema: dict[str, Any] | None = ...,
     *,
     raise_error: Literal[False],
-) -> jsonschema.exceptions.ValidationError | None: ...
+) -> ValidationError | None: ...
 
 
 def validate_jsonschema(
@@ -133,7 +134,7 @@ def validate_jsonschema(
     rootschema: dict[str, Any] | None = None,
     *,
     raise_error: bool = True,
-) -> jsonschema.exceptions.ValidationError | None:
+) -> ValidationError | None:
     """
     Validates the passed in spec against the schema in the context of the rootschema.
 
@@ -398,7 +399,7 @@ def _deduplicate_errors(
     return grouped_errors_deduplicated
 
 
-def _is_required_value_error(err: jsonschema.exceptions.ValidationError) -> bool:
+def _is_required_value_error(err: ValidationError) -> bool:
     return err.validator == "required" and err.validator_value == ["value"]
 
 
@@ -558,7 +559,7 @@ def _resolve_references(
 class SchemaValidationError(jsonschema.ValidationError):
     """A wrapper for jsonschema.ValidationError with friendlier traceback."""
 
-    def __init__(self, obj: SchemaBase, err: jsonschema.ValidationError) -> None:
+    def __init__(self, obj: SchemaBase, err: ValidationError) -> None:
         super().__init__(**err._contents())
         self.obj = obj
         self._errors: GroupedValidationErrors = getattr(
@@ -614,7 +615,7 @@ class SchemaValidationError(jsonschema.ValidationError):
 
     def _get_additional_properties_error_message(
         self,
-        error: jsonschema.exceptions.ValidationError,
+        error: ValidationError,
     ) -> str:
         """Output all existing parameters when an unknown parameter is specified."""
         altair_cls = self._get_altair_class_for_error(error)
@@ -633,9 +634,7 @@ Existing parameter names are:
 See the help for `{altair_cls.__name__}` to read the full description of these parameters"""
         return message
 
-    def _get_altair_class_for_error(
-        self, error: jsonschema.exceptions.ValidationError
-    ) -> type[SchemaBase]:
+    def _get_altair_class_for_error(self, error: ValidationError) -> type[SchemaBase]:
         """
         Try to get the lowest class possible in the chart hierarchy so it can be displayed in the error message.
 
--- a/tools/schemapi/schemapi.py
+++ b/tools/schemapi/schemapi.py
@@ -81,7 +81,9 @@ _DEFAULT_JSON_SCHEMA_DRAFT_URL: Final = "http://json-schema.org/draft-07/schema#
 # class-level _class_is_valid_at_instantiation attribute to False
 DEBUG_MODE: bool = True
 
-jsonschema_version_str = importlib_version("jsonschema")
+
+_USING_REFERENCING: Final[bool] = Version(importlib_version("jsonschema")) >= Version("4.18")  # fmt: off
+"""In version 4.18.0, the ``jsonschema`` package deprecated RefResolver in favor of the ``referencing`` library."""
 
 
 def enable_debug_mode() -> None:
@@ -191,7 +193,7 @@ def _get_errors_from_spec(
     if hasattr(validator_cls, "FORMAT_CHECKER"):
         validator_kwargs["format_checker"] = validator_cls.FORMAT_CHECKER
 
-    if _use_referencing_library():
+    if _USING_REFERENCING:
         schema = _prepare_references_in_schema(schema)
         validator_kwargs["registry"] = _get_referencing_registry(
             rootschema or schema, json_schema_draft_url
@@ -538,7 +540,7 @@ def _resolve_references(
     schema: dict[str, Any], rootschema: dict[str, Any] | None = None
 ) -> dict[str, Any]:
     """Resolve schema references until there is no $ref anymore in the top-level of the dictionary."""
-    if _use_referencing_library():
+    if _USING_REFERENCING:
         registry = _get_referencing_registry(rootschema or schema)
         # Using a different variable name to show that this is not the
         # jsonschema.RefResolver but instead a Resolver from the referencing
--- a/tools/schemapi/schemapi.py
+++ b/tools/schemapi/schemapi.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import contextlib
-import copy
 import inspect
 import json
 import sys
@@ -201,7 +200,7 @@ def _get_errors_from_spec(
         validator_kwargs["format_checker"] = validator_cls.FORMAT_CHECKER
 
     if _USING_REFERENCING:
-        schema = _prepare_references_in_schema(schema)
+        schema = _prepare_references(schema)
         validator_kwargs["registry"] = _get_referencing_registry(
             rootschema or schema, json_schema_draft_url
         )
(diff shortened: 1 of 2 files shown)


tools/schemapi/schemapi.py
```diff
@@ -584,7 +736,10 @@ def indent_second_line_onwards(message: str, indent: int = 4) -> str:
         error_messages: list[str] = []
         # Only show a maximum of 3 errors as else the final message returned by this
         # method could get very long.
-        for errors in list(self._errors.values())[:3]:
+        # ^^^^^^^^^^
+        # CORRECTION: Only show 3 **json_paths**
+
+        for errors in islice(_group_errors_by_json_path(self._errors).values(), 3):
```
dangotbanned, line 742: The call to `_group_errors_by_json_path` is purely to satisfy the additional error message format handling.

Ideally recollecting into a `dict[str, list[ValidationError]` would be avoided, especially since:
- Only the first 3 groups are needed
- The keys (json path) are discarded

tools/schemapi/schemapi.py
```diff
@@ -334,6 +384,109 @@ def _get_leaves_of_error_tree(
...
+
+def _lazy_additional_properties(
+    iterable: Iterable[ValidationError], /
+) -> Iterator[ValidationError]:
+    it = iter(iterable)
+    first = next(it)
+    if (
+        parent := cast("ValidationError", first.parent)
+    ) and parent.validator == "anyOf":
+        yield min(_rechain(first, it), key=lambda x: len(x.message))
```
dangotbanned, line 466: TODO: rewrite this lambda

tools/schemapi/schemapi.py
```diff
@@ -334,6 +384,109 @@ def _get_leaves_of_error_tree(
...
+    # Need to do an eager pass, as this skips intersections of non-overlapping enums
+    # yield reduce(_enum_inner, iterable)
+    yield from _deduplicate_enum_errors(list(iterable))
+
+
+def _enum_inner(prev: ValidationError, current: ValidationError, /) -> ValidationError:
+    """**Disabled**."""
+    longest = set(cast("list[str]", prev.validator_value))
+    contender = set(cast("list[str]", current.validator_value))
+    return current if contender.issuperset(longest) else prev
```
dangotbanned, line 487: # Note
Originally added to use like below.
However this was naive, due to this skipping disjoint enums.

```py
from functools import reduce

def _lazy_enum(iterable: Iterable[ValidationError], /) -> Iterator[ValidationError]:
    yield reduce(_enum_inner, iterable)
```

# Possible alternative
Rather than preserving each enum, we could just extract the intersection of all of them.

It would produce a different error message than the original, so for the below:

https://github.com/vega/altair/blob/591bf40fc0158ddd8d2fad2774d5230eb73ab734/tests/utils/test_schemapi.py#L779-L788

We could adapt the formatting from [`SchemaValidationError._format_params_as_table`](https://github.com/vega/altair/blob/591bf40fc0158ddd8d2fad2774d5230eb73ab734/tools/schemapi/schemapi.py#L660) to print something like this:

<details><summary>Large Screenshot</summary>
<p>


image

</p>
</details> 

Which would just read as:

> 'invalid_value' is an invalid value for `timeUnit`. Valid values are: 
> - One of [...]
> - Of type 'object'

This would also be easier to control the formatting than relying on the groups provided by the enums alone

tools/schemapi/schemapi.py
```diff
@@ -334,6 +384,109 @@ def _get_leaves_of_error_tree(
...
+
+    """
+    rev_sort = sorted(json_path_errors, key=_fn_path, reverse=True)
+    keeping: dict[str, Iterable[ValidationError]] = {}
+    for unique_path, grouped_errors in groupby(rev_sort, key=_fn_path):
+        if any(seen.startswith(unique_path) for seen in keeping):
+            continue
+        else:
+            keeping[unique_path] = [err for _, err in grouped_errors]
+    yield from reversed(keeping.values())
```
dangotbanned, line 431: TODO: Test rewriting this as:

```py
yield from islice(reversed(keeping.values()), 3)
```

Thinking it would skip some work that is discarded in `SchemaValidationError._get_message()`.

tools/schemapi/schemapi.py
```diff
@@ -334,6 +384,109 @@ def _get_leaves_of_error_tree(
...
+    first = next(it)
+    if (
+        parent := cast("ValidationError", first.parent)
+    ) and parent.validator == "anyOf":
+        yield min(_rechain(first, it), key=lambda x: len(x.message))
+    else:
+        yield first
+
+
+def _lazy_enum(iterable: Iterable[ValidationError], /) -> Iterator[ValidationError]:
```
dangotbanned, line 471: Relinking https://github.com/vega/altair/pull/3547#discussion_r1723012326 here, as that was inside an unreferenced function

tools/schemapi/schemapi.py
dangotbanned: Fixed in [`bbf6f14` (#3547)](https://github.com/vega/altair/pull/3547/commits/bbf6f142ac6a4df0dc6d7e3ca326b9a8d32b413e)

tools/schemapi/schemapi.py
dangotbanned: Fixed in [`c261cb4` (#3547)](https://github.com/vega/altair/pull/3547/commits/c261cb46510ce680d99930bfc9a9f1f4762d1d9b)

See the last *"rules"* list in https://docs.python.org/3/howto/functional.html#small-functions-and-the-lambda-expression

tools/schemapi/schemapi.py
dangotbanned: Fixed in [`ad06d08` (#3547)](https://github.com/vega/altair/pull/3547/commits/ad06d080d561009d6ce0fddd228364f6e82c5171)

dangotbanned:
Pulled out alternative error message idea from https://github.com/vega/altair/pull/3547#discussion_r1723012326 out for later discussion.

The issue this was attached to has been resolved, but I still think this is an area the `SchemaValidationError` message could be refined.

---

## Note
Originally added to use like below.
However this was naive, due to this skipping disjoint enums.

```py
from functools import reduce

def _lazy_enum(iterable: Iterable[ValidationError], /) -> Iterator[ValidationError]:
    yield reduce(_enum_inner, iterable)
```

## Possible alternative
Rather than preserving each enum, we could just extract the **union** of them.

It would produce a different error message than the original, so for the below:

https://github.com/vega/altair/blob/591bf40fc0158ddd8d2fad2774d5230eb73ab734/tests/utils/test_schemapi.py#L779-L788

We could adapt the formatting from [`SchemaValidationError._format_params_as_table`](https://github.com/vega/altair/blob/591bf40fc0158ddd8d2fad2774d5230eb73ab734/tools/schemapi/schemapi.py#L660) to print something like this:

<details><summary>Large Screenshot</summary>
<p>


image

</p>
</details> 

Which would just read as:

> 'invalid_value' is an invalid value for `timeUnit`. Valid values are: 
> - One of [...]
> - Of type 'object'

This would also be easier to control the formatting than relying on the groups provided by the enums alone

tests/utils/test_schemapi.py
```diff
@@ -871,10 +878,92 @@ def test_chart_validation_errors(chart_func, expected_error_message):
...
+    # Pre-`SchemaBase.from_dict` refactor (3.12.3)
+    108.16s call     tests/utils/test_schemapi.py::test_chart_validation_benchmark[to_dict-validate]
+    84.62s call     tests/utils/test_schemapi.py::test_chart_validation_benchmark[from_dict]
+    66.71s call     tests/utils/test_schemapi.py::test_chart_validation_benchmark[to_dict]
+
+    # Post-`SchemaBase.from_dict` refactor (3.12.3)
+    107.84s call     tests/utils/test_schemapi.py::test_chart_validation_benchmark[to_dict-validate]
+    50.43s call     tests/utils/test_schemapi.py::test_chart_validation_benchmark[from_dict]
+    67.07s call     tests/utils/test_schemapi.py::test_chart_validation_benchmark[to_dict]
+    ```
```
dangotbanned, line 920: If anyone stumbles upon this PR while in draft, this is the headline:

- `SchemaBase.from_dict` is seeing a **1.67x** speedup on this benchmark so far
- Across the whole test suite, `python<3.12` has seen a huge improvement from fewer forced exceptions [`933c045` (#3547)](https://github.com/vega/altair/pull/3547/commits/933c045f54a9987ca6a6c43f22e6a1e33fe842d9)

tools/schemapi/schemapi.py
```diff
@@ -875,8 +987,10 @@ class SchemaBase:
     the _rootschema class attribute) which is used for validation.
     """
 
+    # TODO: Implement `ClassVar` validation using https://peps.python.org/pep-0487/
+
     _schema: ClassVar[dict[str, Any] | Any] = None
-    _rootschema: ClassVar[dict[str, Any] | None] = None
+    _rootschema: ClassVar[dict[str, Any] | Any] = None
```
dangotbanned, line 993: TODO: 

Mockup/implement `ClassVar` validation using Implement `ClassVar` validation using https://peps.python.org/pep-0487/ AKA `__init_subclass__`

dangotbanned:
_Originally posted by @binste in https://github.com/vega/altair/issues/3545#issuecomment-2297061291_

> Thanks! I've also noticed that validation can take up a significant amount of time in contexts where milliseconds matter such as in web applications. 
> In an application at work, I had to disable validation of charts when it's deployed to production. 
> Maybe there are things in Python we could improve but I think part of it is also just how the `jsonschema` library works. Curious to see what you find out @dangotbanned! 
> There are other libraries such as https://github.com/horejsek/python-fastjsonschema and https://github.com/Stranger6667/jsonschema-rs but we'd need something which is also available in pyodide and from which we can continue to produce user-friendly error messages.

Thought I'd hold off on responding to this until I had a better understanding of the issue(s). 

# So far
Changes I've made have fallen into several categories:
- Reducing usage of `list|dict`, preferring iterators
- Generally deferring as much work as possible (eager -> lazy), to limit potential resource usage
- Making conditional logic **static**, where the result of the condition will not change between each evaluation (e.g. `jsonschema` version)
- Caching immutable or *effectively* immutable resources (e.g. `jsonschema|referencing` objects, `dict`(s) that are generated deterministically)
- Removing redundant checks for states that cannot be reached since at least `v5.0.0`
- Avoid (*avoidable*) exception handling, where there is an alternative (usually faster) way to achieve the same goal

## Non-performance changes
- Moving long comments to docstrings, to: allow collapsing long blocks; viewing on hover of symbol; markdown for easier reading
	- Below is just the first instance of this.
	- [docs: Move `_get_errors_from_spec` comment into docstring (#3547)](https://github.com/vega/altair/pull/3547/commits/f4a4e0e4b0be55ade5e1602cfb65010767739ddf)
- Renaming **private** symbols
	- [refactor: Renaming, docs to align with `jsonschema` (#3547)](https://github.com/vega/altair/pull/3547/commits/56be98401591a17a2a1c0857c1a47078dbdaa159)

# What I've learned
- Limiting `try-except-...` had a suprisingly big impact on older `python`
	- See for clear explanation (https://docs.astral.sh/ruff/rules/try-except-in-loop/)
- I've flagged and/or fixed related *kinds* of potential performance issues, external to `schemapi.py`
	- [perf(ruff): Add & ignore some performance rules (#3547)](https://github.com/vega/altair/pull/3547/commits/677017702c5548ffba2f0a0b6aee7bf2860f944e)
	- [perf(ruff): Fix some `"PERF"` violations (#3547)](https://github.com/vega/altair/pull/3547/commits/0ddf19eb23b18308cf7da6250f9b25a5c7311711)
	- [perf: Remove unreachable `ValidationError` except (#3547)](https://github.com/vega/altair/pull/3547/commits/cace782a11c2fee7e4efa06d4e5a6a626a474dee)
	- [docs(perf): Note some areas that may impact performance (#3547)](https://github.com/vega/altair/pull/3547/commits/ac9993907c78011881cff863dc73afa5e7d6b595)
- Possibly overhead that can be reduced by writing a [`SchemaBase.__init_subclass__`](https://github.com/vega/altair/pull/3547#discussion_r1731037778)
	- E.g. we check for `cls._schema` and `cls._rootschema` a huge number of times, which if this can be performed once must have some performance effect

tools/schemapi/schemapi.py
dangotbanned: - Partial Implementation [`06ca339` (#3547)](https://github.com/vega/altair/pull/3547/commits/06ca33929c5f2a5f54d4f15f872844a8c540f605)
- Tests/Demo [`56a43cb` (#3547)](https://github.com/vega/altair/pull/3547/commits/56a43cb244537dc51b28d866e4d6c6b9781ac98c)

dangotbanned:
Suggested edit:
```diff
diff --git a/tests/utils/test_schemapi.py b/tests/utils/test_schemapi.py
index cba7e2b2..0af849e4 100644
--- a/tests/utils/test_schemapi.py
+++ b/tests/utils/test_schemapi.py
@@ -199,7 +199,34 @@ def test_init_subclasses_hierarchy(dummy_rootschema) -> None:
         )
 
 
-class _TestSchema(SchemaBase):
+class _TestSchema(
+    SchemaBase,
+    schema={},
+    rootschema={
+        "$schema": _JSON_SCHEMA_DRAFT_URL,
+        "definitions": {
+            "StringMapping": {
+                "type": "object",
+                "additionalProperties": {"type": "string"},
+            },
+            "StringArray": {"type": "array", "items": {"type": "string"}},
+        },
+        "properties": {
+            "a": {"$ref": "#/definitions/StringMapping"},
+            "a2": {"type": "object", "additionalProperties": {"type": "number"}},
+            "b": {"$ref": "#/definitions/StringArray"},
+            "b2": {"type": "array", "items": {"type": "number"}},
+            "c": {"type": ["string", "number"]},
+            "d": {
+                "anyOf": [
+                    {"$ref": "#/definitions/StringMapping"},
+                    {"$ref": "#/definitions/StringArray"},
+                ]
+            },
+            "e": {"items": [{"type": "string"}, {"type": "string"}]},
+        },
+    },
+):
     @classmethod
     def _default_wrapper_classes(cls):
         return schemapi._subclasses(_TestSchema)
@@ -221,8 +248,9 @@ class _TestSchema(SchemaBase):
         return converter.from_dict(dct, cls)
 
 
-class MySchema(_TestSchema):
-    _schema = {
+class MySchema(
+    _TestSchema,
+    schema={
         "$schema": _JSON_SCHEMA_DRAFT_URL,
         "definitions": {
             "StringMapping": {
@@ -245,21 +273,43 @@ class MySchema(_TestSchema):
             },
             "e": {"items": [{"type": "string"}, {"type": "string"}]},
         },
-    }
+    },
+    rootschema={
+        "$schema": _JSON_SCHEMA_DRAFT_URL,
+        "definitions": {
+            "StringMapping": {
+                "type": "object",
+                "additionalProperties": {"type": "string"},
+            },
+            "StringArray": {"type": "array", "items": {"type": "string"}},
+        },
+        "properties": {
+            "a": {"$ref": "#/definitions/StringMapping"},
+            "a2": {"type": "object", "additionalProperties": {"type": "number"}},
+            "b": {"$ref": "#/definitions/StringArray"},
+            "b2": {"type": "array", "items": {"type": "number"}},
+            "c": {"type": ["string", "number"]},
+            "d": {
+                "anyOf": [
+                    {"$ref": "#/definitions/StringMapping"},
+                    {"$ref": "#/definitions/StringArray"},
+                ]
+            },
+            "e": {"items": [{"type": "string"}, {"type": "string"}]},
+        },
+    },
+): ...
 
 
-class StringMapping(_TestSchema):
-    _schema = {"$ref": "#/definitions/StringMapping"}
-    _rootschema = MySchema._schema
+class StringMapping(_TestSchema, schema={"$ref": "#/definitions/StringMapping"}): ...
 
 
-class StringArray(_TestSchema):
-    _schema = {"$ref": "#/definitions/StringArray"}
-    _rootschema = MySchema._schema
+class StringArray(_TestSchema, schema={"$ref": "#/definitions/StringArray"}): ...
 
 
-class Derived(_TestSchema):
-    _schema = {
+class Derived(
+    _TestSchema,
+    schema={
         "$schema": _JSON_SCHEMA_DRAFT_URL,
         "definitions": {
             "Foo": {"type": "object", "properties": {"d": {"type": "string"}}},
@@ -272,45 +322,68 @@ class Derived(_TestSchema):
             "b": {"type": "string"},
             "c": {"$ref": "#/definitions/Foo"},
         },
-    }
+    },
+    rootschema={
+        "$schema": _JSON_SCHEMA_DRAFT_URL,
+        "definitions": {
+            "Foo": {"type": "object", "properties": {"d": {"type": "string"}}},
+            "Bar": {"type": "string", "enum": ["A", "B"]},
+        },
+        "type": "object",
+        "additionalProperties": False,
+        "properties": {
+            "a": {"type": "integer"},
+            "b": {"type": "string"},
+            "c": {"$ref": "#/definitions/Foo"},
+        },
+    },
+): ...
 
 
-class Foo(_TestSchema):
-    _schema = {"$ref": "#/definitions/Foo"}
-    _rootschema = Derived._schema
+class Foo(
+    _TestSchema, schema={"$ref": "#/definitions/Foo"}, rootschema=Derived._rootschema
+): ...
 
 
-class Bar(_TestSchema):
-    _schema = {"$ref": "#/definitions/Bar"}
-    _rootschema = Derived._schema
+class Bar(
+    _TestSchema, schema={"$ref": "#/definitions/Bar"}, rootschema=Derived._rootschema
+): ...
 
 
-class SimpleUnion(_TestSchema):
-    _schema = {
+class SimpleUnion(
+    _TestSchema,
+    schema={
         "$schema": _JSON_SCHEMA_DRAFT_URL,
         "anyOf": [{"type": "integer"}, {"type": "string"}],
-    }
+    },
+): ...
 
 
-class DefinitionUnion(_TestSchema):
-    _schema = {"anyOf": [{"$ref": "#/definitions/Foo"}, {"$ref": "#/definitions/Bar"}]}
-    _rootschema = Derived._schema
+class DefinitionUnion(
+    _TestSchema,
+    schema={"anyOf": [{"$ref": "#/definitions/Foo"}, {"$ref": "#/definitions/Bar"}]},
+    rootschema=Derived._rootschema,
+): ...
 
 
-class SimpleArray(_TestSchema):
-    _schema = {
+class SimpleArray(
+    _TestSchema,
+    schema={
         "$schema": _JSON_SCHEMA_DRAFT_URL,
         "type": "array",
         "items": {"anyOf": [{"type": "integer"}, {"type": "string"}]},
-    }
+    },
+): ...
 
 
-class InvalidProperties(_TestSchema):
-    _schema = {
+class InvalidProperties(
+    _TestSchema,
+    schema={
         "$schema": _JSON_SCHEMA_DRAFT_URL,
         "type": "object",
         "properties": {"for": {}, "as": {}, "vega-lite": {}, "$schema": {}},
-    }
+    },
+): ...
 
 
 _validation_selection_schema = {
@@ -320,18 +393,30 @@ _validation_selection_schema = {
 }
 
 
-class Draft4Schema(_TestSchema):
-    _schema = {
+class Draft4Schema(
+    _TestSchema,
+    schema={
         **_validation_selection_schema,
         "$schema": "http://json-schema.org/draft-04/schema#",
-    }
+    },
+    rootschema={
+        **_validation_selection_schema,
+        "$schema": "http://json-schema.org/draft-04/schema#",
+    },
+): ...
 
 
-class Draft6Schema(_TestSchema):
-    _schema = {
+class Draft6Schema(
+    _TestSchema,
+    schema={
         **_validation_selection_schema,
         "$schema": "http://json-schema.org/draft-06/schema#",
-    }
+    },
+    rootschema={
+        **_validation_selection_schema,
+        "$schema": "http://json-schema.org/draft-06/schema#",
+    },
+): ...
 
 
 def test_construct_multifaceted_schema():
@@ -545,12 +630,10 @@ def test_to_from_pickle(dct):
 
 
 def test_class_with_no_schema():
-    class BadSchema(SchemaBase):
-        pass
+    with pytest.raises(TypeError, match="Cannot instantiate object"):
 
-    with pytest.raises(ValueError) as err:  # noqa: PT011
-        BadSchema(4)
-    assert str(err.value).startswith("Cannot instantiate object")
+        class BadSchema(SchemaBase):
+            pass
 
 
 def test_hash_schema():
diff --git a/tools/generate_schema_wrapper.py b/tools/generate_schema_wrapper.py
index f4b97090..b65edbf6 100644
--- a/tools/generate_schema_wrapper.py
+++ b/tools/generate_schema_wrapper.py
@@ -53,9 +53,8 @@ CHANNEL_MYPY_IGNORE_STATEMENTS: Final = """\
 """
 
 BASE_SCHEMA: Final = """
-class {basename}(SchemaBase):
-    _schema = load_schema()
-    _rootschema = load_schema()
+class {basename}(SchemaBase, schema=load_schema(), rootschema=load_schema()):
+
     @classmethod
     def _default_wrapper_classes(cls) -> Iterator[type[Any]]:
         return _subclasses({basename})
@@ -271,9 +270,8 @@ The parameters ``short``, ``long`` accept the same range of types::
 class SchemaGenerator(codegen.SchemaGenerator):
     schema_class_template = textwrap.dedent(
         '''
-    class {classname}({basename}):
+    class {classname}({basename}, schema={schema!r}):
         """{docstring}"""
-        _schema = {schema!r}
 
         {init_code}
     '''
@@ -317,9 +315,8 @@ class FieldSchemaGenerator(SchemaGenerator):
     schema_class_template = textwrap.dedent(
         '''
     @with_property_setters
-    class {classname}(FieldChannelMixin, core.{basename}):
+    class {classname}(FieldChannelMixin, core.{basename}, valid_at_init=False):
         """{docstring}"""
-        _class_is_valid_at_instantiation = False
         _encoding_name = "{encodingname}"
 
         {method_code}
@@ -333,9 +330,8 @@ class ValueSchemaGenerator(SchemaGenerator):
     schema_class_template = textwrap.dedent(
         '''
     @with_property_setters
-    class {classname}(ValueChannelMixin, core.{basename}):
+    class {classname}(ValueChannelMixin, core.{basename}, valid_at_init=False):
         """{docstring}"""
-        _class_is_valid_at_instantiation = False
         _encoding_name = "{encodingname}"
 
         {method_code}
@@ -349,9 +345,8 @@ class DatumSchemaGenerator(SchemaGenerator):
     schema_class_template = textwrap.dedent(
         '''
     @with_property_setters
-    class {classname}(DatumChannelMixin, core.{basename}):
+    class {classname}(DatumChannelMixin, core.{basename}, valid_at_init=False):
         """{docstring}"""
-        _class_is_valid_at_instantiation = False
         _encoding_name = "{encodingname}"
 
         {method_code}
diff --git a/tools/schemapi/schemapi.py b/tools/schemapi/schemapi.py
index 19326d81..c17bd0f8 100644
--- a/tools/schemapi/schemapi.py
+++ b/tools/schemapi/schemapi.py
@@ -1115,35 +1115,63 @@ class SchemaBase:
 
     # TODO: Implement `ClassVar` validation using https://peps.python.org/pep-0487/
 
-    _schema: ClassVar[dict[str, Any] | Any] = None
-    _rootschema: ClassVar[dict[str, Any] | Any] = None
+    _schema: ClassVar[dict[str, Any]]
+    _rootschema: ClassVar[dict[str, Any]]
     _class_is_valid_at_instantiation: ClassVar[bool] = True
 
     def __init__(self, *args: Any, **kwds: Any) -> None:
-        # Two valid options for initialization, which should be handled by
-        # derived classes:
-        # - a single arg with no kwds, for, e.g. {'type': 'string'}
-        # - zero args with zero or more kwds for {'type': 'object'}
-        if self._schema is None:
+        if (kwds and args) or len(args) > 1:
+            name = type(self).__name__
+            _args = ", ".join(f"{a!r}" for a in args)
+            _kwds = ", ".join(f"{k}={v!r}" for k, v in kwds.items())
             msg = (
-                f"Cannot instantiate object of type {self.__class__}: "
-                "_schema class attribute is not defined."
-                ""
+                f"Expected either:\n"
+                f" - a single arg with no kwds, for, e.g. {{'type': 'string'}}\n"
+                f" - zero args with zero or more kwds for {{'type': 'object'}}\n\n"
+                f"but got: {name}({_args}, {_kwds})"
             )
-            raise ValueError(msg)
-
-        if kwds:
-            assert len(args) == 0
-        else:
-            assert len(args) in {0, 1}
-
+            raise AssertionError(msg)
         # use object.__setattr__ because we override setattr below.
+        self._args: tuple[Any, ...]
+        self._kwds: dict[str, Any]
         object.__setattr__(self, "_args", args)
         object.__setattr__(self, "_kwds", kwds)
 
         if DEBUG_MODE and self._class_is_valid_at_instantiation:
             self.to_dict(validate=True)
 
+    def __init_subclass__(
+        cls,
+        *args: Any,
+        schema: dict[str, Any] | None = None,
+        rootschema: dict[str, Any] | None = None,
+        valid_at_init: bool | None = None,
+        **kwds: Any,
+    ) -> None:
+        super().__init_subclass__(*args, **kwds)
+        if schema is None:
+            if hasattr(cls, "_schema"):
+                schema = cls._schema
+            else:
+                msg = (
+                    f"Cannot instantiate object of type {cls}: "
+                    "_schema class attribute is not defined."
+                )
+                raise TypeError(msg)
+        if rootschema is None:
+            if hasattr(cls, "_rootschema"):
+                rootschema = cls._rootschema
+            elif "$ref" not in schema:
+                rootschema = schema
+            else:
+                msg = "`rootschema` must be provided if `schema` contains a `'$ref'` and does not inherit one."
+                raise TypeError(msg)
+        if valid_at_init is None:
+            valid_at_init = cls._class_is_valid_at_instantiation
+        cls._schema = schema
+        cls._rootschema = rootschema
+        cls._class_is_valid_at_instantiation = valid_at_init
+
     def copy(
         self, deep: bool | Iterable[Any] = True, ignore: list[str] | None = None
     ) -> Self:
@@ -1170,12 +1198,16 @@ class SchemaBase:
                 copy[attr] = _shallow_copy(copy._get(attr))
         return copy
 
-    def _get(self, attr, default=Undefined):
+    @overload
+    def _get(self, attr: str, default: Optional = ...) -> Any | UndefinedType: ...
+    @overload
+    def _get(self, attr: str, default: T) -> Any | T: ...
+    def _get(self, attr: str, default: Optional[T] = Undefined) -> Any | T:
         """Get an attribute, returning default if not present."""
-        attr = self._kwds.get(attr, Undefined)
-        if attr is Undefined:
-            attr = default
-        return attr
+        if (item := self._kwds.get(attr, Undefined)) is not Undefined:
+            return item
+        else:
+            return default
 
     def __getattr__(self, attr):
         # reminder: getattr is called after the normal lookups
@@ -1189,7 +1221,7 @@ class SchemaBase:
     def __setattr__(self, item, val) -> None:
         self._kwds[item] = val
 
-    def __getitem__(self, item):
+    def __getitem__(self, item) -> Any:
         return self._kwds[item]
 
     def __setitem__(self, item, val) -> None:
@@ -1427,24 +1459,12 @@ class SchemaBase:
         cls, instance: dict[str, Any], schema: dict[str, Any] | None = None
     ) -> None:
         """Validate the instance against the class schema in the context of the rootschema."""
-        validate_jsonschema(
-            instance, schema or cls._schema, cls._rootschema or cls._schema
-        )
+        validate_jsonschema(instance, schema or cls._schema, cls._rootschema)
 
     @classmethod
     def resolve_references(cls, schema: dict[str, Any] | None = None) -> dict[str, Any]:
         """Resolve references in the context of this object's schema or root schema."""
-        rootschema = cls._rootschema or cls._schema
-        if rootschema is None:
-            name = type(cls).__name__
-            msg = (
-                f"{name}.resolve_references() provided only `None` values for:\n"
-                f"{schema=}, {cls._schema=}, {cls._rootschema=}.\n\n"
-                f"This variant indicates the class definition {name!r} is invalid."
-            )
-            raise TypeError(msg)
-        else:
-            return _resolve_references(schema or cls._schema, rootschema=rootschema)
+        return _resolve_references(schema or cls._schema, cls._rootschema)
 
     @classmethod
     def validate_property(
@@ -1459,9 +1479,7 @@ class SchemaBase:
         np_opt = sys.modules.get("numpy")
         value = _todict(value, context={}, np_opt=np_opt, pd_opt=pd_opt)
         props = cls.resolve_references(schema or cls._schema).get("properties", {})
-        validate_jsonschema(
-            value, props.get(name, {}), rootschema=cls._rootschema or cls._schema
-        )
+        validate_jsonschema(value, props.get(name, {}), rootschema=cls._rootschema)
 
     def __dir__(self) -> list[str]:
         return sorted(chain(super().__dir__(), self._kwds))
@@ -1663,7 +1681,7 @@ class _FromDict:
             return dct
         elif tp is not None:
             current_schema = tp._schema
-            root_schema: dict[str, Any] = rootschema or tp._rootschema or current_schema
+            root_schema: dict[str, Any] = rootschema or tp._rootschema
             target_tp = tp
         elif schema is not None:
             current_schema = schema

```

tools/schemapi/schemapi.py
```diff
@@ -1279,10 +1528,93 @@ def _is_iterable(
...
+    - `Chart` seems to try to avoid calling this
+        - Using `TopLevelMixin.__subclasses__()` first if possible
+    - It is always called during `Chart.encode()`
+        - Chart.encode()
+        - altair.utils.core.infer_encoding_types
+        -  _ChannelCache.infer_encoding_types
+        - _ChannelCache._wrap_in_channel
+        - SchemaBase.from_dict (recursive, hot loop, validate =False, within a try/except)
+        - _FromDict(cls._default_wrapper_classes())
+        - schemapi._subclasses(schema.core.VegaLiteSchema)
```
dangotbanned, line 1607: TODO: Tidy this up

tools/schemapi/schemapi.py
```diff
@@ -1120,17 +1366,16 @@ def from_dict(
         """
         if validate:
             cls.validate(dct)
-        converter = _FromDict(cls._default_wrapper_classes())
+        converter: type[_FromDict] | _FromDict = (
+            _FromDict
+            if _FromDict.hash_tps
+            else _FromDict(cls._default_wrapper_classes())
+        )
         return converter.from_dict(dct, cls)
```
dangotbanned, line 1374: NOTE: the breadth-first search now occurs only once.

`_FromDict` is purely `ClassVar`/`@classmethod`(s)

tools/schemapi/schemapi.py
```diff
@@ -1393,45 +1699,50 @@ def from_dict(
         """Construct an object from a dict representation."""
         target_tp: Any
         current_schema: dict[str, Any]
+        hash_schema: int
         if isinstance(dct, SchemaBase):
             return dct
         elif tp is not None:
             current_schema = tp._schema
+            hash_schema = _hash_schema(current_schema)
             root_schema: dict[str, Any] = rootschema or tp._rootschema or current_schema
```
dangotbanned, line 1708: NOTE: the `current_schema` branch *only* is triggered for mock schema tests.

These all fail when `.. or current_schema` is removed:

`test_schemapi.py::[test_construct_multifaceted_schema, test_copy_method, test_round_trip, test_copy_module, test_from_dict, test_to_from_json, test_to_from_pickle]`

tools/schemapi/schemapi.py
```diff
@@ -1393,45 +1699,50 @@ def from_dict(
...
-            # If there are multiple matches, we use the first one in the dict.
-            # Our class dict is constructed breadth-first from top to bottom,
-            # so the first class that matches is the most general match.
             current_schema = schema
+            hash_schema = _hash_schema(current_schema)
             root_schema = rootschema or current_schema
-            matches = self.class_dict[self.hash_schema(current_schema)]
-            target_tp = matches[0] if matches else default_class
+            matches = cls.hash_tps[hash_schema]
+            target_tp = next(iter(matches), default_class)
```
dangotbanned, line 1715: FIXME: This is the slow branch.

- Improving the performance of the `tp` one has no effect
  - I tried storing the hash value as a `ClassVar` in `SchemaBase`
  - Conceptually made sense, but this branch is triggered much more often
- Every recursive `from_dict` call that isn't solved hits this

tools/schemapi/schemapi.py
dangotbanned: Resolved in [`fba3c46` (#3547)](https://github.com/vega/altair/pull/3547/commits/fba3c46c0aa69a99d731c6a39a90ce396eae8863)

altair/vegalite/v5/schema/core.py
```diff
@@ -487,6 +487,7 @@ def load_schema() -> dict:
 
 
 class VegaLiteSchema(SchemaBase):
+    _schema = load_schema()
     _rootschema = load_schema()
```
dangotbanned, line 491: Linking this back to:
- [refactor: Ensure every `VegaLiteSchema` has a `._schema`](https://github.com/vega/altair/pull/3547/commits/bd31d7c574820ec3ee3bf72ec874c96545f415ba)

Has the same result for `Root`, since it inherits `_schema`.

The main benefit is a stronger guarantee that every derived class is non-None for both attributes

dangotbanned:
@MarcoGorelli do you have any ideas on what is causing this error for `ibis` on `python==3.8`?

https://github.com/vega/altair/actions/runs/10681007191/job/29608118357?pr=3547

I did a merge update from the GitHub UI, but haven't been able to track down the cause yet

tools/schemapi/schemapi.py
```diff
@@ -1,89 +1,126 @@
...
+    "SchemaBase",  # altair.vegalite.v5.schema.core
+    "Undefined",  # altair.utils
+    "UndefinedType",  # altair.vegalite.v5.schema.core -> (side-effect relied on to propagate to alt.__init__)
+    "_is_valid",  # altair.vegalite.v5.api
+    "_resolve_references",  # tools.schemapi.utils -> tools.generate_schema_wrapper
+    "_subclasses",  # altair.vegalite.v5.schema.core
+    "is_undefined",  # altair.typing
+    "validate_jsonschema",  # altair.utils.display
+    "with_property_setters",  # altair.vegalite.v5.schema.channels
+]
```
dangotbanned, line 86: Keeping track of the refs here, since you can't follow the references from the `tools` copy

tools/schemapi/schemapi.py
dangotbanned: Since [refactor: Remove `_SchemaBasePEP487` & test suite](https://github.com/vega/altair/pull/3547/commits/9a48448e334c53d2c431bb9a79eb397bc7c0e7d7) I've extracted all of the related code and moved to https://github.com/dangotbanned/altair/tree/schema-base-pep487

Not planning to pursue that any time soon, but thought it would be easier to explain all the outdated comments.

binste:
Quite the PR!! Will take me a while to dig through it but thanks for going down all these rabbit holes, hope you came out unharmed! ;) I like jsonschema validation but it also can be a wild place, especially with the mapping of the schema to Altair objects...

Do I understand this correct that the main motivation for this PR is to improve the performance and that the main impact on performance is that `from_dict` is considerably faster based on your benchmark (which is super cool!), where as there is no change in `to_dict` and the validation?

dangotbanned:
> Quite the PR!! Will take me a while to dig through it but thanks for going down all these rabbit holes, hope you came out unharmed! ;) I like jsonschema validation but it also can be a wild place, especially with the mapping of the schema to Altair objects...

Thanks and yeah no rush @binste - I know there is a lot to go through - even after all the stuff I removed over the past few days.

> Do I understand this correct that the main motivation for this PR is to improve the performance and that the main impact on performance is that `from_dict` is considerably faster based on your benchmark (which is super cool!), where as there is no change in `to_dict` and the validation?

So it is tricky. 

I found it quite difficult to benchmark this, since there are so many interconnected parts.
I can confidently say `from_dict` is faster, but of the three in the [benchmark](https://github.com/vega/altair/pull/3547#issue-2474332626) that is the only one which benefitted. 

We also have much better performance on `3.8-3.11` with changes like [`933c045` (#3547)](https://github.com/vega/altair/pull/3547/commits/933c045f54a9987ca6a6c43f22e6a1e33fe842d9).
All versions are pretty much equal on the **second run** of `hatch test --all`.

---

I would like to - in a future PR - develop some more isolated benchmarking tools. 
Ideally, these would be based on specs that have proved to be slow in **real-world use**, so that we can more easily get to the root cause

binste:
I very much appreciate the time and effort you put into this and I believe some/many/maybe all changes here could be great improvements to Altair. However, as it's such a large and complex PR, I'm struggling to review this PR to an extent where I'd feel comfortable approving it.

Although improvements to the performance of `from_dict` are great, it's a feature that is not used by many users. I'm worried that the time someone would need to invest to review this PR could be better used elsewhere as well as the added complexity to the codebase which is a lot (it's hard to find developers familiar enough with jsonschema which want to work on Altair, the more complicated we make this part, the harder this gets).

Would you be willing to split this PR into multiple ones? They could be put up for review and merged one after the other with this PR (in draft mode) being the resource to cherry-pick commits from. I think this would help with reviewing as well as with judging the tradeoff between added complexity of some of the changes vs. improved performance.

* Python 3.8-3.11 performance improvements: You mentioned that this mainly comes from https://github.com/vega/altair/pull/3547/commits/933c045f54a9987ca6a6c43f22e6a1e33fe842d9. This could be a separate PR, easy to merge, and already a big win for users.
* `from_dict` performance improvements: Do you have a feeling for which changes lead to this improvement? Is it everything in the current PR or can it be boiled down to fewer more isolated changes?

I hope this is ok for you and that you understand my position, i.e. that it does not feel random. We can also do more pre-alignment on rough drafts before starting to work on detailed PRs. Also something we can pick up in our upcoming call.

dangotbanned:
https://github.com/vega/altair/pull/3547#issuecomment-2366831122
> I hope this is ok for you and that you understand my position, i.e. that it does not feel random.

@binste
Honestly I would say the main benefit of this PR was for me to understand this part of `altair` a little better. 
That is, the process of working on it I see as most valuable - so please don't worry I totally understand that it isn't an easy review.

> Would you be willing to split this PR into multiple ones? 

Yeah that sounds like a good idea.
I will need to have a think about how best to do that, but at the very least I could get to the same outcome with a cleaner history and less experimentation.

I'll loop back to this at some point and see what smaller PRs I can make out of it. 

Appreciate your feedback on this, regardless of the outcome

Issue #3549, opened 2024-08-21
regression: `'csv'` data transformer no longer works for Polars


MarcoGorelli:
### What happened?

I ran
```python
import polars as pl
import altair as alt
import pandas as pd

alt.data_transformers.enable('csv')
df = pl.DataFrame({"a": [1, 2, 3], "b": [4,5,6]})
alt.Chart(df).mark_line().encode(x="a", y="b").to_json()
```

On Altair 5.3.0, it produces a file with
```
"a","b"
1,4
2,5
3,6
```

On Altair 5.4.0:
```
Traceback (most recent call last):
  File "/home/marcogorelli/scratch/t.py", line 7, in <module>
    alt.Chart(df).mark_line().encode(x="a", y="b").to_json()
  File "/home/marcogorelli/scratch/.venv/lib/python3.11/site-packages/altair/vegalite/v5/api.py", line 1893, in to_json
    spec = self.to_dict(
           ^^^^^^^^^^^^^
  File "/home/marcogorelli/scratch/.venv/lib/python3.11/site-packages/altair/vegalite/v5/api.py", line 3772, in to_dict
    return super().to_dict(**kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marcogorelli/scratch/.venv/lib/python3.11/site-packages/altair/vegalite/v5/api.py", line 1798, in to_dict
    copy.data = _prepare_data(data, context)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marcogorelli/scratch/.venv/lib/python3.11/site-packages/altair/vegalite/v5/api.py", line 275, in _prepare_data
    data = func(data)
           ^^^^^^^^^^
  File "/home/marcogorelli/scratch/.venv/lib/python3.11/site-packages/altair/utils/data.py", line 288, in to_csv
    data_str = _data_to_csv_string(data)
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marcogorelli/scratch/.venv/lib/python3.11/site-packages/altair/utils/data.py", line 412, in _data_to_csv_string
    raise NotImplementedError(msg)
NotImplementedError: to_csv only works with data expressed as a DataFrame or as a dict
```

### What would you like to happen instead?

It should still write the csv file

This is related to https://github.com/vega/altair/discussions/3548#discussioncomment-10404821

### Which version of Altair are you using?

5.4.0

PR #3550, opened 2024-08-21
fix: Pass native dataframe to data transformers

MarcoGorelli:
closes https://github.com/vega/altair/issues/3549

Also addresses the issue brought up in https://github.com/marimo-team/marimo/pull/2006

### DIFF
--- a/altair/utils/_vegafusion_data.py
+++ b/altair/utils/_vegafusion_data.py
@@ -13,8 +13,6 @@ from typing import (
 )
 from weakref import WeakValueDictionary
 
-import narwhals.stable.v1 as nw
-
 from altair.utils._importers import import_vegafusion
 from altair.utils.core import DataFrameLike
 from altair.utils.data import (
@@ -71,10 +69,6 @@ def vegafusion_data_transformer(
     data: DataType | None = None, max_rows: int = 100000
 ) -> Callable[..., Any] | _VegaFusionReturnType:
     """VegaFusion Data Transformer."""
-    # Vegafusion does not support Narwhals, so if `data` is a Narwhals
-    # object, we make sure to extract the native object and let Vegafusion handle it.
-    # `strict=False` passes `data` through as-is if it is not a Narwhals object.
-    data = nw.to_native(data, strict=False)
     if data is None:
         return vegafusion_data_transformer
     elif isinstance(data, DataFrameLike) and not isinstance(data, SupportsGeoInterface):
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -353,29 +353,29 @@ def _compute_data_hash(data_str: str) -> str:
 def _data_to_json_string(data: DataType) -> str:
     """Return a JSON string representation of the input data."""
     check_data_type(data)
-    # `strict=False` passes `data` through as-is if it is not a Narwhals object.
-    data_native = nw.to_native(data, strict=False)
-    if isinstance(data_native, SupportsGeoInterface):
-        if _is_pandas_dataframe(data_native):
-            data_native = sanitize_pandas_dataframe(data_native)
-        data_native = sanitize_geo_interface(data_native.__geo_interface__)
-        return json.dumps(data_native)
-    elif _is_pandas_dataframe(data_native):
-        data = sanitize_pandas_dataframe(data_native)
-        return data_native.to_json(orient="records", double_precision=15)
-    elif isinstance(data_native, dict):
-        if "values" not in data_native:
+    if isinstance(data, SupportsGeoInterface):
+        if _is_pandas_dataframe(data):
+            data = sanitize_pandas_dataframe(data)
+        data = sanitize_geo_interface(data.__geo_interface__)  # type: ignore[arg-type]
+        return json.dumps(data)
+    elif _is_pandas_dataframe(data):
+        data = sanitize_pandas_dataframe(data)
+        return data.to_json(orient="records", double_precision=15)
+    elif isinstance(data, dict):
+        if "values" not in data:
             msg = "values expected in data dict, but not present."
             raise KeyError(msg)
-        return json.dumps(data_native["values"], sort_keys=True)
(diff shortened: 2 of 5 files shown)


altair/utils/_vegafusion_data.py
```diff
@@ -71,10 +69,6 @@ def vegafusion_data_transformer(
     data: DataType | None = None, max_rows: int = 100000
 ) -> Callable[..., Any] | _VegaFusionReturnType:
     """VegaFusion Data Transformer."""
-    # Vegafusion does not support Narwhals, so if `data` is a Narwhals
-    # object, we make sure to extract the native object and let Vegafusion handle it.
-    # `strict=False` passes `data` through as-is if it is not a Narwhals object.
-    data = nw.to_native(data, strict=False)
```
MarcoGorelli, line 77: if native objects are being passed to data transformers now, then this isn't necessary any more

altair/utils/data.py
```diff
@@ -398,18 +398,12 @@ def _data_to_csv_string(data: dict | pd.DataFrame | DataFrameLike) -> str:
...
-        return csv_buffer.getvalue().to_pybytes().decode()
-    else:
-        msg = "to_csv only works with data expressed as " "a DataFrame or as a dict"
-        raise NotImplementedError(msg)
+    try:
+        data_nw = nw.from_native(data, eager_only=True)
+    except TypeError as exc:
+        msg = "to_csv only works with data expressed as a DataFrame or as a dict"
+        raise NotImplementedError(msg) from exc
+    return data_nw.write_csv()
```
MarcoGorelli, line 406: in order to be able to use `write_csv` in Narwhals, `v1.5.2+` is required - is that OK?

I'm not aware of anyone putting upper bounds on Narwhals (in theory that shouldn't ever be necessary) so this shouldn't cause issues

altair/utils/data.py
```diff
@@ -353,29 +353,29 @@ def _compute_data_hash(data_str: str) -> str:
...
+    if isinstance(data, SupportsGeoInterface):
+        if _is_pandas_dataframe(data):
+            data = sanitize_pandas_dataframe(data)
+        data = sanitize_geo_interface(data.__geo_interface__)  # type: ignore[arg-type]
+        return json.dumps(data)
+    elif _is_pandas_dataframe(data):
+        data = sanitize_pandas_dataframe(data)
+        return data.to_json(orient="records", double_precision=15)
+    elif isinstance(data, dict):
+        if "values" not in data:
```
MarcoGorelli, line 365: the diff looks deceptively big - it's mostly just renaming `data_native` to `data`, because now the transformer is starting with a native object (not a Narwhals one)

altair/utils/data.py
dangotbanned: > in order to be able to use `write_csv` in Narwhals, `v1.5.2+` is required - is that OK?

Yeah seems fine to me.
Upgrading (my env) silenced this warning from `pylance` for me: 

```
> Cannot access attribute "write_csv" for class "narwhals.stable.v1.DataFrame[NativeFrame | narwhals.dataframe.DataFrame[Any] | narwhals.typing.DataFrameLike | altair.utils.core.DataFrameLike]"
> > Attribute "write_csv" is unknown

```

altair/utils/data.py
```diff
@@ -350,14 +345,27 @@ def _compute_data_hash(data_str: str) -> str:
...
+    Notes
+    -----
+    Split out to resolve typing issues related to:
+    - Intersection types
+    - ``typing.TypeGuard``
+    - ``pd.DataFrame.__getattr__``
+    """
+    if _is_pandas_dataframe(data):
+        data = sanitize_pandas_dataframe(data)
+    return sanitize_geo_interface(data.__geo_interface__)
```
dangotbanned, line 361: @MarcoGorelli I've somewhat bailed on the typing for this. 
IMO there's too many interacting things that it all just becomes noise.

Hopefully with this split into a new function, no one needs to worry about that again 😅

MarcoGorelli: nice one, thanks @dangotbanned !

altair/utils/data.py
```diff
@@ -350,32 +345,45 @@ def _compute_data_hash(data_str: str) -> str:
...
-        return json.dumps(data.rows(named=True))
-    else:
-        msg = "to_json only works with data expressed as " "a DataFrame or as a dict"
-        raise NotImplementedError(msg)
+        return json.dumps(data["values"], sort_keys=True)
+    try:
+        data_nw = nw.from_native(data, eager_only=True)
+    except TypeError as exc:
+        msg = "to_json only works with data expressed as a DataFrame or as a dict"
+        raise NotImplementedError(msg) from exc
```
dangotbanned, line 381: I'm having a hard time writing a test that would trigger this. 

What does the traceback of this exception look like?

altair/utils/data.py
```diff
@@ -398,18 +406,12 @@ def _data_to_csv_string(data: dict | pd.DataFrame | DataFrameLike) -> str:
...
-        pa_csv.write_csv(pa_table, csv_buffer)
-        return csv_buffer.getvalue().to_pybytes().decode()
-    else:
-        msg = "to_csv only works with data expressed as " "a DataFrame or as a dict"
-        raise NotImplementedError(msg)
+    try:
+        data_nw = nw.from_native(data, eager_only=True)
+    except TypeError as exc:
+        msg = "to_csv only works with data expressed as a DataFrame or as a dict"
+        raise NotImplementedError(msg) from exc
```
dangotbanned, line 413: Same as https://github.com/vega/altair/pull/3550#discussion_r1725383011 but an example of either would work

dangotbanned (approved):
The comments I've pulled out are minor things, overall looks good to me. 
Thanks @MarcoGorelli for putting this together so quickly

altair/utils/data.py
MarcoGorelli: I think this was just added defensively

PR #3543, opened 2024-08-18
refactor: Simplify `SchemaBase.copy`

dangotbanned:
# Related
- #3531 

# Description
This PR addresses one [complex-structure](https://docs.astral.sh/ruff/rules/complex-structure/) violation, adds some typing, and gives a little performance boost in `SchemaBase.copy`.

See [each commit message](https://github.com/vega/altair/pull/3543/commits) for more detail on specific changes.

### Last commit message (auto-added)
- Initialize an empty set **once** at origin, rather than creating a new list per iteration
  - Renamed to `by_ref` in the new private function, to better describe the operation.
  - No change to public API.
- Define a partial `copy` to reduce repetition
- Use a genexpr for `args`, to avoid unpacking twice

PR #3553, opened 2024-08-21
fix: Resolve `Then` copy `TypeError`

dangotbanned:
Fixes 
- https://github.com/vega/altair/pull/3394#issuecomment-2302995453
- https://github.com/vega/altair/pull/3394#issuecomment-2303028521
- https://github.com/vega/altair/actions/runs/10497592890/job/29080773958?pr=3394

---

I originally thought https://github.com/vega/altair/pull/3543 introduced this regression.

Haven't tested on earlier branches, but @mattijn can you check if you can repro on `v5.4.0`?
I think it would be present on there as well

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -854,6 +854,9 @@ def _deep_copy(obj: Any, by_ref: set[str]) -> Any: ...
 def _deep_copy(obj: _CopyImpl | Any, by_ref: set[str]) -> _CopyImpl | Any:
     copy = partial(_deep_copy, by_ref=by_ref)
     if isinstance(obj, SchemaBase):
+        if copier := getattr(obj, "__deepcopy__", None):
+            with debug_mode(False):
+                return copier(obj)
         args = (copy(arg) for arg in obj._args)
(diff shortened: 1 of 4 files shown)


joelostblom (approved):
Thanks for fixing this @dangotbanned. I tried it out and as you mentioned, the error is raised on 5.4.0 as well. In fact, it is present all the way back to the commit that merged #3427, so it seems to have been there since the new `alt.when` functionality was added.

The fix here looks good to me :+1:

dangotbanned:
> Thanks for fixing this @dangotbanned. I tried it out and as you mentioned, the error is raised on 5.4.0 as well. In fact, it is present all the way back to the commit that merged #3427, so it seems to have been there since the new `alt.when` functionality was added.
> 
> The fix here looks good to me 👍

Thanks for the review @joelostblom 

For anyone searching in the future, what I added is a *very basic* implementation of `copy.deepcopy`.
Which has a `memo` parameter, and that is why `Then.__deepcopy__` does as well.

https://github.com/python/cpython/blob/74ff496dce9f7db1dd1e2184d4316cc21d78ee52/Lib/copy.py#L119-L179

PR #3555, opened 2024-08-23
test: Adds `test-(slow|fast)` options

dangotbanned:
While working on #3547, I've been running the test suite a bunch to identify bottlenecks.

## This PR provides
- tools to identify, skip or run only slow tests
- some new, previously repeated mark decorators
- misc refactoring of tests using the above
- docs explaining intended usage

## Impact?
- Locally, this covers most tests that take 2-10 secs.
- In total, **354**/1671 tests are skipped
- reduced runtime 40  -> 20 secs

## Important
The **default** behaviour is **unchanged**:

```bash
# Runs all tests
>>> hatch test 

# Runs only slow tests
>>> hatch run test-slow

# Runs all tests (that are not marked @slow)
>>> hatch run test-fast
```

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -20,72 +20,21 @@ Cache the calls to `compile` in `altair.utils.execeval`
 from __future__ import annotations
 
 import io
-import pkgutil
-import re
-import sys
-from typing import Any, Iterable, Iterator
-
-import pytest
+from typing import Any
 
 import altair as alt
 from altair.utils.execeval import eval_block
-from tests import examples_arguments_syntax, examples_methods_syntax
-
-try:
-    import vl_convert as vlc  # noqa: F401, RUF100
-except ImportError:
-    vlc = None
-
-
-VL_CONVERT_AVAILABLE = "vl_convert" in sys.modules
-
-
-def iter_examples_filenames(syntax_module) -> Iterator[str]:
-    for _importer, modname, ispkg in pkgutil.iter_modules(syntax_module.__path__):
-        if not (
-            ispkg
-            or modname.startswith("_")
-            # Temporarily skip this test until https://github.com/vega/altair/issues/3418
-            # is fixed
-            or modname == "interval_selection_map_quakes"
-        ):
-            yield f"{modname}.py"
-
-
-def _distributed_examples() -> Iterator[tuple[Any, str]]:
-    # `pytest.mark.parametrize` over 2 modules produces 2x workers
-    # - This raises the total jobs from 400 -> 1200
-    # - Preventing the three tests from blocking everything else
-    RE_NAME: re.Pattern[str] = re.compile(r"^tests\.(.*)")
-
-    for module in [examples_arguments_syntax, examples_methods_syntax]:
-        for filename in iter_examples_filenames(module):
-            name = module.__name__
-            source = pkgutil.get_data(name, filename)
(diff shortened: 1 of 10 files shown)


tests/test_examples.py
```diff
@@ -131,19 +78,17 @@ def test_from_and_to_json_roundtrip(source, filename) -> None:
...
-@pytest.mark.skipif(
-    not VL_CONVERT_AVAILABLE,
-    reason="vl_convert not importable; cannot run mimebundle tests",
-)
-def test_render_examples_to_png(source, filename) -> None:
+@slow
+@ignore_DataFrameGroupBy
+@distributed_examples
+@skip_requires_vl_convert
+def test_render_examples_to_png(source: Any, filename: str) -> None:
```
dangotbanned, line 85: Just highlighting somewhere these new markers improve readability.

The difference quite significant in an editor, as you can see the docstrings on hover - but no clutter at the test site

tests/test_examples.py
```diff
@@ -131,19 +78,17 @@ def test_from_and_to_json_roundtrip(source, filename) -> None:
...
     chart = eval_block(source)
     if chart is None:
         msg = f"Example file {filename} should define chart in its final statement."
         raise ValueError(msg)
     out = io.BytesIO()
     chart.save(out, format="png", engine="vl-convert")
-    assert out.getvalue().startswith(b"\x89PNG")
+    buf = out.getbuffer()
+    prefix = buf[:4].tobytes()
+    assert prefix == b"\x89PNG"
```
dangotbanned, line 94: Was an attempt to speed up these tests that were identified as slow, but I don't think it made an observable improvement.

The thinking was that it avoids copying the full contents of the buffer, since all we need are the first four bytes.

See [BytesIO.getbuffer](https://docs.python.org/3/library/io.html#io.BytesIO.getbuffer)

tests/utils/test_utils.py
```diff
@@ -36,13 +31,8 @@ def _check(arr, typ):
     _check(nulled, "quantitative")
     _check(["a", "b", "c"], "nominal")
 
-    if hasattr(pytest, "warns"):  # added in pytest 2.8
```
dangotbanned, line 39: [pytest 2.8](https://pypi.org/project/pytest/2.8.0/) was released Sep 18, 2015

dangotbanned:
- https://github.com/vega/altair/pull/3555#commits-pushed-a4f6c20

These were some drive-by fixes of typing issues, fixing them here in case this stays in review for a while
They would show on `mypy` as well, but are hidden due to https://microsoft.github.io/pyright/#/mypy-comparison?id=type-checking-unannotated-code

tests/utils/test_utils.py
binste: 😃 Wow, some parts of this code base are old..

binste (approved):
Really neat! :) I learned a lot while reading this. Also great to have more convenient decorators for skipping tests re pyarrow and vlc.

dangotbanned:
> Really neat! :) I learned a lot while reading this. Also great to have more convenient decorators for skipping tests re pyarrow and vlc.

Thanks @binste!

I've found that reading the `pytest` docs has a lot of info - but is overwhelming at times. 
The docs I've added here are as much for my own benefit; as they are to simplify the test writing of others

PR #3559, opened 2024-08-27
docs: Change remote nick to `origin` and capitalize version commit

joelostblom:
Since this will always be read by a contributor who has push rights to main, the assumption would be that they are working in the upstreams repo already, so shouldn't we reference `origin` instead of `upstream`?

I capitalized the commits to be consistent with how we write PR titles.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -6,7 +6,7 @@
 
 2. Make certain your branch is in sync with head:
  
-       git pull upstream main
+       git pull origin main
 
 3. Do a clean doc build:
 
@@ -25,13 +25,13 @@
 5. Commit change and push to main:
 
        git add . -u
-       git commit -m "chore: bump version to 5.0.0"
-       git push upstream main
+       git commit -m "chore: Bump version to 5.0.0"
+       git push origin main
 
 6. Tag the release:
 
        git tag -a v5.0.0 -m "version 5.0.0 release"
-       git push upstream v5.0.0
+       git push origin v5.0.0
 
 7. Build source & wheel distributions:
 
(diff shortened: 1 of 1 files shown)


dangotbanned:
This isn't blocking btw, but thought I had some insight as my current situation is an example where your assumption doesn't hold. 

I have several PRs that were opened before I became a maintainer, which are still in my fork https://github.com/dangotbanned/altair/branches/yours

I think I've only done one/two since in `upstream`. 
IIRC, I had some issues with the change in workflow but eventually would want to transition to working in `upstream` more often

joelostblom:
Good point. If someone already have a fork once they become a maintainer (which is probably the case for everyone), they might keep working there and then `upstream` is more appropriate than `origin`. I would expect most contributors transitioning to working on branches in the repo directly, but I understand that doesn't have to be true and there could be a preference to keep working in the fork. Do you prefer that I revert it back to `upstream`?

joelostblom:
By default after cloning a repo (whether it is a fork or not), I believe there is only one remote and it is named `origin` and pointing to the location of the cloned repo (regardless of whether it is a fork or not). You would have to explicitly add an additional remote and name it `upstream` that points to the main altair repo for it to work from both the fork and main repo (but this would be unnecessary if you cloned the main repo directly since then `origin` would already point to this remote).

dangotbanned (approved):
Seems a reasonable change given https://github.com/vega/altair/pull/3559#issuecomment-2313015775

Thanks @joelostblom

binste:
I don't have a preference as it's easy to exchange it :) Btw, I'm still working on my fork, just kind of like the additional layer of separation it gives so I don't accidentally push anything to main as we have not protected it against direct pushs.

joelostblom:
> so I don't accidentally push anything to main as we have not protected it against direct pushs.

That raises a good point, maybe we should? We can still merge our own PRs I believe, so the version bumps that we do when releasing could still be done by one person.

PR #3564, opened 2024-09-01
refactor: Remove `channels` parameter in `infer_encoding_types`

dangotbanned:
Was kept, but only needed for tests since #3444.

This PR factors out the test dependency, by using `pytest.MonkeyPatch` - to *temporarily* replace the global cache.

As `alt.utils.core.infer_encoding_types` is not public API - this is a safe remove, no need for deprecation.

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -33,7 +33,6 @@ else:
 
 if TYPE_CHECKING:
     import typing as t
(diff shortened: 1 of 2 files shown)


binste (approved):
LGTM! Agree that monkey patching is a better way of handling this instaed of having code in Altair which is only used in tests

PR #3565, opened 2024-09-01
fix(typing): Resolve multiple `@utils.use_signature` issues

dangotbanned:
# Description
This PR addresses problems with this decorator that cause the wrong arguments to bind in some cases.

I've tried (and failed) to fix this multiple times over the past 1-2 months, but until now, hadn't come up with a solution that:
- fixed **all** of the cases
- satisfied both `mypy` and `pyright`

After one signature fails in a chain of methods - an IDE (*not `ipython`*) can't understand much that follows. 

Therefore, I consider the issue quite signficant.

# Tasks
- [x] Handle methods & functions correctly in `use_signature` (https://github.com/vega/altair/pull/3565/commits/53057b781ca47349e361a52df50fd322c86a7b32)
- [x] Add a repro for all the call sites this fixes
- [x] Update `use_signature` return type
	- `pyright` is inferring `Overload[...]`, which isn't a form I've seen before
	- `Overload[(cb: (T@decorate, ...) -> R@decorate, /) -> ((T@decorate, **P@use_signature) -> R@decorate), (cb: (...) -> R@decorate, /) -> ((**P@use_signature) -> R@decorate)]`
	- ~~Whatever the solution is, will most likely be a comment~~
	- *Inserted into docstring to avoid breakage*
- [x] Fix bug w/ `Concatenate[T, ...]` inside `TypeAliasType` on `< (3, 11)`
	- `typing.Concatenate` for `3.10`
	- `typing_extensions.Concatenate` for `3.8`, `3.9`

# Repro

```py
from typing_extensions import reveal_type
import altair as alt

# Would also include the function equivalent(s),
# if they did not already have `# pyright: ignore`(s)
concat = alt.ConcatChart()
hconcat = alt.HConcatChart()
vconcat = alt.VConcatChart()
layer = alt.LayerChart()

conf_facet = alt.Chart().configure_facet()
conf_concat = alt.Chart().configure_concat()
conf_tooltip_format = alt.Chart().configure_tooltipFormat()
conf_view = alt.Chart().configure_view()

reveal_type(conf_facet)
reveal_type(conf_concat)
reveal_type(conf_tooltip_format)
reveal_type(conf_view)

chained = conf_facet.configure_arc()
reveal_type(chained)
```

<details>
<summary>Warnings on <code>main</code></summary>

image

</details>

<details><summary>Fixed on <code>fix-use-signature-2</code></summary>
<p>

image

</p>

</details> 

# Examples that fail type checking on `main`
- https://github.com/vega/altair/blob/5b5877903bce55392c66d9dd08b21e0da58a5fda/tests/examples_methods_syntax/ridgeline_plot.py
- https://github.com/vega/altair/blob/5b5877903bce55392c66d9dd08b21e0da58a5fda/tests/examples_methods_syntax/us_population_over_time.py
- https://github.com/vega/altair/blob/5b5877903bce55392c66d9dd08b21e0da58a5fda/tests/examples_methods_syntax/airport_connections.py

## Screenshots

<details>
<summary>Issues in <code>api.py</code></summary>

I added these ignore comments in response to this issue during https://github.com/vega/altair/commit/72bd84df76b405e2e6fc96f2c323470b0cf4bd02

https://github.com/vega/altair/blob/5b5877903bce55392c66d9dd08b21e0da58a5fda/altair/vegalite/v5/api.py#L4168

https://github.com/vega/altair/blob/5b5877903bce55392c66d9dd08b21e0da58a5fda/altair/vegalite/v5/api.py#L4272

https://github.com/vega/altair/blob/5b5877903bce55392c66d9dd08b21e0da58a5fda/altair/vegalite/v5/api.py#L4378

https://github.com/vega/altair/blob/5b5877903bce55392c66d9dd08b21e0da58a5fda/altair/vegalite/v5/api.py#L4504

image
</details>

<details><summary>Enabling type checking for <code>tests/</code></summary>
<p>

image

</p>
</details>

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -12,7 +12,16 @@ from collections.abc import Mapping, MutableMapping
 from copy import deepcopy
 from itertools import groupby
 from operator import itemgetter
-from typing import TYPE_CHECKING, Any, Callable, Iterator, Literal, TypeVar, cast
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    Iterator,
+    Literal,
+    TypeVar,
+    cast,
+    overload,
+)
 
 import jsonschema
 import narwhals.stable.v1 as nw
@@ -26,9 +35,9 @@ if sys.version_info >= (3, 12):
 else:
     from typing_extensions import Protocol, runtime_checkable
 if sys.version_info >= (3, 10):
-    from typing import ParamSpec
+    from typing import Concatenate, ParamSpec
 else:
-    from typing_extensions import ParamSpec
+    from typing_extensions import Concatenate, ParamSpec
 
 
 if TYPE_CHECKING:
@@ -41,9 +50,10 @@ if TYPE_CHECKING:
     from altair.utils._dfi_types import DataFrame as DfiDataFrame
     from altair.vegalite.v5.schema._typing import StandardType_T as InferredVegaLiteType
 
-V = TypeVar("V")
 P = ParamSpec("P")
 TIntoDataFrame = TypeVar("TIntoDataFrame", bound=IntoDataFrame)
+T = TypeVar("T")
+R = TypeVar("R")
 
 
 @runtime_checkable
@@ -709,31 +719,41 @@ def infer_vegalite_type_for_narwhals(
         raise ValueError(msg)
 
 
-def use_signature(obj: Callable[P, Any]):  # -> Callable[..., Callable[P, V]]:
-    """Apply call signature and documentation of `obj` to the decorated method."""
+def use_signature(tp: Callable[P, Any], /):
+    """
+    Use the signature and doc of ``tp`` for the decorated callable ``cb``.
 
-    def decorate(func: Callable[..., V]) -> Callable[P, V]:
-        # call-signature of func is exposed via __wrapped__.
-        # we want it to mimic obj.__init__
+    - **Overload 1**: Decorating method
+    - **Overload 2**: Decorating function
+    """
 
-        # error: Accessing "__init__" on an instance is unsound,
-        # since instance.__init__ could be from an incompatible subclass  [misc]
-        wrapped = (
-            obj.__init__ if (isinstance(obj, type) and issubclass(obj, object)) else obj  # type: ignore [misc]
-        )
-        func.__wrapped__ = wrapped  # type: ignore[attr-defined]
-        func._uses_signature = obj  # type: ignore[attr-defined]
+    @overload
+    def decorate(
+        cb: Callable[Concatenate[T, ...], R], /
+    ) -> Callable[Concatenate[T, P], R]: ...
 
-        # Supplement the docstring of func with information from obj
-        if doc_in := obj.__doc__:
-            doc_lines = doc_in.splitlines(keepends=True)[1:]
-            # Patch in a reference to the class this docstring is copied from,
-            # to generate a hyperlink.
-            line_1 = f"{func.__doc__ or f'Refer to :class:`{obj.__name__}`'}\n"
-            func.__doc__ = "".join((line_1, *doc_lines))
-            return func
+    @overload
+    def decorate(cb: Callable[..., R], /) -> Callable[P, R]: ...
+
+    def decorate(
+        cb: Callable[..., R], /
+    ) -> Callable[Concatenate[T, P], R] | Callable[P, R]:
+        """
+        Raises when no doc was found.
+
+        Notes
+        -----
+        - Reference to ``tp`` is stored in ``cb.__wrapped__``.
+        - The doc for ``cb`` will have a ``.rst`` link added, referring  to ``tp``.
+        """
+        cb.__wrapped__ = getattr(tp, "__init__", tp)  # type: ignore[attr-defined]
+
+        if doc_in := tp.__doc__:
+            line_1 = f"{cb.__doc__ or f'Refer to :class:`{tp.__name__}`'}\n"
+            cb.__doc__ = "".join((line_1, *doc_in.splitlines(keepends=True)[1:]))
+            return cb
         else:
-            msg = f"Found no doc for {obj!r}"
+            msg = f"Found no doc for {tp!r}"
             raise AttributeError(msg)
 
     return decorate
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -31,9 +31,9 @@ from narwhals.typing import IntoDataFrame
 from altair.utils.schemapi import SchemaBase, Undefined
 
 if sys.version_info >= (3, 12):
-    from typing import Protocol, runtime_checkable
+    from typing import Protocol, TypeAliasType, runtime_checkable
 else:
-    from typing_extensions import Protocol, runtime_checkable
+    from typing_extensions import Protocol, TypeAliasType, runtime_checkable
 if sys.version_info >= (3, 10):
     from typing import Concatenate, ParamSpec
 else:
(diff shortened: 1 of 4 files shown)


altair/utils/core.py
```diff
@@ -41,9 +50,21 @@
...
     from altair.vegalite.v5.schema._typing import StandardType_T as InferredVegaLiteType
 
-V = TypeVar("V")
-P = ParamSpec("P")
 TIntoDataFrame = TypeVar("TIntoDataFrame", bound=IntoDataFrame)
+T = TypeVar("T")
+P = ParamSpec("P")
+R = TypeVar("R")
+
+WrapsFunc = TypeAliasType("WrapsFunc", Callable[..., R], type_params=(R,))
```
binste, line 58: Could you elaborate on the benefit of using `TypeAliasType` instead of `WrapsFunc: TypeAlias = Callable[..., R]`? I can't find this way of using it in https://docs.python.org/3/library/typing.html#type-aliases which either recommends `type WrapsFunc = Callable[..., R]` for Python 3.12 (hence we of course can't use it) or the `TypeAlias` way.

altair/utils/core.py
```diff
@@ -41,9 +50,21 @@
...
+WrappedFunc = TypeAliasType("WrappedFunc", Callable[P, R], type_params=(P, R))
+# NOTE: Requires stringized form to avoid `< (3, 11)` issues
+# See: https://github.com/vega/altair/actions/runs/10667859416/job/29567290871?pr=3565
```
binste, line 61: Will this type alias still work as expected in Python < 3.11 or are we just ignoring the error by passing it as a string?

binste (commented):
Nice! A lot for me to learn here. I've added some questions where I could use your help so I can review this further.

altair/utils/core.py
dangotbanned: Apologies if I end up repeating things you've already understood. 

https://docs.python.org/3/library/typing.html#type-aliases states

> A type alias is defined using the [`type`](https://docs.python.org/3/reference/simple_stmts.html#type) statement, which creates an instance of [`TypeAliasType`](https://docs.python.org/3/library/typing.html#typing.TypeAliasType). In this example, `Vector` and `list[float]` will be treated equivalently by static type checkers:

The benefit of using `TypeAliasType` is to have a [**generic**](https://typing.readthedocs.io/en/latest/spec/aliases.html#type-statement) alias (over the `type_params` parameter).
I think the doc for [`typing_extensions.TypeAliasType`](https://github.com/python/typing_extensions/blob/d9509f902c20e3d51c62b8abe522809b4760c0ff/src/typing_extensions.py#L3455-L3481) is a bit more useful to demonstrate this. 

I've used this for [`OneOrSeq`](https://github.com/vega/altair/blob/5207768b6e533c0509218376942309d1c7bac22f/altair/vegalite/v5/schema/_typing.py#L73-L87) as well, which let us specialize its `type_params` when defining [`channels.EncodeKwds`](https://github.com/vega/altair/blob/5207768b6e533c0509218376942309d1c7bac22f/altair/vegalite/v5/schema/channels.py#L25645).

[PEP 695](https://peps.python.org/pep-0695/#generic-type-alias) also has some examples of generic aliases - it is that syntax which this functionality is replicating - but in a backwards-compatible way

altair/utils/core.py
dangotbanned: So [this error](https://github.com/vega/altair/actions/runs/10667859416/job/29567290871?pr=3565
) (*that is no longer present*) 
<details><summary>Log traceback</summary>
<p>

```py
   File "/home/runner/work/altair/altair/altair/utils/core.py", line 61, in <module>
    "WrapsMethod", Callable[Concatenate[T, ...], R], type_params=(T, R)
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/typing.py", line 312, in inner
    return func(*args, **kwds)
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/typing.py", line 403, in __getitem__
    return self._getitem(self, parameters)
  File "/opt/hostedtoolcache/Python/3.10.14/x64/lib/python3.10/typing.py", line 599, in Concatenate
    raise TypeError("The last parameter to Concatenate should be a "
TypeError: The last parameter to Concatenate should be a ParamSpec variable.

```

</p>
</details> 

was being triggered at runtime, but it was not an issue when I used the equivalent as an annotation in an [earlier commit](https://github.com/vega/altair/pull/3565/commits/53057b781ca47349e361a52df50fd322c86a7b32).

Writing it as a string simply gives us the same semantics as writing in an annotation; which seems to be all that was needed in this case.

dangotbanned: To confirm this works, you can try the [**Repro**](https://github.com/vega/altair/pull/3565#issue-2499636422) on `3.10` which should still be able to use this type to resolve the signatures

dangotbanned:
> Nice! A lot for me to learn here. I've added some questions where I could use your help so I can review this further.

Thanks for the review @binste!

Tried my best to answer everything so far.
This was definitely quite a complex one to solve, so I'm happy to explain more if you have any other questions.

altair/utils/core.py
binste: Maybe I'm missing something but defining a type alias the "old" way also produces generic aliases, see [mypy docs]. Or here a small demo:

```python
from typing import List, Tuple, Union
from typing_extensions import TypeAlias, TypeAliasType, TypeVar

T = TypeVar("T")
ListOrTuple = TypeAliasType("ListOrTuple", Union[List[T], Tuple[T]], type_params=(T,))
ListOrTuple2: TypeAlias = Union[List[T], Tuple[T]]


def x(a: ListOrTuple[str], b: ListOrTuple2[str]) -> None:
    pass


x(a=["1"], b=["s"])
x(a=[1], b=["s"])
x(a=["s"], b=[1])

```

Mypy on both Python 3.8 and and 3.11 gives me the expected errors:

```bash
hatch shell hatch-test.py3.8   
mypy mypy_test.py  # <- assuming the above code is stored in such a file

hatch shell hatch-test.py3.11
mypy mypy_test.py
```
image


Not saying that we should not use `TypeAliasType`, I'm just wondering when we want to use it and when we should use `TypeAlias` or if, for our purposes, it does not make a difference.

altair/utils/core.py
binste: Got it, thanks :)

altair/utils/core.py
dangotbanned: Hmm, you're right @binste and I can't seem to find a clear answer on this I'm afraid. 
Quite odd that the doc gives an example that isn't motivating for usage.

I've been using `TypeAliasType` when any `TypeVar`(s) are involved. 
AFAIK, if we followed that as a convention we'd only need to update the definition for `Optional`.

https://github.com/vega/altair/blob/f542e9e8f1e4bf0fa4fe64e89dc5e72b94881cd2/tools/schemapi/schemapi.py#L804-L808

`TypeAlias` is technically [deprecated](https://docs.python.org/3/library/typing.html#typing.TypeAlias) as of `3.12` but until we have `3.12` as a lower bound - I would prefer using `TypeAlias` for simple aliases that are non-generic - as `TypeAliasType` is quite verbose.

Then once we reach `3.12` they can be replaced with `type` statements.

---

I did find that the `...` runtime issue I encountered is known:
- https://github.com/python/typing_extensions/issues/448
- https://github.com/python/typing_extensions/issues/48
- https://github.com/python/typing_extensions/issues/110

binste: I can't find something more definite either but was an interesting journey :) Thanks for all the comments! `TypeAlias` for simple and `TypeAliasType` for more complex syntax is ok with me

binste (approved):
Thanks again for digging through it! :)

dangotbanned:
> Thanks again for digging through it! :)

No problem, thanks for taking the time to review @binste

PR #3566, opened 2024-09-03
fix: dont materialise Ibis table to PyArrow if using vegafusion data transformer

MarcoGorelli:
### Background

In https://github.com/vega/altair/pull/3452#pullrequestreview-2176310770, @jonmmease pointed out that

> To avoid regressing performance in the VegaFusion+Ibis case, I think we'll need to avoid loading the full DataFrame into memory as pyarrow at this point, and instead use the __dataframe__ protocol directly to introspect the column types.

I'd originally done that in the linked PR, but then in a subsequent refactor (in the later commit from the same PR) must have regressed it, because currently, `arrow_table_from_dfi_dataframe` does get called, even with the vegafusion transformer

This PR fixes it

---

I noticed this whilst looking into supporting DuckDB, which is now also supported:

image


jonmmease (approved):
Thanks @MarcoGorelli, lgtm!

PR #3573, opened 2024-09-06
fix(typing): Relax `dict` annotations in `channels.py`

### LINKED ISSUE #3572: `channels.py` parameters, properties types too strict
### What happened?

While rewriting some `alt.condition`(s) to `alt.when`(s) in #3544 I came across some [examples](https://github.com/vega/altair/pull/3544#discussion_r1745791727) that work correctly - but upset a type checker.

# Minimal Repro

<p>

```py
import altair as alt

source = data.cars()
color = (
    alt.when(alt.datum.Weight_in_lbs >= 3500)
    .then(alt.value("hotpink"))
    .otherwise(alt.value("goldenrod"))
)

chart = (
    alt.Chart(source, width=300, height=150)
    .mark_rect()
    .encode(
        x=alt.X("Cylinders:N").axis(labelColor=color),
        y=alt.Y("Origin:N", axis=alt.Axis(tickColor=color)),
        color=color,
    )
)
chart
```

</p>
</details> 

<details><summary>Screenshot</summary>
Didn't promise it looks pretty, but the spec validates
<p>

image

</p>
</details> 

# Solution
The issue is unrelated to #3552, and can instead reuse the solution from #3458.

Not expecting this will require any runtime changes.

However, it will very likely create a **huge diff** since so much of the generated code will have changes like below:

```py
Map: TypeAlias = Mapping[str, Any]

before: dict | SchemaBase
after: Map | SchemaBase
```

Creating an issue for visibility and to note that it will need to be merged before #3544 to avoid needing to ignore typing for examples

Not require `# type: ignore` comments for expected use of public API.

`5.4.0`, `5.4.1`, `5.5.0dev`

dangotbanned:
Fixes #3572

- `mypy` fails in https://github.com/vega/altair/actions/runs/10737803168/job/29780043433?pr=3573
- `mypy` passes in https://github.com/vega/altair/actions/runs/10737971560/job/29780584984?pr=3573


dangotbanned:
Hmm, may move this back to draft. 
Wasn't planning to modify the docstrings

altair/vegalite/v5/schema/channels.py
```diff
@@ -24641,45 +24487,45 @@ def field(
...
-            | Sequence[str | dict | float | Parameter | SchemaBase | Sequence[float]]
+            | Sequence[Map | str | float | Parameter | SchemaBase | Sequence[float]]
             | RangeEnum_T
         ] = Undefined,
-        rangeMax: Optional[str | dict | float | Parameter | SchemaBase] = Undefined,
-        rangeMin: Optional[str | dict | float | Parameter | SchemaBase] = Undefined,
-        reverse: Optional[bool | dict | Parameter | SchemaBase] = Undefined,
-        round: Optional[bool | dict | Parameter | SchemaBase] = Undefined,
-        scheme: Optional[dict | Parameter | SchemaBase | ColorScheme_T] = Undefined,
+        rangeMax: Optional[Map | str | float | Parameter | SchemaBase] = Undefined,
```
binste, line 24522: Thanks @dangotbanned!

Any chance we could have `Map` later on before `Parameter or SchemaBase`? Maybe this was also not done consistently before, there is some code in `tools/schemapi/utils.py:488`.

I'm bringing it up in this PR as `Map` is now first in some cases where it was not before which is a slight deterioration in UX.

dangotbanned: Yeah should be able to sort that out, so `Map` after all lowercase?

binste: I think that would make sense, maybe even after Parameter? To me, it seems like the basic types (`str`, `float`, ...) and `Parameter` are what users are interested in.

dangotbanned: I can get this without needing to special-case anything:

<details>
<summary>Huge output diff screenshot</summary>

image

</details>

Essentially moves `Map` into the aliases group, and since it is the shortest - it appears first

<details>
<summary>Required changes diff</summary>

image


</details>

Would that be a reasonable fix @binste?

Trying it visually here - since the actual diff this produces is gigantic

binste:
Great fix regarding the de-prioritization of `Map`!

dangotbanned:
> Great fix regarding the de-prioritization of `Map`!

Thanks for reviewing @binste

PR #3580, opened 2024-09-06
feat: Add 'olli' renderer to generate accessible text structures for screen reader users

### LINKED ISSUE #3575: improve accessibility for screen reader users
### What is your suggestion?

Olli is a library for converting web visualizations into accessible text structures for blind and low-vision screen reader users. See the [website](https://mitvis.github.io/olli/) and GitHub [repository](https://github.com/mitvis/olli).

We might be able to utilise the Altair’s Renderer Framework to render Altair specifications as Olli charts (eg. `alt.renderers.enable("olli")`)

Inspiration for implementation might come from the recent introduced `alt.renderers.enable("jupyter")` in https://github.com/vega/altair/pull/3283.

For improving accessibility within Altair we are free to request input from @jonathanzong if needed.

There is also Umwelt (see the [paper](https://jonathanzong.com/assets/papers/umwelt-chi-2024.pdf)), which treats visualization, sonification, and textual description as coequal representations, where Olli is merely focussing on providing information for screen-readers. 

In the paper of Umwelt it is phrased that:

> Umwelt extends Vega-Lite’s concept of view composition to express multi-view textual structures and sonifcations.

I'm not sure if Umwelt also (will) support Vega-Lite specifications as-is, or if it require prior changes in the Vega-Lite repository. We could start with Olli, because it can be fed by Vega-Lite specifications as are currently created by Altair.

binste:
> [Olli](https://mitvis.github.io/olli/) is an open-source library for converting data visualizations into accessible text structures for screen reader users. Starting with an existing visualization specification created with a supported toolkit, Olli produces a keyboard-navigable tree view with descriptions at varying levels of detail. Users can explore these structures both to get an initial overview, and to dive into the data in more detail.

Closes #3575. A possible use case of this is that in a university course where Altair is used in a Jupyter notebook, with a single line of code, `alt.renderer.enable("olli")`, all charts in a notebook can be made accessible for screen reader users.

I've tested this in JupyterLab and VS Code. In JupyterLab, all [Olli controls](https://mitvis.github.io/olli/tutorial) worked, in VS Code, the following do not:
- T: Open table
- O: Jump to Olli tree
- F: Data filters

This might be an issue with how VS Code intercepts those key strokes. Might be configurable in VS Code but for now, I'm happy if it works in JupyterLab.

### DIFF
--- a/altair/utils/html.py
+++ b/altair/utils/html.py
@@ -161,7 +161,10 @@ HTML_TEMPLATE_UNIVERSAL = jinja2.Template(
         .catch(err => showError(`Javascript Error: ${err.message}<br>This usually means there's a typo in your chart specification. See the javascript console for the full traceback.`));
       {%- if use_olli %}
       olliAdapters.VegaLiteAdapter(spec).then(olliVisSpec => {
-        const olliRender = olli.olli(olliVisSpec);
+        // It's a function if it was loaded via maybeLoadScript below.
+        // If it comes from require, it's a module and we access olli.olli
+        const olliFunc = typeof olli === 'function' ? olli : olli.olli;
+        const olliRender = olliFunc(olliVisSpec);
         olliDiv.append(olliRender);
       });
       {%- endif %}
@@ -183,7 +186,11 @@ HTML_TEMPLATE_UNIVERSAL = jinja2.Template(
         .then(() => maybeLoadScript("olli-adapters", "{{olli_adapters_version}}"))
         {%- endif %}
         .catch(showError)
+        {%- if use_olli %}
+        .then(() => displayChart(vegaEmbed, olli, OlliAdapters));
+        {%- else %}
         .then(() => displayChart(vegaEmbed));
+        {%- endif %}
     }
   })({{ spec }}, {{ embed_options }});
 </script>
--- a/altair/utils/html.py
+++ b/altair/utils/html.py
@@ -7,7 +7,7 @@ import jinja2
 
 from altair.utils._importers import import_vl_convert, vl_version_for_vl_convert
 
-TemplateName = Literal["standard", "universal", "inline"]
+TemplateName = Literal["standard", "universal", "inline", "olli"]
 RenderMode = Literal["vega", "vega-lite"]
 
 HTML_TEMPLATE = jinja2.Template(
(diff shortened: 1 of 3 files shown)


binste:
@jonathanzong Would you mind reviewing the JavaScript code? It's in an ugly Jinja template without syntax highlighting but maybe you notice something when skimming through it. It works in JupyterLab and VS Code so it can't be super wrong.

jonathanzong:
looks good to me!

altair/utils/html.py
```diff
@@ -293,6 +326,12 @@ def spec_to_html(
         vlc = import_vl_convert()
         vl_version = vl_version_for_vl_convert()
         render_kwargs["vegaembed_script"] = vlc.javascript_bundle(vl_version=vl_version)
+    elif template == "olli":
+        olli_version = "2"
+        olli_adapters_version = "2"
```
dangotbanned, line 331: Should these be defined as global constants or exposed as parameters?

dangotbanned (changes requested):
Seems like a really interesting feature

Just had one question, since your js got the okay in https://github.com/vega/altair/pull/3580#issuecomment-2338489353

altair/utils/html.py
binste: Thought process was that as it's a very new feature, it's better to keep them internal for now. There are also only used in this one place. Regarding parameters, olli is currently being rewritten in Solid.js, it's API might change. Hence, version 3 might require changes to the rendering template -> Don't expect users to be able to just change a parameter to upgrade, we probably have to do it in Altair. Once that interface is more stable, having it as a renderer parameter sounds good!

altair/utils/html.py
```diff
@@ -293,6 +326,12 @@ def spec_to_html(
...
+        render_kwargs["olli_version"] = olli_version
+        render_kwargs["olli_adapters_version"] = olli_adapters_version
```
dangotbanned, line 333:
```suggestion
        OLLI_VERSION = "2"
        OLLI_ADAPTERS_VERSION = "2"
        render_kwargs["olli_version"] = OLLI_VERSION
        render_kwargs["olli_adapters_version"] = OLLI_ADAPTERS_VERSION
```

dangotbanned (approved):
https://github.com/vega/altair/pull/3580/files#r1750840828 

Sounds good to me, thanks for this @binste

altair/utils/html.py
dangotbanned: This isn't blocking for me btw, just thought it would help keep the constants easy to find if/when they need updating

binste:
Thank you both for the reviews!

Issue #3581, opened 2024-09-06
`selection_point` fails schema validation


jpn--:
### What happened?

I am trying to create an interactive figure that filters on data by iteration, but I am getting a SchemaValidationError:

`SchemaValidationError: '2000' is an invalid value for `iteration`. Valid values are of type 'array'.`

If I copy the published example from [here](https://altair-viz.github.io/gallery/us_population_over_time.html), and run it, it works fine.  But if I simply change the name of the data column "year" to "iteration", I can reproduce the failure.

```
import altair as alt
from vega_datasets import data
import pandas as pd

# read to pandas DataFrame 
df = pd.read_json(data.population.url)

# change column name "year" to "iteration"
df = df.rename(columns={"year": "iteration"})

# code below otherwise identical to example except changing "year" to "iteration"
select_year = alt.selection_point(
    name="Year",
    fields=["iteration"],
    bind=alt.binding_range(min=1900, max=2000, step=10, name="Year"),
    value={"iteration": 2000},
)

alt.Chart(df).mark_bar().encode(
    alt.X("sex:N").title('').axis(labels=False, ticks=False),
    alt.Y("people:Q").scale(domain=(0, 12000000)).title("Population"),
    alt.Color("sex:N")
        .scale(domain=("Male", "Female"), range=["steelblue", "salmon"])
        .title("Sex"),
    alt.Column("age:O").title("Age")
).properties(
    width=20,
    title="U.S. Population by Age and Sex"
).add_params(
    select_year
).transform_calculate(
    "sex", alt.expr.if_(alt.datum.sex == 1, "Male", "Female")
).transform_filter(
    select_year
).configure_facet(
    spacing=8
)
```

### What would you like to happen instead?

The selection should work and not depend on the variable name literally being "year".

### Which version of Altair are you using?

5.4.1

dangotbanned:
```py
# change column name "year" to "iteration"
# df.rename(columns={"year": "iteration"})
source = df.rename(columns={"year": "iteration"})

```

jpn--:
@dangotbanned Apologies, I had a bad copy-and-paste job into the issue.  I've corrected the code above, the error still persists; and actually, I've found the error occur even before my (now fixed) bug.  I can reproduce the `SchemaValidationError` with just this:

```
import altair as alt

select_year = alt.selection_point(
    name="Year",
    fields=["iteration"],
    bind=alt.binding_range(min=1900, max=2000, step=10, name="Year"),
[...]

PR #3584, opened 2024-09-09
fix: Update docs to use correct init value for `selection_point`

jpn--:
Fixes for #3581


mattijn:
Thanks @jpn--! Very much appreciated that you took the time to create a PR!

Issue #3587, opened 2024-09-10
Preparing for Python 3.13


dangotbanned:
https://www.python.org/downloads/release/python-3130rc2/

> This release, 3.13.0rc2, is the final release preview. This release is expected to become the final 3.13.0 release, barring any critical bugs being discovered.
> The official release of `3.13.0` is scheduled for Tuesday, **2024-10-01**.

> There will be no ABI changes from this point forward in the 3.13 series.
> Call to action
> We **strongly encourage maintainers of Python projects to prepare their projects for 3.13 compatibilities during this phase**, and where necessary publish Python 3.13 wheels on PyPI to be ready for the final release of 3.13.0. 
> Any binary wheels built against Python 3.13.0rc1 and later will work with future versions of Python 3.13. As always, report any issues to [the Python bug tracker](https://github.com/python/cpython/issues).

# Tasks
- [x] Resolve possible `ImportError`(s) https://github.com/vega/altair/discussions/3548#discussioncomment-10407366
    - https://github.com/vega/altair/pull/3593
- [x] Add `"3.13"` to
	- [x] [build.yml#L10](https://github.com/vega/altair/blob/a171ce8cb2f0b0cb0c944ddbd0c0623282570c0c/.github/workflows/build.yml#L10)
	- [x] [pyproject.toml#L147-L148](https://github.com/vega/altair/blob/a171ce8cb2f0b0cb0c944ddbd0c0623282570c0c/pyproject.toml#L147-L148)
	- [x] [pyproject.toml#L48](https://github.com/vega/altair/blob/a171ce8cb2f0b0cb0c944ddbd0c0623282570c0c/pyproject.toml#L48)
- [x] Update `typing_extensions` bound [pyproject.toml#L18](https://github.com/vega/altair/blob/a171ce8cb2f0b0cb0c944ddbd0c0623282570c0c/pyproject.toml#L18)

# Related
- https://github.com/python-jsonschema/jsonschema/pull/1282
- https://github.com/pallets/jinja/pull/1977
- https://github.com/pypa/packaging/pull/783
- https://github.com/astral-sh/uv/pull/7263
- https://github.com/pypa/hatch/releases/tag/hatch-v1.13.0
- https://github.com/geopandas/geopandas/issues/3400
  - https://github.com/geopandas/pyogrio/issues/470
  - https://github.com/pyproj4/pyproj/issues/1439
- https://github.com/narwhals-dev/narwhals/pull/1094
[...]

Issue #3586, opened 2024-09-10
Make it easier for downstream libraries to *safely* contribute themes


dangotbanned:
### What is your suggestion?

<details><summary><i>Originally posted by @dangotbanned in <a href="https://github.com/pola-rs/polars/pull/18609#discussion_r1749814694">discussion</a> w/ @MarcoGorelli </i></summary>
<p>

As I understand, the `alt.Chart.configure_` calls are being used to avoid registering + enabling a theme - which could override a user's custom theme.

These work fine in isolation, but AFAIK would have issues if a user were to layer/concat/facet the result - since `config` is only valid at the top-level. 

You might want to add tests to see if [these ops](https://github.com/vega/altair/blob/df14929075b45233126f4cfe579c139e0b7f0559/tests/vegalite/v5/test_api.py#L332-L361) would still be possible

- https://altair-viz.github.io/user_guide/customization.html#global-config
- https://altair-viz.github.io/user_guide/configuration.html#top-level-chart-configuration

Using a theme would have the benefit of deferring these config settings until the `Chart` is rendered - placing them in the [top-level only](https://github.com/vega/altair/blob/df14929075b45233126f4cfe579c139e0b7f0559/altair/vegalite/v5/api.py#L1837). 

---

It might be worth seeing if we can come to a good solution to this as part of https://github.com/vega/altair/issues/3519 since we have already [discussed](https://github.com/pola-rs/polars/pull/17995#issuecomment-2282795126) [issues](https://github.com/pola-rs/polars/pull/17995#issuecomment-2282802508) with the theme route

</p>
</details> 

## Problem

A library like `polars` may wish to provide a default theme, but not override a user-defined or user-enabled theme.

AFAIK, the "best" solution for this right now would be to override our `"default"` theme.
However, this would be a **destructive** action and wouldn't scale well to multiple 3rd-parties each doing so:

<details>
<summary>Code block</summary>

https://github.com/vega/altair/blob/df14929075b45233126f4cfe579c139e0b7f0559/altair/vegalite/v5/theme.py#L56-L74

</details>

## Solution(s)
We could extend `ThemeRegistry` to support priority levels.

Either when registering/enabling a theme a level will be set corresponding to the party.

```py
from enum import IntEnum

class ThemePriority(IntEnum):
    USER = 1
    THIRD_PARTY = 2
    DEFAULT = 3 # alternatives: `ALTAIR`, `STANDARD`, `BUILTIN`
```

[...]

PR #3544, opened 2024-08-18
docs: Promote `when-then-otherwise` in User Guide

dangotbanned:
Will close #3500

# Status
All outstanding work should be scoped to `parameters.rst`.
*Corrections/suggestions elsewhere are still welcome*

The [linked version](https://github.com/dangotbanned/altair/blob/390d59935bf1cf81ce4988fdc6099b9d9a50b790/doc/user_guide/interactions/parameters.rst) represents **Draft 1**.
- Kept things mostly as they were for `alt.condition`
- Hoping to get feedback on
	- **How much** of [alt.when](https://altair-viz.github.io/user_guide/generated/api/altair.when.html#altair.when) feels **missing**?
	- If something is missing, can a link to an example improve the read?
- Potentially missing
	- Chaining, reuse e.g. https://github.com/vega/altair/pull/3544/files#r1746885701
	- Simply repeating these in order https://github.com/vega/altair/pull/3492#pullrequestreview-2205990789

# Tasks
- [x] Use `alt.when` in [waterfall chart](https://github.com/vega/altair/issues/3500#issuecomment-2312826920)
- [x] Fill out tasks following [discussion](https://github.com/vega/altair/issues/3500)
- [x] Add callout re: `v5.4.0` change [comment](https://github.com/vega/altair/issues/3500#issuecomment-2295148166)
	- Probably makes most sense here   <details><summary><code>parameters.rst</code></summary>  https://github.com/vega/altair/blob/5207768b6e533c0509218376942309d1c7bac22f/doc/user_guide/interactions/parameters.rst?plain=1#L148-L155  </details> 
- [x] `when(..., empty=False)` example in *User Guide* [comment](https://github.com/vega/altair/pull/3544#issuecomment-2329471941)
	- Possibly link to [example](https://altair-viz.github.io/gallery/interactive_bar_select_highlight.html)
	- Similar to the *zorder* link at the end of [conditional-encodings](https://altair-viz.github.io/user_guide/interactions.html#conditional-encodings)
	- Resolved in [`df5555c` (#3544)](https://github.com/vega/altair/pull/3544/commits/df5555cd869c4b4b4b7bf638da08850346110915)
- [x] `parameters.rst`
	- [x] Draft 1
	- [x] Draft 2
- [x] Replace `condition` references with `when` [comment](https://github.com/vega/altair/issues/3500#issuecomment-2295148166)
	- [x] User Guide 
	    - [x]  Interactions
	    	- [x] `bindings_widgets.rst`
[...]

PR #3591, opened 2024-09-14
ci: Support Python 3.13

dangotbanned:
Resolves https://github.com/vega/altair/issues/3587

Minimal/fallback support for `3.13`.

Ideally I'd like to minimise hard dependencies we have on `typing_extensions` before October 2024. However, this would be the simpler option if it passes CI

- [x] https://github.com/pyproj4/pyproj/issues/1439
- [x] https://github.com/apache/arrow/issues/43519
- [x] https://github.com/apache/arrow/pull/44440
- [x] #3672


dangotbanned:
> I think it is noble to wait for support for python 3.13 for all our upstream soft dependencies. But I'm also not against releasing before all our soft dependencies provide support (in my understanding pyarrow). What you think @dangotbanned?

@mattijn nothing noble here 😄 

I'm happy with releasing before `pyarrow>=18.0.0` in principle.
The issue is we'd need to remove `pyarrow` from `altair[all]` (https://github.com/vega/altair/actions/runs/11448782729/job/31853033909?pr=3591)

Also maybe do some decoupling in `tests/.`, if there are any imports that assume `pyarrow` is always installed.

Doing either of those is fine with me if you'd like to give it a try?

---

Personally though, I've got a few issues/PRs that I'd rather have ready for `5.5.0` (_+ some others that are less pressing_):
- https://github.com/vega/altair/issues/3650
	- https://github.com/vega/altair/pull/3654
- https://github.com/vega/altair/pull/3544
- https://github.com/vega/altair/issues/3519
	- https://github.com/vega/altair/issues/3645
	- https://github.com/vega/altair/pull/3618
	- https://github.com/vega/altair/issues/3619
	- https://github.com/vega/altair/issues/3586
- https://github.com/vega/altair/issues/3657

I think there's more value in delivering some/all of these vs a slightly earlier `5.5.0`, if that makes sense?

mattijn:
Hmpf. Decoupling `altair[all]` is not something I was aiming for. Let's wait.

dangotbanned:
> Hmpf. Decoupling `altair[all]` is not something I was aiming for. Let's wait.

No worries @mattijn 

Also, IIRC `pyarrow` was blocking for `geopandas` - which I think is relevant to you?

mattijn:
Yeah, I assume that geospatial users of Altair are the minority, so if that's causing the delay, also feel a bit responsible for that.

dangotbanned:
> Yeah, I assume that geospatial users of Altair are the minority, so if that's causing the delay, also feel a bit responsible for that.

Its not on you @mattijn! 
just giving you a heads up if you were eager to use `3.13` is all

dangotbanned:
I'm all out of ideas on this one 😞 

@mattijn @jonmmease @joelostblom @binste 
Pinging in case any of you guys can make sense of the logs

jonmmease:
It looks like pip is first installing pyarrow 18, but then installs 17, which it tries to build from source since there aren't wheel published for Python 3.13. I think the issue is that ibis 9.5.0 has an upper bound on pyarrow of `<18`. See https://github.com/ibis-project/ibis/blob/9.5.0/pyproject.toml. This was relaxed on their man branch, but a new version hasn't been published yet.

I don't recall offhand which of our tests need ibis with the polars backend, but we could see if theres a way to rework them without it.

dangotbanned:
Thanks @jonmmease 

> I don't recall offhand which of our tests need ibis with the polars backend, but we could see if theres a way to rework them without it.

If that is the issue, these tests are the one's that need it:

https://github.com/vega/altair/blob/211f7c517d7c005af521b242e51e414a084b8350/tests/vegalite/v5/test_api.py#L1614-L1656

I had to change those a bit when dropping `3.8` to fix https://github.com/vega/altair/pull/3647#issuecomment-2426312642

Interesting if they cause problems in both cases

dangotbanned:
@MarcoGorelli 
I was wondering if you had any ideas on *if/how* we could move this forward, prior to `ibis>=10.0.0`?

I'd been following the `3.13` PR you had (https://github.com/narwhals-dev/narwhals/pull/1094) but I must say I'm feeling a bit out of my depth on this one

MarcoGorelli:
hey - I'd suggest removing Ibis from CI, I think we can use duckdb for that test, I'll make a PR

dangotbanned:
> hey - I'd suggest removing Ibis from CI, I think we can use duckdb for that test, I'll make a PR
> 
> here you go #3672
> 
> (note that we continue to test Ibis in CI in Narwhals, so support for it should stay strong)

@MarcoGorelli looks to have done the job, thanks again 🙏

mattijn:
Indeed thanks @MarcoGorelli! Appreciated👍

Issue #3603, opened 2024-09-20
When using the browser renderer, no charset is set in the generated HTML


davidgroves:
### What happened?

1. Generate any chart with a large label that will be truncated.
2. Observe the truncation on a browser shows `â€¦` instead of …

Confirm this is seen on Firefox 130.0 and Chromium 128.0.6613.119

### What would you like to happen instead?

Have the Unicode ellipsis character … (U+2026) instead, as intended.

Current Output.
![image](https://github.com/user-attachments/assets/63750ead-dc17-4379-b2b9-b32ffa063be9)

Desired Output.
![image](https://github.com/user-attachments/assets/2acaf596-15e1-4ae3-bd7c-e59b6dcd5794)

### Which version of Altair are you using?

5.4.1 and 5.5.0dev

PR #3604, opened 2024-09-20
fix: Set charset=UTF-8 in HTML templates.

davidgroves:
- Fix https://github.com/vega/altair/issues/3603

I'm not sure if you want a more sophisticated approach, or even if this is user-error rather than a defect, but this at least fixed my immediate problem and may be "good enough" for the project.


binste (approved):
Makes sense to me, thank you @davidgroves for the PR!

PR #3609, opened 2024-09-24
ci: Moved `vl-convert-python` to `save` dependency group

### LINKED ISSUE #3608: Define `vl-convert-python` dependency outside of `altair[all]`
### What is your suggestion?

The `vl-convert-python` package is necessary for saving plots in PNG / SVG / PDF formats. I believe this is a rather critical function, so I'm wondering why `vl-convert-python` needs to be manually installed for this capability. (I see that `altair[all]` is an alternative, but I think `vega_datasets`, etc. are less critical than the save functionality.)

`vl-convert-python` has zero dependencies. So I think this would be a simple addition. Maybe that is the reason Is this dependency optional because the built wheels are ~30 MB? Could this size be reduced?

Thanks so much for your consideration!

Alternatively, could this (at least) be added to a new optional dependency group `altair[save]`, as suggested [here](https://github.com/vega/altair/issues/3309#issuecomment-1953600671)?

apoorvkh:
Simple PR! As in title: proposing `altair[save]` optional dependency group to enable saving to static files (without installing `altair[all]`).

Fixes #3608

### DIFF
--- a/doc/getting_started/installation.rst
+++ b/doc/getting_started/installation.rst
@@ -24,6 +24,12 @@ and non-notebook IDEs, see :ref:`displaying-charts`.
 If you wish to install Altair with only the required dependencies,
 you can omit the ``[all]``/``-all`` suffix.
 
+Altair can also be installed with only the dependencies necessary for saving charts to PNG, SVG, or PDF formats, using:
+
+.. code-block:: bash
+
+    pip install "altair[save]"
+
 Development Installation
 ========================
 
--- a/doc/getting_started/installation.rst
+++ b/doc/getting_started/installation.rst
@@ -24,7 +24,7 @@ and non-notebook IDEs, see :ref:`displaying-charts`.
 If you wish to install Altair with only the required dependencies,
 you can omit the ``[all]``/``-all`` suffix.
 
-Altair can also be installed with only the dependencies necessary for saving charts to PNG, SVG, or PDF formats, using:
+Altair can also be installed with just the dependencies necessary for saving charts to offline HTML files or PNG/SVG/PDF formats, using:
 
 .. code-block:: bash
 
--- a/doc/user_guide/saving_charts.rst
+++ b/doc/user_guide/saving_charts.rst
@@ -157,7 +157,7 @@ As an alternative, the ``inline=True`` keyword argument may be provided to ``cha
 
 .. note::
 
-   Calling ``chart.save`` with ``inline=True`` requires the :ref:`install-vl-convert` package.
+   Calling ``chart.save`` with ``inline=True`` requires installation via ``pip install "altair[save]"`` or ``pip install "altair[all]"``.
 
 
 .. _saving-png:
@@ -172,25 +172,9 @@ To save an Altair chart object as a PNG, SVG, or PDF image, you can use
     chart.save('chart.svg')
     chart.save('chart.pdf')
 
-Saving these images requires an additional extension vl-convert_ to run the
-javascript code necessary to interpret the Vega-Lite specification and output
-it in the form of an image.
+.. note::
 
-.. _install-vl-convert:
-
-vl-convert
-^^^^^^^^^^
-The vl-convert_ package can be installed with::
-
-    conda install -c conda-forge vl-convert-python
-
-or::
-
-    pip install vl-convert-python
-
-vl-convert_ does not require any external dependencies.
-See the vl-convert documentation for information and for known
-`limitations <https://github.com/vega/vl-convert#limitations>`_.
+   Saving these images requires installation via ``pip install "altair[save]"`` or ``pip install "altair[all]"``.
 
 altair_saver
 ^^^^^^^^^^^^
--- a/doc/user_guide/saving_charts.rst
+++ b/doc/user_guide/saving_charts.rst
@@ -157,7 +157,7 @@ As an alternative, the ``inline=True`` keyword argument may be provided to ``cha
 
 .. note::
 
-   Calling ``chart.save`` with ``inline=True`` requires installation via ``pip install "altair[save]"`` or ``pip install "altair[all]"``.
+   Calling ``chart.save`` with ``inline=True`` requires :ref:`additional-dependencies`.
 
 
 .. _saving-png:
@@ -174,7 +174,9 @@ To save an Altair chart object as a PNG, SVG, or PDF image, you can use
 
 .. note::
 
-   Saving these images requires installation via ``pip install "altair[save]"`` or ``pip install "altair[all]"``.
+   :ref:`additional-dependencies` are required to save charts as images by running the javascript
+   code necessary to interpret the Vega-Lite specification and output it in the form of an image.
+
 
 altair_saver
 ^^^^^^^^^^^^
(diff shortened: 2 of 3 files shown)


mattijn:
Thanks for this PR! Makes sense to add this as a separate extras group. It would be nice to update the docs on the installation page as well: https://altair-viz.github.io/getting_started/installation.html.

Do you see a chance to add this as commit to this PR @apoorvkh? 

Also @joelostblom, is it possible to have this working within the conda feedstock as well?

Again, appreciated for the PR!

apoorvkh:
Sure! How are those changes? @mattijn

mattijn:
Thanks for adding info to the docs. I understand that you want to emphasize the possibility of this routine, but I was hoping not to remove the references to the vl-convert extension for saving and the method how conda can be used to to install vl-convert-python. As long as there is not yet an option like `conda install altair-save -c conda-forge` we have to keep information how this should currently be done using conda within the documentation. Can you update that? Almost there! Thanks!

apoorvkh:
Okay -- I added the information back but with small modifications to those docs. Good now?

joelostblom:
Thanks for your contribution @apoorvkh.

From my side, I'm not convinced that we need a dependency group for a single package. Typing out `pip install altair vl-convert-python` is not notably more inconvenient than `pip install "altair[save]"`, and it doesn't solve the issue you reported in #3608 regarding it being unexpected that saving was not included by default. I'm on board with making it clearer in the docs how to install the saving packages though, like [you did here](https://github.com/vega/altair/pull/3609/files#diff-d0877f6f8f9ef83d5f63719d6e590c91ef6c55c3e50a2ae769ad1cfff7e9aabdR27-R32) (but for `vl-convert-python`).

mattijn:
Thanks for joining @joelostblom! Just at the right moment as I'd already activated auto-merge :). 

I have to admit, I often cannot remember the dependency that is required to save altair charts to images (repo name vl-convert, pypi pacakge vl-convert-python, import vl_convert as vlc) . I'm quite in favor to have something that facilitate this for people that don't care the names of the packages that are used under the hood, but just want to save a chart. It is not totally uncommon to have a single dependency in the extras-group, see eg. https://pandas.pydata.org/docs/getting_started/install.html#xml.

apoorvkh:
I typed the following at the same time as @mattijn's comment, so echo'ing some things here:

My motivation is to simplify. If I have a project with an Altair dependency, my `pyproject.toml` might look like:

```tool
dependencies = [
    "altair==5.4.1",
    "vl-convert-python==1.6.1",
]
```

It is then not obvious that `vl-convert-python` is a dependency for `altair` and particularly for saving charts to images, etc. What if a future version of `altair` decides to use a different package as the save dependency? The onus is on the user (who didn't really care about `vl-convert-python` in the first place) to notice their project breaks and track that change.

You could say then that the user should specify the `altair[all]` dependency, but we are back at the problem where it seems silly to install 8 dependencies so I can save my plot as a PNG (again, imo, a rather critical feature).

apoorvkh:
I think the long term solution is to build this feature directly into `altair`, without `vl-convert` (or similar overhead)! But @jonmmease elaborated on [why that was hard](https://github.com/vega/altair/issues/3608#issuecomment-2367887119) lol

joelostblom:
Ok I don't feel strongly about this one and it sound like you both have some use cases where it would be beneficial, so I'm fine with merging unless someone else wants to chime in.

It should be straight forward to add the conda package, although it might be abit unclear what does what in conda since the packages are hyphenated, so we will have `altair-save` which might mean Altair + saving dependencies, but I could also see someone interpret it as just the saving dependencies without altair itself, just like we used to have `altair_saver`.

mattijn:
I'd forgotten about `altair_saver`.., but yes `altair-save` will be different in this case. I can see some incoming confusion there. Let's leave the PR open for a while to see if we might come up with something better. Not merging this PR won't be a blocker for other pull requests at the moment.

apoorvkh:
So any better ideas? If not, shall we merge this?

apoorvkh:
If interested in my two cents: I think `altair[save]` is by far most intuitive, because these are the extra dependencies for `.save()`.

I would point out that `altair_saver` compatibility was dropped in Altair 5.0 (May 2023). I think most people would be introduced to the new `altair-save` package via the official [installation page](https://altair-viz.github.io/getting_started/installation.html) (just as `altair-all` is currently). So I think a good strategy would just be to make that page really clearly state that `altair-save` includes both `altair` and additional saving capabilities.

If that is too opinionated: `altair[export]` sounds good too. Indicating export _to_ other file format(s).

In any case, would be great if we can push any of these! Thanks!

dangotbanned:
As always, I'm here with what does `polars` do?

https://github.com/pola-rs/polars/blob/425e251fa3fed25f1956fb5486ed1e41f51ce170/py-polars/pyproject.toml#L39-L81

<details><summary><code>[project.optional-dependencies]</code></summary>
<p>

```toml
[project.optional-dependencies]
# NOTE: keep this list in sync with show_versions() and requirements-dev.txt
# Interop
numpy = ["numpy >= 1.16.0"]
pandas = ["pandas", "polars[pyarrow]"]
pyarrow = ["pyarrow >= 7.0.0"]
pydantic = ["pydantic"]

# Excel
calamine = ["fastexcel >= 0.9"]
openpyxl = ["openpyxl >= 3.0.0"]
xlsx2csv = ["xlsx2csv >= 0.8.0"]
xlsxwriter = ["xlsxwriter"]
excel = ["polars[calamine,openpyxl,xlsx2csv,xlsxwriter]"]

# Database
adbc = ["adbc-driver-manager[dbapi]", "adbc-driver-sqlite[dbapi]"]
connectorx = ["connectorx >= 0.3.2"]
sqlalchemy = ["sqlalchemy", "polars[pandas]"]
database = ["polars[adbc,connectorx,sqlalchemy]", "nest-asyncio"]

# Cloud
fsspec = ["fsspec"]

# Other I/O
deltalake = ["deltalake >= 0.15.0"]
iceberg = ["pyiceberg >= 0.5.0"]

# Other
async = ["gevent"]
cloudpickle = ["cloudpickle"]
graph = ["matplotlib"]
plot = ["altair >= 5.4.0"]
style = ["great-tables >= 0.8.0"]
timezone = ["backports.zoneinfo; python_version < '3.9'", "tzdata; platform_system == 'Windows'"]

# GPU Engine
gpu = ["cudf-polars-cu12"]

# All
all = [
  "polars[async,cloudpickle,database,deltalake,excel,fsspec,graph,iceberg,numpy,pandas,plot,pyarrow,pydantic,style,timezone]",
]
```

</p>
</details> 

I think @apoorvkh's case for `altair[save]` in https://github.com/vega/altair/pull/3609#issuecomment-2436555064 is sound.

---

Regarding `altair_saver`, looking at https://github.com/conda-forge/altair_saver-feedstock?tab=readme-ov-file#installing-altair_saver

I think these commands are different enough to avoid confusion

### v4 (unmaintained)

```
conda install altair_saver
```

### >= v5 (proposed)

```
conda install altair altair-save
```

mattijn:
Thanks for your comment! Adding a (sphinx) note below this option to the installation page should help users make an informed decision. Just to clarify, there’s no need for extra packages if you're simply saving your chart to an HTML file with Vega packages loaded from a CDN. But if you want to convert an Altair specification to a static image (SVG, PNG, JPEG, or PDF), or a Vega JSON file, or generate a URL for opening the Vega editor with the chart specification (LZ-compressed), you’ll need `vl-convert-python`. You can also use it to create an HTML file with all dependencies bundled for offline use.

apoorvkh:
Then sounds like we are in agreement! And thanks for the elaboration.

Shall we merge? I imagine the conda feedstock maintainers would add `altair-save` and relevant docs after this is merged.

dangotbanned (approved):

PR #3600, opened 2024-09-19
feat: Generate `expr` method signatures, docs

dangotbanned:
Closes #3563 

# Description
This PR builds on the work of #3466, and provides accurate signatures for `alt.expr` methods.

## Benefits
*Originally listed in https://github.com/vega/altair/issues/3563#issue-2494365499*
- Raising a `TypeError` at the time of `expr` definition
	- Rather than silently passing until validated within a `ChartType`
	- Importantly, this is within `python` and **not** `javascript`
- Providing a meaningful traceback, identifying both the `expr` and the missing argument
- Opens the door for defining default argument values
	-  E.g. in [waterfall_chart](https://altair-viz.github.io/gallery/waterfall_chart.html), there are repeat uses of `""` for the `elseValue`
	- If this is a common case, we could use that as a default and it would be clear by the signature alone
- The current docstrings - which refer to parameters by name - would be easier to understand
	- Right now, you'd need to refer to [Vega Expressions](https://vega.github.io/vega/docs/expressions/) to determine the order they must appear in
- The change **would not** add any new requirements to currently valid specs
	- Therefore, suitable as a [**MINOR**](https://altair-viz.github.io/about/versioning.html#version-numbers) version feature

## Screenshots
image
image
image

## Implementation

In #3563 I proposed authoring these changes manually.
However, after revisiting a [conversation](https://github.com/vega/altair/pull/3466#issuecomment-2239361425) I had with @jonmmease I thought it would be worth at least *trying* to generate the definitions.

I found that https://vega.github.io/vega/docs/expressions/ seemed to be the only place they are documented in a single place.
From my understanding, this document is manually maintained but [updates occur infrequently](https://github.com/vega/vega/commits/main/docs/docs/expressions.md).

There was *enough* consistency in the markup used to work out how various parameters aligned with `python` syntax https://docs.python.org/3/tutorial/controlflow.html#special-parameters.

I also used this opportunity to clean up the docs themselves, and get them *closer* to the rest of `altair`.
There is definitely more that could be explored on that front, and some of the changes here could be utilised to improve the existing generated docs.

# Related 
- https://github.com/vega/altair/pull/3466

# Tasks
- [x] Parse https://vega.github.io/vega/docs/expressions/
- [x] Generate `py`-equivalent signatures
- [x] Generate docs
	- [x] Find target content
	- [x] Clean up result
[...]

PR #3614, opened 2024-09-26
docs: bar chart with labels coloured by measured luminance

mattijn:
This PR adds an example that demonstrates how to change the color of text labels based on measured luminance. 

The example looks as such:

image

Usage of luminance for the text color of the labels was created using the following code (method-based syntax):

```python
import altair as alt

source = data.barley()

    x=alt.X('sum(yield):Q').stack('zero'),
    y=alt.Y('site:O').sort('-x'),
    text=alt.Text('sum(yield):Q', format='.0f')
)

bars = base.mark_bar(
    tooltip=alt.expr("luminance(scale('color', datum.sum_yield))")
).encode(
    color='sum(yield):Q'
)

text = base.mark_text(
    align='right',
    dx=-3,
    color=alt.expr("luminance(scale('color', datum.sum_yield)) > 0.5 ? 'black' : 'white'")
)

bars + text
```

Especially this line
```python
color=alt.expr("luminance(scale('color', datum.sum_yield)) > 0.5 ? 'black' : 'white'")
```

- The lighter the bar, the higher the luminance. If the bar is light we like the text overlay be black. 
- The darker the bar, the lower the luminance. If the bar is dark, we like the text overlay be white.

In the expression above this is written as a predicate. The text appears black if the luminance is above `0.5` and white when the luminance is below `0.5`. The luminance is computed using the `color` scale in combination with the internal computed data field `datum.sum_yield`. Inspecting the luminance can be done through the tooltip by hovering the bars.

Currently this expression is a `str` with plain Vega Expression syntax. It would be nice to use the equivalent python syntax instead.

### DIFF
--- a/tests/examples_arguments_syntax/bar_chart_with_labels_measured_luminance.py
+++ b/tests/examples_arguments_syntax/bar_chart_with_labels_measured_luminance.py
@@ -1,0 +1,30 @@
+"""
+Bar Chart with Labels based on Measured Luminance
+=================================================
+This example shows a basic horizontal bar chart with labels where the measured luminance to decides if the text overlay is be colored ``black`` or ``white``.
+"""
+# category: bar charts
+import altair as alt
+from vega_datasets import data
+
+source = data.barley()
+
+base = alt.Chart(source).encode(
+    x=alt.X('sum(yield):Q', stack='zero'),
+    y=alt.Y('site:O', sort='-x'),
+    text=alt.Text('sum(yield):Q', format='.0f')
+)
+
+bars = base.mark_bar(
+    tooltip=alt.expr("luminance(scale('color', datum.sum_yield))")
+).encode(
+    color='sum(yield):Q'
+)
+
+text = base.mark_text(
+    align='right',
+    dx=-3,
+    color=alt.expr("luminance(scale('color', datum.sum_yield)) > 0.5 ? 'black' : 'white'")
+)
+
+bars + text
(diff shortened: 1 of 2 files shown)


joelostblom:
Love this example! I have been doing similar things with a manual threshold as I didn't know there was a function in Vega to compute the luminance already!

dangotbanned (approved):
@mattijn I'm truly impressed by how quickly you identified a new example, and this looks great 🎉. 

However, this example has identified a bug with the `python` wrappers, and while it can be worked around - it wouldn't be good content for an example.

Therefore, I'm approving your solution but would ask if you could please open a new issue so that we can track a fix?

# Current non `str` version
```py
import altair as alt

source = data.barley()

    x=alt.X("sum(yield):Q").stack("zero"),
    y=alt.Y("site:O").sort("-x"),
    text=alt.Text("sum(yield):Q", format=".0f"),
)

# NOTE: Original `str` expressions
# tooltip = alt.expr("luminance(scale('color', datum.sum_yield))")
# color = alt.expr("luminance(scale('color', datum.sum_yield)) > 0.5 ? 'black' : 'white'")

luminance = alt.expr.luminance(alt.expr.scale("color", alt.datum.sum_yield))

# BUG: These definitions alone *should* work, but don't evaluate as expressions
tooltip = luminance
color = alt.expr.if_(luminance > 0.5, "black", "white")

# HACK: Actual additional steps required
tooltip = alt.expr(repr(luminance))
color = alt.expr(repr(color))

bars = base.mark_bar(tooltip=tooltip).encode(color="sum(yield):Q")
text = base.mark_text(align="right", dx=-3, color=color)

bars + text
```

---

Sadly #3600 won't resolve this issue, but this did help me identify a different regression on that branch - very grateful for that thank you

PR #3626, opened 2024-10-03
feat: adding observable10 color scheme

apoorvkh:
I saw that Vega supports the [observable10](https://vega.github.io/vega/docs/schemes/#observable10) color scheme, which I think looks quite nice. It was added to Vega in: https://github.com/vega/vega/pull/3843

I am adding this scheme to Altair's list as well.

Since, `alt.Scale(scheme="observable10")` currently raises a SchemaValidationError on `altair==5.4.1`.

### DIFF
--- a/altair/vegalite/v5/schema/_typing.py
+++ b/altair/vegalite/v5/schema/_typing.py
@@ -481,6 +481,7 @@ ColorScheme_T: TypeAlias = Literal[
     "category20",
     "category20b",
     "category20c",
+    "observable10",
     "dark2",
     "paired",
     "pastel1",


apoorvkh:
Ah, looks like that file is automatically generated by: https://github.com/vega/altair/blob/main/tools/generate_schema_wrapper.py

Before I dig further, is there any obvious reason that `observable10` isn't included in the Python schema validation / typing file? Maybe Altair depends on an older version of the vega/vegalite schema?

dangotbanned:
> Ah, looks like that file is automatically generated by: [`main`/tools/generate_schema_wrapper.py](https://github.com/vega/altair/blob/main/tools/generate_schema_wrapper.py?rgh-link-date=2024-10-03T18%3A53%3A33Z)
> 
> Before I dig further, is there any obvious reason that `observable10` isn't included in the Python schema validation / typing file? Maybe Altair depends on an older version of the vega/vegalite schema?

Appreciate the PR @apoorvkh!

So I did a search of https://github.com/vega/altair/blob/main/altair/vegalite/v5/schema/vega-lite-schema.json, but couldn't find any mention of `"observable10"`.
We're currently on [vega-lite/v5.20.1.json](https://github.com/vega/schema/blob/ef61166f3f95154465c4b3ebdca88e3c2d25b005/vega-lite/v5.20.1.json), which isn't the latest version.
However, I didn't any references to that color scheme in the past few releases https://github.com/vega/vega-lite/releases

It might be worth searching for/opening an issue in https://github.com/vega/vega-lite/issues to see if it could be added to the schema?

apoorvkh:
Okay, I am circling back to this: I found "observable10" at this line in `vega-lite-schema.json`. It was added in v5.21.0.

https://github.com/vega/vega-lite/blob/802d661e561068d702459b0b14a15736a6e8ea18/build/vega-lite-schema.json#L4756

It would be great if altair could bump the vega-lite version soon! But I presume that will happen inevitably and this issue will be resolved then, so I will close this issue for now.

Thanks!

PR #3638, opened 2024-10-12
Add upper version bound on VegaFusion

jonmmease:
This PR is looking ahead toward the future release of [VegaFusion 2](https://github.com/vega/vegafusion/discussions/433). It adds an upper version bound in the `all` group, and makes a tweak to how the VegaFusion versions are validated.

I'm hoping there won't be breaking changes wrt to Altair's usage of VegFusion, but want to be safe.

VegaFusion 2 will only have a single Python package named VegaFusion (`vegafusion-python-embed` will be no more). If I can make things backward compatible, I'll patch things so that calling `importlib_version("vegafusion-python-embed")` will return the same version as `importlib_version("vegafusion")` so that our validation passes. To make this possible in the future, I moved the vegafusion import above these checks.

### DIFF
--- a/altair/utils/_importers.py
+++ b/altair/utils/_importers.py
@@ -12,6 +12,8 @@ if TYPE_CHECKING:
 def import_vegafusion() -> ModuleType:
     min_version = "1.5.0"
     try:
+        import vegafusion as vf  # type: ignore
+
         version = importlib_version("vegafusion")
         embed_version = importlib_version("vegafusion-python-embed")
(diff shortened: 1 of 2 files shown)


dangotbanned (approved):
Seems like a good idea to me, thanks @jonmmease
We should probably consider setting upper bounds more often.

I was planning to suggest some changes to fix the `# type: ignore`, but realised that it ended up touching 4 other files.

Looks good to me

PR #3644, opened 2024-10-19
docs: adds info with step size/independent scale

daylinmorgan:
It was not clear from the existing docs how to set individual chart size when using an independent scale with faceted charts and an unequal number of observations.

This adds one more chart example to `user_guide/customization.rst` to show how to vary the size using `alt.Step` to achieve this.

### DIFF
--- a/doc/user_guide/customization.rst
+++ b/doc/user_guide/customization.rst
@@ -687,7 +687,7 @@ the subchart rather than to the overall chart:
        height=100
    )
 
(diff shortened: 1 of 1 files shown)


joelostblom:
Thanks @daylinmorgan ! I agree that this example could be clearer. I changed your suggestion to use the same data as earlier in the same section and cleaned up the page a bit. Will merge when tests are passing

daylinmorgan:
Perfect!

PR #3653, opened 2024-10-22
feat: Support `datetime.(date|datetime)` as a `SchemaBase` parameter

dangotbanned:
Closes #3651

# Example
This is now a working solution for #3643.

No user conversions, or dependency on `pandas`/`polars` required:

<details><summary>Code block</summary>
<p>

```py
from datetime import date

import altair as alt
from vega_datasets import data

source = data.sp500.url

# Create a brush (interval) selection with initial range
brush = alt.selection_interval(
    encodings=["x"], value={"x": (date(2005, 1, 1), date(2009, 1, 1))}
)

base = (
    alt.Chart(source, width=600, height=200).mark_area().encode(x="date:T", y="price:Q")
)
upper = base.encode(alt.X("date:T").scale(domain=brush))
lower = base.properties(height=60).add_params(brush)
chart = upper & lower
chart
```

</p>
</details> 

<details><summary>Screenshot</summary>
<p>

image

</p>
</details> 

<details><summary>Schema output</summary>
<p>

```py
{'config': {'view': {'continuousWidth': 300, 'continuousHeight': 300}},
 'vconcat': [{'mark': {'type': 'area'},
   'encoding': {'x': {'field': 'date',
     'scale': {'domain': {'param': 'param_9'}},
     'type': 'temporal'},
    'y': {'field': 'price', 'type': 'quantitative'}},
   'height': 200,
   'width': 600},
  {'mark': {'type': 'area'},
   'encoding': {'x': {'field': 'date', 'type': 'temporal'},
    'y': {'field': 'price', 'type': 'quantitative'}},
   'height': 60,
   'name': 'view_3',
   'width': 600}],
 'data': {'url': 'https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/sp500.csv'},
 'params': [{'name': 'param_9',
   'select': {'type': 'interval', 'encodings': ['x']},
   'value': {'x': [{'year': 2005, 'month': 1, 'date': 1},
     {'year': 2009, 'month': 1, 'date': 1}]},
   'views': ['view_3']}],
 '$schema': 'https://vega.github.io/schema/vega-lite/v5.20.1.json'}
```

</p>
</details> 

# Typing Examples
Utilizing #3656, and making some progress towards resolving #3643 - these functions now have much more useful annotations.

This wasn't straightforward to piece together, but the types (*and their docs*) are informed by these sources:
- https://vega.github.io/vega-lite/docs/selection.html#current-limitations
[...]

PR #3655, opened 2024-10-23
fix(typing): Revise generated annotation order

dangotbanned:
# Description
The doc stated that `None` would appear after builtins. 

https://github.com/vega/altair/blob/2ba89d508f3eac2d0bfaea53dc74a13c105a0452/tools/schemapi/utils.py#L855-L864

However, this assertion doesn't hold for any `builtin` with a name length greater than 4.

The issue is not new, but became more apparent to me while writing https://github.com/vega/altair/pull/3653#discussion_r1810884346

# Example
Below `float` appears after `None`

## Before
https://github.com/vega/altair/blob/2ba89d508f3eac2d0bfaea53dc74a13c105a0452/altair/vegalite/v5/schema/_config.py#L5004

## After
https://github.com/vega/altair/blob/e656661e6abe0c84dc548289f62b38dfc7663091/altair/vegalite/v5/schema/_config.py#L5004

# Note
I was unable to work out a sorting strategy to achieve what the doc had intended.

Instead, I've opted to always place `None` **last in the context it appears**.
This is much easier to do, and is generally what most people would write (*and expect to see*).

In the non-generated code, this is already the pattern we use:

https://github.com/vega/altair/blob/2ba89d508f3eac2d0bfaea53dc74a13c105a0452/altair/vegalite/v5/api.py#L2091

https://github.com/vega/altair/blob/2ba89d508f3eac2d0bfaea53dc74a13c105a0452/altair/vegalite/v5/api.py#L3494

https://github.com/vega/altair/blob/2ba89d508f3eac2d0bfaea53dc74a13c105a0452/altair/vegalite/v5/api.py#L3859

## Preventing regressions
I've started small, but have included a few doctests (https://github.com/vega/altair/pull/3655/commits/f9f0a53fd1f5ca8bfc7a3c8b10a01a8be6e325f5) that are now run alongside (https://github.com/vega/altair/pull/3655/commits/20d5c17e8862d8711d12630d34dda98217c76728) those in `altair`, `tests`.

### DIFF
--- a/altair/vegalite/v5/schema/_config.py
+++ b/altair/vegalite/v5/schema/_config.py
@@ -515,7 +515,7 @@ class AreaConfigKwds(TypedDict, total=False):
     dy: float
     ellipsis: str
     endAngle: float
-    fill: None | ColorHex | LinearGradientKwds | RadialGradientKwds | ColorName_T
+    fill: ColorHex | LinearGradientKwds | RadialGradientKwds | ColorName_T | None
     fillOpacity: float
     filled: bool
     font: str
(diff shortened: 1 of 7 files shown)


dangotbanned:
Pinging @binste so that you see this when you're available again.

Happy to revert later if you have any objections 👍

PR #3659, opened 2024-10-26
refactor: Simplify `channels.py` overloads

dangotbanned:
This PR significantly reduces the number of `@overload` definitions for channel property setters.
At the same time, it makes many of them **more permissive** (`Sequence` vs `list`) and adds support for `(datetime|date)`.

Supporting `(datetime|date)` was my original aim, following #3653. 
However, once I had a better understanding of the prior [`codegen.py`](https://github.com/vega/altair/pull/3659/files#diff-4ed6d13a6ff1bb029ccb8835ed2aed8e3d790f817e42629850b42a59e97df468) these changes came together.

# Tasks
- [x] Core reimplementation [commits](https://github.com/vega/altair/pull/3659#commits-pushed-bdb2423)
- [x] Add test suite for typing/schema validation parity
- [x] Correctly generate positional/keyword-only signatures (https://github.com/vega/altair/pull/3659/commits/898142b9c8faa3620a618fac0904b184089f89f1)
- [x] Improve docs for new additions to `tools/*`
	- [x] `.schemapi.utils` [commits](https://github.com/vega/altair/pull/3659#commits-pushed-5a72c43)
	- [x] `.schemapi.codegen` [commit](https://github.com/vega/altair/pull/3659/commits/971b39fcbbae67d80a5b1e8926c4f153dbc4150f)
- [x] Pick a solution for solving new typing warnings in examples ([thread](https://github.com/vega/altair/pull/3659#discussion_r1818159550))
	- https://github.com/vega/altair/pull/3659#commits-pushed-426bc29

# Deferred
- Replacing 45x `list[core.ConditionalValueDef...]`
	- `api._Conditions` would be the equivalent

# Related
- https://github.com/vega/altair/pull/2795


tests/examples_methods_syntax/diverging_stacked_bar_chart.py
```diff
@@ -262,9 +262,9 @@ def compute_percentages(
 
 y_axis = alt.Axis(title="Question", offset=5, ticks=False, minExtent=60, domain=False)
 
-alt.Chart(source).mark_bar().encode(
+alt.Chart(source).mark_bar().encode( # pyright: ignore[reportArgumentType]
```
dangotbanned, line 265: ### Note
- Remove this comment before merging
- It isn't related, would have been needed regardless of overloads

tests/examples_methods_syntax/scatter_marginal_hist.py
```diff
@@ -17,8 +17,8 @@
 yscale = alt.Scale(domain=(1.9, 4.55))
 
 points = base.mark_circle().encode(
-    alt.X("sepalLength").scale(xscale),
-    alt.Y("sepalWidth").scale(yscale),
+    alt.X("sepalLength").scale(xscale),  # type: ignore[call-overload]
+    alt.Y("sepalWidth").scale(yscale), # type: ignore[call-overload]
```
dangotbanned, line 21: Using this thread as a catch-all for every other new `examples_methods_syntax` ignore comment.

dangotbanned: # Why the new warnings?
Introducing the `*` keyword-only marker highlights that these examples were only passing type checking via a quite strange soundness hole.

Let's look at the `@overload`s for `alt.X.scale`:
```py
@overload
def scale(self, _: None, /) -> X: ...
@overload
def scale(
    self,
    *,
    align: Optional[float | Parameter | SchemaBase | Map] = Undefined,
    base: Optional[float | Parameter | SchemaBase | Map] = Undefined,
...
```
Previously, a constructed `alt.Scale` was being passed to `alt.X.scale(align=...)`.
This *"passes"* the check since `align` accepts any `SchemaBase` type.

However, now that this `@overload` states you cannot use `align` without a keyword - the previous *"match"* no longer works.

# But why does the example still work?
Here it is switching between 1 positional, or many keywords

https://github.com/vega/altair/blob/6beec7bf0819d961d3d32823d7e658429c306b97/tools/schemapi/schemapi.py#L1602-L1606

The right hand side below uses an unpacked `dict` for the *technically* correct behavior:

<p>

```py
import altair as alt

color = alt.Color("team:N")
scale = alt.Scale(
    domain=["Germany", "Mexico", "South Korea", "Sweden"],
    range=["#000000", "#127153", "#C91A3C", "#0C71AB"],
)
color.scale(scale), color.scale(**scale.to_dict())
```

</p>
</details> 

image

dangotbanned: @mattijn tagging you as I've seen that you peeked at this PR early 😉 

I hope I've explained the issue above clearly. 
If not, feel free to ask for any clarifications.

I can see **2** quite different solutions to this problem and was curious if you had any thoughts?

# 1. Add to the `_` overloads
For the exact behavior that was previously taking place, it would look like:

```py
@overload
def scale(self, _: SchemaBase | None, /) -> X: ...
```

Since all of these examples are only for `.axis()` and `.scale()`, a slightly more specific solution might be:

```py
@overload
def axis(self, _: Axis | None, /) -> X: ...
@overload
def scale(self, _: Scale | None, /) -> X: ...
```

It wouldn't be as clean though for all cases.
E.g. `.aggregate()` -> `Aggregate` provides a meaningless wrapper:
```py
class Aggregate(VegaLiteSchema):
    """Aggregate schema wrapper."""

    _schema = {"$ref": "#/definitions/Aggregate"}

    def __init__(self, *args, **kwds):
        super().__init__(*args, **kwds)
```

# 2. Rewrite the examples
I don't think most of these actually need to be written in the way that they are.

For example [us_population_pyramid_over_time](https://altair-viz.github.io/gallery/us_population_pyramid_over_time.html) 

After formatting, it is much clearer that there is *a lot of repetition* that can be reduced.
:

<details><summary>Reformatted code block</summary>
<p>

```py
import altair as alt

source = data.population.url

slider = alt.binding_range(min=1850, max=2000, step=10)
select_year = alt.selection_point(name="year", fields=["year"], bind=slider, value=2000)

base = (
    alt.Chart(source)
    .add_params(select_year)
    .transform_filter(select_year)
    .transform_calculate(gender=alt.expr.if_(alt.datum.sex == 1, "Male", "Female"))
    .properties(width=250)
)


color_scale = alt.Scale(domain=["Male", "Female"], range=["#1f77b4", "#e377c2"])

left = (
    base.transform_filter(alt.datum.gender == "Female")
    .encode(
        y=alt.Y("age:O").axis(None),
        x=alt.X("sum(people):Q").title("population").sort("descending"),
        color=alt.Color("gender:N").scale(color_scale).legend(None),  # type: ignore[call-overload]
    )
    .mark_bar()
    .properties(title="Female")
)

middle = (
    base.encode(alt.Y("age:O").axis(None), alt.Text("age:Q"))
    .mark_text()
    .properties(width=20)
)

right = (
    base.transform_filter(alt.datum.gender == "Male")
    .encode(
        y=alt.Y("age:O").axis(None),
        x=alt.X("sum(people):Q").title("population"),
        color=alt.Color("gender:N").scale(color_scale).legend(None),  # type: ignore[call-overload]
    )
    .mark_bar()
    .properties(title="Male")
)

alt.concat(left, middle, right, spacing=5)
```

</p>
</details> 


Defining `color_scale` differently is only a small part of what I'd change personally.

Also, after reading https://github.com/vega/altair/pull/2795#issuecomment-1369692487 - I got the feeling that these examples aren't really using the feature as intended.

mattijn:
Reducing code lines while still passing tests is always something to cheer for👍

tests/examples_methods_syntax/scatter_marginal_hist.py
dangotbanned: @mattijn hoping that this demonstrates the *kind* of changes I was proposing in [**2**](https://github.com/vega/altair/pull/3659#discussion_r1818172455)
- [docs(DRAFT): Rewrite `us_population_pyramid_over_time` to fix warning](https://github.com/vega/altair/pull/3659/commits/0a26f276e4d532a80a2461af4d00d8f8da920265)


Either way, I'm planning to revert it on this PR - but thought the diff (including the removal of ignore comments) would be helpful.

If we went with that option, I could either do a PR per example or all of them in one.
Probably would also need to update any *attribute syntax* equivalents - since the structure changes quite a bit.

tests/vegalite/v5/schema/test_channels.py
```diff
@@ -0,0 +1,82 @@
...
+    # - `"ascending"` is silently ignored when positional
+    # - Caught as invalid statically, but not at runtime
+    bad = angle.sort("x", "ascending").to_dict()  # type: ignore[call-overload]
+    good = angle.sort(encoding="x", order="ascending").to_dict()
+    assert isinstance(bad, dict)
+    assert isinstance(good, dict)
+    with pytest.raises(
+        AssertionError, match=r"'x' == {'encoding': 'x', 'order': 'ascending'}"
+    ):
+        assert bad["sort"] == good["sort"]
```
dangotbanned, line 82: # Updated context

The bug I found here was that `_PropertySetter.__call__`  doesn't align with `SchemaBase.__init__`:

https://github.com/vega/altair/blob/6beec7bf0819d961d3d32823d7e658429c306b97/tools/schemapi/schemapi.py#L1602-L1606

Above, we silently drop any positional args *after* the first.

Below, we **raise** if there are more than 1 positional args

https://github.com/vega/altair/blob/6beec7bf0819d961d3d32823d7e658429c306b97/tools/schemapi/schemapi.py#L1018-L1038

# Solution?
I'd recommend fixing this in another PR, since it may have been a source for unexpected behavior since `5.0.0`
- https://github.com/vega/altair/blame/c273c0a031cf1bfdfc9f8ea18ebeab9399581363/tools/schemapi/schemapi.py#L1097

tests/examples_methods_syntax/scatter_marginal_hist.py
mattijn: Hm. If I understand correctly. The issue is that when defining a scale or axis upfront and use that within the method-based syntax the types are not happy anymore with this PR? 

```python
import polars as pl

source = pl.DataFrame(data={'gender':['M', 'M', 'F', 'F', 'F']})

# define your scale
color_scale = alt.Scale(domain=['M', 'F'], range=['#1FC3AA', '#8624F5'])

# add it in the chart
    y='gender', 
    x='count()', 
    color=alt.Color('gender').scale(color_scale)
)
```
image

Honestly, I quite like the style to be able to define these upfront and just pass them and it works. Especially if you have a complex chart where the same scale is being reused multiple times.

Maybe it was not intended originally, but always have been happy that this works too. It would be a pity if the type-folks will get a type-warning or type-error for this style. 

So then I prefer option 1 I guess?

dangotbanned: @mattijn so this is on `main`, demonstrating what a type checker thinks is happening: 

https://github.com/user-attachments/assets/112e2abf-467f-436e-a81b-9da0fd0e5d51

As a result, clearly obvious errors like this are not flagged:

image

mattijn: We send the type checker into the bush there :). Probably better if other people respond here. I liked this option, but maybe not type-proof🤷

dangotbanned: Resolved with docs update as of [`9cf9458` (#3659)](https://github.com/vega/altair/pull/3659/commits/9cf9458b7a4220ce02590e35cb6b187b9019795f)

tests/vegalite/v5/schema/test_channels.py
dangotbanned: Marking as resolved, tracking in:
- https://github.com/vega/altair/issues/3661

mattijn (approved):

dangotbanned:
Thanks for reviewing @mattijn!

Issue #3643, opened 2024-10-19
Clarify usage of initial values in selection intervals with dates


dsmedia:
### What is your suggestion?

When setting initial values for a selection interval based on date values, users must manually convert dates to timestamps, as discussed [here](https://stackoverflow.com/a/61470024/25835793). This requirement does not appear to be clearly documented or exemplified in this repository. Solutions could be one or more of the following:

#### Potential Solutions
- [ ] **Example Gallery:** Update [this](https://altair-viz.github.io/gallery/interval_selection.html) example to include an initial value for the interval (see below)
- [ ] **API Reference:** Briefly reference in the description of the value paramater in the API Reference [here](https://altair-viz.github.io/user_guide/generated/api/altair.selection_interval.html)
- [ ] **Docs:** Incorporate into the [docs](https://altair-viz.github.io/user_guide/interactions.html#selections-capturing-chart-interactions) on selections

#### Questions:
1. Which, if any, of these suggestions should be implemented?
2. The linked stackoverflow [discussion](https://stackoverflow.com/a/61470024/25835793) above highlights an important distinction between datetime.datetime and numpy.datetime64 in this context. Should this point be referenced in any documentation, and if so, how?

#### Sample diff for current example gallery [item](https://altair-viz.github.io/gallery/interval_selection.html)

``` diff
import altair as alt
from vega_datasets import data

source = data.sp500.url

+ # Define an initial date range as timestamps
+ x_init = pd.to_datetime(['2005-01-01', '2009-01-01']).astype(int) / 1E6

+# Create a brush (interval) selection with initial range
+brush = alt.selection_interval(
+    encodings=['x'], 
+   value={'x': list(x_init)}  # Initialize selection
+)

brush = alt.selection_interval(encodings=['x'])

base = alt.Chart(source, width=600, height=200).mark_area().encode(
    x = 'date:T',
    y = 'price:Q'
)

upper = base.encode(
    alt.X('date:T').scale(domain=brush)
)

lower = base.properties(
    height=60
).add_params(brush)

upper & lower
```

### Have you considered any alternative solutions?

I've proposed three possible solutions above.

dangotbanned:
Thanks for raising this @dsmedia 

> When setting initial values for a selection interval based on date values, users must manually convert dates to timestamps, as discussed [here](https://stackoverflow.com/a/61470024/25835793). This requirement does not appear to be clearly documented or exemplified in this repository. 

[...]

PR #3667, opened 2024-10-31
docs: Add initial date range to interval selection example

dsmedia:
This is a partial implementation of #3643, demonstrating how to set initial date ranges in interval selections through a concrete example. This improvement was made possible by the recent datetime support added in #3653 and #3662 by @dangotbanned.

Rather than creating a separate example, I opted to modify the existing interval selection example because:

1. Setting initial ranges is a core feature of interval selections, not a separate use case. Including it in the primary example helps users discover this capability alongside basic usage.

2. The change adds minimal complexity while showing the idiomatic way to set initial values using native datetime objects.

3. A single progressive example provides a clearer learning path than splitting this into two separate examples.

Changes made:
- Added demonstration of initial date range setting using datetime objects
- Updated example title and description to reflect the addition
- Added clarifying comments about the selection range

Remaining work in #3643:
- Add information about initial ranges to the selection documentation
- Consider documenting datetime handling in times_and_dates section

### DIFF
--- a/tests/examples_arguments_syntax/interval_selection.py
+++ b/tests/examples_arguments_syntax/interval_selection.py
@@ -1,6 +1,6 @@
 """
-Interval Selection with Initial Selection
-=========================================
+Interval Selection with Initial Date Range
+==========================================
 
 This is an example of creating a stacked chart for which the domain of the
 top chart can be selected by interacting with the bottom chart. The initial
--- a/tests/examples_arguments_syntax/interval_selection.py
+++ b/tests/examples_arguments_syntax/interval_selection.py
@@ -3,15 +3,20 @@ Interval Selection
 ==================
 
 This is an example of creating a stacked chart for which the domain of the
-top chart can be selected by interacting with the bottom chart.
+top chart can be selected by interacting with the bottom chart. The initial
+selection range is set using Python's native datetime objects.
 """
 # category: area charts
 import altair as alt
 from vega_datasets import data
+import datetime as dt
 
 source = data.sp500.url
 
-brush = alt.selection_interval(encodings=['x'])
+date_range = (dt.date(2007, 6, 30), dt.date(2009, 6, 30))
+
+brush = alt.selection_interval(encodings=['x'],
+                               value={'x': date_range})
 
 base = alt.Chart(source).mark_area().encode(
     x = 'date:T',
--- a/tests/examples_arguments_syntax/interval_selection.py
+++ b/tests/examples_arguments_syntax/interval_selection.py
@@ -1,6 +1,6 @@
 """
-Interval Selection
-==================
+Interval Selection with Initial Selection
+=========================================
 
 This is an example of creating a stacked chart for which the domain of the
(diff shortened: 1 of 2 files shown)


dangotbanned:
@dsmedia thanks for opening this PR!

> Rather than creating a separate example, I opted to modify the existing interval selection example because:
> 
> 1. Setting initial ranges is a core feature of interval selections, not a separate use case. Including it in the primary example helps users discover this capability alongside basic usage.
> 2. The change adds minimal complexity while showing the idiomatic way to set initial values using native datetime objects.
> 3. A single progressive example provides a clearer learning path than splitting this into two separate examples.

I agree with these points. 
It makes more sense to me now that setting the value is simpler than the [stackoverflow](https://stackoverflow.com/questions/61469922/how-to-have-init-values-for-a-selection-interval-in-altair-when-using-dates/61470024#61470024) solution.

---

One question, do you think this example would make more sense in a different category?
I've noticed these two which have some similarities:

- https://altair-viz.github.io/gallery/select_mark_area.html
- https://altair-viz.github.io/gallery/interval_selection_map_quakes.html

But they are catgorised as [interactive charts](https://altair-viz.github.io/gallery/index.html#interactive-charts). 
This one is under [area charts](https://altair-viz.github.io/gallery/index.html#area-charts)

There is also [VL-interactive_overview_detail](https://vega.github.io/vega-lite/examples/interactive_overview_detail.html)

I know that this wasn't a change that you introduced, but I think it would further improve discoverability of this functionality

tests/examples_arguments_syntax/interval_selection.py
```diff
@@ -1,17 +1,22 @@
...
-==================
+Interval Selection with Initial Date Range
+==========================================
 
 This is an example of creating a stacked chart for which the domain of the
-top chart can be selected by interacting with the bottom chart.
+top chart can be selected by interacting with the bottom chart. The initial
+selection range is set using Python's native datetime objects.
 """
 # category: area charts
```
dangotbanned, line 9:
```suggestion
# category: interactive charts
```

tests/examples_methods_syntax/interval_selection.py
```diff
@@ -1,17 +1,22 @@
...
-==================
+Interval Selection with Initial Date Range
+==========================================
 
 This is an example of creating a stacked chart for which the domain of the
-top chart can be selected by interacting with the bottom chart.
+top chart can be selected by interacting with the bottom chart. The initial
+selection range is set using Python's native datetime objects.
 """
 # category: area charts
```
dangotbanned, line 9:
```suggestion
# category: interactive charts
```

dangotbanned (approved):
Thanks again for the PR @dsmedia 

I'm approving now, will leave open for a bit to see if anyone else wants to weigh in on the [suggestions](https://github.com/vega/altair/pull/3667#issuecomment-2450249670)

mattijn:
Applied suggestions from review and activated auto-merge. Thanks @dsmedia! Much appreciated!

mattijn (approved):

PR #3670, opened 2024-11-03
chore: use more robust dtype comparisons, use Narwhals stable API in tests

MarcoGorelli:
Since Narwhals 1.9.0 (release last month), `nw.Datetime` has `time_unit` and `time_zone` metadata, which means that `hash(nw.Datetime)` has changed.
This change only affects the main Narwhals namespace - the `narwhals.stable.v1` namespace (which you're using 🙌 ) is unaffected

The fact that having `narwhals.stable.v1` allowed us to make this change without it affecting any users is a pretty good feeling 😄 

I've rewritten the dtype comparison from
```
elif dtype in {nw.Datetime, nw.Date}:
```
to
```
elif dtype == nw.Datetime or dtype == nw.Date:
```
so that it will also work in the same way if and when you decide to move to `narwhals.stable.v2` (when we get there). There'll be no obligation to perform such a move of course, we'll keep `stable.v1` working as-is, so either way, **there's no impact on users**

---

See the second bullet point at https://narwhals-dev.github.io/narwhals/backcompat/#after-stablev1 for more context

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -709,11 +709,14 @@ def infer_vegalite_type_for_narwhals(
         and not (categories := column.cat.get_categories()).is_empty()
     ):
         return "ordinal", categories.to_list()
-    if dtype in {nw.String, nw.Categorical, nw.Boolean}:
+    if dtype == nw.String or dtype == nw.Categorical or dtype == nw.Boolean:  # noqa: PLR1714
         return "nominal"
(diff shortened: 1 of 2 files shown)


dangotbanned (approved):
Nothing to add here at all, your explanation was super helpful

Thanks for this @MarcoGorelli

PR #3680, opened 2024-11-12
chore: Prep for VegaFusion 2.0

jonmmease:
Small update that supersedes https://github.com/vega/altair/pull/3638.  VegaFusion 2.0 isn't going to break any of the functionality that Altair depends on, so the only update needed is to not check for the `vegafusion-python-embed` package when the `vegafusion` package is >2 (becuase the functionality this is being merged into the vegafusion package).

In terms of sequencing, I want this update to be released in Altair before I publish VegaFusion 2.0 so that installing `altair[all]` will never pull in a version of VegaFusion that's not compatible with altair.

### DIFF
--- a/altair/utils/_importers.py
+++ b/altair/utils/_importers.py
@@ -15,18 +15,21 @@ def import_vegafusion() -> ModuleType:
         import vegafusion as vf
 
         version = importlib_version("vegafusion")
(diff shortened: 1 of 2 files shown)


mattijn:
Thanks @jonmmease! Looking forward for a new release of Altair and VegaFusion!

Issue #3577, opened 2024-09-06
Switch from `hatch` to `uv` for package and environment management


binste:
### What is your suggestion?

`uv` is now a Python packaging manager https://astral.sh/blog/uv-unified-python-packaging. For a while already it had a lot of useful functionality but now it has all the features we'd need to replace `hatch`. Paraphrasing @mattijn (correct me if I missed anything!), switching to `uv` might help us attract new developers. It's very likely that uv is much more popular than hatch, judging based on GitHub stars but also development activity and feature set. Due to this:

* It makes it easier for new Altair contributors to get started as they might already know it from other projects
* Or if a new contributor has not yet used it, it's likely that they are more interested in learning it

### Have you considered any alternative solutions?

Poetry would be an alternative. uv seems closer to hatch so less of a switching cost and it just has a super fast solver and we have it already installed for the linting and formatting

dangotbanned:
FYI we're already using `uv` for dependencies.

I'd support switching over fully if there was no loss of functionality.

A first step would be trying to replace `pip` and `hatch` in all the GitHub workflows.
They now have proper documentation - which was missing when I tried that out before

joelostblom:
> I'd support switching over fully if there was no loss of functionality.

I'm also on board with this

jonmmease:
I think moving to uv sounds like a fine direction. One thing to throw out there is that I'm using [pixi](https://pixi.sh/latest/) for VegaFusion, and will soon be using it for vl-convert. Pixi is a very similar idea, but has a few advantages for certain kinds of projects.

 - It can pull dependencies from PyPI using en embedded copy of uv, so this is just as fast as plain uv
 - It can also pull dependencies from conda-forge. This is really useful for non-python development dependencies. For example, in VegaFusion I configure Pixi to install Rust, Java, and Node.js from conda-forge so that these don't need to be installed at the system level.
 - Pixi includes a cross-platform shell (based on [Deno's shell](https://github.com/denoland/deno_task_shell)) that it uses to evaluate tasks. This shell includes a bunch of [basic unix commands](https://pixi.sh/latest/features/advanced_tasks/#built-in-commands) like (`cp`, `rm`, `pwd`, `cat`, etc.) and it has access to any CLI dependencies installed from conda-forge.  So you can write tasks as if you're on linux, and they work the same on Windows.
[...]

PR #3719, opened 2024-12-20
ci: Switch from `hatch`, `pip` to `uv` in GitHub Actions

### LINKED ISSUE #3718: Switch from `hatch`, `pip` to `uv` in GitHub Actions
### What is your suggestion?

 _Originally posted by @dangotbanned in [#3577](https://github.com/vega/altair/issues/3577#issuecomment-2334214226)_

A first step would be trying to replace `pip` and `hatch` in all the GitHub workflows.
They now have proper documentation - which was missing [when I tried](https://github.com/vega/altair/pull/3431#issuecomment-2168654915) that out before

_No response_

dangotbanned:
Closes #3718

One step towards #3577

This PR moves to `uv` in each of:
- [`build.yml`](https://github.com/dangotbanned/altair/blob/8d8c95254ca73c37e8e791307cca625071c107ed/.github/workflows/build.yml)
- [`docbuild.yml`](https://github.com/dangotbanned/altair/blob/8d8c95254ca73c37e8e791307cca625071c107ed/.github/workflows/docbuild.yml)
- [`lint.yml`](https://github.com/dangotbanned/altair/blob/8d8c95254ca73c37e8e791307cca625071c107ed/.github/workflows/lint.yml)

I tried this a while back, but gave up on (https://github.com/vega/altair/pull/3719/commits/27999c1b041cd2f685dc05d37cf613eb2dfc4020).
Since then, I've used `uv` some more as I mentioned in (https://github.com/vega/altair/issues/3577#issuecomment-2548280136) - which helped unblock this part 😅


.github/workflows/docbuild.yml
```diff
@@ -2,6 +2,9 @@ name: docbuild
 
 on: [push, pull_request]
 
+env:
+  UV_SYSTEM_PYTHON: 1
```
mattijn, line 6: Great PR @dangotbanned! Do you have a reference why this is environment definition is necessary? Thanks!

mattijn:
Regarding feedback on parallel running of sphinx docs:

```
WARNING: the sphinxext_altair.altairplot extension does not declare if it is safe for parallel reading, assuming it isn't - please ask the extension author to check and make it explicit
```

Is this something worthwhile pursuing for?

.github/workflows/docbuild.yml
dangotbanned: Thanks @mattijn!
Yeah I can explain.

I've enabled [`UV_SYSTEM_PYTHON`](https://docs.astral.sh/uv/configuration/environment/#uv_system_python) as a [workflow environment variable](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-environment-variables-for-a-single-workflow) to add the `--system` argument on all relevant `uv` commands.
You can see in [`narwhals/.github/workflows/downstream_tests.yml`](https://github.com/narwhals-dev/narwhals/blob/7316b2eaa35801559876f42bc911d1285f28c3ef/.github/workflows/downstream_tests.yml) that argument is [used **25** times](https://github.com/search?q=repo%3Anarwhals-dev%2Fnarwhals%20path%3A.github%2Fworkflows%2Fdownstream_tests.yml%20%22--system%22&type=code). 

So this is just to keep it simple, making sure it is there when needed - but isn't something you need to worry about if updating the workflow in the future 🙂

dangotbanned:
> Regarding feedback on parallel running of sphinx docs:
> 
> ```
> WARNING: the sphinxext_altair.altairplot 
> extension does not declare if it 
> is safe for parallel reading, 
> assuming it isn't - please ask 
> the extension author to check 
> and make it explicit
> ```
> 
> Is this something worthwhile pursuing for?

@mattijn 

It has the *potential* to speed this CI job up, but I think I'd struggle with testing [`--jobs`](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-j) locally:

> This feature only works on systems supporting “fork”. Windows is not supported. 

I'm happy to support if anyone else wants to take the lead on it though. 
I just imagine getting it working on native `posix` will be easier than [`wsl`](https://learn.microsoft.com/en-us/windows/wsl/install) or [`MINGW64`](https://www.mingw-w64.org/)

.github/workflows/docbuild.yml
dangotbanned: I really need to transition away from windows at some point 😔

Think it is causing problems in https://github.com/vega/vega-datasets/pull/648#issuecomment-2558086569 as well

mattijn:
Thanks for providing context regarding the environment variable of `uv`!

In fact I'm quite happy that we have maintainers on multiple operating systems as it increase the robustness of the codebase.

PR #514, opened 2018-02-26
Deviation ellipses example

essicolo:
This is an example showing bivariate deviation ellipses of petal length and width of three iris species. In Jupyter lab, it works only if chart_deviation is printed before printing chart_points + chart_deviation (error message below). A more complete example could include a pairs plot including petal and sepal dimensions, with ellipses. Is it feasible? 

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
~/bin/anaconda3/envs/altair-env/lib/python3.6/site-packages/pandas/core/indexes/base.py in get_loc(self, key, method, tolerance)
   2524             try:
-> 2525                 return self._engine.get_loc(key)
   2526             except KeyError:

pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc()

pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc()

pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.PyObjectHashTable.get_item()

pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.PyObjectHashTable.get_item()

KeyError: 'index'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
~/bin/anaconda3/envs/altair-env/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj, include, exclude)
    968 
    969             if method is not None:
--> 970                 return method(include=include, exclude=exclude)
    971             return None
    972         else:

~/Documents/GitHub/altair/altair/vegalite/v2/api.py in _repr_mimebundle_(self, include, exclude)
     56     def _repr_mimebundle_(self, include, exclude):
     57         """Return a MIME bundle for display in Jupyter frontends."""
---> 58         return renderers.get()(self.to_dict())
     59 
     60 

~/Documents/GitHub/altair/altair/vegalite/v2/api.py in to_dict(self, *args, **kwargs)
     39             context['data'] = original_data
     40         kwargs['context'] = context
---> 41         return super(TopLevelMixin, self).to_dict(*args, **kwargs)
     42 
     43     # Layering and stacking

~/Documents/GitHub/altair/altair/utils/schemapi.py in to_dict(self, validate, ignore, context)
    143             result = _todict(self._args[0])
    144         elif not self._args:
--> 145             result = _todict({k: v for k, v in self._kwds.items()
    146                               if k not in ignore})
    147         else:

[...]

PR #3631, opened 2024-10-04
feat(RFC): Adds `altair.datasets`

dangotbanned:
# Related
- https://github.com/vega/altair/discussions/3150
- https://github.com/narwhals-dev/narwhals/issues/876#issuecomment-2323036762

# Tracking
Waiting on the next `vega-datasets` release.
~~Once there is a stable [`datapackage.json`](https://github.com/vega/vega-datasets/blob/369b462f7505e4ef3454668793e001e3620861ff/datapackage.json) available - there is quite a lot of [`tools/datasets`](https://github.com/vega/altair/tree/6080116447590f4c036275ab8c144b1c1b18f57a/tools/datasets) that can be simplified/removed.~~
- [x] https://github.com/vega/vega-datasets/issues/654
- [x] [feat: Update for `vega-datasets@3.0.0-alpha.1`](https://github.com/vega/altair/pull/3631/commits/51a967aef63e6c934b1b227cc9214776c3a5c699)

Discovered a bug that makes some handling of expressions a little less efficient.
- [x] https://github.com/narwhals-dev/narwhals/issues/1897

Upstreaming some `nw.Schema` stuff to `narwhals`
- [x] https://github.com/narwhals-dev/narwhals/issues/1912
- [x] https://github.com/vega/altair/pull/3631#discussion_r1937953187

Improve user-facing interface
- [ ] https://github.com/vega/altair/pull/3631#issuecomment-2493064370

# Description
Providing a minimal, but up-to-date source for https://github.com/vega/vega-datasets.

This PR takes a different approach to that of https://github.com/altair-viz/vega_datasets, notably:
- No datasets are included in the package
	- Instead, [several metadata files](https://github.com/vega/altair/tree/2bd89aaf6848fad1ef116af03c24e44a38db8667/altair/datasets/_metadata) form a dense summary of [`vega-datasets/datapackage.json`](https://github.com/vega/vega-datasets/blob/06a3ade07259101587dafa3f6f8a862afd217093/datapackage.json)
	- 3 files provide a reduction (*70-150kb* -> *15kb*) and optimized views for this particular use-case
	- Includes redundancies for missing dependencies
- Strong support for typing
	- Annotations are generated from the metadata itself
	- https://github.com/vega/altair/blob/9e9deeb95668d2c4e7d30311e85a8f9f6acdc88c/altair/datasets/_typing.py
- So far, **4** backends have been implemented, instead of only `pandas`
[...]

PR #3747, opened 2025-01-04
docs(example): Adds Confidence Interval Ellipses

### LINKED ISSUE #3715: Support/document Confidence Interval Ellipse
### What is your suggestion?

Hello author, can Altair add confidence interval ellipses for different groups of points in the plot after drawing PCA scatter plot?
The following is a PCA scatter plot drawn in R language ggplot2.
Image
Here is the PCA scatter plot I am drawing now with altair and the implementation code.
Image
```
chart = alt.Chart(data).mark_circle(size=60).encode(
    x=alt.X(pc1_column, title='Principal Component 1'),
    y=alt.Y(pc2_column, title='Principal Component 2'),
    color=alt.Color('Environment1', title='Environment'), 
#    tooltip=['Index', pc1_column, pc2_column]
)
chart.display()
```
Is there any solution at the moment? thx.

Image
That was one of my attempt, but it obviously didn't work.
I don't have any good ideas right now.
```
kmeans = KMeans(n_clusters=3, random_state=0).fit(data[[pc1_column, pc2_column]])
data['Cluster'] = kmeans.labels_

chart = alt.Chart(data).mark_circle(size=60).encode(
    x=alt.X(pc1_column, title='Principal Component 1'),
    y=alt.Y(pc2_column, title='Environment'),
    color=alt.Color('Environment1', title='Environment'),
    tooltip=[pc1_column, pc2_column, 'Environment1']
)

def plot_confidence_ellipses_altair(df, chart):
    for cluster in df['Cluster'].unique():
        cluster_data = df[df['Cluster'] == cluster]
        mean_x = cluster_data[pc1_column].mean()
        mean_y = cluster_data[pc2_column].mean()
        cov_matrix = np.cov(cluster_data[pc1_column], cluster_data[pc2_column])
        eigvals, eigvecs = np.linalg.eigh(cov_matrix)
        order = eigvals.argsort()[::-1]
        eigvals, eigvecs = eigvals[order], eigvecs[:, order]
        theta = np.degrees(np.arctan2(*eigvecs[:, 0][::-1]))
        width, height = 2 * np.sqrt(chi2.ppf(0.95, 2)) * np.sqrt(eigvals)

        ellipse_df = pd.DataFrame({
            'x': mean_x + width / 2 * np.cos(np.linspace(0, 2 * np.pi, 100)) * np.cos(np.radians(theta)) -
                 height / 2 * np.sin(np.linspace(0, 2 * np.pi, 100)) * np.sin(np.radians(theta)),
            'y': mean_y + width / 2 * np.sin(np.linspace(0, 2 * np.pi, 100)) * np.cos(np.radians(theta)) +
                 height / 2 * np.cos(np.linspace(0, 2 * np.pi, 100)) * np.sin(np.radians(theta))
        })

        chart += alt.Chart(ellipse_df).mark_line(color='red', opacity=0.5, strokeWidth=2).encode(
            x=alt.X('x:Q', title='Principal Component 1'),
            y=alt.Y('y:Q', title='Environment')
        )

    return chart

chart = plot_confidence_ellipses_altair(data, chart)
chart.display()
```

dangotbanned:
Will close #3715

# Description
Adds an example inspired by ([`ggplot2`](https://ggplot2.tidyverse.org/reference/stat_ellipse.html#ref-examples)|[`plotnine`](https://plotnine.org/reference/stat_ellipse.html#plotnine.stat_ellipse))`.stat_ellipse()`.

As can be seen in the [first commit](https://github.com/vega/altair/commit/a22e8dcb64c7136bc346e638d1f6923c60cbbaf0), this PR began by **rebasing a closed PR** from [*almost* 7 years ago](https://github.com/vega/altair/pull/514#issue-300358957).
- #514

Relevant info from (https://github.com/vega/altair/issues/3715#issuecomment-2571334087):

> I believe [`plotnine.stat_ellipse`](https://plotnine.org/reference/stat_ellipse.html#plotnine.stat_ellipse) would be an example of implementing this with `numpy`, `scipy`.
> [Source code](https://github.com/has2k1/plotnine/blob/250ae83bc9bdd9ec32ccded781ff929e489d5229/plotnine/stats/stat_ellipse.py#L19-L256)
> 
> I also found an old closed PR ([#514](https://github.com/vega/altair/pull/514) by [@essicolo](https://github.com/essicolo?rgh-link-date=2025-01-04T15%3A29%3A00.000Z)) that _would have_ added an example for this.
> The [blocker at the time](https://github.com/vega/altair/pull/514#issuecomment-368637430) is no longer an issue as ([#3202](https://github.com/vega/altair/pull/3202) by [@joelostblom](https://github.com/joelostblom?rgh-link-date=2025-01-04T15%3A29%3A00.000Z)) added `scipy` as a docs dependency.

# Example
image

# Tasks
- [x] Fix `sphinxext.altairgallery` parsing ([CI Run](https://github.com/vega/altair/actions/runs/12612637008/job/35149593128?pr=3747#step:6:25))
  - (https://github.com/vega/altair/pull/3747/commits/4944b80d2ed26be6831f28a00410314107c51ff9)
  - (https://github.com/vega/altair/pull/3747/commits/7cd2a775566891d5be861252ac0f87b351ac6351)
- [x] Finish tidying up from (https://github.com/vega/altair/pull/3747/commits/1983ede5160d5f71d374066f3f59645261545e18)
  - [x] [`numpy` and `scipy`](https://github.com/vega/altair/pull/3747#commits-pushed-1623629)
  - [x] `pandas` (see [thread](https://github.com/vega/altair/pull/3747#discussion_r1903144682))
- [x] Add `examples_methods_syntax` version
  - Only `alt.(X|Y)` changed to chain `.scale(zero=False)`
- [x] [docs: rename to 'Confidence Interval Ellipses'](https://github.com/vega/altair/pull/3747/commits/bd4e30f42dfafe35c79e16698832e13bb65f5899)
- [x] Provide more detail in the description (https://github.com/vega/altair/pull/3747/commits/62927af736569f7ad906a5106a976ce110c86983)
  - Reference [`plotnine.stat_ellipse`](https://plotnine.org/reference/stat_ellipse.html#plotnine.stat_ellipse)
  - Attributing author of original PR **Serge-Étienne Parent** (@essicolo)

# Future Work
I think a more generalized version of this would be a good fit for https://github.com/vega/altair_ally.
An issue might be the `scipy` dependency, which I *really* was hoping to be able to avoid here.
The [dendrogram](https://github.com/vega/altair/blob/be5e9ecd61d099c847dca44fb3a4283940c1e3b8/tests/examples_arguments_syntax/dendrogram.py) example shows some kind of *inlining* from `scipy` - but I have no idea if that is possible for:
- https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.rv_continuous.ppf.html#scipy.stats.rv_continuous.ppf
- https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.f.html#scipy-stats-f


dangotbanned:
Cannot express how relieved I am to see the CI *finally* green 😅
https://github.com/vega/altair/pull/3747/commits/be087d25cb46d3a0ea459ab5ee69c5b78add4d40

tests/examples_arguments_syntax/deviation_ellipses.py
```diff
@@ -0,0 +1,89 @@
...
+    cols = col_x, col_y
+    groups = []
+    # TODO: Rewrite in a more readable way
+    categories = df[col_group].unique()
+    for category in categories:
+        sliced = df.loc[df[col_group] == category, cols]
+        ell_df = pd.DataFrame(np_ellipse(sliced.to_numpy()), columns=cols) # type: ignore
+        ell_df[col_group] = category
+        groups.append(ell_df)
+    return pd.concat(groups).reset_index()
```
dangotbanned, line 69: ## TODO
- Figure out a more ergonomic way of applying the function to each group

tests/examples_arguments_syntax/deviation_ellipses.py
```diff
@@ -0,0 +1,89 @@
...
+    arr: np.ndarray[tuple[int, int], np.dtype[np.float64]],
+    conf_level: float = 0.95,
+    method: Literal["deviation", "error"] = "deviation",
+    segments: int = 100,
+):
+    """
+    Arguments:
+        arr:  2D numpy array with 2 columns
+        level: confidence level
+        method: either 'deviation' (swarning data) or 'error (swarning the mean)'
```
dangotbanned, line 29: ## TODO
- Find out what was meant by *swarning data* and *swarning the mean*

tests/examples_arguments_syntax/deviation_ellipses.py
dangotbanned: @MarcoGorelli not an urgent one.

Do you know of a *more idiomatic* way to write this `pandas` code?

Based on [this](https://github.com/vega/altair/pull/514) from **7 years ago**:

https://github.com/vega/altair/blob/a22e8dcb64c7136bc346e638d1f6923c60cbbaf0/tests/examples_arguments_syntax/deviation_ellipses.py#L37-L45

---

Personally I'd rather use `polars`, but the `pandas` dependency is already there due to https://github.com/altair-viz/vega_datasets

dangotbanned: ### `polars` version
Could probably be reduced a bit further.

- Using [`pl.DataFrame.partition_by`](https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.partition_by.html) *works*
	- But needing to handle `dict[tuple[str, ...], pl.DataFrame]` for a single key seems like a code smell

```py
def pl_ellipse(
    df: pl.DataFrame, col_x: str, col_y: str, col_group: str
) -> pl.DataFrame:
    parts = df.select(col_x, col_y, col_group).partition_by(
        col_group, as_dict=True, include_key=False
    )
    return pl.concat(
        pl.DataFrame(np_ellipse(group.to_numpy()), [col_x, col_y])
        .with_columns(pl.lit(k[0]).alias(col_group))
        .with_row_index()
        for k, group in parts.items()
    )
```

tests/examples_arguments_syntax/deviation_ellipses.py
dangotbanned: No longer relevant, following (https://github.com/vega/altair/pull/3747/commits/dcb9fa565adb28a06938a3d8fe6be9810d90ffb3)

tests/examples_arguments_syntax/deviation_ellipses.py
MarcoGorelli: hey - i haven't looked into `ellipse`, but the pattern of creating a list of dataframes and then concatenating is what pandas recommends (as opposed to continuously concatenating in the for loop)

dangotbanned: > hey - i haven't looked into `ellipse`, but the pattern of creating a list of dataframes and then concatenating is what pandas recommends (as opposed to continuously concatenating in the for loop)

Thanks @MarcoGorelli
I mean -  if nothing jumped out at you as a `pandas` anti-pattern - then that's a good sign at least 🙂 

Re: (`ellipse`|`np_ellipse`) the only relevant parts would be the signature:

```py
from typing import TypeAlias
import numpy as np

_2DArray: TypeAlias = np.ndarray[tuple[int, int], np.dtype[np.float64]]

def np_ellipse(arr: _2DArray, segments: int = 50) -> _2DArray: ...
```

The `segments` parameter controls the number of rows (elements) returned. 
So there's a potential for changing shape before/after `numpy`

tests/examples_arguments_syntax/deviation_ellipses.py
```diff
@@ -0,0 +1,91 @@
...
+    return center + radius * (circle @ np.linalg.cholesky(cov_mat).T)
+
+
+def pd_ellipse(
+    df: pd.DataFrame, col_x: str, col_y: str, col_group: str
+) -> pd.DataFrame:
+    cols = col_x, col_y
+    groups = []
+    # TODO: Rewrite in a more readable way
+    categories = df[col_group].unique()
```
dangotbanned, line 65:
```suggestion
    groups = []
    categories = df[col_group].unique()
```

tests/examples_arguments_syntax/deviation_ellipses.py
dangotbanned: Resolved in (https://github.com/vega/altair/pull/3747/commits/fa280f1fc8d71a6e12fa891ab1d0907143075866)

mattijn:
Thanks for this example! Can we add this to the category case studies instead of distributions? If the uncertainty was computed within Altair (not yet possible) than it was a good example for the distribution section.

On-the-fly computation of these confidence regions by doing selections would be really great though. Continue dreaming for https://github.com/vega/vega-lite/issues/6043.

Can we remove this line regarding the import annotation from `__future__`? Probably also requires removing other types from the function, but that is fine for this example.

Can we rename the functions? Eg can we use `confidence_region_2d` and `grouped_confidence_regions` instead?

I also tried to reduce the usage of pandas by using dicts directly. I think that can work, but the readability is slightly less than current situation. But doing more with dicts as containers is always good 😊

dangotbanned:
> Thanks for this example!

No problem @mattijn, although I feel my role in this was mostly making the connection between #3715, #514 and [`ggplot2.stat_ellipse`](https://ggplot2.tidyverse.org/reference/stat_ellipse.html#ref-examples).
It was interesting though brushing up on some `scipy`/`numpy` - as I don't usually interact with these packages directly.

> Can we add this to the category case studies instead of distributions? If the uncertainty was computed within Altair (not yet possible) than it was a good example for the distribution section.
> On-the-fly computation of these confidence regions by doing selections would be really great though. Continue dreaming for [vega/vega-lite#6043](https://github.com/vega/vega-lite/issues/6043).

Huh, I guess I'd never considered *where* the computation occured being a factor in the category.
Yeah I can recategorise if you think that would be a better fit

> Can we remove this line regarding the import annotation from `__future__`? Probably also requires removing other types from the function, but that is fine for this example.

Yeah I think we *might* not need that line at all actually.
I added it out of habit when I used `tuple[...]` here, but that should be valid since https://docs.python.org/3.9/whatsnew/3.9.html#type-hinting-generics-in-standard-collections

I'll take a look today and see what works 👍 

https://github.com/vega/altair/blob/c578ac1b42139e83a4b990fb309a6cea4afb2850/tests/examples_arguments_syntax/deviation_ellipses.py#L28

> Can we rename the functions? Eg can we use `confidence_region_2d` and `grouped_confidence_regions` instead?

Yeah those seem like much better names!
Bikeshedding a little here, but maybe `confidence_region_2d` and `confidence_region(s?)_grouped`?.
I don't feel too strongly about it since this is an example - but I'd usually try to match the prefix for related functionality

> I also tried to reduce the usage of pandas by using dicts directly. I think that can work, but the readability is slightly less than current situation. But doing more with dicts as containers is always good 😊

Interesting, I hadn't thought to try this without a dataframe at all 🤔 
I'm not a fan of `pandas`, but since we're already using it (`vega_datasets`) for `iris` - I think it is simpler just to keep the data in that container.
We'll also be able to easily adapt this example to `penguins` (https://github.com/vega/altair/pull/2231) after (https://github.com/vega/altair/pull/3631) and in #2213

dangotbanned:
@mattijn hopefully I've addressed all of your points in (https://github.com/vega/altair/pull/3747#commits-pushed-3debf11)

dangotbanned:
Thanks for your help with this @mattijn

Issue #2470, opened 2021-06-01
Feature Request for Andrews Curves


Make42:
I would like to have Andrews Curves in Altair. It seems it is possible to draw them with PdVega, which is is build on top of vega-lite, so https://altair-viz.github.io/pdvega/plotting.html#andrews-curves seems to suggest that this should be possible to implement in altair.

jakevdp:
Hi - you can create Andrews curves with Altair, it just takes a bit of boilerplate. The pdvega version just constructs a dataframe with the appropriate values; see https://github.com/altair-viz/pdvega/blob/e3f1fc9730f8cd9ad70e7ba0f0a557f41279839a/pdvega/plotting.py#L107

In Altair you could probably do this more efficiently using transforms: it would be akin to the parallel coordinates example here: https://altair-viz.github.io/gallery/normed_parallel_coordinates.html

Make42:
curves = np.outer(vals[0], np.ones_like(t))
    for i in range(1, len(vals)):
        ft = ((i + 1) // 2) * t
        if i % 2 == 1:
            curves += np.outer(vals[i], np.sin(ft))
        else:
            curves += np.outer(vals[i], np.cos(ft))

is the part that calculates the values for the Andrews Curve. But the first lines does not contain a `/np.sqrt(2)` as I would expect from https://en.wikipedia.org/wiki/Andrews_plot - is that a bug? I understand that it might not make too much of a difference in the plot, but I would think it would still be better to stick to the standard, right?

jakevdp:
I don’t know - it plots what it plots, if that’s not what you want to plot, you should change it.

PR #3768, opened 2025-01-11
feat: Show `user_rows` in `MaxRowsError` message

dangotbanned:
Came across this error while looking into [`pdvega.plotting.andrews_curves()`](https://github.com/altair-viz/pdvega/blob/e3f1fc9730f8cd9ad70e7ba0f0a557f41279839a/pdvega/plotting.py#L107-L186) for #2470

I think this is pretty helpful if you've performed a transformation that *significantly* alters the shape of your data.

In this case, I was confused as the source data had `150` rows, but after transforming this exploded to `30000`.

## Current

```
MaxRowsError: The number of rows in your dataset is greater than the maximum allowed (5000) ...
```

## Changed
Now the number of rows is clearly displayed in the first set of parentheses:

```
MaxRowsError: The number of rows in your dataset (30000) is greater than the maximum allowed (5000).

Try enabling the VegaFusion data transformer which raises this limit by pre-evaluating data
transformations in Python.
    >> import altair as alt
    >> alt.data_transformers.enable("vegafusion")

Or, see https://altair-viz.github.io/user_guide/large_datasets.html for additional information
on how to plot large datasets.
```

The rest of the message is left unchanged

### DIFF
--- a/altair/utils/data.py
+++ b/altair/utils/data.py
@@ -106,6 +106,26 @@ class DataTransformerRegistry(PluginRegistry[DataTransformerType, R]):
 class MaxRowsError(Exception):
     """Raised when a data model has too many rows."""
 
+    def __init__(self, message: str, /) -> None:
+        self.message = message
+        super().__init__(self.message)
+
+    @classmethod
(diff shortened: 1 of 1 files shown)


jonmmease (approved):
looks good. thanks!

PR #3827, opened 2025-04-17
feat: Allow numpy scalars in altair expressions

stanmart:
## Use old numpy scalar representation in expressions.

Numpy 2.0 has changed the representation of scalars (e.g., `repr(np.int64(0))` used to be `'0'` but became `np.int64(0)` ([related issue](https://github.com/h5py/h5py/issues/2456)). Due to this change, numpy constants can no longer be used in altair expressions for, e.g., filtering. This is inconvenient, as iterating over a numpy array returns such scalars.

Given that numpy is such a fundamental part of the scientific Python ecosystem, and is often used in conjunction with plotting tasks, I propose that Altair could special-case numpy scalars and unpack them into Python scalars to re-enable the numpy 1.x behavior of numpy constants being usable in altair expressions. This PR implements a simple solution for this.

### DIFF
--- a/altair/expr/core.py
+++ b/altair/expr/core.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 import datetime as dt
 from typing import TYPE_CHECKING, Any, Literal, Union
 
+import numpy as np
+
 from altair.utils import SchemaBase
 
 if TYPE_CHECKING:
@@ -50,6 +52,8 @@ def _js_repr(val) -> str:
         return val._to_expr()
     elif isinstance(val, dt.date):
         return _from_date_datetime(val)
+    elif isinstance(val, np.generic):
+        return repr(val.item())
     else:
         return repr(val)
 
--- a/altair/expr/core.py
+++ b/altair/expr/core.py
@@ -1,15 +1,12 @@
 from __future__ import annotations
 
 import datetime as dt
+import sys
 from typing import TYPE_CHECKING, Any, Literal, Union
-
-import numpy as np
 
 from altair.utils import SchemaBase
 
 if TYPE_CHECKING:
-    import sys
-
     from altair.vegalite.v5.schema._typing import Map, PrimitiveValue_T
 
     if sys.version_info >= (3, 10):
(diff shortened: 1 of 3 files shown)


mattijn:
Thanks for this PR! Altair has no hard dependency on numpy anymore, but it is delegating this through narwhals. That requires a change to this PR.

I see you have added a test, which is much appreciated. Do you also have an example that can be easily copied and run in a Jupyter notebook that highlights the issue?

stanmart:
Ah, sorry, I did not notice that numpy is not a hard dependency. In that case Narwhals has a `narwhals.dependencies.is_numpy_scalar` function that seems to do what we need, but it does not seem to work that way. (I am also unsure if it is part of their public API.) I've opened [an issue](https://github.com/narwhals-dev/narwhals/issues/2394) there to clarify it. I'll get back to this PR once that's resolved.

> Do you also have an example that can be easily copied and run in a Jupyter notebook that highlights the issue?

Sure, please see [this gist](https://gist.github.com/stanmart/8583aaa30413a53fbd2c49a41d58f413) for a minimal example, and [this other one](https://gist.github.com/stanmart/257f30684362e45eb39b2ac730030462) for a more realistic use case where it also happens. Unfortunately that the actual error is not displayed on GH, so I'll also include a screenshot here:
image

altair/expr/core.py
```diff
@@ -3,6 +3,8 @@
 import datetime as dt
 from typing import TYPE_CHECKING, Any, Literal, Union
 
+import numpy as np
```
MarcoGorelli, line 6: rather than importing numpy, you could make a little helper `is_numpy_generic` which does
```python
return (np := sys.modules.get('numpy')) is not None and isinstance(val, np.generic)
```
then use
```python
    elif is_numpy_generic(val):
        return repr(val.item())
```
on L55

MarcoGorelli: we could (and should!) add/change this in Narwhals, but given that it's a one-liner, I think you'd be better off just defining a helper function here than bumping the minimum narwhals version

stanmart: Thank you for the tip @MarcoGorelli, I updated the PR. I could make the signature of `is_numpy_generic` more precise by making the return type a `TypeGuard`, but that would mean that we have to import numpy in the `if TYPE_CHECKING` block. I don't think it's worth it, but me know if you'd prefer to add it @mattijn.

mattijn:
Activated auto-merge (if tests pass do PR merge). Thanks again for this PR, appreciated!

PR #3857, opened 2025-07-15
test: update test_datasets.py

### LINKED ISSUE #3854: improve robustness of `altair.datasets`
### What is your suggestion?

In https://github.com/vega/altair/pull/3848 `altair.datasets` was added. 

As was pointed out by @dangotbanned on slack, we should doublecheck that we haven't introduced regressions in the tests.

This [commit](https://github.com/vega/altair/commit/f6789e351bca96f581e2a1cce6233e62763d5ece#diff-3b63cb613626e7d787b8e63ced54267d5d18966f5fbc3795812573b4e50a942d) shows shows the partial diff between https://github.com/vega/altair/commit/4b348c9000570459509da889d44f62085db3b02b and https://github.com/vega/altair/commit/628ca56bcc71633cc14561203766329b7436e7ce of only the changes in the `altair/datasets/*` directory and the `tests/test_datasets.py` on a separate branch.

Hope this issue helps in checking these.

_No response_

mattijn:
This PR improves the datasets test suite based on feedback in #3854. Now, the automated tests reflect the implementation, solid work. Namely:
- Fixed test failures: corrected the regex pattern in `test_all_datasets` to properly match error messages (the test was looking for `icon_7zip.png` but the actual error shows `7zip.png`).
- Added data API tests: integrated comprehensive tests for the new data accessor API into the main test file.
- Streamlined test coverage: removed redundant test `is_url` function that validates with vega-datasets 3.2.0 solely precomputed URLs.
- Maintained core functionality: preserved all essential tests for backend functionality, error handling, caching, and data parsing.

The test suite now provides better coverage and the tests are reflecting by their names. No changes to the code in altair were required.

Could you have a look if this addresses your concerns @dangotbanned? 

Close #3854


dangotbanned:
Thanks @mattijn, I've only skimmed through but this is looking better! 🎉 

> [!NOTE]
> I'll aim to do a proper review soon

One thing that stands out to me on a first look:

> No changes to the code in altair were required.

Apologies for not being clearer, **I'm absolutely fine with further changes being made in both `altair` and `tools`.**

## `test_pandas_date_parse`

I think that *at least* (https://github.com/vega/altair/issues/3854#issuecomment-3073518839) will require some changes outside of the test suite - since that issue seems **very related** to (https://github.com/vega/vega-datasets/pull/702).

My guess is that because that PR introduced a disconnect between the `Resource.name` and `Resource.path`, like below, the way I was previously handling these fields probably needs to be updated in some way:


<table>
  <thead>
    <tr>
      <th width="500px">Before</th>
      <th width="500px">After (vega/vega-datasets#702)</th>
    </tr>
  </thead>
  <tbody>
  <tr width="600px">
      <td>


```py
{"name": "7zip.png", "path": "7zip.png"}
{"name": "annual-precip.json", "path": "annual-precip.json"}
{"name": "flights-200k.arrow", "path": "flights-200k.arrow"}
{"name": "flights-20k.json", "path": "flights-20k.json"}
```

</td>
<td>


```py

{"name": "icon_7zip", "path": "7zip.png"}
{"name": "annual_precip", "path": "annual-precip.json"}
{"name": "flights_200k_arrow", "path": "flights-200k.arrow"}
{"name": "flights_20k", "path": "flights-20k.json"}
```

</td>
</tr>

  </tbody>
</table>


> [!IMPORTANT]
> Would you be able to restore (`test_pandas_date_parse`), with the only change being the names used in the `@pytest.mark.parametrize` please?

For reference:
- (https://github.com/vega/altair/commit/909e7d05e57718b2f634a7e6781cb4e58a835837) was the commit that introduced using schema data type information for `pandas`, in response to @joelostblom's early experience (https://github.com/vega/altair/pull/3631#issuecomment-2480816377)
- (https://github.com/vega/altair/pull/3631/commits/a776e2fd5a74dd21917b884e2b418ade18773895) was the last major change to `SchemaCache`
- Also here's the docstring for `SchemaCache` https://github.com/vega/altair/blob/78a57f2d9e973930fbc859175ac6c2d3f16c2881/altair/datasets/_cache.py#L206-L220

Does that help demonstrate why this is a regression?

- I added functionality to solve an issue that came up in the PR
- I added a test to make sure the data types were *at least* all temporal data types
- But now they aren't, and the issue is only showing up in `pandas` - which is where the problem was in the first place 😔

tools/datasets/datapackage.py
```diff
@@ -95,7 +95,7 @@ def core(self) -> pl.LazyFrame:
     def schemas(self) -> Mapping[Dataset, Mapping[str, FlFieldStr]]:
         """Reduce all datasets with schemas to a minimal mapping."""
         m: Any = {
-            Path(rsrc["path"]).stem: {f["name"]: f["type"] for f in s["fields"]}
+            rsrc["name"]: {f["name"]: f["type"] for f in s["fields"]}
```
dangotbanned, line 98: Ah I remember why I did this now!

For reference, you can see all the schemas like this:

```py
from altair.datasets import Loader


load = Loader.from_backend("polars")
>>> schemas = load._reader._schema_cache.mapping
# very big output
```

But just the names will do for now:

```py
>>> list(schemas)
```

<details><summary><b>Show all dataset file stems with a schema</b></summary>
<p>

```py
['airports',
 'anscombe',
 'barley',
 'birdstrikes',
 'budget',
 'budgets',
 'burtin',
 'cars',
 'co2-concentration',
 'countries',
 'crimea',
 'disasters',
 'driving',
 'flare-dependencies',
 'flare',
 'flights-10k',
 'flights-200k',
 'flights-20k',
 'flights-2k',
 'flights-3m',
 'flights-5k',
 'flights-airport',
 'football',
 'gapminder-health-income',
 'gapminder',
 'github',
 'global-temp',
 'income',
 'iowa-electricity',
 'jobs',
 'la-riots',
 'londonCentroids',
 'lookup_groups',
 'lookup_people',
 'monarchs',
 'movies',
 'normal-2d',
 'obesity',
 'ohlc',
 'penguins',
 'platformer-terrain',
 'political-contributions',
 'population',
 'population_engineers_hurricanes',
 'seattle-weather-hourly-normals',
 'seattle-weather',
 'sp500-2000',
 'sp500',
 'species',
 'stocks',
 'udistrict',
 'unemployment-across-industries',
 'unemployment',
 'uniform-2d',
 'us-employment',
 'us-state-capitals',
 'weather',
 'wheat',
 'windvectors',
 'zipcodes']
```

</p>
</details>

For any of the datasets which have multiple file formats (e.g. `flights-200k.json`, `flights-200k.arrow`) - this let's us use the same schema for both.

IIRC, the `.json` datasets don't get inferred correctly, but in that case the `.arrow` one can be used instead.

On `main`, this works:

```py
from altair.datasets import Loader

load = Loader.from_backend("polars")
>>> load._reader._schema_cache.schema("flights-200k")

Schema([('delay', Int64), ('distance', Int64), ('time', Float64)])
```

mattijn: Yup, but we need the dataset name which is a valid Python identifier and not the url stem

dangotbanned: > Yup, but we need the dataset name which is a valid Python identifier

At the boundary of `data.<...>`? Sure, we need a valid identifer *there*.

But that doesn't mean we **need** to use that name everywhere else - especially if there are benefits to *not* doing that 🙂 

I haven't thought too long on this yet, but considering we have a row like this per-dataset - we *could* do something like adding a `stem` column there - but retrieve it using the snake case name


https://github.com/vega/altair/blob/49a8d770c8aad8d13e932a70002376771164b048/altair/datasets/_typing.py#L191-L203

mattijn: Sure, with the `dataset_name` you can lookup. Isn't it the `file_name` what you are after?

dangotbanned:
(https://github.com/vega/altair/pull/3857/commits/42236c210ee1b8b35a7f5987026ccff5165577f8)

@mattijn did you generate this?

mattijn:
Yes, but now ruff and or mypy aren't yet happy. Or you don't think we should make pylance happy too?

dangotbanned:
> Yes, but now ruff and or mypy aren't yet happy. Or you don't think we should make pylance happy too?

No I mean did you write this yourself?

mattijn:
Not a type expert, let me leave this to you, but the stubs are now derived from the DataPackage and not the other type files.

mattijn:
Thanks for reviewing @dangotbanned! I've included one raise using the `AltairDatasetsError` in case for the PyArrow backend when the date type column is in non-ISO date format, related pyarrow issue: https://github.com/apache/arrow/issues/41488.

This is specifically for the 'stocks' dataset example (date contains values like this:`"Jan 1 2000"`):
```python

data.stocks(engine="pyarrow")
```

> AltairDatasetsError: PyArrow cannot parse date format in dataset 'stocks'. This is a known limitation of PyArrow's CSV reader for non-ISO date formats.

PR #3858, opened 2025-07-16
ci: add nightly builds system

### LINKED ISSUE #3576: adopt nightly builds in accordance with SPEC-4
### What is your suggestion?

In accordance with [SPEC 4 — Using and Creating Nightly Wheels](https://scientific-python.org/specs/spec-0004/) we should explore how we can create nightly (_semi regular interval_) wheels.

Reasons why this is a good idea can be found in SPEC-4:

> _Regularly running your project’s tests while using the nightly version of your dependencies allows you to spot problems caused by upstream changes before a new release is made. This way potential issues can be resolved before they find their way into a release, at which point it becomes much harder to change or revert something._
> 
> _Regularly creating nightly wheels for your project allows projects that depend on you to give feedback about upcoming changes. As with testing against nightlies of your dependencies this gives your dependents a chance to report problems before they find their way into a release._

_No response_

mattijn:
This PR adds a nightly build system for Altair that creates and publishes nightly versions when there are new commits since the last nightly build.

The system:
- Only runs when there are commits since the last nightly release
- Signs wheel files using sigstore/cosign
- Generates dependency diffs and binary file checksums
- Uses GitHub Actions and releases
- Uses version format YYYYMMDD.dev0+commit

Close #3576


mattijn:
The workflow runs 3 times per week (only if commits since the last nightly release) on:
- Monday at 3:00 AM UTC
- Wednesday at 3:00 AM UTC
- Friday at 3:00 AM UTC

Issue #3868, opened 2025-09-08
`JupyterChart` accesses `.changed` property in callbacks


azjps:
### What happened?

In `JupyterChart` #3119, [`.changed`](https://github.com/vega/altair/blob/5e3a0c7bcb2ebc74fbcaec47d4c9b6ec7828b1d9/altair/jupyter/js/index.js#L94) I think is an [implementation detail in Jupyter's backbone.js patch](https://github.com/jupyter-widgets/ipywidgets/blob/57ff79dce80b38b6666f66d1c7436153fe7c85b5/packages/base/src/backbone-patch.ts#L83), and isn't supported in other [AFM](https://anywidget.dev/en/afm/) implementations such as `marimo`'s:

<img width="1205" height="848" alt="Image" src="https://github.com/user-attachments/assets/db321a6d-5012-4c07-ad45-e345c7a2502d" />

This is [using the `JupyterChart` example in the docs](https://altair-viz.github.io/user_guide/interactions/jupyter_chart.html#accessing-variable-params).

If I overwrite the anywidget ESM with
```
class AltairChart(altair.JupyterChart):
     _esm = """
            // ... rest of https://github.com/vega/altair/blob/5e3a0c7bcb2ebc74fbcaec47d4c9b6ec7828b1d9/altair/jupyter/js/index.js
            // for (const [param, value] of Object.entries(new_params.changed._params)) {
            for (const [param, value] of Object.entries(new_params)) {
     """
```
these errors go away and appears to work in both `jupyter` & `marimo`. I'm not sure if there's some other impact of this change though.

I came across this while trying to debug an AFM implementation in `nicegui`: https://github.com/zauberzeug/nicegui/discussions/5096#discussioncomment-14332020. Separately, I'm having trouble getting the binded JS interactions to work and don't have any leads, if someone would have advice on that? 

### What would you like to happen instead?

No console errors.

### Which version of Altair are you using?

Tried with both `altair=6` and `altair=5.5`.

mattijn:
Fixed by #3871

PR #3871, opened 2025-09-13
fix: Make `.changed` property optional in `JupyterChart` #3868

azjps:
Refer to #3868 for details.

`value.changed` is an implementation detail in Jupyter's backbone.js patch, and isn't supported in other AFM implementations such as marimo's (or streamlit, nicegui, etc).

To silence the errors and allow updates from python to JS happen, for now signaling all params instead of just changed ones, if `.changed` is undefined.


joelostblom:
Thanks for submitting a fix for the issue you reported earlier @azjps !

@jonmmease Do you have the bandwidth to review if this change makes sense?

mattijn:
I checked if I could observe any negative consequence of this PR, but I couldn't find any. Merging for now. Thanks @azjps!

PR #3882, opened 2025-10-10
docs: Remove whitespace after '*' so that text is rendered italic.

thomascamminady:
Fixes extra whitespace that prevents text from being rendered as italic in the transform regression docs.


mattijn:
Thanks @thomascamminady 😊

thomascamminady:
Always feels a bit stupid with these trivial PRs :D 
Thanks for all the work you're doing!

PR #3888, opened 2025-10-13
fix(types): Change the height and width types of `alt.Chart()` from `int` to `float`

### LINKED ISSUE #3873: Partially incorrect type for the `width` and `height` of `altair.Chart()`s
### What happened?

Hi! 👋 

I'm migrating a codebase to Altair 5.5.0 and I noticed that `width`s and `height`s defined with `float`s to create basic Altair charts (`altair.Chart()`) triggered an error when type checking with mypy: `error: Argument "width" to "Chart" has incompatible type "float"; expected "int | dict[Any, Any] | Step | Literal['container'] | UndefinedType"  [arg-type]`

Looking at the source code, the error is expected given that `float`s are not expected. However, I would expect to be able to pass a `float` here, since they are supported upstream. References:

- https://github.com/vega/altair/blob/v5.5.0/altair/vegalite/v5/api.py#L3929-L3938
- https://github.com/vega/altair/blob/v5.5.0/altair/vegalite/v5/schema/core.py#L25094
- https://github.com/vega/altair/blob/5e3a0c7bcb2ebc74fbcaec47d4c9b6ec7828b1d9/altair/vegalite/v6/api.py#L4042-L4051
- https://github.com/vega/altair/blob/v5.5.0/altair/vegalite/v5/schema/_config.py#L7767-L7777 (`altair.theme.ThemeConfig`)

Therefore, I think this type definition is incorrect (it should be `float` instead of `int`), but I'm not certain. Could you please validate this?

If it is indeed a small bug, please let me know if I can open a PR to fix it. 

Thanks!

Pass `height`s and `width`s as `float`s without type errors.

5.5.0

joaopalmeiro:
Closes #3873 

Hi! 👋 

This PR updates the height and width types for `alt.Chart()` from `int` to `float`, since these values can contain decimal places.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -16,7 +16,7 @@ body:
       label: What happened?
       description: |
         Describe what happened and how to reproduce the bug.
-        Include the full code and data to reproduce it. Use a simple toy data set, e.g. from `vega_datasets`.
+        Include the full code and data to reproduce it. Use a simple toy data set, e.g. from `altair.datasets`.
     validations:
       required: true
   - type: textarea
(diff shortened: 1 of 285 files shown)


joaopalmeiro:
Thanks, @mattijn!

mattijn:
Thank you for the PR @joaopalmeiro!

PR #3889, opened 2025-10-13
fix: Repair vega export

bobot:
Even if not advertised, it is better to not fail when format is "vega". Currently it is failing since the svg case is reached.


mattijn:
Thanks @bobot!

PR #3558, opened 2024-08-27
docs: Unstack area to render cumulative chart correctly

joelostblom:
This has likely been broken since Vega-Lite changed the default behavior in
https://github.com/vega/vega-lite/releases/tag/v5.14.1

You can see the current broken chart here https://altair-viz.github.io/gallery/cumulative_count_chart.html

After this PR:
image

PR #3890, opened 2025-10-13
docs: Unstack area to render cumulative chart correctly

joelostblom:
We forgot this location when updating the gallery example in https://github.com/vega/altair/pull/3558.

### DIFF
--- a/doc/user_guide/transform/window.rst
+++ b/doc/user_guide/transform/window.rst
@@ -14,14 +14,15 @@ For example, consider the following cumulative frequency distribution:
 
     import altair as alt
     from vega_datasets import data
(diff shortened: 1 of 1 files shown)


mattijn:
I noticed I resolved this differently in #3859. Can you check what is currently in the main branch in the doc folder [here](https://github.com/vega/altair/blob/main/doc/user_guide/transform/window.rst#window) if the fix I applied is appropriate? I followed the spec from the corresponding Vega-lite page: https://vega.github.io/vega-lite/docs/window.html.

joelostblom:
That looks fine too, thanks!

PR #3902, opened 2025-11-01
feat: add warning for deduplications

### LINKED ISSUE #3891: Altair incorrectly combines parameters with the same properties instead of assigning them unique IDs
### What happened?

The example from https://github.com/vega/altair/pull/3628 does still not work correctly in the docs, but now for a new reason. It seems that when two parameters in the same spec have the same properties, Altair assigns them the same name so that they are combined to the same parameter in the VL spec.

As an example, for the altair code below, I would expect there to be two different param IDs generate, so that I need to make a selection in both charts to trigger the bars to show up. Instead, I only need to select in one chart, since it is the same param name that is used as you can see in [the corresponding VL spec](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAZgAY2AacaYsiygAlMiRoVYcAvpO5oCYAIbjkAbVCN5SlGpABbBQCcA1nhCEAngAdMKEJYglx3SnABeN5AEY2MkJiiQdCRIqDwANhAGpvCYYXS2APIGiCQg3BbWtlAQuiQKYSC+AB7RWHG2gpGUmPbYmFHpVh4gAI6kCkRwhEpwaDa+5qWx8cggALJwYZiUAPrWBjMA4vkRUGlmTbZtHcTdxH2FvlAKus0YODMAbABGdADsLJiYbGCYAKyYAJxsMNczbIVOKB9MZTBlmvZHOsXO4UN5fP5AsFTJAIlFQjFyqMkik1o1MqNsrljgVikMsSAAIJgV5TAw9aDrcFbdqdPa9frcQYYsojcaTaZzepLFaM-HNbZsnoHaTcY6nWznbAzAAc7EuDzoABYYJcYAprpc6JgZp5CgBdGTAwwmULM0bXQyAvwBCBBKAhfAQNHkvk4hCpcVZHJ5UncEqhBQIBAGTAIJTNSCkIhMzajSW7aWckDc0CYv3JAN4jYEkBE0OHdL0qCUGCRXTaPOTQj1UwdEY6SyGE62Lv03Sqy6XFiXABM10dKqnb0+11H60wuksFhQ+rC1Stdm7DdGfZOg+HY4nCinKpnc4XS5XyDXG8tkkt5u4dCUClMpAMBVGYkIlkoyAAegAsA6CgAA6AArShjTCXoDDAqBMEIACoEsXQAL6eMAFoX26apCEoAABNAWDA0cwM8ADcIUYDDEoSDKEZXw910ShGzLE4IW3A8R3HSdp1nednFiTAwGUUB7RARx6jQfJnSVNjVBAJUrluB4nhed4vh+P4AVkLBlTVNgNRYbVdX1Q1jVNEBH24AASSgwEYRc32-Qhf3-IDMIUMCAyYUhrjAuAIAAxznP0DC4wULDYJbDDLgosC2AYpigA).

```py
import altair as alt

cars = data.cars.url

# empty=False ensure that no points are selected before a selection is drawn
brush = alt.selection_interval(empty=False)
brush2 = alt.selection_interval(empty=False)

points = alt.Chart(cars).mark_point(size=10).encode(
    x='Horsepower:Q',
    y='Miles_per_Gallon:Q',
    color='Origin:N'
)

points2 = points.encode(
    x='Acceleration:Q',
    y='Miles_per_Gallon:Q',
)

bars = alt.Chart(cars).mark_bar().encode(
    x='count()',
    y='Origin:N',
    color='Origin:N'
).transform_filter(
    brush & brush2
)

(points.add_params(brush) | points2.add_params(brush2)) & bars
```

If the second param is changed to the following, two unique parameters are created and [the example works as expected](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAZgAY2AacaYsiygAlMiRoVYcAvpO5oCYAIbjkAbVCN5SlGpABbBQCcA1nhCEAngAdMKEJYglx3SnABeN5AEY2MkJiiQdCRIqDwANhAGpvCYYXS2APIGiCQg3BbWtlAQuiQKYSC+AB7RWHG2gpGUmPbYmFHpVh4gAI6kCkRwhEpwaDa+5qWx8cggALJwYZiUAPrWBjMA4vkRUGlmTbZtHcTdxH2FvlAKus0YODMAbABGdADsLJiYbGCYAKyYAJxsMNczbIVOKB9MZTBlmvZHOsXO4UN5fP5AsFTJAIlFQjFyqMkik1o1MqNsrljgVikMsSAAIJgV5TAw9aDrcFbdqdPa9frcQYYsojcaTaZzepLFaM-HNbZsnoHaTcY6nWznbAzAAc7EuDzoABYYJcYAprpc6JgZp5CgBdGTAwwmULM0bXQyAvwBCBBKAhfAQNHkvk4hCpcVZHJ5UncEqhBQIBAGTAIJTNSCkIhMzajSW7aWckDc0CYv3JAN4jYEkBE0OHdL0qCUGCRXTaPOTQj1UwdEY6SyGE62Lv03Sqy6XFiXABM10dKqnb0+11H60wuksFhQ+rC1Stdm7DdGHQgTHqAAI+z3uIvl9y1xvLZJLebuHQlApTKQDAVRmJCJZKMgAPS-sA6CgAA6AArShjTCXoDGAqBMEIX8oEsXRfz6eMAFpH26apCEoAABNAWGA0dgM8X8sIUADDEoMDKEZXwT10ShGzLE4IW3QdhzHCcFCnFUZznaFYkwMBlFAe0QEceo0HyZ0lWY1QQCVK5bgeJ4XneL4fj+AFLSBViFV3bIDwMY9tyEqZRLBNNJKIaTZN8eTtCUrBlTVNgNRYbVdX1Q1jVNEA724AASSgwEYRdnw-Qgvx-f80IUYCAyYUhrmAuAIF-MKIv0VC4wUdCoJbVDLlI4C2Fo+igA):

```py
brush2 = alt.selection_interval(empty=False, name='another param')
```

Altair should always assign unique IDs to distinct parameters instead of combining them into the same parameter name.

main

mattijn:
Attempt to fix #3891.

This PR now emits a warning when identical parameters are merged during concatenation to alert users to unintended linking.

With #3851, we now use a hash-based naming for parameters and views for deterministic names and thread safety, but I said in this comment https://github.com/vega/altair/pull/3851#issuecomment-3080396020 that duplicate explicit names will raise an error to catch potential mistakes.

> - Identical parameters gets deduplicated only when a hash-based name was derived
> - Explicit parameter names are still supported, but duplicate explicit names will raise an error to catch potential mistakes.

But I found that this only works for a single chart (via `add_params`): hash-based duplicates are dropped; explicit duplicates raise. During concatenation (via `_combine_subchart_params`) all identical parameters (hash-based and explicit) are deduplicated and a warning is raised, so when using explicit names there is no raising of error.

Eg. this (explicit name for interactive selection paramater, that becomes identical during concatenation):

```python
import altair as alt

cars = data.cars.url

chart = (
    alt.Chart(cars)
    .mark_point()
    .encode(x="Horsepower:Q", y="Miles_per_Gallon:Q")
    .interactive(name="MY_CHART")
)
(chart & chart).to_dict()
```
Returns
```bash
/var/folders/b3/pctp5dwn6t7fm1t4sc4tknb00000gp/T/ipykernel_88674/3413242154.py:12: 
UserWarning: Automatically deduplicated selection parameter with identical configuration. 
If you want independent parameters, explicitly name them differently (e.g., name='param1', name='param2'). 
See https://github.com/vega/altair/issues/3891
  (chart & chart).to_dict()
```
```python
{'config': {'view': {'continuousWidth': 300, 'continuousHeight': 300}},
 'vconcat': [{'mark': {'type': 'point'},
   'encoding': {'x': {'field': 'Horsepower', 'type': 'quantitative'},
    'y': {'field': 'Miles_per_Gallon', 'type': 'quantitative'}},
   'name': 'view_6e7cfb454e831ee6_0'},
  {'mark': {'type': 'point'},
   'encoding': {'x': {'field': 'Horsepower', 'type': 'quantitative'},
    'y': {'field': 'Miles_per_Gallon', 'type': 'quantitative'}},
   'name': 'view_6e7cfb454e831ee6_1'}],
 'data': {'url': 'https://cdn.jsdelivr.net/npm/vega-datasets@v3.2.1/data/cars.json'},
 'params': [{'name': 'MY_CHART',
   'select': {'type': 'interval', 'encodings': ['x', 'y']},
   'bind': 'scales',
   'views': ['view_6e7cfb454e831ee6_0', 'view_6e7cfb454e831ee6_1']}],
 '$schema': 'https://vega.github.io/schema/vega-lite/v6.1.0.json'}
```

And this (explicit identical name for parameters):

```python

cars = data.cars.url
param_opacity = alt.param(value=1, name='MY_PARAM')
param_size = alt.param(value=1, name='MY_PARAM')

# Create chart with same parameter added twice
chart = (
    alt.Chart(cars)
    .mark_circle(opacity=param_opacity, size=param_size)
    .encode(x="Horsepower:Q", y="Miles_per_Gallon:Q", color="Origin:N")
    .add_params(param_opacity, param_size)
)
```
Returns
```bash
ValueError: Duplicate explicit parameter name: MY_PARAM
```

I'm merely documenting this. I don't know what to do with it, since I'm not really sure if we can improve this situation any better.

It would be great if you could have a look to @joelostblom!


joelostblom (approved):
Thanks for tackling this @mattijn ! I think just informing the user about what happens with a warning is a good solution since my biggest concern was that the deduplication happened silently and it was hard to figure out what was going on. Now the user can decide how they want to resolve the situation.

My only question is whether the lock file needs to be updated as part of this PR or if that should happen separately? You can decide and then go ahead and merge.

mattijn:
Thanks @joelostblom! Changes to `uv.lock` file seems to have disappeared by pulling in latest changes from main into this branch. Activated auto-merge.

liquidcarbon:
I have vertically concatenated charts that share X axis that can be zoomed together.  This generates a warning as I'm reusing the parameters on both layers of the chart.

Is the warning necessary in this case?  I was fighting them for a while and had to just suppress them.

```
    # vertically concatenate and add zoom
    wheel_zoom_x = alt.selection_interval(
        bind="scales",
        encodings=["x"],
        zoom="wheel![!event.shiftKey]",
    )
    wheel_zoom_y = alt.selection_interval(
        bind="scales",
        encodings=["y"],
        zoom="wheel![!event.altKey]",
    )
        warnings.simplefilter("ignore")
        c = (c0 & c1).add_params(wheel_zoom_x, wheel_zoom_y)
```

joelostblom:
@liquidcarbon We have an issue here https://github.com/vega/altair/issues/3916 tracking how we can reduce false positives for this warning. Feel free to comment there with a complete example spec and share any ideas you might have to how we can reduce these warnings when they are not needed

liquidcarbon:
Seems like it's more than just a nuisance with warnings, but rather a breaking change.

I can't make my two panels respond to zoom in sync like before.  I tried various combinations of wheel params (same names, different names) and the best I could do was to have the synced zoom respond only from the bottom panel.

```python
wheel_zoom_x_a = alt.selection_interval(
    bind="scales",
    encodings=["x"],
    zoom="wheel![!event.shiftKey]",
    name="zx-a"
)
wheel_zoom_x_b = alt.selection_interval(
    bind="scales",
    encodings=["x"],
    zoom="wheel![!event.shiftKey]",
    name="zx-b"
)


c = alt.vconcat(
    c0.add_params(wheel_zoom_x_a),
    c1.add_params(wheel_zoom_x_b),
).resolve_scale(
    x="shared",  # without this they're not synced
    y="independent"
)
```

liquidcarbon:
this is better but still not like before

```python
def zoomable(axis="x", name=""):
    key = "shift" if axis == "x" else "alt"
    selection_interval = alt.selection_interval(
        bind="scales",
        encodings=[axis],
        zoom=f"wheel![!event.{key}Key]",
        name=f"zoom_{name}"
    )
    return selection_interval
    # no warnings are thrown
    c = alt.vconcat(
        c0.add_params(zoomable("x", "xa"), zoomable("y", "ya")),
        c1.add_params(zoomable("y", "yb")),
    )
```

PR #3920, opened 2025-11-14
fix: Correct tooltip field for faceted choropleth example

### LINKED ISSUE #3919: faceted choropleth has wrong tooltip reference: `County ID: undefined`
### What happened?

See screenshot:

Image

From this page: https://altair-viz.github.io/gallery/maps_faceted_species.html.

The tooltip has to reference `county_id` and not `id` as is currently. 

version 6.0.0

Sparlitu:
Fix #3919.

This PR aims to provide the required changes to the tooltip reference in order to display the correct County id

### DIFF
--- a/tests/examples_arguments_syntax/maps_faceted_species.py
+++ b/tests/examples_arguments_syntax/maps_faceted_species.py
@@ -25,7 +25,7 @@ chart = alt.Chart(csv_data).mark_geoshape().encode(
         legend=alt.Legend(format='.0%')
     ),
     tooltip=[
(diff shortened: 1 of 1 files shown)


Sparlitu:
Hi, @joelostblom! I just noticed that my latest commit doesn't contain the necessary modifications for the examples_methods_syntax, only for the arguments one. The correct commit is the aefb97f one. Can we do something about this?

joelostblom:
Ah I didn't notice that. Could you open a new PR from that commit (or a new one with the remaining modification)?

PR #3933, opened 2025-12-27
style: Make chained chart methods work with type checkers other than pyright

hamdanal:
The return type of `use_signature` that wraps most `Chart` methods was not set. Only pyright worked[^1] with this decorator because it supports function return type inference. mypy and other type checkers do not work with this decorator.

`use_signature` appears to only be used with methods nowadays. This means we can simplify its type annotations so that it is understood by mypy and other type checkers.

Unfortunately ty does not work yet because it does not understand the `Concatenate` special form https://github.com/astral-sh/ty/issues/1535. The new typing will work with ty as soon as `Concatenate` support is added.

[^1]: Well, kinda. It needed some `pyright: ignore` comments because it couldn't distinguish the overloads used.

### DIFF
--- a/altair/utils/core.py
+++ b/altair/utils/core.py
@@ -769,7 +769,7 @@ def use_signature(tp: Callable[P, Any], /) -> _MethodSignatureCopier[P]:
 
 
 class _FunctionSignatureCopier(Protocol[P]):
-    def __call__(self, cb: WrapsFunc[R], /) -> WrappedFunc[P, R]: ...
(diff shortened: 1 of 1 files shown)


hamdanal:
Turns out my grep for the decorator usage was wrong, there still are some functions that need the decorator. I created `use_signature_func` for these cases which shares the implementation with `use_signature` but has the correct typing for functions. `use_signature_func` works with mypy, ty and pyright.

mattijn:
Really nice PR @hamdanal! Great that we found a way to improve the method chaining approach for type checkers. Thanks!

PR #3934, opened 2025-12-31
fix: type issues with Chart save methods.

### LINKED ISSUE #3870: Partially unknown type for some Chart methods.
### What happened?

I'm getting type checking (basedpyright) `reportUnknownMemberType` warnings for some Chart methods, such as `mark_*` and `save`. Others, like `encode`, are fine. Does not affect functionality, just trips up the type-checker.

Example:
```python
from altair import Chart

Chart(df.unpivot(on=matches(regex), index=idx, value_name=ylab))
        .mark_line()
        .encode(x=idx, y=ylab, color="variable")
```

basedpyright output:
```sh
Type of "mark_line" is partially unknown
  Type of "mark_line" is "(align: Parameter | SchemaBase
          ...
          **kwds: Unknown) -> Chart" (basedpyright reportUnknownMemberType)
```

No warnings.

6.0.0

alec-bike:
`Chart.save()` methods trigger type-check warnings with ty. This PR adds some type annotations to avoid these warnings. 

Resolves #3870.

To test run ty:

```sh
uvx ty check altair/vegalite/v6/api.py
uvx ty check altair/utils/save.py
```

`uv run task test` reveals no change in test results compared to main.

Notes:
- many more type issues still remain in these 2 files
- (reverted) ~~replacing Union with | requires min python version 3.10~~

### DIFF
--- a/altair/vegalite/v6/api.py
+++ b/altair/vegalite/v6/api.py
@@ -243,7 +243,7 @@ def _consolidate_data(
 
     This function will modify context in-place, and return a new version of data
     """
-    values: dict[str, Any] | list | InlineDataset | None = None
+    values: Any = Undefined
     kwds = {}
 
     if isinstance(data, core.InlineData):
@@ -258,7 +258,7 @@ def _consolidate_data(
         values = data["values"]
         kwds = {k: v for k, v in data.items() if k != "values"}
 
-    if values is not None:
+    if not utils.is_undefined(values):
         name = _dataset_name(values)
         data = core.NamedData(name=name, **kwds)
         context.setdefault("datasets", {})[name] = values


alec-bike:
@mattijn could you please approve the workflow for the latest commit? The original commit failed due to incompatibility with python 3.9 so I reverted that change and rebased.

mattijn:
Approved the workflow! Looks good! And great that you work on this PR🙌. 

I've one question: do we need the change regarding this line:

Before 
```python
values: Any = Undefined
```

After 
```python
values: dict[str, Any] | list | InlineDataset | None = None
```

There have been long discussions in the past on the difference between `None` and `Undefined` (also type-wise) and at that time end up with the conclusion that there is a difference between these.

alec-bike:
Thanks for approving, the PR passes now. Let me look more into `Undefined` vs `None` and revert or update as needed.

alec-bike:
Neither ty nor basedpyright complained about the (original) use of Undefined so I reverted that change. Might be worth adding a unit test to illustrate the difference.

mattijn:
Nice! Can this PR be moved out of draft status?

alec-bike:
Done. PR is live and ready to review.

mattijn:
Thanks @alec-bike! Looks great

PR #3948, opened 2026-02-06
feat: add ensure_ascii option to to_json() with default False

### LINKED ISSUE #3925: [feature-request] ensure_ascii=False as default in to_json() and expose this option
## Enhancement Description

I want to use to_json() to study vega. There are zh_TW characters in my chart, and they are encoded into ascii. I suppose utf8 is quite standard now everywhere. It would help international users if to_json() sets ensure_ascii=False as default when calling json.dumps(), and then make this option exposed for users to turn it on if they need to avoid utf8 characters and create pure ascii output.
<details>
  <summary><h2>Checklist</h2></summary>

- [v ] I checked for duplicate issues and found none.
</details>

RamiNoodle733:
## Summary

Fixes #3925

This PR adds an `ensure_ascii` parameter to the `to_json()` method that:
- Defaults to `False` to preserve UTF-8 characters (better for international users)
- Allows users to explicitly set `ensure_ascii=True` for ASCII-only output

## Changes

- Added `ensure_ascii: bool = False` parameter to `SchemaBase.to_json()` method
- Updated docstring to document the new parameter  
- Passed the parameter to `json.dumps()`
- Updated both the runtime module (`altair/utils/schemapi.py`) and the template (`tools/schemapi/schemapi.py`)

## Example Usage

```python
import altair as alt

# Create a chart with non-ASCII characters (e.g., Chinese)
chart = alt.Chart(...).encode(
    alt.X('field').title('中文标题')  # Chinese title
)

# Default: UTF-8 characters preserved
json_output = chart.to_json()

# Explicit ASCII-only output
json_ascii = chart.to_json(ensure_ascii=True)
```

## Benefits

- International users can now see their native characters in JSON output
- Maintains backward compatibility (users can still get ASCII-only if needed)
- Aligns with modern practices where UTF-8 is standard everywhere

## Testing

- Tested with Chinese, Japanese, and emoji characters
- Verified that `ensure_ascii=True` still produces escaped output when needed

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -1245,6 +1245,7 @@ class SchemaBase:
         indent: int | str | None = 2,
         sort_keys: bool = True,
         *,
+        ensure_ascii: bool = False,
         ignore: list[str] | None = None,
         context: dict[str, Any] | None = None,
         **kwargs,
@@ -1260,6 +1261,9 @@ class SchemaBase:
             The number of spaces of indentation to use. The default is 2.
         sort_keys : bool, optional
             If True (default), sort keys in the output.
+        ensure_ascii : bool, optional
+            If False (default), allow non-ASCII characters in the output.
+            If True, escape all non-ASCII characters.
         ignore : list[str], optional
             A list of keys to ignore.
         context : dict[str, Any], optional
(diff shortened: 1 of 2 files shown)


mattijn:
Hi @RamiNoodle733, why did you close the PR?

PR #3952, opened 2026-02-07
feat: add ensure_ascii option to to_json() with default False

RamiNoodle733:
Add ensure_ascii parameter to to_json() method with default False to preserve UTF-8 characters for better international user support.

This re-creates the functionality from the original PR #3948.

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -1286,7 +1286,9 @@ class SchemaBase:
         if context is None:
             context = {}
         dct = self.to_dict(validate=validate, ignore=ignore, context=context)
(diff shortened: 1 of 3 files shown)


mattijn:
Thanks for this PR @RamiNoodle733! I'm working in https://github.com/vega/altair/pull/3950 to fix the GitHub actions CI due to new releases upstream.

mattijn:
Thanks again for this PR @RamiNoodle733! Altair is used all around the world and non-ASCII characters should just work. Nice one for fixing this, and if you see more places where we can do better, please do not hesitate to contribute or raise an issue👍

PR #3964, opened 2026-02-23
fix: Ensure name uniqueness for concatenated facets

### LINKED ISSUE #3954: Error when concatenating interactive faceted charts
### What happened?

Altair produces an error when I try to concatenate interactive faceted charts:

```python
import altair as alt
import numpy as np
import pandas as pd

n = 100
rng = np.random.default_rng(42)

df = pd.DataFrame({"x": rng.random(n), "z": rng.integers(0, 2, n)}).assign(y=lambda x: x["x"])

chart_0 = alt.Chart(df).mark_line().encode(x="x", y="y").facet("z")
chart_1 = alt.Chart(df).mark_line().encode(x="x", y="y").facet("z")

# Both charts fail

(chart_0 & chart_1).add_params(alt.selection_interval(bind="scales"))

(
    chart_0.add_params(alt.selection_interval(bind="scales"))
    & chart_1.add_params(alt.selection_interval(bind="scales"))
)
```

This example used to work before Altair 6. It also works if I remove the selection parameter or the faceting.

_No response_

Altair 6.

joelostblom:
This is a suggestion for how to resolve https://github.com/vega/altair/issues/3954.

I found that when the facet charts were assigned names initially, they were not part of the concatenation so there was only one subchart to loop over for some reason. Since they both already had names (both ending in "_0"), the `_needs_name` check would fail and they would not be assigned a new name when they later went through the loop as part of the concatenation (ie i>1).

There might be more appropriate solutions if we dig more into how the current iteration over `subcharts` work and why each chart come through there multiple times, both as single charts and as part of the concatenation as faceted charts. This works though, and does not trigger failure of any existing test so maybe it is good enough.


mattijn:
Nice @joelostblom! If it is not to complicated, can you still see how these signals got named within Altair 5?

Styling: is this definition also possible?:

```python
f"{subchart.spec.name}_{i}"
```

Instead of

```python
f"{subchart.spec.name[0:-1]}{i}"
```

joelostblom:
Sure! It seems like the v5 type naming was there until #3851 and it appears that previously the charts were named `view_1`, `view_2`, etc. There [was a get_name class with a counter](https://github.com/vega/altair/pull/3851/changes#diff-9bb4b3e6736266a4e9a6238a50c744810e9d0d7aa85b7548cb9b1effe33f6c12L3966-L3969) which was [called after what seems to be the same _needs_name check we have currently](https://github.com/vega/altair/pull/3851/changes#diff-9bb4b3e6736266a4e9a6238a50c744810e9d0d7aa85b7548cb9b1effe33f6c12L5002).

Regarding the syntax, `subchart.spec.name` is already suffixed with `_{i}` at the point I'm doing the replacement here, so changing to `f"{subchart.spec.name}_{i}"` would results in names with two underscores (except for the first name since the replacement only activated if `i > 0`):

```
'view_7d5fdb59e52741c4_0'
'view_7d5fdb59e52741c4_0_1'
'view_7d5fdb59e52741c4_0_2'
```

mattijn:
Added this commit to split on underscore instead of trailing: https://github.com/vega/altair/pull/3964/commits/e3dac48303b52f13f7355f1b212887d7d7ac7277. I don't think there will be charts with views in the double digits, but just to be sure. You have added new tests, and all tests still pass, so maybe indeed good enough. Activated auto-merge👍

mattijn:
@joelostblom:
I was reviewing https://github.com/vega/altair/pull/3965: and was able to introduce an error that is originating from this PR:

```python
import altair as alt
import pandas as pd

data = pd.DataFrame({"x": [1, 2], "y": [3, 4], "row": ["a", "b"], "col": ["c", "d"]})

base = alt.Chart(data).encode(x="x:Q", y="y:Q", row="row:N", column="col:N")
selection = alt.selection_interval(bind="scales")
(
    (base.mark_point() + base.mark_line()) | (base.mark_point() + base.mark_line())
).add_params(selection)
```

> ```python
> ---------------------------------------------------------------------------
> AttributeError                            Traceback (most recent call last)
> File ~/vega/altair/altair/vegalite/v6/api.py:2108, in TopLevelMixin.to_dict(self, validate, format, ignore, context)
>    2105 context.setdefault("datasets", {})
>    2106 is_top_level = context.get("top_level", True)
> -> 2108 copy = _top_schema_base(self).copy(deep=False)
>    2109 original_data = getattr(copy, "data", Undefined)
>    2110 if not utils.is_undefined(original_data):
> 
> File ~/vega/altair/altair/utils/schemapi.py:1135, in SchemaBase.copy(self, deep, ignore)
>    1133     return cast("Self", _deep_copy(self, set(ignore) if ignore else set()))
>    1134 with debug_mode(False):
> -> 1135     copy = self.__class__(*self._args, **self._kwds)
>    1136 if _is_iterable(deep):
>    1137     for attr in deep:
> 
> File ~/vega/altair/altair/vegalite/v6/api.py:4613, in HConcatChart.__init__(self, data, hconcat, **kwargs)
>    4611 self.data: Optional[ChartDataType]
>    4612 self.data, self.hconcat = _combine_subchart_data(self.data, self.hconcat)
> -> 4613 self.params, self.hconcat = _combine_subchart_params(self.params, self.hconcat)
> 
> File ~/vega/altair/altair/vegalite/v6/api.py:5240, in _combine_subchart_params(params, subcharts)
>    5238 # Increment names properly for concatenated facet charts to ensure uniqueness
>    5239 elif i > 0 and isinstance(subchart, FacetChart):
> -> 5240     name_prefix, _ = subchart.spec.name.rsplit("_", 1)
>    5241     subchart.spec.name = f"{name_prefix}_{i}"
>    5243 for param in subchart.params:
> 
> AttributeError: 'UndefinedType' object has no attribute 'rsplit'
> ```
> ```cmd
> alt.HConcatChart(...)
> ```

mattijn:
Probably issue in just this PR and not related to #3965, see my [comment](https://github.com/vega/altair/pull/3965#issuecomment-3961147937) there:

> I think it is unrelated to this PR, since I also can introduce it through the conventional matter of layering facetted charts:
> 
> ```python
> import altair as alt
> import pandas as pd
> 
> data = pd.DataFrame({"x": [1, 2], "y": [3, 4], "row": ["a", "b"], "col": ["c", "d"]})
> 
> manual = alt.layer(
>     alt.Chart(data).mark_point().encode(x="x:Q", y="y:Q"),
>     alt.Chart(data).mark_line().encode(x="x:Q", y="y:Q"),
> ).facet(row="row:N", column="col:N")
> 
> selection = alt.selection_interval(bind="scales")
> 
> (manual | manual).add_params(selection)
> ```
> 
> > ```python
> > AttributeError: 'UndefinedType' object has no attribute 'rsplit'
> > ```

PR #3966, opened 2026-02-25
fix: typing issues for mypy 1.18 and above.

### LINKED ISSUE #3962: with mypy ≥ 1.18 some mytype errors appear
### What happened?

With mypy ≥ 1.18, mypy reports: *Argument "filename" to "compile" has incompatible type "str | Buffer | PathLike[Any]"; expected "str | bytes | PathLike[Any]"*.

On eg code like here:

https://github.com/vega/altair/blob/f5cad857074158b4fa7182bd874e54de11b595dd/altair/utils/execeval.py#L81

and

https://github.com/vega/altair/blob/f5cad857074158b4fa7182bd874e54de11b595dd/altair/utils/execeval.py#L87

Error is reproducible with this snippet (`repro_mypy.py`):

```python
import ast
from typing import Any
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from _typeshed import ReadableBuffer
    from os import PathLike
filename: str | ReadableBuffer | PathLike[Any] = "<string>"
compile(ast.Module([], []), filename=filename, mode="exec")
```

- reproduce:

```bash
uv run --with "mypy==1.19.0" mypy repro_mypy.py
# error

uv run --with "mypy==1.17.0" mypy repro_mypy.py
# success
```

To also support higher versions of mypy. If you like to work on this, please do not hesitate in doing so. It might be a good idea to first check mypy repository https://github.com/python/mypy. Maybe you can find the PR that introduced this behaviour, with hopefully a proposed migration path.

_No response_

alec-bike:
Upgrading mypy to 1.18 or higher results in the following warnings:

```sh
altair/utils/execeval.py:80: error: Argument "filename" to "compile" has incompatible type:
   "str | Buffer | PathLike[Any]"
expected:
   "str | bytes | PathLike[Any]"  [arg-type]

   compiled = compile(ast.Module([node], []), filename=filename, mode="exec")
                                                       ^~~~~~~~
```

This PR address this issue. Fixes #3962.


alec-bike:
I updated `uv.lock` to remove the upper-bound on mypy. I also bumped the `arro3-core` dependency to `0.8.0` (`0.7.0` had been yanked).

@mattijn this PR is ready to review.

mattijn:
Looks great @alec-bike! Thanks for the PR!

PR #3936, opened 2026-01-08
fix: type issues with Chart mark methods.

alec-bike:
Chart `mark_line` and similar methods trigger a basedpyright type check warning for unknown argument. This PR adds type annotations to remove the warning. It also fixes some minor type warnings raised by ty and basedpyright.

Resolves #3870.

PR #3967, opened 2026-02-26
chore: Simplify `ruff` config, fix minor type issues

alec-bike:
This PR tidies up some ruff rules and comments in `pyproject.toml`:
- ~~increase max-complexity to 15 (allows removal of some `noqa: C901` annotations)~~  reverted
- remove commented-out preview-only ruff rules[^1]

Also, the experimental `tool.pyright` section has been simplified (and tested against pyright and basedpyright).

[^1]: ruff preview was turned off in #3936

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -354,20 +354,13 @@ disable_error_code = ["annotation-unchecked"]
 
 [tool.pyright]
 enableExperimentalFeatures = true
-extraPaths = ["./tools"]
 pythonPlatform = "All"
 pythonVersion = "3.10"
 reportTypedDictNotRequiredAccess = "none"
 reportIncompatibleMethodOverride = "none"
 reportUnusedExpression = "none"
 reportUnsupportedDunderAll = "none"
-include = [
-    "./altair/**/*.py",
-    "./doc/*.py",
-    "./tests/**/*.py",
-    "./tools/**/*.py",
-    "./sphinxext/**/*.py",
-]
+include = ["altair", "doc", "tests", "tools", "sphinxext"]
 ignore = [
     "./altair/vegalite/v6/schema/channels.py",  # 716 warns
     "./altair/vegalite/v6/schema/mixins.py",    # 1001 warns


pyproject.toml
```diff
@@ -183,16 +180,16 @@ ignore = [
...
-    "E501",   # line-too-long (https://docs.astral.sh/ruff/formatter/#conflicting-lint-rules)
+    "E501",   # line-too-long
     "RUF012", # mutable-class-default
     "RUF052", # used-dummy-variable
-    "SIM105", # suppressible-exception (https://github.com/vega/altair/pull/3431#discussion_r1629808660)
+    "SIM105", # suppressible-exception (https://github.com/vega/altair/pull/3431)
     "W505",   # doc-line-too-long
 ]
-mccabe.max-complexity = 10
+mccabe.max-complexity = 15
```
dangotbanned: Please revert this change 🙏 

10 is the [default](https://docs.astral.sh/ruff/settings/#lint_mccabe_max-complexity) for ([`C901`](https://docs.astral.sh/ruff/rules/complex-structure/)), and increasing the value encourages writing unmaintainable code.

> Functions with a high complexity are hard to understand and maintain.

The *moment of pause* that adding an ignore comment gives is the aim of lint rules.
Do you need to adhere to every rule every time? Nope!
But being aware of these issues gives you a chance to say one of:
> Oh hey, thanks yeah I'll fix that
> You've got a point, but not right now
> Stop being pedantic

It also highlights things to discuss during review. 
For this rule a reviewer is less likely to understand a function with that much context to hold in their head

alec-bike:
Thanks for the review @dangotbanned.

> Could you set up `pyright` to run in CI _first_, before making changes like this please?

I've been wondering about the status of pyright (and multiple type checkers) in altair. Is the goal for altair to support both mypy and pyright in CI?

I've been running ty for my testing, using a local `ty.toml` config to keep it separate from `pyproject.toml`. Would altair also want to run ty in CI?

dangotbanned:
https://github.com/vega/altair/pull/3967#issuecomment-3998816779

> I've been wondering about the status of pyright (and multiple type checkers) in altair. Is the goal for altair to support both mypy and pyright in CI?

I guess this is a history of the progress I made with `pyright` support:

- https://github.com/vega/altair/pulls?q=sort%3Aupdated-desc+is%3Apr+author%3Adangotbanned+is%3Amerged+typing

Before that, there was a long road to supporting `mypy`:

- https://github.com/vega/altair/pulls?q=sort%3Aupdated-desc+is%3Apr+mypy+is%3Amerged+author%3Abinste

This section of the config was where there were still gaps, but IIRC some test modules were an issue as well:

https://github.com/vega/altair/blob/356c78bbe8cd41118ff3ac963737c844584cd032/pyproject.toml#L374-L378


> I've been running `ty` for my testing, using a local `ty.toml` config to keep it separate from `pyproject.toml`. Would altair also want to run ty in CI?

I'm very much looking forward to using `ty` when it is ready, but there are many gaps that impact an upstream project I contribute to:

- [ ] https://github.com/narwhals-dev/narwhals/issues/3409


Using `ty` locally is probably the best move for now - I have no doubt it will play a really important role in moving python forward.

Running `pyright` in `ci` is quite doable - with the right amount of effort.
If you look at [Python Type System Conformance Test Results](https://htmlpreview.github.io/?https://github.com/python/typing/blob/main/conformance/results/results.html) - it really is the bar for following the rules 😄 
Wish it was faster though

dangotbanned (approved):
Thanks @alec-bike, sorry for the delay!

cat-ptain on a narwhal

PR #3969, opened 2026-03-01
docs: add identity line (x=y diagonal) example to rule mark docs

### LINKED ISSUE #3317: Include example of identity line (ie infinite diagonal line of x=y)
I [answered this on SO](https://stackoverflow.com/questions/77859278/how-to-draw-a-diagonal-line-with-mark-rule-that-stretches-to-infinity-in-altair/77862151#77862151) and thought it could be good to include this in the docs

Feel free to add a PR to the [rule page](https://altair-viz.github.io/user_guide/marks/rule.html#width-height-spanning-rules)

## Answer

Yes, by using `alt.value` you can anchor the line to a pixel value rather than a value of the data axis. So even if you zoom out and the axis extent change, the line will span the same pixel distance. You can either hardcode the heigth/width of the chart or use the special signals for this like so:

```python
import altair as alt

source = data.cars.url

points = alt.Chart(source).mark_circle(size=60).encode(
    x='Horsepower:Q',
    y='Miles_per_Gallon:Q',
)

line_black = alt.Chart().mark_rule().encode(
    x=alt.value('width'),
    x2=alt.value(0),
    y2=alt.value(0),
    y=alt.value('height')
)

line_red = alt.Chart().mark_rule(color='red').encode(
    x=alt.value(0),
    x2=alt.value('width'),
    y=alt.value('height'),
    y2=alt.value(0)
)


(points + line_black + line_red).interactive()
```

Zoomed out:

[![enter image description here][2]][2]

The above would create the line x=y only if the x and y scale's domain spans the same range. If that is not the case (like for the data above), you have two options:

## 1. Manually change the scale domain to be the same

We can set the y=axis to have the same max as the xaxis:

```py
points = alt.Chart(source).mark_circle(size=60).encode(
    x='Horsepower:Q',
    y=alt.Y('Miles_per_Gallon:Q').scale(domainMax=240),
)
```

[![enter image description here][3]][3]

## 2. Compute the lines as the ratio of the original domains

I'm not sure if there is a way to automatically find the max value of the y and x domain, but that would make this more robust (maybe look around in the Vega-Lite issue tracker or ask there). For now, I just plug in the values manually:

```python
line_black = alt.Chart().mark_rule().encode(
    x=alt.value(0),
    y=alt.value(0),
    x2=alt.value(alt.expr('width * 50/240')),
    y2=alt.value('height'),
)

line_red = alt.Chart().mark_rule(color='red').encode(
    x=alt.value(0),
    y=alt.value('height'),
    x2=alt.value(alt.expr('width * 50/240')),
    y2=alt.value(0),
)
```

[![enter image description here][4]][4]

  [1]: https://i.sstatic.net/zaw37.png
  [2]: https://i.sstatic.net/ZiDUi.png
  [3]: https://i.sstatic.net/ZunH1.png
  [4]: https://i.sstatic.net/SnGDw.png

stbiadmin:
Adds a new example to the rule mark documentation page showing how to draw an identity line (x=y diagonal) that spans the full chart.

The example uses `alt.value` with the special `'width'` and `'height'` signals to anchor the line to pixel coordinates, so it spans correctly regardless of data extent. A dashed line style differentiates it from data marks.

The example is placed under a new "Identity Line" subsection after the existing "Ranged Rules" section, as suggested in the issue.

Closes #3317

### DIFF
--- a/altair/datasets/__init__.py
+++ b/altair/datasets/__init__.py
@@ -77,7 +77,7 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING
 
-from altair.datasets._loader import Loader as Loader
+from altair.datasets._loader import Loader
 
 if TYPE_CHECKING:
     import sys
@@ -91,6 +91,8 @@ if TYPE_CHECKING:
     from altair.datasets._data import DataObject
     from altair.datasets._loader import _Load
     from altair.datasets._typing import Dataset, Extension
+
+__all__ = ["Loader", "data", "load", "url"]
 
 
 load: _Load[Any, Any]
@@ -166,18 +168,17 @@ def url(
     return url
 
 
-def __getattr__(name):
-    if name == "data":
-        from altair.datasets._data import data
+if not TYPE_CHECKING:
 
-        return data
-    elif name == "load":
-        from altair.datasets._loader import load
+    def __getattr__(name):
+        if name == "data":
+            from altair.datasets._data import data
 
-        return load
-    elif name == "__all__":
-        # Define __all__ dynamically to avoid ruff errors
-        return ["Loader", "data", "load", "url"]
-    else:
-        msg = f"module {__name__!r} has no attribute {name!r}"
-        raise AttributeError(msg)
+            return data
+        elif name == "load":
+            from altair.datasets._loader import load
+
+            return load
+        else:
+            msg = f"module {__name__!r} has no attribute {name!r}"
+            raise AttributeError(msg)
--- a/altair/datasets/_cache.py
+++ b/altair/datasets/_cache.py
@@ -286,7 +286,7 @@ class SchemaCache(CompressedCache["_Dataset", "_FlSchema"]):
                     # For pyarrow CSV reading, use the schema as intended
                     # This will fail for non-ISO date formats, but that's the correct behavior
                     # Users can handle this by using a different backend or converting dates manually
-                    return {"convert_options": ConvertOptions(column_types=schema)}  # pyright: ignore[reportCallIssue]
+                    return {"convert_options": ConvertOptions(column_types=schema)}
                 elif suffix == ".parquet":
                     return {"schema": schema}
 
--- a/altair/datasets/_loader.py
+++ b/altair/datasets/_loader.py
@@ -3,10 +3,8 @@ from __future__ import annotations
 import typing as t
 from typing import Generic, final, overload
 
-from narwhals.stable.v1.typing import IntoDataFrameT
-
 from altair.datasets import _reader
-from altair.datasets._reader import IntoFrameT
+from altair.datasets._reader import IntoDataFrameT, IntoLazyFrameT
 
 if t.TYPE_CHECKING:
     import sys
@@ -30,7 +28,7 @@ if t.TYPE_CHECKING:
 __all__ = ["Loader", "load"]
 
 
-class Loader(Generic[IntoDataFrameT, IntoFrameT]):
+class Loader(Generic[IntoDataFrameT, IntoLazyFrameT]):
     """
     Load example datasets *remotely* from `vega-datasets`_, with caching.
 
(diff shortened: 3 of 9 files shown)


mattijn:
Thanks for working on this @stbiadmin! Really valuable addition to the docs! 
When I copy+paste the example you added, it shows for me like this:

image

Would it be possible to use a dataset where the scatter follows more or less the identity line? Maybe @joelostblom has some ideas as well, as author of https://github.com/vega/altair/issues/3317 and answering https://stackoverflow.com/questions/77859278/how-to-draw-a-diagonal-line-with-mark-rule-that-stretches-to-infinity-in-altair/77862151#77862151. 🙌

stbiadmin:
Oops, yeah that doesn't look good. I've updated the example to use a predicted vs actual scatter so the points cluster around the identity line and the diagonal is more meaningful. HTH.

joelostblom:
Thanks for working on this @stbiadmin! I think the new example data you added is a much more natural fit than the previous one. As I revisited this example, (a)I noticed another way of drawing the line that creates a true identity line (x=y) that is robust to both zooming and panning the chart:

```py
line = alt.Chart().mark_rule().encode(
    x=alt.value(alt.expr("scale('x', domain('x')[0])")),
    y=alt.value(alt.expr("scale('y', domain('x')[0])")),
    x2=alt.value(alt.expr("scale('x', domain('x')[1])")),
    y2=alt.value(alt.expr("scale('y', domain('x')[1])")),
)
```

This is robust to both zooming and panning of interactive charts,  since it uses an expression that re-computes the scale of where to draw the line as the axis domains change.

I think create a diagonal line in pixel units (like the current approach) can still be useful in some situation so maybe we can include two examples/headings here: the first one showing how to draw a diagonal line in a chart using the current approach in the PR (not sure what is an appropriate data set for this...) and the next one showing how to draw an identity line which is more appropriate when comparing e.g. actual vs predicted values. What do you think?

stbiadmin:
Yeah that makes way more sense as two separate examples. I split it up, first one is a basic diagonal using pixel values with the cars data, second uses alt.expr scale expressions so the line tracks x=y through zoom/pan. Let me know if that looks right.

doc/user_guide/marks/rule.rst
```diff
@@ -142,3 +142,57 @@ For example, we can use ``y`` and ``y2`` show the ``"min"`` and ``"max"`` values
...
+        y=alt.value("height"),
+        x2=alt.value("width"),
+        y2=alt.value(0),
+    )
+
+    (points + diagonal).interactive()
+
+Identity Line (x=y)
+^^^^^^^^^^^^^^^^^^^^
+When comparing two variables on the same scale (e.g. predicted vs actual values), you can draw a true identity line using ``alt.expr`` with scale expressions. This line stays on x=y even when zooming and panning.
```
joelostblom:
```suggestion
When comparing two related variables (e.g. predicted vs actual values), drawing a line representing a perfect relationship (i.e. x=y) can facilitate comparisons between the variables. In contrast to the diagonal line above that was drawn in pixel space, we know need to draw the line in data space. Therefore we use ``alt.datum`` together with ``alt.expr``. If our chart is static, we could simply put in values for the line manually, but if we want to draw an identiy line that is robust to zooming and panning, we can instead compute the line dynamically using the ``domain`` expression, so that each time the range of the x or y domain changes, the line is redrawn.
```

doc/user_guide/marks/rule.rst
```diff
@@ -142,3 +142,57 @@ For example, we can use ``y`` and ``y2`` show the ``"min"`` and ``"max"`` values
...
+
+.. altair-plot::
+    import altair as alt
+    import pandas as pd
+    import numpy as np
+
+    rng = np.random.RandomState(42)
+    actual = rng.uniform(10, 90, 50)
+    predicted = actual + rng.normal(0, 8, 50)
```
joelostblom:
```suggestion
    predicted = actual + rng.normal(0, 12, 50)
```

Just to make the max value different between the axes to further emphasize that the line doesn't go from corner to corner

joelostblom (changes requested):
This is looking great @stbiadmin ! I suggest some clarifications and elaborations in my suggestions. Let me know if you have any questions or disagree with something I wrote. If not, you can just apply the suggestions and we can merge after.

PR #3970, opened 2026-03-04
chore: bump `narwhals>=2.4.0`

### LINKED ISSUE #3960: Narwhals typing breakage after lock file update (narwhals 2.2 / 2.9)
### What happened?

The following issue was generated by a code generator after updating my uv lock file.

---

When trying to update the lock file (`uv lock --upgrade`), I ran into narwhals-related mypy errors in the datasets layer.

**What happened**

After upgrading the lock file (including narwhals to a recent 2.x version), the test suite started failing with 21 mypy errors in the datasets backend.

**Where the errors appear**

- **`altair/datasets/_readimpl.py`**  
  - The `IntoFrameT` (and `R`) `TypeVar` bound that references `nwt.NativeFrame` and `nwt.DataFrameLike` (around lines 52–57).  
  - The `into_scan` helper and its use of `BaseImpl` / `Scan` with those type parameters (around lines 214–228).

- **`altair/datasets/_reader.py`**  
  - Uses of `Scan[IntoFrameT]` and `Reader[..., IntoFrameT]` (e.g. `_scan`, `scan_fn`, overloads and `reader` implementation).  
  - `_metadata_frame` and `_scan_metadata` return types and the `nw.from_native(...).lazy()` call.  
  - `_NoParquetReader._metadata_frame`, where `nw.from_dict(data, backend=impl)` is called (around line 390).  
  - `_steal_eager_parquet` return type (around line 542).

**Root cause: two narwhals typing changes**

1. **`from_dict(..., backend=impl)`**  
   From **narwhals 2.2.0**, the type stub for `from_dict`’s `backend` no longer accepts `nw.Implementation` (only `Literal[...] | Module | None`). Runtime is unchanged; only type-checking breaks.

2. **`nwt.NativeFrame` / `nwt.DataFrameLike`**  
   From **narwhals 2.9.0**, these names are no longer in the public `narwhals.stable.v1.typing` API. Code that uses them in a `TypeVar` bound gets a “name-defined” error.

**Minimal repro**

Save as `repro.py` and run:

```bash
uv run --with narwhals==2.16.0 mypy repro.py
```

```python
from __future__ import annotations
from typing import TYPE_CHECKING, Any

from narwhals.stable import v1 as nw
from narwhals.stable.v1.typing import IntoDataFrameT

if TYPE_CHECKING:
    from typing import TypeVar
    from narwhals.stable.v1 import typing as nwt
    IntoFrameT = TypeVar(
        "IntoFrameT",
        bound="nwt.NativeFrame | nw.DataFrame[Any] | nw.LazyFrame[Any] | nwt.DataFrameLike",
    )

def from_dict_with_impl(data: dict[str, Any], impl: nw.Implementation) -> nw.DataFrame[Any]:
    return nw.from_dict(data, backend=impl)
```

With narwhals 2.16.0 this yields:

- `name-defined`: `nwt.NativeFrame` is not defined  
- `arg-type`: `backend` to `from_dict` has incompatible type `Implementation`  

With narwhals 1.46.0 or 2.1.0, mypy reports no issues.

I was hoping to get no type issues when bumping the version of narwhals

main

dangotbanned:
Closes #3960

## Related
- https://github.com/narwhals-dev/narwhals/issues/3013
- https://github.com/narwhals-dev/narwhals/pull/2944
- https://github.com/narwhals-dev/narwhals/pull/2971
- https://github.com/narwhals-dev/narwhals/pull/3016
- https://github.com/narwhals-dev/narwhals/pull/3086
- https://github.com/narwhals-dev/narwhals/issues/3498

## Description
Bumps `narwhals>=1.27.1` -> `2.4.0` and resolves the typing yells mentioned in the linked issues.

I merged (#3973) ahead of this since it fixed some other issues that were on main already

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -313,12 +313,10 @@ norecursedirs = [
     "tests/examples_arguments_syntax",
     "tests/examples_methods_syntax"
 ]
+testpaths = ["tests","altair","tools"]
 addopts = [
     "--numprocesses=logical",
     "--doctest-modules",
-    "tests",
-    "altair",
-    "tools",
     "-m not datasets_debug and not no_xdist",
 ]
 # https://docs.pytest.org/en/stable/how-to/mark.html#registering-marks


altair/datasets/__init__.py
```diff
@@ -166,18 +168,17 @@ def url(
     return url
 
 
-def __getattr__(name):
-    if name == "data":
-        from altair.datasets._data import data
+if not TYPE_CHECKING:
```
dangotbanned: This was something I learned later, see me spreading the gospel here https://github.com/apache/arrow/pull/48618#discussion_r2720997888

mattijn:
Great work @dangotbanned! Out of curiosity, if we move to narwhals v2+, would it make sense to also switch to `narwhals.stable.v2`? Not suggesting we do it here, just wondering.

dangotbanned:
> Great work @dangotbanned!

Thanks @mattijn 


> Out of curiosity, if we move to narwhals v2+, would it make sense to also switch to `narwhals.stable.v2`?

Yeah definitely! 
There is some `ruff` config we need to update to `v2`, along with the actual imports

> Not suggesting we do it here, just wondering.

Yeah agreed, should definitely be it's own PR - but one that I'd welcome 😊

Issue #3024, opened 2023-04-17
naming parameter views of repeated charts + interactivity


mattijn:
I was re-reading this comment https://github.com/vega/vega-lite/issues/8446#issuecomment-1437517310, but **I'm not** able to get
- _repeat charts + interactivity_

to work anymore.

The following `views` definition:  `<view-name>child__<repeat-row>_<data-field><repeat-column>_<data-field>` does not give interactivity in the charts with VL5.7.

See also this example where the interactivity (here zooming through `"bind": "scales"`) does not work anymore [Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUATJQAMlAFYQ2AOxAAaEFBUAzGmTShiNOAHcDmlZhrKGbBhAAScPVjQBmGTI1blVm3YgAdRoAE3p3TwBfSI0cJAAnJAQINABtUCZrELQQaCRBOBSNZSS4HLjEhAB9AEZ1XLgCqExzOGUtEOsyFNRUkAAPeoBPEABdDUwhnDL0a0w4eOJ8kBiQI1MevvWTWtgaQRCqqvi2HenMfIAZNrJ6LUEGBGUqiDg4wWvlW5h67d2YfaHY6nKrnK43O5sB5PF5vfLBMI-DR-Gp7A5HE5nOAXQQIyHQ56vd6fb6-Yw7VEA9HArE4vEwe6PQlw3GhcKjSLjEDxN5wJAtVCgRlPNINYkQpFi+Fsn5czGisEfCX1RX0sarCDTKDmEoIGZrcm1eohfkoQUgBjxQQ5LC4fBEKAhZQKCAhRo0YjxSjKbEEZQ4BC8Cj8E0XV6YCAAAWINSkAE5ZARQ0gCDR4jQIC6VCsNMh4gBrcpsOb1NodLrme5seLmXSNbLoTVwKDGIogSbTHLKNgIazLVaDc11g7mHnTfk5YWqVYd-UARwYSF8dH5HrKqxGQ+MI-NY75Au5pxz7am88Xy4uVlIK1W8BcApqHg0JhlaBqADYZNEgA)

[...]

Issue #2849, opened 2023-01-23
naming parameter views of layered repeated charts + interactivity


mattijn:
As discussed in this comment: https://github.com/vega/vega-lite/issues/8348#issuecomment-1399528866. It is not really clear how we implemented naming of view in top-level parameters for a layered repeater.

This was partly discussed before here: https://github.com/altair-viz/altair/pull/2684#issuecomment-1287596272, and potentially initially implemented in this commit: https://github.com/altair-viz/altair/pull/2684/commits/f5473234a43b054e778541665544086895d65d9f (this commit was not merged), but this PR https://github.com/altair-viz/altair/pull/2702 was merged where the actual change is included.

Moreover, the behavior we implemented previously does work in the following situation, -example 1- (using altair from main branch on github):

* * *
_Example 1_
```python
import altair as alt
from vega_datasets import data
iris = data.iris.url

alt.Chart(iris).mark_point().encode(
    alt.X(alt.repeat("column"), type='quantitative'),
    alt.Y(alt.repeat("row"), type='quantitative'),
    color='species:N'
).properties(
    width=160,
    height=130
).repeat(
    row=['petalLength', 'petalWidth'],
    column=['sepalLength', 'sepalWidth']
).interactive().to_dict()
```
```json
{
  "$schema": "https://vega.github.io/schema/vega-lite/v5.2.0.json",
  "config": {
    "view": {
      "continuousHeight": 300,
      "continuousWidth": 300
    }
  },
  "params": [
    {
      "bind": "scales",
      "name": "param_20",
      "select": {
        "encodings": [
          "x",
          "y"
        ],
        "type": "interval"
      },
      "views": [
        "view_24child__row_petalLengthcolumn_sepalLength",
        "view_24child__row_petalLengthcolumn_sepalWidth",
        "view_24child__row_petalWidthcolumn_sepalLength",
        "view_24child__row_petalWidthcolumn_sepalWidth"
      ]
    }
  ],
  "repeat": {
    "column": [
      "sepalLength",
      "sepalWidth"
    ],
    "row": [
      "petalLength",
      "petalWidth"
    ]
  },
  "spec": {
    "data": {
      "url": "https://cdn.jsdelivr.net/npm/vega-datasets@v1.29.0/data/iris.json"
    },
    "encoding": {
      "color": {
        "field": "species",
        "type": "nominal"
      },
      "x": {
        "field": {
          "repeat": "column"
        },
        "type": "quantitative"
      },
      "y": {
[...]

PR #3974, opened 2026-03-04
fix: unique view names per concat cell and correct param.views

### LINKED ISSUE #3954: Error when concatenating interactive faceted charts

mattijn:
After some hours digging.

**What’s going wrong (v6.0.0):** View and param names in v6.0.0 use content-based hashing for thread-safety. When you concatenate two (or more) **identical** faceted charts, they get the **same** content hash, so they all receive the **same** view name (e.g. `view_9f1740faf6e66b88_0`). As a result, each param’s `views` array ends up pointing at that single view instead of one view per concat cell. The compiler then sees duplicate or incorrect bindings and can error or behave wrongly.

**What v5 did:** v5 used a simple counter for view names (`view_5`, `view_6`, …), so each concat cell had a **distinct** name and each param could correctly list its views (e.g. `param.views = ["view_5", "view_6"]` for a vconcat of two faceted charts).

**Fix direction:** Keep content-based hashing, but in the **concat** case (multiple sibling subcharts), disambiguate by **position**: assign each subchart a unique view name, e.g. `f"{base}_{i}"` where `i` is the concat index. For a FacetChart whose inner spec is a single Chart, put the name on `spec.name`; for a FacetChart whose inner spec is a LayerChart, put it on the **first layer** (so `param.views` can point at those). In `_combine_subchart_params`, when creating a new param entry (APPEND) we use only this subchart’s view for that entry’s `views` list (not the param’s existing `views`), so params that share or reuse view lists don’t get the wrong views. That restores the intended one-view-per-concat-cell behaviour and fixes the error.

Fix #3954

### DIFF
--- a/altair/vegalite/v6/api.py
+++ b/altair/vegalite/v6/api.py
@@ -5204,44 +5204,30 @@ def _remove_duplicate_params(layer: list[ChartType]) -> list[ChartType]:
     return subcharts
 
 
-def _view_name_base(name: str | None) -> str:
-    """Strip a trailing _<digits> from a view name to get a reusable base."""
-    if not name:
-        return "view_0"
+def _view_base_for_chart(obj: Any) -> str:
+    """Return a base view name for a chart/layer (for building position-based names)."""
+    name = obj.name
     parts = name.rsplit("_", 1)
     if len(parts) == 2 and parts[1].isdigit():
         return parts[0] or name
     return name
 
 
-def _view_base_for_chart(obj: Any) -> str:
-    """Return a base view name for a chart/layer (for building position-based names)."""
-    name = getattr(obj, "name", None)
-    if name is None or name is Undefined:
-        name = (
-            obj._get_view_hash_name() if hasattr(obj, "_get_view_hash_name") else None
-        )
-    return _view_name_base(name)
-
-
-def _view_name_for_param(
-    subchart: ChartType, param_idx: int, is_concat: bool
-) -> str | None:
-    """View name for this subchart to add to a param's views, or None."""
+def _view_name_for_param(subchart: ChartType, is_concat: bool) -> str:
+    """View name for this subchart to add to a param's views."""
     if isinstance(subchart, Chart):
         return subchart.name
     if (
         is_concat
         and isinstance(subchart, FacetChart)
         and subchart.spec is not Undefined
-        and param_idx == 0
     ):
         spec = subchart.spec
         if isinstance(spec, Chart):
-            return getattr(spec, "name", None)
+            return spec.name
         if isinstance(spec, LayerChart) and spec.layer:
-            return getattr(spec.layer[0], "name", None)
-    return None
+            return spec.layer[0].name
+    return ""
 
 
 def _combine_subchart_params(  # noqa: C901
@@ -5288,7 +5274,7 @@ def _combine_subchart_params(  # noqa: C901
             elif isinstance(spec, Chart):
                 spec.name = f"{_view_base_for_chart(spec)}_{i}"
 
-        for param_idx, param in enumerate(subchart.params):
+        for param in subchart.params:
             p = _prepare_to_lift(param)
             pd = _viewless_dict(p)
 
@@ -5304,7 +5290,7 @@ def _combine_subchart_params(  # noqa: C901
 
             # At this stage in the loop, p must be a TopLevelSelectionParameter.
             # Get this subchart's view name from the subchart only (not p.views: params can share lists).
-            view_to_add = _view_name_for_param(subchart, param_idx, is_concat)
+            view_to_add = _view_name_for_param(subchart, is_concat)
             # MERGE: start from param's views; APPEND: start from [] so we don't pull in another param's views.
             views_after = list(p.views or []) if found else []
             if view_to_add and view_to_add not in views_after:
--- a/altair/vegalite/v6/api.py
+++ b/altair/vegalite/v6/api.py
@@ -5219,9 +5219,7 @@ def _view_base_for_chart(obj: Any) -> str:
     name = getattr(obj, "name", None)
     if name is None or name is Undefined:
         name = (
-            obj._get_view_hash_name()
-            if hasattr(obj, "_get_view_hash_name")
-            else None
+            obj._get_view_hash_name() if hasattr(obj, "_get_view_hash_name") else None
         )
     return _view_name_base(name)
 
@@ -5261,7 +5259,9 @@ def _combine_subchart_params(  # noqa: C901
         views = (
             []
             if isinstance(p, core.VariableParameter)
-            else list(p.views) if p.views else []
+            else list(p.views)
+            if p.views
+            else []
         )
         param_info.append((p, _viewless_dict(p), views))
 
--- a/altair/vegalite/v6/api.py
+++ b/altair/vegalite/v6/api.py
@@ -5217,11 +5217,7 @@ def _view_name_for_param(subchart: ChartType, is_concat: bool) -> str:
     """View name for this subchart to add to a param's views."""
     if isinstance(subchart, Chart):
         return subchart.name
-    if (
-        is_concat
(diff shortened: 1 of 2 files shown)


mattijn:
Willing to review @joelostblom🙃? Its getting a bit complicated with hashing and adding parameters and do proper lifting..

mattijn:
A bit of explanation using VL-JSON snippets.

#### Test 1: three faceted charts, three params (p1, p2, p3), 
test-name: `test_concat_faceted_three_params_unique_views_per_param_issue_3954`

**v5 (correct, but no hashing)** - each concat cell has a distinct view name; each param lists only its cell’s view:

```json
"params": [
  {
    "name": "p1",
    "select": { "type": "point" },
    "views": ["view_1"]
  },
  {
    "name": "p2",
    "select": { "type": "point" },
    "views": ["view_2"]
  },
  {
    "name": "p3",
    "select": { "type": "point" },
    "views": ["view_3"]
  }
]
```

vconcat cells: `spec.name` is `"view_1"`, `"view_2"`, `"view_3"` respectively.

---

**v6 before fix (incorrect)** - all three cells get the same view name; all three params list that one view:

```json
"params": [
  {
    "name": "p1",
    "select": { "type": "point" },
    "views": ["view_7d5fdb59e52741c4_0"]
  },
  {
    "name": "p2",
    "select": { "type": "point" },
    "views": ["view_7d5fdb59e52741c4_0"]
  },
  {
    "name": "p3",
    "select": { "type": "point" },
    "views": ["view_7d5fdb59e52741c4_0"]
  }
]
```

vconcat cells: every `spec.name` is `"view_7d5fdb59e52741c4_0"`.

---

**v6 with this PR** - one view per cell; each param lists only its cell’s view:

```json
"params": [
  {
    "name": "p1",
    "select": { "type": "point" },
    "views": ["view_7d5fdb59e52741c4_0"]
  },
  {
    "name": "p2",
    "select": { "type": "point" },
    "views": ["view_7d5fdb59e52741c4_1"]
  },
  {
    "name": "p3",
    "select": { "type": "point" },
    "views": ["view_7d5fdb59e52741c4_2"]
  }
]
```

vconcat cells: `spec.name` is `"view_7d5fdb59e52741c4_0"`, `"view_7d5fdb59e52741c4_1"`, `"view_7d5fdb59e52741c4_2"` respectively.

---

####  Test 2: two faceted charts, one shared param (selection_interval bind=scales), 
test-name `test_concat_faceted_shared_param_both_views_issue_3954`

**v5 (correct, but no hashing)** - one param listing both cell views; each cell has a distinct layer view name:

```json
"params": [
  {
    "bind": "scales",
    "name": "param_3",
    "select": { "type": "interval" },
    "views": ["view_5", "view_6"]
  }
]
```

vconcat: first cell `spec.layer[0].name` = `"view_5"`, second = `"view_6"`.

---

**v6 before fix (incorrect)** - both cells share the same view name; param lists only that one view:

```json
"params": [
  {
    "bind": "scales",
    "name": "param_f268f798f41c853e",
    "select": { "type": "interval" },
    "views": ["view_9f1740faf6e66b88_0"]
  }
]
```

vconcat: both cells have `spec.layer[0].name` = `"view_9f1740faf6e66b88_0"`.

---

**v6 with this PR** - each cell has a distinct view name; param lists both:

```json
"params": [
  {
    "bind": "scales",
    "name": "param_f268f798f41c853e",
    "select": { "type": "interval" },
    "views": ["view_9f1740faf6e66b88_0", "view_9f1740faf6e66b88_1"]
  }
]
```

vconcat: first cell `spec.layer[0].name` = `"view_9f1740faf6e66b88_0"`, second = `"view_9f1740faf6e66b88_1"`.

joelostblom (commented):
Thanks for putting the time and effort into digging in to this @mattijn! I think your approach makes sense overall and is definitely more proper than the fix we tried earlier. Just a couple of comments / questions.

mattijn:
There is one thing that I didn't doublecheck, I'm not sure if our test-suite sufficiently covers what happens when facing a chart using repeat + interactivity or a chart with a layered repeat + interactivity, see also https://github.com/vega/vega-lite/issues/6890#issuecomment-1266821631, https://github.com/vega/altair/issues/2849 and https://github.com/vega/altair/issues/3024.

joelostblom (approved):
Great job @mattijn! I think the updates fix the previous issues and it is also cleaner to read.

I'm not sure we need a test for repeat... I guess it's always good to be comprehensive but do we expect something different to happen with the parameters when a repeat chart is involved?

mattijn:
I've double-checked what I liked to do before merging. I was wondering if the following chart still functions, which has some complicated naming of views:

```python
import altair as alt

source = alt.UrlData(
    data.flights_2k.url,
    format={'parse': {'date': 'date'}}
)

brush = alt.selection_interval(encodings=['x'])

# Define the base chart, with the common parts of the
# background and highlights
base = alt.Chart(source).mark_bar().encode(
    x=alt.X(alt.repeat('column'), type='quantitative', bin=alt.Bin(maxbins=20)),
    y='count()'
).properties(
    width=160,
    height=130
)

# gray background with selection
background = base.encode(
    color=alt.value('#ddd')
).add_params(brush)

# blue highlights on the transformed data
highlight = base.transform_filter(brush)

# layer the two charts & repeat
chart = alt.layer(
    background,
    highlight,
    data=source
).transform_calculate(
    "time",
    "hours(datum.date)"
).repeat(column=["distance", "delay", "time"])
chart
```
image

```python
chart.to_dict()
```

> ```python
> {'config': {'view': {'continuousWidth': 300, 'continuousHeight': 300}},
>  'repeat': {'column': ['distance', 'delay', 'time']},
>  'spec': {'layer': [{'mark': {'type': 'bar'},
>     'encoding': {'color': {'value': '#ddd'},
>      'x': {'bin': {'maxbins': 20},
>       'field': {'repeat': 'column'},
>       'type': 'quantitative'},
>      'y': {'aggregate': 'count', 'type': 'quantitative'}},
>     'name': 'view_0b48749199e89fbe_0'},
>    {'mark': {'type': 'bar'},
>     'encoding': {'x': {'bin': {'maxbins': 20},
>       'field': {'repeat': 'column'},
>       'type': 'quantitative'},
>      'y': {'aggregate': 'count', 'type': 'quantitative'}},
>     'transform': [{'filter': {'param': 'param_c97a99acfc68b794'}}]}],
>   'data': {'url': 'https://cdn.jsdelivr.net/npm/vega-datasets@v3.2.1/data/flights-2k.json',
>    'format': {'parse': {'date': 'date'}}},
>   'height': 130,
>   'transform': [{'calculate': 'hours(datum.date)', 'as': 'time'}],
>   'width': 160},
>  'params': [{'name': 'param_c97a99acfc68b794',
>    'select': {'type': 'interval', 'encodings': ['x']},
>    'views': ['child__column_distance_view_0b48749199e89fbe_0',
>     'child__column_delay_view_0b48749199e89fbe_0',
>     'child__column_time_view_0b48749199e89fbe_0']}],
>  '$schema': 'https://vega.github.io/schema/vega-lite/v6.1.0.json'}
> ```

All good from my side as well, merging then:) Thanks for reviewing @joelostblom!

Issue #3981, opened 2026-03-12
## Duplicate auto-generated `name` in vconcat causes line layers to render wrong data


jbloom:
### Description

When vconcatenating multiple layered charts (line + circle) that share the same structural pattern but have **different data**, Altair auto-generates the same `name` for each line layer. Vega-Lite treats `name` as a unique view identifier, so the last chart's data pipeline wins — all line layers render with the bottom row's data.

Circle marks are unaffected because conditional encodings change their spec hash, giving each a unique auto-generated name.

### Minimal reproducible example

See attached `altair_vconcat_duplicate_name_bug.py`, or inline:

```python
import altair as alt
import pandas as pd

df_peak = pd.DataFrame(
    {"site": [1, 2, 3, 4, 5], "escape": [0.1, 0.9, 0.1, 0.2, 0.1]}
)
df_valley = pd.DataFrame(
    {"site": [1, 2, 3, 4, 5], "escape": [0.9, 0.1, 0.9, 0.8, 0.9]}
)

selection = alt.selection_point(fields=["site"], on="mouseover", empty=False)

def make_layered_chart(df, title):
    base = alt.Chart(df).encode(
        x="site:O",
        y=alt.Y("escape:Q", scale=alt.Scale(domain=[0, 1])),
    )
    lines = base.mark_line(size=1)
    points = base.encode(
        size=alt.condition(selection, alt.value(80), alt.value(30)),
    ).mark_circle(filled=True)
    return (lines + points).add_params(selection).properties(
        title=title, width=300, height=150,
    )

chart = alt.vconcat(
    make_layered_chart(df_peak, "Peak at site 2"),
    make_layered_chart(df_valley, "Valley at site 2"),
)
chart.save("bug.html")
```

### Expected behavior

Each row's line layer should display its own data. The top row should show a peak at site 2 and the bottom row should show a valley.

### Actual behavior

Both rows' line layers display the bottom row's data (valley pattern). The circle/point marks display correctly.

Inspecting the generated Vega-Lite spec shows the cause — both line layers receive the same auto-generated `name`:

```
vconcat[0] layer[0] mark=line    name=view_f61e5eb42c4e42eb_0   <-- duplicate
vconcat[1] layer[0] mark=line    name=view_f61e5eb42c4e42eb_0   <-- duplicate
```

### Workaround

Assign a unique `name` to each line layer via `.properties(name=...)`:

```python
lines = base.mark_line(size=1).properties(name=f"line_{unique_suffix}")
```

[...]

PR #3982, opened 2026-03-13
fix: Compute unique hashes for identical specs that use different data

joelostblom:
close https://github.com/vega/altair/issues/3981

This PR is largely from agentic coding; I thought it would be interesting to see how helpful that can be for us in fixing bugs quicker (but still appropriately). I only did minor fixups on the suggested solution and gave some guidance. An alternative approach that was suggested was to go through the entire spec at the end and fix duplicated names by appending a suffix, but I think the approach in this PR is more robust. Another alternative could be that we restructure the flow on a larger scale so that hashes are computed before hoisting data, but that would be a larger restructure.

One potential issue with taking the data into account now is that we could change some hashes computed with the current version of altair, so if someone for some reason was relying on a certain spec giving a particular hash, that hash might now have been updated (but note that the updated hash would be more correct; overall I think this is a minor issue).

Here is a summary of the coding agent's "thoughts":

## Problem

When concatenating multiple layered charts with different data but the same structure, Altair auto-generated the same name for each layer, causing Vega-Lite to treat them as the same view. This resulted in all layers rendering with the last chart's data.

The `_compute_hash()` method computed hashes based on chart spec structure (encodings, mark type, transforms) but didn't include data identity. When `_combine_subchart_data()` hoisted data from subcharts, the data became Undefined for all charts, making them hash to the same value.

## Solution

1. Cache hash before data removal (altair/vegalite/v6/api.py):
   - In `_combine_subchart_data()`, compute and cache the hash in `_kwds["_cached_hash"]` before hoisting and removing data
   - Store in `_kwds` (not as attribute) so it survives copy() operations
2. Use cached hash (altair/vegalite/v6/api.py):
   - In `_compute_hash()`, check for `_cached_hash` first and return it if present
3. Exclude from serialization (altair/utils/schemapi.py):
   - Added `_data_hash` to the exclude list in to_dict() to prevent validation errors


jbloom:
I cannot effectively review the actual code changes, but confirming I have verified that this branch does solve the bug noted in #3981

mattijn (approved):

PR #3990, opened 2026-04-05
feat: Allow signals/params to be passed to alt.datum[...]

### LINKED ISSUE #3366: Support `alt.datum[signal_name]` in addition to `"datum[signal_name]"`
Currently using `alt.datum` in altair does not support passing a signal/parameter whereas writing the expression string directly as JS-like string does (an example of dynamically passing fields via parameters here  https://github.com/altair-viz/altair/pull/3357/files (both the chart title and calculate transform).

Here is another example:

```py
import altair as alt

dropdown = alt.binding_select(
    options=['Horsepower', 'Displacement', 'Weight_in_lbs', 'Acceleration'],
    name='X-axis column '
)
xcol_param = alt.param(
    value='Horsepower',
    bind=dropdown
)

alt.Chart(data.cars.url).mark_circle().encode(
    x=alt.X('x:Q').title(''),
    y='Miles_per_Gallon:Q',
    color='Origin:N'
).transform_calculate(
    x=f'datum[{xcol_param.name}]'
    # x=alt.datum[xcol_param]
).add_params(
    xcol_param
)
hist
```

The reason it fails is that `alt.datum[xcol_param.name]` inserts additional quotation around the parameter name in the VL spec:

```json
"transform": [{"calculate": "datum['param_1']", "as": "x"}],
```

whereas the VL spec for `f'datum[{xcol_param}]'` does not:

```json
"transform": [{"calculate": "datum[param_1]", "as": "x"}],
```

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4Igxg9gdgZglgcxALlANzgUwO4tJKAFzigFcJSBnAdTgBNCALFAZgAY2AacaYsiygAlMiRoVYcAvpO50AhoTl4QpAE4AbFCDGEADpWQB6Q2DpQAdACtKdTOrhpV5qJkKGougLaG0mBHIBaeUVKV0oAATQARnMAJgBOczZDYLkTOVVKK0poEBkQTwyAa2VCAE9dTC0wOFUwdSr8zChIOhIkVB51CFVleDs6LQB5VUQSEG5yyq0oCE8SOU18gA8+rHVB5BBVybhCBq0JkCmqrYBHUjkiPYUHRu4ytYGtAFk4BsoAfUrVT4BxRbdKBHE5aC5XYiKYi+PL5XQZOSeSgoADaoCgiNOIHhqkRnyiRwARiRNqASLpSOItqEGmBxNwILpiNBkcgUSBBD1QroINhML1uAAROCUXTqORgTCeZr0kDUEQIMSfEifdSE5HcACCYElDVxzOBAF1uBjpVoABoBOTLEUAAkg6lInigtry3DQi1IWM5mUwPL5vUkxuOuKglBgPU8qPwizApHFhCxwSdKJxeKihqOclZ2zywYAJJQwIwpUotjp9EYfH45OYEHtGKRCeY4BBDEWS4Vq-4AvZEz4AKzmKIANnMLGyuUkQA)

Personally I think the cleanest would be if we could support just using the variable name without the `.name` suffix when using `alt.datum`, as in [this example from the docs](https://altair-viz.github.io/user_guide/interactions.html#expressions): 

```
```

But even if that doesn't work, just supporting the use of the `.name` suffix without the quotes inserted would be great. Here are a few examples of how these datums and expressions get translated with differential quoting:

image

joelostblom:
close https://github.com/vega/altair/issues/3366, ie it is now possible to e.g. do:

```py
transform_calculate(x=alt.datum[xcol_param])
```

instead of 

```py
transform_calculate(x=f'datum[{xcol_param.name}]')
```

After this, I think we can consider using alt.expr as the default way to represent expressions in the docs instead of the js strings (but that can be a follow up PR).


altair/expr/core.py
```diff
@@ -281,7 +281,7 @@ def __init__(self, group, name) -> None:
         super().__init__(group=group, name=name)
 
     def __repr__(self) -> str:
-        return f"{self.group}[{self.name!r}]"
+        return f"{self.group}[{_js_repr(self.name)}]"
```
joelostblom: `!r` calls Python's repr() unconditionally. For a string "foo" that gives 'foo' (quoted), which is correct for a literal string key. For a Parameter object it gives the full Python repr like `Parameter('xcol', VariableParameter({...}))` — which is meaningless as a Vega expression. Switching to `_js_repr` will instead call `_to_expr()`, which is overridden by `Parameter` to return `self.name` (the bare string "xcol"), which is exactly what Vega expects as a signal reference. For plain strings and numbers, _js_repr falls through to repr(), so those are unchanged.

Neither me, nor my agent can think of any notable caveat here: strings are rendered as before so there is no change there. There are some new nonsensical constructs that are now possible, like passing dictkeys and datetime objects, but none of them represent an actual use case I believe and they still fail as expected on the Vega-Lite level.

<details><summary>Some details agent notes about these cases:</summary>

1. Any OperatorMixin can now be a dict key. Beyond Parameter, things like BinaryExpression or FunctionExpression also extend OperatorMixin. So datum[datum.x + 1] would now render as datum[(datum.x + 1)], which is actually valid Vega — but it's an unusual pattern nobody has likely tested. Previously it would have rendered as datum[<Expression object repr>] which was broken anyway, so this is strictly an improvement.
2. _js_repr on a plain integer returns repr(int) = e.g. "0". So datum[0] renders as datum[0] — same as before since 0!r is also "0". No change.
3. _js_repr on True/False/None returns "true"/"false"/"null". So datum[True] would now render as datum[true] instead of datum[True]. This is arguably more correct JavaScript, but it's an edge case nobody would realistically write.
4. _js_repr on a dt.date object would render it as a datetime(...) Vega call rather than Python's repr. Again, datum[some_date] is not a realistic use case.
None of the caveats are likely to matter in practice — the only realistic index types are strings (field name literals, unchanged) and Parameter/OperatorMixin objects (signal references, now fixed). The existing test suite catches the string case explicitly.

</details>

mattijn:
Lgtm 👍

mattijn (approved):

PR #3997, opened 2026-04-07
docs: Speed up doc build time for faster local development

joelostblom:
I think editing the docs and rebuilding them is somewhat of a frustrating experience due to the long build times. On my machine, a clean build is somewhere around 6 min 30 s and even edits that just include small textual changes on a single page will trigger a complete rebuild of the gallery which results in a build time of a few minutes.

This PR alleviates these issues by offering a "fast" build mode where the gallery and the autogenerated API docs are not built, which shaves 6 min off the build time bringing it down to around 30 s. This PR also speeds up the incremental builds to just a few seconds by only rebuilding gallery examples if they have changed; so editing textual content on a page in the user docs will no longer rebuild the entire gallery.

These options are for fast local development of the docs and the full docs will still rebuild as part of the tests running on each PR, so I think there is a low risk that this would introduce any errors that we don't catch during PR review. I also explored using multi CPU in sphinx, but didn't see as big changes as for the strategy described above. 

I will explore auto triggering incremental build and hot reloading of the webserver in a follow up PR.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -55,8 +55,7 @@ autodoc_typehints = "none"
 # generate autosummary pages (set ALTAIR_AUTOSUMMARY_GENERATE=0 for faster local builds)
 autosummary_generate = os.environ.get("ALTAIR_AUTOSUMMARY_GENERATE", "1") != "0"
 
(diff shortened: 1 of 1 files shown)


joelostblom:
Merging this since it is only updating internal convenience functions.

PR #3999, opened 2026-04-08
docs: Highlight new examples in the gallery

joelostblom:
When we have new gallery examples contributed, I think it's unfortunate that they get lost/buried in the many other examples on the gallery page. This PR adds support for a `# :new:` tag in gallery example files, which is then translated into some CSS highlighting effects plus added to a section on top of the docs called "Recently Added":

image

New examples are also present in their respective categorical sections, where they are highlighted further with a (NEW) pill/tag:

image

I'm hoping this can draw more attention to the new examples and think it can be especially impactful since our gallery page is one of the most frequently visited in the docs. For now, I chose to highlight a few neat newish examples that have been contributed somewhat recently.

### DIFF
--- a/doc/_static/altair-gallery.css
+++ b/doc/_static/altair-gallery.css
@@ -132,25 +132,25 @@ div.bottomnav {
 .gallery .image:hover {
   box-shadow:
     0 0 14px rgba(110, 116, 124, 0.5),
-    0 0 26px rgba(110, 116, 124, 0.3);
+    0 0 28px rgba(110, 116, 124, 0.3);
 }
 
 .gallery .imagegroup-new .image {
   box-shadow:
     0 0 14px rgba(255, 143, 63, 0.56),
-    0 0 26px rgba(255, 143, 63, 0.38);
+    0 0 28px rgba(255, 143, 63, 0.38);
 }
 
 .gallery .imagegroup-new .image:hover {
   box-shadow:
-    0 0 16px rgba(255, 143, 63, 0.7),
-    0 0 32px rgba(255, 143, 63, 0.45);
+    0 0 17px rgba(255, 143, 63, 0.7),
+    0 0 34px rgba(255, 143, 63, 0.45);
 }
 
 .gallery .imagegroup-recent .image:hover {
   box-shadow:
-    0 0 16px rgba(255, 143, 63, 0.7),
-    0 0 32px rgba(255, 143, 63, 0.45);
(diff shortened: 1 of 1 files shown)


mattijn:
Awesomeness😎🤩👏

dsmedia:
Very nice @joelostblom!

Just to flag a related effort: vega/vega-datasets#776 builds a meta-dataset cross-referencing ~400 examples across the Vega, Vega-Lite, and Altair galleries, mapping each to its datasets, categories, and descriptions. The motivation is similar — examples are a huge resource but hard to see as a whole. Having them indexed opens up things like spotting coverage gaps (which features lack gallery examples?), identifying where new datasets could better showcase capabilities, and giving tools a way to surface relevant examples in context. 

For Vega and Vega-Lite the ingestion is straightforward — structured JSON indexes and specs that can be walked mechanically. **Altair is the trickiest leg** since I don't believe there's a structured index of examples: it discovers examples by listing `tests/examples_methods_syntax/` via the GitHub API, then pattern-matches dataset references and parses metadata from docstrings and `# category:` comments in the source. This PR's `# :new:` comments don't affect any of that, but wanted to flag the dependency on that directory's structure and conventions in case it's useful to know about.

joelostblom:
Thank you @dsmedia! It's been exciting to also watch all the great work you have been doing with vega-datasets package and repo to get them into such good shape! It seems like https://github.com/vega/vega-datasets/pull/776 is another intriguing step in that direction and it will be cool to see some visualizations of that gallery metadata as per the different scenarios you have imagined. I replied in more detail over in that PR.

mattijn:
Looks great! Quick question, how do you decide what counts as "recently added"? Based on date?
Also, really like the blur effect, could see that working well on the other examples too!

joelostblom:
For what is recently added, I manually checked which examples were added in the last two years that looked especially cool. I don't think we need a strict cutoff but can choose to have some up there for longer if we think they are especially neat and would be instructive for readers.

As for the background shadow/blur, I was thinking that it makes the new examples stand out from the rest, but maybe that is not needed now that we have a separate section on top for recently added and also the "new" pill. I will try applying the shadow to everything and seeing how it looks

joelostblom:
I updated to add the grey background shadow everywhere and make the shadows a bit more subtle overall. New examples are highlighted in grey that switched to orange in the recently added section, but elsewhere have an orange background from the beginning to make them standout more:

image


image

mattijn:
Looks great!

PR #4000, opened 2026-04-08
docs: Create docs CLI to facilitate different build combinations

joelostblom:
Continuation of #3998. I think the current amount of available build options for the docs becomes more manageable and flexible with a small CLI wrapper like this. Now we can combine different docs options via flags instead of having unique `uv run task ...` commands for each combination. I also made it clearer that the most convenient way to develop locally is via

```
uv run task doc-build -- --clean --no-autosummary --watch
```

Regarding the implementation, I went with argparse since I didn't want to take on an extra dependency and this is just some internal tooling for relatively self-explanatory options when building the docs. We could have gone with something fancier like cyclopts or typer, but I think this is sufficient for this purpose.

### DIFF
--- a/tools/docs_cli.py
+++ b/tools/docs_cli.py
@@ -147,22 +147,22 @@ def parse_args() -> argparse.Namespace:
     build_parser = subparsers.add_parser(
         "build",
         help="Build docs with sphinx",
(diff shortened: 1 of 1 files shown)


joelostblom:
Again, merging since its internal tooling and tests pass. PR/Issue 4000! :tada:

PR #4002, opened 2026-04-08
docs: Insert relevant "Learn more" links in gallery examples

### LINKED ISSUE #4001: Link from each gallery example to the relevant part of the user docs
I think our gallery page is a common landing point for people searching for the altair docs. A common usage scenario is likely that a reader looks for something close to the chart they want to make, and then modifies that example. That's how I started out using visualization libraries myself and how I still often use them, and also also what I see from many students etc. The popularity of the gallery page is also backed up by our Plausible analytics, where it is the second most visited page after the home page.

I think our gallery has many great examples, but that we are not helping readers in regards to where they can learn more about the concepts included in a particular example. We have put noticeable effort into making our docs easy to read and to explain the many concepts in Altair as well as we can, but I think adding links from each gallery example to the relevant sections in the User Docs would help with discoverability and give readers a clear "Where to go next".

joelostblom:
close https://github.com/vega/altair/issues/4001, see there for some more background motivation.

This PR includes both direct backlinks to parts of the user docs that explicitly references a gallery examples. These backlinks will also work with the proposed file-reading capabilities of the `..altair-plot` directive in https://github.com/vega/sphinxext-altair/pull/14.

In addition to links to directly referencing pages, I've also included some heuristics for potentially relevant pages, such as always linking the docs of a mark or transform that is used in the example. I'm sure there are more that can be added than what I thought of but this could at least be a good starting point (especially the interactivity links could become more fine grained).

It will be difficult to evaluate how effective this is at shepherding readers to other relevant pages in the docs, but at least personally, this is something I would find useful if I wanted to learn more about a specific example. I also find it minimally disturbing, since it is at the end of each page:

image

image


dsmedia:
Nice work @joelostblom and I fully spport the discoverability motivation you outlined in #4001.

Since you're already working in `populate_examples()` and computing both the example metadata and the doc links: would you be open to extending this PR to also persist that to a JSON file (e.g., `_data/examples.json`)?

Right now everything is ephemeral — computed during the Sphinx build, rendered into HTML, discarded. Persisting it would make it available to external consumers as a reusable artifact, similar to how VL already publishes [`site/_data/examples.json`](https://github.com/vega/vega-lite/blob/main/site/_data/examples.json).

Two fields in particular would be valuable beyond the Sphinx build:

**`related_docs`** — the user-guide links your PR already computes (both the explicit backrefs and the heuristic links). Persisting these means any consumer — vega-datasets, search tools, LLMs — can map examples to relevant documentation without reimplementing the detection logic.

**`datasets`** — which vega-datasets resources each example uses. [vega/vega-datasets#776](https://github.com/vega/vega-datasets/pull/776) builds a cross-gallery examples registry mapping ~400 Vega, VL, and Altair examples to datasets. For Vega and VL the generator walks JSON specs mechanically. For Altair it currently scrapes via the Trees API, parses docstrings with regex, and extracts dataset references by pattern-matching `data.cars()` / `data.cars.url` against source text — ~180 lines of fragile code. `datasets` is the one field the generator can't reliably derive from outside: the regex catches today's three patterns, but for examples that use `data.cars()` (returns a DataFrame), there's no way to recover the dataset name from the compiled spec since it inlines as `values`.

I'm imagining something like:

```json
[
  {
    "name": "selection_detail",
    "title": "Selection Detail",
    "description": "Shows how to use a selection to filter a detail view.",
    "categories": ["interactive"],
    "path": "tests/examples_methods_syntax/selection_detail.py",
    "related_docs": [
      "user_guide/interactions/index",
      "user_guide/encodings/index#datum-and-value"
    ],
    "datasets": ["cars"]
  }
]
```

The base fields (name, title, description, categories, path) are what `populate_examples()` already knows, and `related_docs` is what this PR already computes. `datasets` would be a new addition that only altair can provide authoritatively — it requires resolving which `data.X()` calls each example makes, which is new logic.

I know adding dataset extraction might be scope creep for this PR. But even just persisting what you've already built here (`related_docs` + base metadata) to a JSON file would lay the exact foundation we need. `datasets` could come in a follow-up.

This connects to your [comment on vega-datasets#776](https://github.com/vega/vega-datasets/pull/776#issuecomment-4224699147) about consuming the gallery examples registry from altair's sphinx extension — a persisted index would work in both directions. On the vega-datasets side I'm thinking about publishing a [Table Schema](https://datapackage.org/standard/table-schema/) contract that defines a standard shape for gallery example metadata across the vega org — so each library's index can be validated in CI and consumed uniformly. cc @domoritz since this touches VL's existing `examples.json` pattern and the vega-datasets schema work.

joelostblom:
Thanks for the detailed comment @dsmedia! I think it is a great idea to expand on this and formalize the connections/links from the gallery examples into a standardized format that can also be used by the gallery discovery features you are working on (and other initiatives like you mentioned). I will have a go at it and see if I can adapt this PR to persist the currently created metadata and also include the dataset used in each example. I did do some changes to the sphinx altair extension to as part of this effort, but I think they would just require some tweaking to work with a persisted JSON file instead.

joelostblom:
Sorry for taking so long to come back to this @dsmedia . I made an attempt to include your suggestion and persist the example metadata to a file. Would you be able to review this PR and let me know what you think of this approach and if you would like to see any changes?

dsmedia:
Looks great, @joelostblom. I compared the `datasets` field against what we currently scrape in vega/vega-datasets#776 and they match on all 185 examples the two files share, so this would let me delete my custom scraping entirely. A few thoughts:

### CI

Consider adding something to check that the committed `_data/examples.json` matches what the docs build produces. If someone adds a gallery example and doesn't rebuild, the file quietly falls behind and nothing notices.

The docs build already regenerates the file, and `docbuild.yml` already runs on every PR that touches the examples, so one line after the build step should cover it:

```yaml
    - name: Check examples index is up to date
      run: git diff --exit-code -- _data/examples.json
```

It just fails if the committed file and the newly generated one differ.

### Shape

I'd suggest something like:

```json
{
  "gallery": {
    "name": "altair",
    "repository": "https://github.com/vega/altair",
    "docs_url": "https://altair-viz.github.io/"
  },
  "examples": [
    {
      "name": "airport_connections",
      "title": "Connections Among U.S. Airports Interactive",
      "url": "https://altair-viz.github.io/gallery/airport_connections.html",
      "path": "tests/examples_arguments_syntax/airport_connections.py",
      "description": "This example shows all the connections between major U.S. airports.",
      "categories": ["case studies"],
      "datasets": ["airports", "flights_airport", "us_10m"],
      "related_docs": {
        "explicit": ["user_guide/encodings/channels#position"],
        "inferred": ["user_guide/marks/circle"]
      }
    }
  ]
}
```

The wrapper gives somewhere to put gallery-level info and to add `$schema` later without flipping the document from an array to an object. And once the shape settles I'd like to propose a small schema shared across the Altair, Vega and VL indexes. That block just has the document's identity plus what's needed to resolve the relative fields inside it: `repository` for `path`, `docs_url` for `related_docs`.

The part I'm least sure about is splitting `related_docs`, or at least whether `explicit` and `inferred` are the right names for the split. But something along these lines seems worth doing. Most of the links in there are generated by scanning the example's code: call `.mark_circle()` and the circle page gets added. Only a small number are pages that actually reference the example, about 28 out of 996 entries. Merged into one list there's no way to tell them apart. You already compute the two separately and only merge at the end, so hopefully it's a small change either way.

The per-example `url` is so consumers don't need to know Altair's URL conventions, and `docs_url` + docname + `.html` resolves the `related_docs` entries, which I checked.

joelostblom:
Thanks for the thoughtful review and for confirming that the dataset assignments match all 185 shared examples in `vega-datasets`!

I think I've included all your suggestions now, here are some details on each one:

### CI

The docs workflow now checks that the committed index matches the output generated by the documentation build:

```yaml
- name: Check examples index is up to date
  run: |
    if ! git diff --exit-code -- _data/examples.json; then
      echo "::error file=_data/examples.json::The examples index is stale. Run 'uv run task doc-build --no-autosummary' and commit the updated _data/examples.json."
      exit 1
    fi
```

I also added `_data/**` and `sphinxext/**` to the workflow paths so changes to either the artifact or its generator trigger the check.

### Document Shape

The index is now wrapped in an object with gallery-level metadata:

```json
{
  "gallery": {
    "name": "altair",
    "repository": "https://github.com/vega/altair",
    "docs_url": "https://altair-viz.github.io/"
  },
  "examples": []
}
```

This provides the context needed to resolve repository paths and documentation references, while leaving room for a future top-level `$schema`.

Each example now also includes its canonical gallery URL:

```json
"url": "https://altair-viz.github.io/gallery/airport_connections.html"
```

### Related Documentation

I split `related_docs` into `explicit` and `inferred`:

```json
"related_docs": {
  "explicit": [
    "user_guide/encodings/channels#position"
  ],
  "inferred": [
    "user_guide/marks/circle"
  ]
}
```

I kept the proposed names because they describe the provenance without exposing implementation details. If a relationship is found through both methods, it is retained only as `explicit`, since that is the stronger signal and avoids requiring consumers to deduplicate them.

dsmedia (approved):

PR #4005, opened 2026-04-13
feat: Update to Vega-Lite 6.4.1

joelostblom:
Pull in the updates that has occurred between Vega-Lite 6.1 and 6.4.1, vlconvert 1.8 to 1.9, vega-embed 7 to 7.0.2


joelostblom:
@jonmmease I'm updating the vlconvert before making a new release for Altair and there seems to be [an issue with the test](https://github.com/vega/altair/actions/runs/24354372695/job/71117249861) `tests/utils/test_mimebundle.py::test_vegalite_to_vega_mimebundle`. The new outputted Vega is a bit different than the old one:

```diff
  -                             'signal': '"a: " + (isValid(datum["a"]) ? datum["a"] : '
  ?                                                                       -------------
  +                             'signal': '"a: " + (isValid(datum["a"]) ? '
  +                             'isArray(datum["a"]) ? join(datum["a"], \' \') : '
  -                             '""+datum["a"]) + "; b: " + (format(datum["b"], '
  ?                                                         --------------------
  +                             'datum["a"] : ""+datum["a"]) + "; b: " + '
  ?                              +++++++++++++
  -                             '""))',
  +                             '(format(datum["b"], ""))',
  ?                              ++++++++++++++++++++
```

Is this an expected change and something we should update in the test on the altair side of things?

mattijn:
It's not uncommon that these test results change in altair when bumping versions of Vega /vegalite. If the Vega output is still functions as the vegalite input in the editor, then I would adapt the test.

jonmmease:
I'm not familiar with the exact background, but this looks like a reasonable change in Vega-Lite where arrays are now accepted and joined on space characters. You're fine to update the baseline

joelostblom:
Great, thanks both for chiming in! It seemed harmless but I wanted to make sure before going ahead with the release. I just pushed a fix to the test and will go ahead and merge this since all tests are now passing. I will try to make the release later today.

PR #4015, opened 2026-04-21
feat: Add ty type-check support

alec-bike:
This PR adds initial ty type-check support: 
- add ty dependency and configuration to pyproject.toml
- add ty to `type-check` task in pyproject.toml
- add ty to GitHub lint workflow
- add some `ty: ignore` to flag issues that do not yet have a fix
- fix minor type issues, remove some `Union`
- **remove some mypy-specific annotations** from `type: ignore` (see note below)

Limitations:
- include only some folders under altair
- exclude generated schema files

## mypy-specific annotations

Annotations such as `type: ignore[arg-type]` are not processed by ty. Removing the annotation (to generic `type: ignore`) does work with ty and should with most other type checkers. 

There are discussions around cross-compatibility between mypy and ty (see https://github.com/astral-sh/ty/issues/3127), which would obviate this issue in some cases. 

Alternatively, ty-specific annotations could be added, eg `type: ignore[arg-type, ty:invalid-argument-type]`, but this verbose and requires annotations for each new type checker.

This PR reverts to the generic `type: ignore` by removing some existing mypy-specific annotations in `altair/vegalite/v6`.


alec-bike:
@dangotbanned and @mattijn, could you take a quick look at this PR? In particular, the removal of some mypy-specific type annotations in `vegalite/v6`. 

I'll plan to merge early next week, if no objections.

mattijn:
Thanks @alec-bike! I went over the changes one by one, I like it that we can use the pipe operator, `|` more often than using the `Union`.

mattijn (approved):

PR #4019, opened 2026-05-06
fix: Propagate params to outer level after add_params in concat charts

### LINKED ISSUE #4018: add_params(selection_interval(bind="scales")) on vconcat of hconcat drops first subchart's views
### What happened?

This is a regression in Altair 6. This worked in Altair 5.

I'm concatenating several plots and binding their scales so I can zoom simultaneously.

```python
import altair as alt
import pandas as pd

data = pd.DataFrame({"x": [1, 2, 3], "y": [1, 2, 3]})
top = alt.hconcat(
    alt.Chart(data).mark_line().encode(x="x:Q", y="y:Q").properties(name="top_1"),
    alt.Chart(data).mark_line().encode(x="x:Q", y="y:Q").properties(name="top_2"),
)
bot = alt.hconcat(
    alt.Chart(data).mark_line().encode(x="x:Q", y=alt.Y("y:Q", title="B")).properties(name="bot_1"),
    alt.Chart(data).mark_line().encode(x="x:Q", y=alt.Y("y:Q", title="B")).properties(name="bot_2"),
)
chart = alt.vconcat(top, bot).add_params(alt.selection_interval(bind="scales", name="zoom"))
print(chart.to_dict()["params"][0]["views"])
# Expected: ['top_1', 'top_2', 'bot_1', 'bot_2']
# Actual:   ['bot_1', 'bot_2']   ← top_* missing
```

The chart looks fine:

Image

However, the missing views mean that some features, e.g. zooming doesn't work for the top charts.

When you run
```python
chart.interactive()
```
you can see that there's only zooming in the bottom row.

I would like to see the top plots also in the views (so that zooming works).

6.1.0

jtilly:
Fixes #4018.

When calling `add_params` on a nested concat (e.g. `vconcat` of `hconcat` charts), params were not propagated to the outer level, causing `bind="scales"` zoom to silently apply only to the innermost charts.

Two changes:

1. For `ConcatChart`, `HConcatChart`, and `VConcatChart`, the `add_params` method now calls `_combine_subchart_params` after delegating to children, so `params` are propagated up to the outer level with the correct `views` list.
2. `_combine_subchart_params` now preserves already-aggregated `views` on a compound subchart (one with no single view name), rather than discarding them.

From the issue:

```python
import altair as alt
import pandas as pd

data = pd.DataFrame({"x": [1, 2, 3], "y": [1, 2, 3]})
top = alt.hconcat(
    alt.Chart(data).mark_line().encode(x="x:Q", y="y:Q").properties(name="top_1"),
    alt.Chart(data).mark_line().encode(x="x:Q", y="y:Q").properties(name="top_2"),
)
bot = alt.hconcat(
    alt.Chart(data).mark_line().encode(x="x:Q", y=alt.Y("y:Q", title="B")).properties(name="bot_1"),
    alt.Chart(data).mark_line().encode(x="x:Q", y=alt.Y("y:Q", title="B")).properties(name="bot_2"),
)
chart = alt.vconcat(top, bot).add_params(alt.selection_interval(bind="scales", name="zoom"))
chart.save("chart.html")

print(chart.to_dict()["params"][0]["views"])
# Expected: ['top_1', 'top_2', 'bot_1', 'bot_2']
# Actual:   ['bot_1', 'bot_2']   ← top_* missing
```

now produces

https://github.com/user-attachments/assets/ce32d589-7cb3-499a-a142-61e13f92f2a2

### DIFF
--- a/altair/vegalite/v6/api.py
+++ b/altair/vegalite/v6/api.py
@@ -4577,6 +4577,7 @@ class ConcatChart(TopLevelMixin, core.TopLevelConcatSpec):
             return self
         copy = self.copy()
         copy.concat = [chart.add_params(*params) for chart in copy.concat]
+        copy.params, copy.concat = _combine_subchart_params(copy.params, copy.concat)
         return copy
 
     @utils.deprecated(version="5.0.0", alternative="add_params")
@@ -4681,6 +4682,7 @@ class HConcatChart(TopLevelMixin, core.TopLevelHConcatSpec):
             return self
         copy = self.copy()
         copy.hconcat = [chart.add_params(*params) for chart in copy.hconcat]
+        copy.params, copy.hconcat = _combine_subchart_params(copy.params, copy.hconcat)
         return copy
 
     @utils.deprecated(version="5.0.0", alternative="add_params")
@@ -4787,6 +4789,7 @@ class VConcatChart(TopLevelMixin, core.TopLevelVConcatSpec):
             return self
         copy = self.copy()
         copy.vconcat = [chart.add_params(*params) for chart in copy.vconcat]
+        copy.params, copy.vconcat = _combine_subchart_params(copy.params, copy.vconcat)
         return copy
 
     @utils.deprecated(version="5.0.0", alternative="add_params")
@@ -5301,8 +5304,10 @@ def _combine_subchart_params(  # noqa: C901
             # At this stage in the loop, p must be a TopLevelSelectionParameter.
             # Get this subchart's view name from the subchart only (not p.views: params can share lists).
             view_to_add = _view_name_for_param(subchart, is_concat)
-            # MERGE: start from param's views; APPEND: start from [] so we don't pull in another param's views.
-            views_after = list(p.views or []) if found else []
+            # MERGE (found=True): start from p.views to accumulate all views for this param.
+            # APPEND (found=False, view_to_add set): start from [] to avoid pulling in sibling views.
+            # COMPOUND (found=False, no view_to_add): subchart already aggregated views into p.views; preserve them.
+            views_after = list(p.views or []) if (found or not view_to_add) else []
             if view_to_add and view_to_add not in views_after:
                 views_after.append(view_to_add)
 
--- a/tests/vegalite/v6/test_api.py
+++ b/tests/vegalite/v6/test_api.py
@@ -1494,6 +1494,24 @@ def test_compound_add_selections(charttype):
     assert chart1.to_dict() == chart2.to_dict()
 
 
+def test_nested_concat_add_params_views():
+    # Regression test for https://github.com/vega/altair/issues/4018
+    base = alt.Chart("data.csv").mark_line()
+    top = alt.hconcat(
+        base.encode(x="x:Q", y="y:Q").properties(name="top_1"),
+        base.encode(x="x:Q", y="y:Q").properties(name="top_2"),
+    )
+    bot = alt.hconcat(
+        base.encode(x="x:Q", y=alt.Y("y:Q", title="B")).properties(name="bot_1"),
+        base.encode(x="x:Q", y=alt.Y("y:Q", title="B")).properties(name="bot_2"),
+    )
+    chart = alt.vconcat(top, bot).add_params(
+        alt.selection_interval(bind="scales", name="zoom")
+    )
+    views = chart.to_dict()["params"][0]["views"]
+    assert views == ["top_1", "top_2", "bot_1", "bot_2"]
+
+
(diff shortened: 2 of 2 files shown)


mattijn:
Thanks for the PR @jtilly! From the code diff it looks great, but will like to try it out in coming week as well.

knowecho:
This PR also fixes a related case: `vconcat` of layered panels (`alt.layer(...)`) with `add_params(selection_interval(bind="scales"))` on an inner layer of each panel. It has the same symptom (first panel's view missing from the top-level `views` list), although it's a bit different from the original reproduction in #4018: there `add_params` is on the outer concat container, while here it's on a layer inside each panel.

Minimal reproduction on altair 6.1.0 — prints `1`, should print `2`:

```python
import altair as alt
import polars as pl
import warnings

frame = pl.DataFrame({"x": [0.0, 1.0, 2.0, 3.0], "y1": [1, 2, 3, 4], "y2": [4, 3, 2, 1]})
sel = alt.selection_interval(name="zoom", bind="scales", encodings=["x"])
panels = []
for col in ("y1", "y2"):
    extent = alt.Chart(frame).mark_rule(opacity=0).encode(alt.X("x"))
    line = alt.Chart(frame).mark_line().encode(alt.X("x"), alt.Y(col)).add_params(sel)
    panels.append(alt.layer(extent, line))
    warnings.simplefilter("ignore")
    spec = alt.vconcat(*panels).to_dict()
print(len(spec["params"][0]["views"]))
```

Tested on this branch (`jtilly:issue-4018`): prints `2` - fixed.

Is it worth adding a regression test for this case too?

knowecho:
Actually I've found one related case that still fails on this branch: wrapping the above in a nested `vconcat` (e.g. `vconcat(vconcat(layered_panels), vconcat(layered_panels))`) collapses the views back down to 1 per selection. Worth covering in this PR, or a follow-up?

Minimal reproduction (prints `1` per selection on this branch; expected `4`):

```python
import altair as alt
import polars as pl

frame = pl.DataFrame({
    "date": [1.0, 2.0, 3.0, 4.0], "other_x": [10.0, 20.0, 30.0, 40.0],
    "y": [1.0, 2.0, 3.0, 4.0],
})

def inner(x_field):
    sel = alt.selection_interval(name=x_field, bind="scales", encodings=["x"])
    panels = []
    for _ in range(2):
        extent = alt.Chart(frame).mark_rule(opacity=0).encode(alt.X(x_field))
        line = alt.Chart(frame).mark_line().encode(alt.X(x_field), alt.Y("y")).add_params(sel)
        panels.append(alt.layer(extent, line))
    return alt.vconcat(*panels)

spec = alt.vconcat(*(alt.vconcat(inner("date"), inner("other_x")) for _ in range(2))).to_dict()
for p in spec["params"]:
    print(p["name"], len(p["views"]))
```

mattijn:
Since Altair version 6, we have been introducing the ability to compile Altair specifications into thread-safe Vega-Lite specifications, meaning that rerunning the code consistently produce the same result.

That said, I’m still unclear if we are now fixing the root cause or symptoms of a bad implemented thread-safe model😓?

Thanks for diving deeper into more cases that are not working @knowecho, if you can find a fix for these, you may add them as a commit or post them as a comment in this thread so @jtilly or me can update the PR. Thanks again!

knowecho:
Thanks for the reply!

Unfortunately I probably won't be able to find a fix for these. I am a relatively new end-user of Altair and don't know much about its internals or Vega-Lite etc. I don't think I will have time to focus on this and be confident I am fixing it in the right way.

But I am happy to test any fixes, and give any other cases I see that don't work.

mattijn:
@knowecho I'm not seeing wrong behaviour in your comment with describing a related case (https://github.com/vega/altair/pull/4019#issuecomment-4485118554), from what I see, with this PR, is that the x-scale zooming behaviour is done per different x-field. And that is what is happening from what I see. But if you can explain me better why you would expect to see different behaviour please open a new issue describing what you mean.

@jtilly again, thanks for the PR! And apologies for the delay in reviewing. I think it's great as it is in current state and with respect to the linked issue #4018.

knowecho:
> @knowecho I'm not seeing wrong behaviour in your comment with describing a related case ([#4019 (comment)](https://github.com/vega/altair/pull/4019#issuecomment-4485118554))

I built altair main and tested this on my code above. The per-x-field zoom behaviour is correct - panels sharing an x-field zoom and pan together as expected.

I was misled before by seeing `len(p["views"]) == 1` - I thought this meant the panels weren't being linked, but I was wrong.

So I don't think there's anything left to fix here - thanks for your help.

Issue #1691, opened 2019-09-10
How to bin data before sending it to a chart?


KoenBal:
Hello,

I'm trying to build a bar chart with on the x-axis a binned result_perc.
Because I have too many records I have to bin them first using pandas.
I do this by using an IntervalIndex:
```
width = 10
bins = pd.IntervalIndex.from_breaks([round(x, 1) for x in list(np.linspace(-width, width, (width*10+1), endpoint=True))])
# aggregate per bin
aggregations = {"amount_eur": ['sum']}
df = df[['result_perc_bin', 'amount_eur']].groupby(by=['result_perc_bin']).agg(aggregations)
df = df.reset_index()
df.columns = df.columns.get_level_values(0)
```
I now have a df with reduced records which I can use to create a bar chart.
But Altair gives the error:

> TypeError: Object of type 'Interval' is not JSON serializable

To fix this I convert the dtype of result_perc_bin (the IntervalIndex) which looks like this
```
IntervalIndex([(-10.0, -9.8], (-9.8, -9.6], (-9.6, -9.4], (-9.4, -9.2], (-9.2, -9.0] ... (9.0, 9.2], (9.2, 9.4], (9.4, 9.6], (9.6, 9.8], (9.8, 10.0]],
              closed='right',
              dtype='interval[float64]')
```
`
df.result_perc_bin = df.result_perc_bin.astype('str').astype('category')
`

But this conversion screws up the order of the labels, so I need a sort_order:

`result_perc_bin_sort_order = list(bins.astype(str))`

The code to create the chart:

```
        chart = alt.Chart(data=df).mark_bar(
                opacity=0.3,
                color=color
            ).encode(
                alt.Y(field='amount_eur',
                      type='quantitative',
                     ),
                alt.X(field= "result_perc_bin",
                      sort= result_perc_bin_sort_order,
                      type = 'ordinal'
                     ),
                tooltip=[
                    alt.Tooltip('amount_eur:Q', format='.2f'),
                ]
            )
```

Question 1:
Is there a better way to work with bins? A way where I don't have to back up the sort_order?

The above code actually solves my problem but it breaks when I try to merge 2 charts.
```
chart = (chart_1 + chart_2).encode(alt.X(field= "result_perc_bin",
                      sort= result_perc_bin_sort_order,
                      type = 'ordinal'
                     )
                ).properties(
                    title = "result_perc_bin - buy(blue), sell(red)",
                    width = 1000, 
                    height = 100,
                ) 
[...]

PR #4027, opened 2026-06-01
chore: update pandas-stubs pin for where typing fix

### LINKED ISSUE #3961: with pandas-stubs ≥ 2.3.3 some mytype errors appear
### What happened?

With pandas-stubs ≥ 2.3.3, mypy reports: *No overload variant of "where" of "Series" matches argument types "Series[bool]", "None"*.

On eg code like this (lines 394, 399, 407, 447, 456 in altair/utils/core.py):
https://github.com/vega/altair/blob/f5cad857074158b4fa7182bd874e54de11b595dd/altair/utils/core.py#L394

Error is reproducible with this snippet (`repro_pdstubs.py`):

```python
from typing import Any
import pandas as pd
col: pd.Series[Any] = pd.Series([1, 2, None])
col.where(col.notnull(), None)
```

```bash
uv run --with "pandas-stubs==3.0.0.260204" mypy repro_pdstubs.py
# error

uv run --with "pandas-stubs==2.3.2.250827" mypy repro_pdstubs.py
# success
```

To also support higher versions of pandas-stubs. If you like to work on this, please do not hesitate in doing so. It might be a good idea to first check pandas-stubs repository https://github.com/pandas-dev/pandas-stubs. Maybe you can find the PR that introduced this behaviour, with hopefully a proposed migration path.

_No response_

axelray-dev:
Fixes #3961

## What changed

Pinned `pandas-stubs` with Python version markers in `pyproject.toml`:

- `pandas-stubs<2.3.3` for `python_version < "3.11"`
- `pandas-stubs>=3.0.3.260530` for `python_version >= "3.11"`

The new pandas-stubs release (3.0.3.260530) requires Python 3.11+, so a single unconditional pin would break CI on Python 3.10. Version markers keep the old pin for 3.10 and pick up the typing fix for 3.11+.

## Why the new lower bound

pandas-stubs 2.3.3 changed `Series.where()` type stubs to no longer accept `None` as the second argument, causing mypy errors on 6 lines in `altair/utils/core.py` that use `col.where(col.notnull(), None)`.

The upstream fix was merged in [pandas-stubs PR #1691](https://github.com/pandas-dev/pandas-stubs/pull/1691) (allow `other=None` in `where`) and released in pandas-stubs 3.0.3.260530 (2026-05-30).

The lower bound of 3.0.3.260530 was chosen because it is the first release containing the upstream fix. Versions between 2.3.3 and 3.0.3 may still have the broken `where()` stubs, so `>=2.3.3` would not be correct.

## Tests/checks run

```
uv sync --all-extras                          # installed pandas-stubs-3.0.3.260530
mypy altair tests                             # no where() errors; 6 pre-existing errors (vl_convert/anywidget imports)
ruff check altair tests                       # all checks passed
ruff format --check altair tests              # 92 files already formatted
```

## Skipped checks

- Full `pytest` not run (heavy test suite, would need full environment setup)
- The 6 remaining mypy errors are pre-existing on `main` (confirmed by running mypy against the unmodified branch, same 6 errors)

## Notes

- No runtime code changed, only the dev dependency pin
- The upstream pandas-stubs fix was confirmed by @MarcoGorelli (author of the original pandas-stubs typing PR) in the issue comments


joelostblom:
Thanks for the suggested fix @axelray-dev ! This looks good to me. Note that I think we can drop support for python 3.10 altogether as per https://github.com/vega/altair/pull/4029, but that can be dealt with after this is merged.

joelostblom (approved):

PR #4010, opened 2026-04-21
docs: Fix broken example and sync with VL

joelostblom:
Fix before 6.1.0.

PR #4029, opened 2026-06-03
build: Drop support for Python 3.10

joelostblom:
3.10 hasn't received bug fixes for years and is [soon out of security fixes](https://devguide.python.org/versions/) as well.

There are two PRs with py3.10 specific workarounds that we could avoid: https://github.com/vega/altair/pull/4010 and https://github.com/vega/altair/pull/4027

We might even want to adopt a stricter policy like that for [the general scientific python ecosystem](https://scientific-python.org/specs/spec-0000/), which dropped py 3.11 in the end of 2025.


altair/vegalite/v6/api.py
```diff
@@ -71,11 +68,9 @@
...
         from typing import Required, TypeIs
     else:
-        from typing_extensions import Required, TypeIs
-    if sys.version_info >= (3, 11):
-        from typing import Never, Self
-    else:
-        from typing_extensions import Never, Self
+        from typing import Required
+        from typing_extensions import TypeIs
+    from typing import Never, Self
```
alec-bike: I think it's cleaner to write:

```py
    if sys.version_info >= (3, 13):
        from typing import TypeIs
    else:
        from typing_extensions import TypeIs
    from typing import Never, Required, Self
```

tools/generate_schema_wrapper.py
```diff
@@ -909,7 +908,7 @@ def generate_vegalite_channel_wrappers(fp: Path, /) -> ModuleDef[list[str]]:
             f"from altair.vegalite.v6.schema.core import {', '.join(TYPING_CORE)}",
             f"from altair.vegalite.v6.api import {', '.join(TYPING_API)}",
             "from typing import TypeAlias",
-            textwrap.indent(import_typing_extensions((3, 11), "Self"), "    "),
+            "from typing import Self",
```
alec-bike:
```py
"from typing import Self, TypeAlias"
```

alec-bike (changes requested):
Looks good. A couple of minor notes to consolidate some dangling "from typing import..." statements.

joelostblom:
Thanks @alec-bike! I updated with your suggestions

alec-bike (approved):

PR #4034, opened 2026-06-04
fix: Write dataset cache files atomically

joelostblom:
I think the issues we're seeing with intermittent failures in CI are due that the dataset cache previously touched the final cache path before writing the downloaded content. With pytest-xdist, multiple workers can request the same uncached dataset at the same time; one worker could see the non-empty cache file while another worker was still writing it and then try to parse incomplete JSON.

This PR avoids exposing partially written dataset cache files to concurrent readers. Instead it writes downloads to a per-process temporary file first, then atomically replaces the final cache path after the full response has been written.


joelostblom:
I'll merge this as I keep running into it and have to manually retrigger workflows

PR #4035, opened 2026-06-04
fix: Preserve existing config in configure

### LINKED ISSUE #3841: Bug: chart `configure_axisBottom()` has no effect if `configure()` is called later
### What happened?

In this example, `configure_axisBottom()` is called put the axis labels on top, and then `configure()` is called to set a tooltip format. The first call has no effect, as it seems to be overwritten by the second.

```python
import pandas as pd
import altair as alt
data = pd.DataFrame(dict(foo=[1.1, 2.2, 3.3], bar=[3, 2, 4]))
chart = (
    alt.Chart(data)
    .mark_circle()
    .encode(x="foo", y="bar")
    .configure_axisBottom(orient="top")
    .configure(tooltipFormat={"numberFormat": ".2f"})
)
```

However, it works as expected if the two calls are switched:

```python
import pandas as pd
import altair as alt
data = pd.DataFrame(dict(foo=[1.1, 2.2, 3.3], bar=[3, 2, 4]))
chart = (
    alt.Chart(data)
    .mark_circle()
    .encode(x="foo", y="bar")
    .configure(tooltipFormat={"numberFormat": ".2f"})
    .configure_axisBottom(orient="top")
)
```

It would be helpful to either support both orders, or else raise an informative exception if the wrong order is used.

5.5.0

puneetdixit200:
Fixes #3841

## Summary
- update `configure()` to merge new config entries into an existing chart config
- keep generated `ConfigMethodMixin` output in sync with the generator template
- add regression coverage for `configure_axisBottom(...).configure(...)` order

## Tests
- `python -m pytest tests/vegalite/v6/test_api.py -k configure_preserves_existing_config_methods -q` failed before the fix because `axisBottom` was missing
- `.venv/bin/python -m pytest tests/vegalite/v6/test_api.py -k "configure_preserves_existing_config_methods or themes" -q`
- `PATH=/tmp/altair-3841/.venv/bin:$PATH UV_CACHE_DIR=/tmp/altair-3841-uv-cache .venv/bin/task generate-schema-wrapper`
- `git diff --check`


altair/vegalite/v6/schema/mixins.py
```diff
@@ -1161,8 +1161,13 @@ class ConfigMethodMixin:
...
     def configure(self, *args, **kwargs) -> Self:
-        copy = self.copy(deep=False)  # type: ignore[attr-defined]
-        copy.config = core.Config(*args, **kwargs)
+        copy = self.copy(deep=["config"])  # type: ignore[attr-defined]
+        config = core.Config(*args, **kwargs)
+        if copy.config is Undefined:
+            copy.config = config
+        else:
+            for key, value in config.to_dict(validate=False).items():
+                copy.config[key] = value
```
joelostblom, line 1170: Do we need `to_dict()` here or would `config._kwds.items()` be more appropriate? 

Take this example:

```py
chart = (
    alt.Chart("data.csv")
    .mark_point()
    .configure_axis(labelColor="red")
    .configure(axis=alt.AxisConfig(titleColor="blue"))
)
```

On `main` `chart.config.axis` is AxisConfig. On the PR, it is instead `dict`. Serialization is the same, but this is a small API gotcha for users who inspect or mutate `chart.config.axis`, which is likely rare but unless we see a downside, maybe `_kwds` (or another way to retain the object type) would be preferred?

altair/vegalite/v6/schema/mixins.py
```diff
@@ -1161,8 +1161,13 @@ class ConfigMethodMixin:
 
     @use_signature(core.Config)
     def configure(self, *args, **kwargs) -> Self:
-        copy = self.copy(deep=False)  # type: ignore[attr-defined]
-        copy.config = core.Config(*args, **kwargs)
+        copy = self.copy(deep=["config"])  # type: ignore[attr-defined]
+        config = core.Config(*args, **kwargs)
```
joelostblom, line 1165: I believe the PR changes the behavior of an empty `config()` from clearing the config value to doing nothing. I don't really have a preference, but I wonder if we should change so that we preserve the old behavior in case someone relies on it (and to avoid this becoming a breaking change).

I wonder if something like this is sufficient?

```py
if not args and not kwargs:
    copy.config = core.Config()
    return copy
```

what do you think?

tests/vegalite/v6/test_api.py
```diff
@@ -1552,6 +1552,24 @@ def test_themes():
         assert "config" not in chart.to_dict()
 
 
+def test_configure_preserves_existing_config_methods():
+    data = pd.DataFrame({"foo": [1.1, 2.2, 3.3], "bar": [3, 2, 4]})
+
+    chart = (
+        alt.Chart(data)
+        .mark_circle()
+        .encode(x="foo", y="bar")
```
joelostblom, line 1561:
```suggestion
        .encode(x="foo", y="bar")
        .configure(
            numberFormat=".1f",
            axisBottom={"orient": "bottom"},
        )
```

Just to ensure the overwrite works both ways and that the same value is overwritten correctly when merged.

tests/vegalite/v6/test_api.py
```diff
@@ -1552,6 +1552,24 @@ def test_themes():
...
+        .configure_axisBottom(orient="top")
+        .configure(tooltipFormat={"numberFormat": ".2f"})
+    )
+
+    assert chart.to_dict()["config"] == {
+        "view": {"continuousWidth": 300, "continuousHeight": 300},
```
joelostblom, line 1567:
```suggestion
        "view": {"continuousWidth": 300, "continuousHeight": 300},
        "numberFormat": ".1f",
```

joelostblom (changes requested):
Thanks for the PR! This looks like a good fix, just a few minor comments.

Issue #4013, opened 2026-04-21
Tweak weekly builds to be less noisy


joelostblom:
I think it is great that we are shipping fixes and new features quicker through the weekly builds. However, I think they are noisier than they need to be. Since we use github releases for our release notes, having multiple pre-release build show up there makes it difficult to find release notes between versions. Currently, it is only possible to see two actual releases on [the github release page for altair](https://github.com/vega/altair/releases/), the rest are autogenerated ones.

It also becomes a bit cluttered on PyPI now:

<img width="802" height="962" alt="Image" src="https://github.com/user-attachments/assets/47c3393d-134b-4069-ab18-41ae1ee742f4" />

Vega-Lite had this issues before and they moved away from publishing the pre-releases, at least in the release notes section. I like the quicker updated that are coming from regular releases and I wonder if we can do something more related to SemVer with our automatic releases. In general, I think we are safe to automatically publish minor and patch releases when they don't involve updating Vega-Lite. I wonder if we can have the bot run weekly/biweekly and make an actual release if there are any `fix` or `feat` (and maybe some other labels) PRs in that time span. And instead of tagging them as `-dev`, it would just bump the relevant part of the version number.

An alternative would be to release on every single `fix` or `feat` PR that is merged which is more strictly following, but that might make it a bit noisy in terms of the number of releases instead and might result in quite large release numbers, which is why I suggested bundling them weekly/biweekly.

joelostblom:
@mattijn and @alec-bike I saw that you both discussed and improved this previously, so I would be curious to what your thoughts are here

mattijn:
It's supposed to do some cleanup, original intent was to only keep last x versions of the prerelease. That is not working correctly at the moment, so nothing is cleaned up after releasing a new prerelease. We can fix that, and/or release it biweekly. It currently only prerelease if there has been done commits in that period since previous (pre)release, but checking for feats might be interesting as well to reduce noice. Open for all of it.

mattijn:
(btw, prerelease is done weekly at the moment - on mondays)

alec-bike:
I tested using pre-releases for a bit, rather than pointing to my local altair fork. Turns out to be less convenient as they don't trigger as updated with `uv tree --outdated`. So I like the idea of a semver patch release every 2 weeks.

joelostblom:
[...]

PR #4036, opened 2026-06-04
ci: Semi-automate releases

joelostblom:
Merge after https://github.com/vega/altair/pull/4030

@mattijn @alec-bike This is an attempt to replace the weekly builds as discussed in https://github.com/vega/altair/issues/4013. The PR adds a semi-automated release flow that prepares draft releases to be reviewed by a maintainer. After a release is published, the corresponding files will also be released on PyPI.

## Details

This PR replaces the old weekly/dev prerelease publishing flow with two release workflows:

- `Prepare Release Draft`
  - Manually triggered for now, with a commented schedule that can be enabled later. Manual triggers mean that we can release quicker too if there is some feature/fix we want to publish right away.
  - Uses Cocogitto to inspect conventional commits since the latest `v*` tag.
  - Skips release preparation if no SemVer-relevant changes are found.
  - Runs the test suite.
  - Creates the release commit and `vX.Y.Z` tag.
  - Creates a draft GitHub release using GitHub-generated release notes.
  - Builds a preview of the documentation for the draft release so that we don't need to build it locally each time.
  - Opens a release review issue assigned to maintainers with links to the draft release and release notes.

- `Publish Release to PyPI`
  - Triggered when a GitHub release is published.
  - Only publishes stable non-prerelease tags matching `vX.Y.Z`.
  - Builds from the release tag.
  - Publishes to PyPI using trusted publishing.

This also removes the old weekly/dev prerelease workflow and nightly version helper.

This branch assumes Altair now derives package versions from VCS tags via `versioningit`, so the release automation no longer edits version files or bumps `main` back to a development version after release.

Cocogitto only determines the next SemVer release and creates the release commit/tag. Package version metadata is derived from the tag during build.


mattijn:
Interesting approach! In my experience it takes a bit of trial and error to get it working as you want (few PRs), so please go ahead👍

mattijn:
Btw, not sure if I understand you correctly, you don’t want to publish prereleases anymore?

mattijn:
Ah, now I see in https://github.com/vega/altair/issues/4013 again. More like an automated two weekly semver patch release

joelostblom:
> Ah, now I see in #4013 again. More like an automated two weekly semver patch release

Exactly, with an option to review the release notes before the release is published in case we want to elaborate on something with screenshots, additional info, etc.

And yes, I imagine it might take a few PRs to get it exactly right.

alec-bike (approved):
Looks good to me. I don't have much experience with the details of github workflows and pypi releases, but the concept and documentation reads well.

Once the first semver release is pushed, I can test with my uv-based project where altair is a dependency.

joelostblom:
Thanks both! I will merge this in an iterate in (hopefully small) follow up PRs if it is not working as intended

alec-bike:
@joelostblom I was seeing errors on `uv sync --all-extras` after the recent version updates, when working on my altair fork. Turns out release tags are now required, so I needed to pull all tags from upstream into my fork:


```sh
git fetch --tags git@github.com:vega/altair.git
```

May need to add this to `CONTRIBUTING.md`.

joelostblom:
Thanks for reporting @alec-bike. I put up https://github.com/vega/altair/pull/4051 with a note on how to get the correct version and also the addition of a base tag to hopefully avoid the error you were seeing. Let me know if you think that's sufficient or feel free to suggest any changes you would like.

alec-bike:
Thanks @joelostblom, the note in #4051 looks good.

Issue #3906, opened 2025-11-06
improve tests for python 3.14


mattijn:
### What is your suggestion?

- Re enable altair tests for python 3.14 that include geopandas https://github.com/vega/altair/pull/3876/commits/cc64b1183ae991537fe7587ed85e60ae927b4194
- Re enable altair tests for python 3.14 that include duckdb https://github.com/vega/altair/pull/3876/commits/bf2dfe2e749582b802b0b37d6246189a16548f7e
- Enable tests for python 3.14t (threaded) if pyarrow is supported https://github.com/vega/altair/pull/3876/commits/bd7b91d32a5680f7714f2b76ca2c0648da7a9a6f

cclauss:
To be confident about free-threaded compatibility, it is recommended to test with [`pytest-run-parallel`](https://github.com/Quansight-Labs/pytest-run-parallel).
* https://py-free-threading.github.io/testing

Something like:
% `uvx --with=pytest-run-parallel pytest --iterations=8 --parallel-threads=auto`

mattijn:
Thanks @cclauss! I have added another workflow to test the free-threaded version of python 3.14 in https://github.com/vega/altair/pull/3907. The workflow is currently, as expected, failing, see https://github.com/vega/altair/actions/runs/19181574029/job/54839362746#step:6:1. But hopefully in coming period can check which dev dependencies are the culprit and which tests shall be skipped in this process. Do you have a free-threaded version of python locally to test this a bit more? Otherwise will continue testing through the github action workflow, albeit more cumbersome.

cclauss:
% `uv run pytest -o addopts= --pyargs --doctest-modules --doctest-ignore-import-errors --iterations=8 --parallel-threads=auto tests`
--> # Add the dependencies needed...
% `uv run --python=3.14t --with=numpy,pandas pytest -o addopts= --pyargs --doctest-modules --doctest-ignore-import-errors --iterations=8 --parallel-threads=auto tests`

axelray-dev:
I took a look at the current state of this.

The first two bullets seem to be handled already by #3975: geopandas and duckdb are no longer pinned away from Python 3.14 in pyproject.toml.

The remaining open question looks like the free-threaded Python 3.14 workflow from #3907. It is currently manual-only, and several optional dependencies are still commented out. From current wheel availability, pyarrow, pandas, and numpy appear to have 3.14t wheels, while duckdb and polars still do not.

[...]

PR #4038, opened 2026-06-05
ci: enable partial Python 3.14t coverage

axelray-dev:
Addresses #3906.

## Summary

This updates the manual free-threaded Python 3.14t workflow to install the optional dev dependencies that now support cp314t or are pure Python and install cleanly on free-threaded builds.

A few compiled dependencies remain disabled with comments linking to the relevant upstream blockers.

## What is enabled

- hatch, ipython, ipykernel, pandas, pyarrow-stubs, pytest-xdist, mistune, mypy, pandas-stubs, types-jsonschema, types-setuptools, geopandas, taskipy, tomli

## What remains blocked

- **ruff**: no cp314t wheels available yet
- **duckdb**: no free-threaded builds ([duckdb/duckdb-python#76](https://github.com/duckdb/duckdb-python/issues/76))
- **polars**: pyo3/numpy crate incompatibility ([pola-rs/polars#21889](https://github.com/pola-rs/polars/issues/21889))

## Files changed

- `.github/workflows/build-free-threaded.yml` (1 file, +25/-15)

## Verification

- YAML syntax validated
- git diff --check clean
- No non-ASCII or smart punctuation in diff
- Only the free-threaded workflow file changed; main CI and weekly workflows untouched
- Python 3.14t is not available in the standard CI runners, so dependency resolution for the newly enabled packages will be validated when the workflow is triggered manually via workflow_dispatch


.github/workflows/build-free-threaded.yml
```diff
@@ -28,24 +32,30 @@ jobs:
           uv pip install pytest
           uv pip install pytest-cov
           uv pip install pytest-run-parallel
-          # uv pip install hatch>=1.13.0
+          uv pip install hatch>=1.13.0
```
joelostblom: Don't we need quotes around all these version identifiers? 

```
uv pip install 'hatch>=1.13.0'
uv pip install 'pandas>=1.1.3'
uv pip install 'geopandas>=0.14.3'
uv pip install 'taskipy>=1.14.1'
uv pip install 'tomli>=2.2.1'
```

joelostblom (changes requested):
Thanks for putting this up, just one comment

axelray-dev:
Thanks, fixed by quoting the version specifiers in the uv pip install commands.

joelostblom (approved):
Thanks @axelray-dev !

PR #4050, opened 2026-06-05
docs: Add an interactive bin-size example to the gallery

### LINKED ISSUE #3355: Add gallery example of adjustable bin sizes
I saw this cool example from @dwootton and think it could be cool to have in the altair example gallery (it's even cooler when the adjusting is bound to zooming the chart, but my understanding is that this is not possible in VL directly)

image

[Open the Chart in the Vega Editor](https://vega.github.io/editor/#/url/vega-lite/N4IgJAzgxgFgpgWwIYgFwhgF0wBwqgegIDc4BzJAOjIEtMYBXAI0poHsDp5kTykBaADZ04JAKyUAVhDYA7EABoQAEySYUqUAwBOgtBmx5CBbUgDu1OoyYMIcbVDmY4szJUcJeFLwNXq7mBAEyDSyBH5IwWzENHAQUjLyAL5KOEimCBBoANqgskgIcPpMoVlKxEiCDEWoAIwADEolsspooKE4DJj6prJkRUoIoWi1g0gAHiP1jSAQzjgjSUkAukqYvRAAZmzaCDmgmzSCztptIEgt+yCHcIKt6ACSALIAIgBCAAQASmqhZIogCrCe7raopA6xO76L5sbAuD4AFTYyEwbDi31+fQBQJoIO0YOWS1WIGQ2gA1j04FBukozLj6GgAMzTJTwGhkLBoABMLJALkcyj+Z0mmhAzTOyHGzSyoog7PyenQ0pASyUNyhj1enx+mCFawAnjgaiAAI4MC669S60gqpT6s7i0WS5Wy+WVYqlFUpa6Q+4gGFw2SI5FqNEQDG6rEGo36M0Wui-G3exyCHZnJBkMjaPjOfSOBiuAGYQ3GuOuBPWoqqkCOWSHf6imJwMxnObaNhk43rC4QNLZwtLJJAA)

kyo5uke:
Closes #3355.

### Description

Adds a **Histogram with Adjustable Bin Size** example to the example gallery (category *Distributions*), as requested in #3355 (labelled *good first issue* / *help wanted*).

The example is a 2D histogram heatmap of the `movies` dataset (`IMDB Rating` × `Rotten Tomatoes Rating`, coloured by `count()`) with a slider that changes the **number of bins** (`maxbins`) live.

Vega-Lite's `BinParams.maxbins` is a plain number and cannot be driven by a parameter expression, so a slider can't be bound to `maxbins` directly (this is the blocker the issue ran into). The example works around it by drawing **one `mark_rect` layer per available bin count** and using `transform_filter("maxbins === N")` so that only the layer matching the slider value is shown. The slider steps (`10, 20, 30, 40, 50, 60`) line up exactly with the layers, so there is no intermediate slider position that renders an empty chart, and the default (`30`) is shown on load. One layer is materialized per bin count, so the spec carries all six binnings even though only one is visible at a time — an acceptable cost for the six options here, and a direct consequence of `maxbins` not being parameter-bindable.

Added in both syntaxes, as the contributing guide requires:

- `tests/examples_arguments_syntax/interactive_bin_size.py`
- `tests/examples_methods_syntax/interactive_bin_size.py`

### How this differs from the existing `histogram_responsive.py`

There is already a *"Histogram with Responsive Bins"* example in the same *Distributions* category, which implements @jonmmease's `alt.Bin(maxbins=20, extent=interval_selection)` idea. That one keeps `maxbins` **fixed** and lets a brush change the bin **extent** (the value range that gets binned). This new example does the opposite: the extent is fixed and the slider changes the **bin count**. They demonstrate two different things, so I kept them as separate gallery entries.

I also deliberately avoided the `opacity`-condition layering that @dsmedia sketched in the issue. With that approach every layer is always present and only its opacity is toggled, which (as @dsmedia noted) leaves the colour scale's domain static. By filtering with `transform_filter` instead, only the matching layer has data, so the `count()` colour scale adapts to the current bin count. The trade-off is that the colour scale and its legend rescale when the bin count changes — which is the intended behaviour here, but worth calling out.

### Tests & Docs

The example is itself a gallery/documentation item, covered by the existing `tests/test_examples.py` suite. I verified locally (altair 6.1.0, vl-convert installed) that, for **both** files, the harness flow passes:

- `eval_block` returns a `LayerChart` (the final statement is a bare expression);
- `chart.to_dict()` succeeds and produces 6 layers with the `maxbins` parameter;
- the `to_json` → `from_json` → `to_json` round-trip is stable;
- `chart.save(format="png", engine="vl-convert")` renders a valid PNG.

I did not run the full Sphinx docs/thumbnail build locally.


mattijn:
Thanks for this PR!.
Currently this renders as follows:

image


Something is off with the titles. You also might have a look to https://altair-viz.github.io/gallery/interactive_aggregation.html which is a bit similar, in this reference example the x-axis starts at zero and the x-axis has no decimals (equally to the y-axis). Hopefully this is something that you are able to improve.

Thanks again for the PR!

kyo5uke:
Thanks @mattijn! The axis titles were being concatenated across the six filtered layers, so I set an explicit title on each layer's x/y axis and they now render once (IMDB Rating / Rotten Tomatoes Rating). I also made the x-axis zero-based with integer ticks (`format='d'`) to match interactive_aggregation, and applied the same integer formatting to the y-axis.

mattijn:
Add a commit so it is default width/height. Thanks again!

PR #4067, opened 2026-06-11
fix: disambiguate inner layer names when identical LayerCharts are concatenated

### LINKED ISSUE #4065: Layered charts in a concat still collide on auto-generated view names (regression of #3981 for `LayerChart`)
## Summary

When two structurally identical layered `(line + circle)` charts are concatenated
(`alt.vconcat` / `alt.hconcat`), their inner `mark_line` layers are given the
**same** auto-generated content-hash view name (e.g. `view_<hash>_0`). Vega-Lite
then deduplicates the two same-named views, so both line marks are rendered from a
single data pipeline. The result: in one of the panels the line is drawn from the
*other* panel's data.

This is the same class of bug as #3981 (fixed by #3982), but that fix does not
cover this case. #3982 folds the data source into the view-name hash so that
panels backed by *different* DataFrames get different names. Here both panels
share **one** DataFrame and differ only by transforms/encodings that live on the
**parent** unit spec — so the inner line layers carry no data of their own and
hash identically again.

The circle marks escape the collision because a selection-based conditional
encoding (`size=alt.condition(...)`) makes their specs differ, so they are *not*
named and *not* deduplicated. That asymmetry is what makes the bug visible: the
**points** render at the correct positions while the **line** does not pass
through them.

- Reproduced on **altair 6.2.1** (vl-convert-python 1.9.0, Vega-Lite 5.x).
- **Not** reproduced on altair 5.5.0 (there the inner layers get sequential
  names `view_1`/`view_2` rather than colliding content hashes), so this is a 6.x
  regression.

Related: #3981, #3982, #3891 (selection-param deduplication).

## Minimal reproducible example

```python
import altair as alt
import pandas as pd

# one dataframe, two "antibodies" per site
df = pd.DataFrame({
    "site":     [1, 2, 3, 4, 5] * 2,
    "antibody": ["a"] * 5 + ["b"] * 5,
    "escape":   [0.2, 0.8, 0.2, 0.3, 0.2, 0.4, 1.0, 0.4, 0.5, 0.4],
})

hover = alt.selection_point(fields=["site"], on="mouseover", empty=False)
base = alt.Chart(df).encode(x="site:O", y=alt.Y("escape:Q", scale=alt.Scale(domain=[0, 1])))
lines = base.mark_line(size=1)
points = base.encode(size=alt.condition(hover, alt.value(120), alt.value(40))).mark_circle(filled=True)
layered = lines + points

# top panel: mean over antibodies (aggregate lives on the PARENT unit)
mean_panel = layered.transform_aggregate(escape="mean(escape)", groupby=["site"])
# bottom panel: individual antibodies (detail lives on the PARENT unit)
indiv_panel = layered.encode(detail="antibody:N")

chart = alt.vconcat(mean_panel, indiv_panel).add_params(hover)
```

Inspecting the produced spec, both `mark_line` layers receive the same name:

```
line    name=view_61fc0cabe14d30bc_0     <- vconcat[0].layer[0]   (mean panel)
circle  name=None
line    name=view_61fc0cabe14d30bc_0     <- vconcat[1].layer[0]   (individual panel)
circle  name=None
```

Compiling to Vega (`vl_convert.vegalite_to_vega`) confirms the consequence — two
line marks with the **same** Vega mark name pulling from **different** datasets:

```
line  view_61fc0cabe14d30bc_0_marks  <- data-56d8751c...                      (mean)
line  view_61fc0cabe14d30bc_0_marks  <- faceted_path_view_61fc0cabe14d30bc_0_main  (individual)
```

Since Vega mark names must be unique within a group, one of these is effectively
clobbered and the panels share a line, while the (uniquely named) circle marks
stay correct.

## Expected behavior

Each panel's line is drawn from that panel's own data pipeline; the points lie on
the line.

## Actual behavior

The two line layers share a view name, are deduplicated, and one panel's line is
rendered from the other panel's data — the points no longer sit on the line.

## Root cause

`_combine_subchart_params` in `altair/vegalite/v5/api.py` already disambiguates
concatenated subcharts by appending the positional index `_{i}`, but only for
`Chart` and `FacetChart` subcharts:

```python
if _needs_name(subchart):
    base_name = subchart._get_view_hash_name()
    subchart.name = f"{base_name}_{i}"

if is_concat and isinstance(subchart, FacetChart):
    ...
    spec.layer[0].name = f"{_view_base_for_chart(spec.layer[0])}_{i}"
```

For a `LayerChart` concat subchart neither branch fires:

1. `_needs_name` returns `False` for anything that isn't a `Chart`.
2. The `FacetChart` branch obviously doesn't match a `LayerChart`.
3. Worse, the loop body is guarded by an early
   `if ... is_undefined(subchart.params): continue`, and these `LayerChart`
   subcharts carry no params of their own (their params were already lifted to
   the parent during the layer-level `_combine_subchart_params`), so the loop
   `continue`s before reaching any naming code.

Meanwhile, the inner line `Chart` of each panel *did* get named
`view_<hash>_0` during the panel's own layer-level `_combine_subchart_params`
(it holds the lifted hover param, so `_needs_name` is `True` there). Because the
two panels' inner line charts are byte-for-byte identical (data + transforms are
on the parent), they hash to the same name. Nothing later disambiguates them by
concat position, so they collide.

## Proposed fix

Add positional disambiguation for `LayerChart` concat subcharts, in a pass that
runs regardless of whether the subchart carries params (mirroring the existing
`FacetChart` handling). See `proposed_fix.diff` in this directory:

```python
# in _combine_subchart_params, right after `is_concat = len(subcharts) > 1`
if is_concat:
    for i, subchart in enumerate(subcharts):
        if not (isinstance(subchart, LayerChart) and subchart.layer):
            continue
        subchart.layer = [lyr.copy() for lyr in subchart.layer]
        for lyr in subchart.layer:
            if isinstance(lyr, Chart) and lyr.name is not Undefined:
                old_name = lyr.name
                new_name = f"{_view_base_for_chart(lyr)}_{i}"
                lyr.name = new_name
                for _p, _d, _views in param_info:
                    if old_name in _views and new_name not in _views:
                        _views.append(new_name)
```

With this patch the two line layers become `view_<hash>_0` and `view_<hash>_1`,
compile to two distinct Vega marks bound to their own datasets, and the rendering
is correct. Verified on altair 6.2.1, and checked for regressions against:

- the original #3981 two-DataFrame case (still distinct, via the data hash),
- a plain single `LayerChart` (unchanged),
- a 3-panel concat of identical layered charts (`_0`/`_1`/`_2`).

### Caveat / open question for maintainers

The snippet attempts to keep selection params bound to each renamed layer by
extending their `views`, but in practice the param `views` list is finalized in a
later pass, so the `views`-extension part is a no-op as written and the top-level
param ends up referencing only `view_<hash>_0` (this matches the pre-patch
behavior, so it is not a regression). If you want the selection to attach to every
renamed sibling layer, the rename and the `views` accumulation likely need to be
folded into the existing param loop / `_view_name_for_param`, which you know
better than I do. The primary rendering corruption is fixed independently of that.

## Environment

- altair 6.2.1
- vl-convert-python 1.9.0
- pandas 3.0.3
- Python 3.12

A self-contained script (`reproduce_bug.py`) accompanies this report.

[reproduce_bug.py](https://github.com/user-attachments/files/28806169/reproduce_bug.py)

axelray-dev:
Fixes #4065

## Problem

When two structurally identical layered `(line + circle)` charts are concatenated via `alt.vconcat` / `alt.hconcat`, their inner `mark_line` layers receive the same auto-generated content-hash view name. Vega-Lite then deduplicates the two same-named views, so both line marks render from a single data pipeline. The result: in one panel the line is drawn from the other panel's data.

This is the same class of bug as #3981 (fixed by #3982), but that fix only handles the case where DataFrames differ. Here both panels share one DataFrame and differ only by transforms/encodings on the parent unit spec, so the inner layers still hash identically.

## Fix

In `_combine_subchart_params`, add positional disambiguation for `LayerChart` subcharts in concat. When `is_concat and isinstance(subchart, LayerChart)`, copy the inner layers and append `_{i}_{j}` to each layer's name, matching the existing `FacetChart` handling pattern.

## Changes

- `altair/vegalite/v6/api.py`: 11 lines added. New branch in `_combine_subchart_params` that disambiguates LayerChart inner layer names by position.
- `tests/vegalite/v6/test_params.py`: 28 lines added. Regression test reproducing the exact scenario from the issue (two identical `(line + circle)` charts sharing one DataFrame, concatenated via `vconcat`).

## Tests

```
uv run pytest tests/vegalite/v6/test_params.py -k "vconcat_identical_layered_charts_unique_inner_layer_names or vconcat_different_data_unique_names" -v
```

2 passed. Full `uv run task test` passes formatting and pytest; mypy reports pre-existing import-not-found errors for optional modules (`vl_convert`, `anywidget`) unrelated to this change.

### DIFF
--- a/altair/vegalite/v6/api.py
+++ b/altair/vegalite/v6/api.py
@@ -5289,6 +5289,17 @@ def _combine_subchart_params(  # noqa: C901
             elif isinstance(spec, Chart):
                 spec.name = f"{_view_base_for_chart(spec)}_{i}"
 
+        if is_concat and isinstance(subchart, LayerChart) and subchart.layer:
+            subchart.layer = [layer.copy() for layer in subchart.layer]
+            for j, layer in enumerate(subchart.layer):
+                if isinstance(layer, Chart):
+                    base_name = (
+                        layer._get_view_hash_name()
+                        if layer.name is Undefined
+                        else _view_base_for_chart(layer)
+                    )
+                    layer.name = f"{base_name}_{i}_{j}"
+
         for param in subchart.params:
             p = _prepare_to_lift(param)
             pd = _viewless_dict(p)
--- a/tests/vegalite/v6/test_params.py
+++ b/tests/vegalite/v6/test_params.py
@@ -413,3 +413,31 @@ def test_vconcat_different_data_unique_names():
             f"Layers should have unique names when data differs, "
             f"but both got name: {line_names[0]}"
         )
+
+
+def test_vconcat_identical_layered_charts_unique_inner_layer_names():
+    df = pd.DataFrame({"site": [1, 2, 3, 4, 5], "escape": [0.1, 0.9, 0.1, 0.2, 0.1]})
+    selection = alt.selection_point(fields=["site"], on="mouseover", empty=False)
+
+    def make_layered_chart():
+        base = alt.Chart(df).encode(
+            x="site:O",
+            y=alt.Y("escape:Q", scale=alt.Scale(domain=[0, 1])),
+        )
+        lines = base.mark_line(size=1)
+        points = base.encode(
+            size=alt.condition(selection, alt.value(80), alt.value(30)),
+        ).mark_circle(filled=True)
+        return (lines + points).add_params(selection)
+
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", UserWarning)
+        spec = alt.vconcat(make_layered_chart(), make_layered_chart()).to_dict()
+
+    layers0 = spec["vconcat"][0]["layer"]
+    layers1 = spec["vconcat"][1]["layer"]
+    names0 = [layer["name"] for layer in layers0]
+    names1 = [layer["name"] for layer in layers1]
+
+    assert names0[0] != names1[0]
+    assert names0[1] != names1[1]


axelray-dev:
Closing this in favor of #4066, which was opened first and addresses the same issue. Thanks.

joelostblom:
Thanks for working on this @axelray-dev , and yes it's unfortunate that we worked on the same fix; don't hesitate to drop a line on a n issue when you're starting to work on it and I can leave that issue for you!

PR #4068, opened 2026-06-15
fix: raise TypeError for invalid property setter arguments

### LINKED ISSUE #3661: `_PropertySetter` silently ignores multiple positional arguments
### What happened?

Adapted from https://github.com/vega/altair/pull/3659#discussion_r1820574348

# Minimal Repro
These are all valid:
```py
import altair as alt

valid_scale = alt.Scale(domain=["M", "F"], range=["#1FC3AA", "#8624F5"])

valid_color_1 = alt.Color().scale(valid_scale)
valid_color_2 = alt.Color().scale(domain=["M", "F"], range=["#1FC3AA", "#8624F5"])
```

Both of these cases naively just ignore everything after the first argument:
```py
alt.Color().scale(["M", "F"], ["#1FC3AA", "#8624F5"])
# Color({
#  scale: ['M', 'F']
#})

alt.Color().scale(valid_scale, valid_scale, valid_scale)
# Color({
#   scale: Scale({
#     domain: ['M', 'F'],
#     range: ['#1FC3AA', '#8624F5']
#   })
# })
```

## Expected
Behavior similar the wrapped class:
```py
invalid_scale = alt.Scale(["M", "F"], ["#1FC3AA", "#8624F5"])
```
```log
SchemaValidationError: Multiple errors were found.

Error 1: '['M', 'F']' is an invalid value for `align`. Valid values are of type 'number' or 'object'.

Error 2: '['#1FC3AA', '#8624F5']' is an invalid value for `base`. Valid values are of type 'number' or 'object'.
```

# Cause
`_PropertySetter.__call__`  doesn't align with `SchemaBase.__init__`:

https://github.com/vega/altair/blob/6beec7bf0819d961d3d32823d7e658429c306b97/tools/schemapi/schemapi.py#L1602-L1606

Above, we silently drop any positional args *after* the first (see [recording](https://github.com/vega/altair/pull/3659#discussion_r1820870591))

Below, we `raise` an `AssertionError` if there is more than 1 positional argument

https://github.com/vega/altair/blob/6beec7bf0819d961d3d32823d7e658429c306b97/tools/schemapi/schemapi.py#L1018-L1038
   
## Note
Appears to have been present in all `5.*` versions
- https://github.com/vega/altair/blame/c273c0a031cf1bfdfc9f8ea18ebeab9399581363/tools/schemapi/schemapi.py#L1097

I think we should `raise` for all cases that look like this:

```py
alt.ChannelsClass().property_setter(arg1, arg...)
```

`5.5.0dev`

suzunn:
## Problem

`_PropertySetter.__call__` silently accepted multiple positional arguments and discarded every value after the first. It also ignored keyword arguments whenever a positional value was present, which made invalid method-style channel configuration appear successful.

## Solution

Argument validation now rejects multiple positional arguments and mixed positional/keyword calls with a clear `TypeError`. The message is prefixed with the owner class so the offending call is easy to locate (e.g. `X.scale()`). Both the schema generator source and the generated runtime module are updated, and the regression tests are parametrized over several setters (`scale`, `axis`, `sort`) to confirm the behaviour applies to every property setter.

This is a behavioural (breaking) change: calls that previously "succeeded" by silently dropping arguments now raise.

Closes #3661.

## Before / After

**Before** — extra positional arguments were silently dropped, and keyword arguments were ignored when a positional value was present, so an invalid call appeared to succeed:

```python
import altair as alt

# second argument silently discarded, no error
alt.X("field:Q").scale(["a", "b"], ["#1FC3AA", "#8624F5"])

# keyword arguments silently ignored, no error
alt.X("field:Q").axis(alt.Axis(), title="hi")
```

**After** — the invalid call raises immediately, naming the owner class and setter:

```python
alt.X("field:Q").scale(["a", "b"], ["#1FC3AA", "#8624F5"])
# TypeError: X.scale() accepts at most one positional argument, but 2 were given

alt.X("field:Q").axis(alt.Axis(), title="hi")
# TypeError: X.axis() cannot combine a positional argument with keyword arguments
```

If you hit this error, either pass a single value positionally (e.g. `.scale(alt.Scale(...))`) or use keyword arguments only (e.g. `.scale(domain=[...], range=[...])`).

### DIFF
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -1662,6 +1662,19 @@ class _PropertySetter:
         return self
 
     def __call__(self, *args: Any, **kwargs: Any):
+        if len(args) > 1:
+            msg = (
+                f"{self.prop}() accepts at most one positional argument, "
+                f"but {len(args)} were given"
+            )
+            raise TypeError(msg)
+        if args and kwargs:
+            msg = (
+                f"{self.prop}() cannot combine a positional argument "
+                "with keyword arguments"
+            )
+            raise TypeError(msg)
+
         obj = self.obj.copy()
         # TODO: use schema to validate
         obj[self.prop] = args[0] if args else kwargs
--- a/altair/utils/schemapi.py
+++ b/altair/utils/schemapi.py
@@ -1662,16 +1662,16 @@ class _PropertySetter:
         return self
 
     def __call__(self, *args: Any, **kwargs: Any):
+        name = f"{type(self.obj).__name__}.{self.prop}"
         if len(args) > 1:
             msg = (
-                f"{self.prop}() accepts at most one positional argument, "
+                f"{name}() accepts at most one positional argument, "
                 f"but {len(args)} were given"
             )
             raise TypeError(msg)
         if args and kwargs:
             msg = (
-                f"{self.prop}() cannot combine a positional argument "
-                "with keyword arguments"
+                f"{name}() cannot combine a positional argument with keyword arguments"
             )
             raise TypeError(msg)
 
--- a/tests/utils/test_schemapi.py
+++ b/tests/utils/test_schemapi.py
@@ -1151,14 +1151,14 @@ def test_property_setter_rejects_multiple_positional_arguments() -> None:
     with pytest.raises(
         TypeError, match=r"scale\(\) accepts at most one positional argument"
     ):
-        alt.Color().scale(["M", "F"], ["#1FC3AA", "#8624F5"])
+        alt.Color().scale(["M", "F"], ["#1FC3AA", "#8624F5"])  # type: ignore[call-overload]
 
 
 def test_property_setter_rejects_positional_and_keyword_arguments() -> None:
     with pytest.raises(
         TypeError, match=r"scale\(\) cannot combine a positional argument"
     ):
-        alt.Color().scale(alt.Scale(), domain=["M", "F"])
+        alt.Color().scale(alt.Scale(), domain=["M", "F"])  # type: ignore[call-overload]
 
 
 def test_to_dict_datetime_typing() -> None:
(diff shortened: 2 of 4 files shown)


suzunn:
Gentle nudge on this one whenever there's review bandwidth. It's a small, self-contained change (rejecting invalid positional/keyword arguments to property setters instead of silently mis-assigning them), it's conflict-free, and CI is green across the full 3.10–3.14 matrix. Happy to adjust the approach if a maintainer has a preferred direction.

mattijn:
Thanks for the PR, and sorry for the wait. On mobile, so apologies if I missed something.

Fix itself looks right, but this covers every property setter. Both new tests use only `Color().scale`, so this is not visible there. Can you parametrize over a few setters (`scale`, `axis`, `sort`) to show it applies everywhere?

Btw, our release notes are inferred from the PR titles, and it's a breaking change, so please change to something as such: "fix!: raise TypeError for invalid property setter arguments"

Can you add a few before/after example including raised error in the PR description  to help users if they hit this new error and search for this (section on testing can be removed).

 Also is it possible to add the owner class in the error message? (Maybe this works: `{type(self.obj).__name__}.{self.prop}()` instead of only the `prop`

Thanks again for this PR!

suzunn:
Thanks for the review! All addressed:

- **Parametrized the tests** over `scale`, `axis`, and `sort` (via `alt.X`) so both rejection cases are exercised for several setters, not just `Color().scale`.
- **Owner class in the message**: the `TypeError` now reads e.g. `X.scale() accepts at most one positional argument, but 2 were given` / `X.axis() cannot combine a positional argument with keyword arguments`, using `type(self.obj).__name__`.
- **Title** changed to `fix!: raise TypeError for invalid property setter arguments`.
- **Description** now has before/after examples with the raised errors, and the testing section is removed.

Both `schemapi.py` copies are updated in sync; ruff check/format, mypy, and the parametrized tests all pass locally. I also rebased onto latest `main`.

mattijn:
Thanks @suzunn! Much appreciated for this PR! Keep on pushing 👍👍🤗

PR #4069, opened 2026-06-15
fix: Include layered concat views in selection params

joelostblom:
This fixes selection param `views` accumulation for layered charts inside concat charts. After #4066, layered concat children receive distinct generated view names, which fixes duplicate view-name collisions. However, a related issue mentioned in #4065 remained: when a selection parameter was lifted/deduplicated from layered concat children, the generated top-level param could still reference only the first renamed view.

For example, before this PR a spec could produce:

```json
"views": ["view_..._0"]
```

even though the concat panels contained:

```json
["view_..._0", "view_..._1"]
```

That meant hover/click selections could be scoped to only the first panel. This change updates param view collection so layered concat children can contribute all relevant named child views.

Taking the following spec as an example:

```py
import altair as alt
import polars as pl

df = pl.DataFrame(
    {
        "x": [1, 2, 3, 1, 2, 3],
        "group": ["a", "a", "a", "b", "b", "b"],
        "y": [1, 3, 2, 2, 4, 3],
    }
)

hover = alt.selection_point(fields=["x"], on="mouseover", empty=False)

base = alt.Chart(df).encode(
    x="x:O",
    y="y:Q",
)

line = base.mark_line()

points = (
    base.mark_circle()
    .encode(size=alt.condition(hover, alt.value(150), alt.value(50)))
    .add_params(hover)
)

layered = line + points

left = layered.transform_filter(alt.datum.group == "a")
right = layered.transform_filter(alt.datum.group == "b")

chart = alt.hconcat(left, right)

chart
```

Before this PR, the params were only attached to one chart (ie hover only possible in the right-most chart):

https://github.com/user-attachments/assets/bf0ebd01-afcc-41ce-abbf-a7873da3dae7

After this PR, either chart can be hovered, as expected:

https://github.com/user-attachments/assets/6f67bdc6-1dcb-4caf-b910-f641e41d649a

### DIFF
--- a/altair/vegalite/v6/api.py
+++ b/altair/vegalite/v6/api.py
@@ -5232,17 +5232,23 @@ def _view_base_for_chart(obj: Any) -> str:
     return name
 
 
-def _view_name_for_param(subchart: ChartType, is_concat: bool) -> str:
-    """View name for this subchart to add to a param's views."""
+def _view_names_for_param(subchart: ChartType, is_concat: bool) -> list[str]:
+    """View names for this subchart to add to a param's views."""
     if isinstance(subchart, Chart):
-        return subchart.name
+        return [subchart.name]
+    if is_concat and isinstance(subchart, LayerChart):
+        return [
+            layer.name
+            for layer in subchart.layer
+            if isinstance(layer, Chart) and layer.name is not Undefined
+        ]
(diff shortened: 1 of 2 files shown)


joelostblom:
Merging to get param and view naming synced up with that #4067 was merged

PR #4078, opened 2026-06-30
fix: preserve user-set names on subcharts in concat

### LINKED ISSUE #4070: User set names are overridden by automatic
### What happened?

When a user explicitly sets `name=` on a chart that ends up as an inner spec inside a `FacetChart` or `LayerChart` concat panel, Altair silently renames it by appending `_0`, `_1`, etc. The user-supplied name is not preserved in the output spec.

This affects two code paths in `_combine_subchart_params` in `altair/vegalite/v6/api.py`:

1. **`FacetChart`** — renaming happens inside the param-processing loop when the `FacetChart` subchart carries params.
2. **`LayerChart` (introduced in PR #4066)** — renaming happens in a new pre-pass that runs unconditionally for all `is_concat` cases, regardless of whether params are present.

The two cases also have an asymmetry: the `FacetChart` rename only fires when that subchart has params attached; the `LayerChart` pre-pass renames even when there are no params at all.

## Minimal reproducible examples

**Case 1 — `FacetChart`**

```python
import altair as alt
import pandas as pd

df = pd.DataFrame({"x": [1, 2, 3], "y": [1, 4, 9], "category": ["a", "b", "c"]})
hover = alt.selection_point(fields=["x"], on="mouseover", empty=False)

chart = alt.Chart(df, name="my_panel").encode(x="x:Q", y="y:Q").mark_line()
faceted = chart.add_params(hover).facet("category:N")

spec = alt.vconcat(faceted, faceted).to_dict()
print(spec["vconcat"][0]["spec"]["name"])  # "my_panel_0"  ← expected "my_panel"
print(spec["vconcat"][1]["spec"]["name"])  # "my_panel_1"  ← expected "my_panel"
```

**Case 2 — `LayerChart`**

```python
import altair as alt
import pandas as pd

df = pd.DataFrame({"x": [1, 2, 3], "y": [1, 4, 9]})
hover = alt.selection_point(fields=["x"], on="mouseover", empty=False)

base = alt.Chart(df, name="my_panel").encode(x="x:Q", y="y:Q")
lines = base.mark_line()
points = base.encode(size=alt.condition(hover, alt.value(120), alt.value(40))).mark_circle()
layered = lines + points

p1 = layered.transform_filter("datum.x > 1")
p2 = layered.transform_filter("datum.x < 3")

spec = alt.vconcat(p1, p2).add_params(hover).to_dict()
print(spec["vconcat"][0]["layer"][0]["name"])  # "my_panel_0"  ← expected "my_panel"
print(spec["vconcat"][1]["layer"][0]["name"])  # "my_panel_1"  ← expected "my_panel"
```

The `LayerChart` case also renames when no params are present at all (the `FacetChart` case does not):

```python
# No params — LayerChart still renames on PR #4066 branch
base = alt.Chart(df, name="my_panel").encode(x="x:Q", y="y:Q")
layered = base.mark_line() + base.mark_circle()
p1 = layered.transform_filter("datum.x > 1")
p2 = layered.transform_filter("datum.x < 3")

spec = alt.vconcat(p1, p2).to_dict()
print(spec["vconcat"][0]["layer"][0]["name"])  # "my_panel_0"  ← expected "my_panel"
```

The renaming is necessary — preserving a colliding user-set name would recreate the original rendering bug. But when Altair modifies a **user-supplied** name, maybe a `UserWarning` could be emitted so the user knows to update any downstream references. No warning is needed when renaming auto-generated content-hash names (those matching `view_<16 hex chars>`), since those are an internal implementation detail.

This mirrors the existing pattern in the codebase, where Altair already warns for analogous silent fixups:

```
UserWarning: Automatically deduplicated selection parameter with identical configuration.
If you want independent parameters, explicitly name them differently ...
```

We also don't want Altair to be too noisy, so I'm not 100% sure on this fix, but I'm recording it here for visibility.

## Proposed fix direction

In `_combine_subchart_params` (`altair/vegalite/v6/api.py`):

- The **`FacetChart`** branch (`spec.layer[0].name = f"{_view_base_for_chart(spec.layer[0])}_{i}"`) runs inside the param loop and does not distinguish between user-set and auto-generated names.
- The **`LayerChart`** pre-pass added in PR #4066 (`layer.name = f"{_view_base_for_chart(layer)}_{i}"`) has the same issue, and additionally runs even when there are no params.

Add a helper that distinguishes auto-generated names from user-set ones, and emit a warning only for the latter:

```python
import re
_AUTO_NAME_RE = re.compile(r'^view_[0-9a-f]{16}(_\d+)?$')

def _is_auto_name(name: str) -> bool:
    return bool(_AUTO_NAME_RE.match(name))
```

Then wrap the rename with a warning for user-set names:

```python
if isinstance(layer, Chart) and layer.name is not Undefined:
    new_name = f"{_view_base_for_chart(layer)}_{i}"
    if not _is_auto_name(layer.name) and layer.name != new_name:
        warnings.warn(
            f"Chart name {layer.name!r} was automatically renamed to {new_name!r} "
            "to avoid view name collisions in concat. If you need to reference this "
            "view by name, use the renamed form.",
            UserWarning,
        )
    layer.name = new_name
```

The same guard could be applied to the existing `FacetChart` branch.

_No response_

eugen-goebel:
## Problem
Closes #4070.

A user-supplied `name=` on a chart that becomes an inner spec inside a concatenated `FacetChart` or `LayerChart` was silently overwritten with a position suffix (`my_panel` becomes `my_panel_0`, `my_panel_1`). Both rename paths in `_combine_subchart_params` append `_i` unconditionally to disambiguate identical charts, which also caught names the user set on purpose.

## Fix
Add a small helper `_has_user_set_name`. Auto-generated view names come from the content hash (`_get_view_hash_name`), which deliberately excludes the name, so a name that does not match that hash was set by the user. The two rename paths now only apply the position suffix to auto-generated names and leave user-set names untouched. This also handles a user name that happens to end in `_<digit>` (for example `report_0`), since its base still does not match the content hash.

## Tests
Adds `test_facet_chart_preserves_user_set_name_in_concat` and `test_layer_chart_preserves_user_set_name_in_concat`, mirroring the two minimal reproductions from the issue. Both fail on main (`my_panel_0`, `my_panel_1`) and pass with the fix. The existing auto-disambiguation tests still pass, so identical unnamed charts are still disambiguated.

I commented on the issue to claim it before opening this.


joelostblom:
Thanks for the PR @eugen-goebel! Would preserving user names like this not reintroduce the rendering bug fixed by #4066?  I'm not sure about the best fix. Maybe we can only preserve user names only when they do not collide and omit a warning otherwise?

eugen-goebel:
Thanks @joelostblom, you are right, and I confirmed it. If two subcharts in a concat share the same user-set name, preserving it unchanged does bring back the #4066 collision. I reproduced it with two panels named "my_panel" that differ only by a parent-level transform, and the compiled spec ends up with two identically named views. My current tests actually assert that colliding output, so they were wrong.

I will switch to the approach you suggested: keep a user-set name only when it is unique across the concatenated subcharts, and fall back to the positional `_0`/`_1` suffix with a `UserWarning` on collision. Unique names stay untouched so the original case still works, and a collision renders correctly again. I will update the tests to match (unique names preserved with no warning, colliding names disambiguated and warned).

One scope question: detecting collisions among the immediate concat siblings is straightforward, but the same name reused across two separate inner concats that are later hconcat-ed together would need a top-level uniqueness sweep. Would you prefer I scope this PR to the single-level sibling case and open a follow-up for the nested case, or handle both here?

eugen-goebel:
Implemented the collision-aware version: a user-set name is kept when it is unique among the concatenated subcharts, and falls back to the positional `_0`/`_1` suffix with a `UserWarning` when two subcharts share one. Updated the tests accordingly, since the previous ones asserted the colliding output.

I scoped this to collisions among immediate siblings. Names reused across separate inner concats that are later combined would need a top-level uniqueness sweep, which I left out here, happy to add it if you would rather have it in this PR.

joelostblom (changes requested):
Thanks for the updates @eugen-goebel,
It seems like there are still some issues. For example, if I try to run the following code:

```py
import altair as alt
import pandas as pd

df = pd.DataFrame(
    {
        "x": [1, 2, 3],
        "y": [1, 4, 9],
        "category": ["a", "b", "c"],
    }
)

hover = alt.selection_point(
    fields=["x"],
    on="mouseover",
    empty=False,
)


def panel(name):
    return (
        alt.Chart(df, name=name)
        .mark_line()
        .encode(
            x="x:Q",
            y="y:Q",
        )
        .add_params(hover)
        .facet("category:N")
    )


chart = alt.vconcat(
    panel("report"),
    panel("report"),
    panel("report_0"),
)

spec = chart.to_dict()

print("Final view names:")
print([item["spec"]["name"] for item in spec["vconcat"]])

print("Parameter views:")
print(spec["params"][0]["views"])

chart
```

The output shows that it is still possible to end up with duplicate names:

```
Final view names:
['report_0', 'report_1', 'report_0']

Parameter views:
['report_0', 'report', 'report_1']
```

The first two report views are disambiguated to report_0 and report_1, but the generated report_0 collides with the third chart’s explicitly supplied report_0.
The relevant generated Vega-Lite structure is:
```json
{
  "params": [
    {
      "views": [
        "report_0",
        "report",
        "report_1"
      ]
    }
  ],
  "vconcat": [
    {
      "spec": {
        "name": "report_0"
      }
    },
    {
      "spec": {
        "name": "report_1"
      }
    },
    {
      "spec": {
        "name": "report_0"
      }
    }
  ]
}
```

The final view names should be unique, and the parameter’s views should match those final names.

eugen-goebel:
You were right, thank you for the example. Two separate problems.

The first is the one your snippet shows. My collision check only compared user-set names against each other, so a generated replacement name was never checked against anything. With panels named `report`, `report` and `report_0`, the first two were disambiguated into `report_0` and `report_1`, and that generated `report_0` then collided with the third panel's own name. Unique user-set names are now reserved before any positional name is generated, and the positional name counts up until it finds a free one. `{base}_{i}` is still tried first, so nothing changes in the case without a collision.

The second is the parameter views. `.facet()` runs each panel through `_combine_subchart_params` once on its own, before the concat renames anything, and that pre-fills `p.views` with the original name. The outer merge then carried that stale name along. The views are now taken from the renamed target, and `p.views` is only consulted when the subchart is a compound with no direct target.

Running your script on the branch now gives:

```
Final view names:
['report_1', 'report_2', 'report_0']
Parameter views:
['report_1', 'report_2', 'report_0']
```

Worth pointing out: the first panel is now `report_1`, not `report_0`, because `report_0` is taken by the third panel. That follows from reserving the user-set names first, but tell me if you would rather have the generated names not depend on what the other panels are called.

Added your case as a test, asserting that the final names are pairwise distinct and that the parameter views are exactly those names.

PR #4083, opened 2026-07-17
fix: Expression objects not working in alt.value()

### LINKED ISSUE #4082: Encoding with an expression object but not string requires wrapping in `{"expr": ...}`
### What happened?

Using a string expression, e.g. `alt.expr("scale()")` for the value of an encoding channel works as intended, but creating the same expression using dot notation, e.g., `alt.expr.scale()`, requires wrapping in an `{"expr": ...}` dict.

Example:

```python
import pandas as pd
import altair as alt

df = pd.DataFrame({"group": ["A", "B"], "value": [1, 2]})
annotations = pd.DataFrame(
    {"c1": ["A"], "c2": ["B"], "y": [2.5], "text": ["p<0.05"]}
)

_expr_str = alt.expr("(scale('x', datum.c1)+bandwidth('x')/2)")
_expr_obj = alt.expr.scale("x", alt.datum.c2) + alt.expr.bandwidth("x") / 2

_bars = alt.Chart(df).mark_bar().encode(x="group", y="value")
_layer_expr_str = (
    alt.Chart(annotations)
    .mark_point(color="green")
    .encode(x=alt.X(value=_expr_str), y="y")
)

_layer_expr_obj = (
    alt.Chart(annotations)
    .mark_point(color="red")
    # Why does this require wrapping in a "expr" dict, but the other one doesn't?
    .encode(x=alt.X(value={"expr": _expr_obj}), y="y")
)
_chart = alt.layer(_bars, _layer_expr_str, _layer_expr_obj)
_chart
```

Chart

I think it makes the most sense for the following code to work:

```python
_layer_expr_obj = (
    alt.Chart(annotations)
    .mark_point(color="red")
    # No wrapping with a dict
    .encode(x=alt.X(value=_expr_obj), y="y")
)
```

Or better yet, support both the expression object and string in the shorthand syntax `alt.X(_expr_obj)` (though I would be happy with the former)

6.2.2

genrichez:
Closes #4082

## Problem

Passing an `Expression` object (from `alt.expr.*`) to `alt.value()` fails with a schema validation error. The string form (`alt.expr("...")`) works fine, but the object form does not:

```python
import altair as alt

# This works:
alt.value(alt.expr("scale('x', datum.c1)"))

# This fails with SchemaValidationError:
alt.value(alt.expr.scale("x", alt.datum.c1) + alt.expr.bandwidth("x") / 2)
```

## Cause

`Expression.to_dict()` returns a plain string (the JS expression text). When `alt.value(expr_obj)` creates `{"value": <Expression>}` and this gets serialized, the value becomes a raw string like `"scale('x',datum.c1,null)"`. But the Vega-Lite schema expects `value` to be either a number or a mapping like `{"expr": "..."}`, so validation rejects it.

The `alt.expr("...")` form works because it returns an `ExprRef` object whose `to_dict()` produces `{"expr": "..."}` directly.

## Fix

In `alt.value()`, detect `Expression` instances and wrap them in `ExprRef` before storing. This produces the correct `{"value": {"expr": "..."}}` output.

```python
def value(value, **kwargs):
    if isinstance(value, _expr_core.Expression):
        value = core.ExprRef(expr=repr(value))
    return _Value(value=value, **kwargs)
```

## Testing

- Verified the exact reproduction from the issue works
- Tested with BinaryExpression, FunctionExpression, UnaryExpression, datum expressions
- Confirmed SelectionExpression (which uses a different path) still works
- Confirmed plain numeric and string values are unaffected
- All 257 runnable existing tests pass

### DIFF
--- a/altair/vegalite/v6/api.py
+++ b/altair/vegalite/v6/api.py
@@ -1397,6 +1397,8 @@ def when(
 
 def value(value: Any, **kwargs: Any) -> _Value:
     """Specify a value for use in an encoding."""
+    if isinstance(value, _expr_core.Expression):
+        value = core.ExprRef(expr=repr(value))
     return _Value(value=value, **kwargs)  # type: ignore
 
 


mattijn:
Thanks @genrichez! Definitions of expressions become very pythonic like this. Nice!

PR #4085, opened 2026-07-18
fix: allow categorical values with >2 items in selection_interval

### LINKED ISSUE #3896: `value` of `selection_interval` should be list, not tuple
### What happened?

I wanted to create a visualization of a 2D grid, with an initial selection defined as a subgrid of the larger grid.
Here is a minimal reproducible example (y axis only):
```
import polars as pl
import altair as alt
# Create data
data = pl.DataFrame([
    {"x": "1", "y": "A", "value": 1},
    {"x": "2", "y": "A", "value": 2},
    {"x": "3", "y": "A", "value": 3},
    {"x": "1", "y": "B", "value": 4},
    {"x": "2", "y": "B", "value": 5},
    {"x": "3", "y": "B", "value": 6},
    {"x": "1", "y": "C", "value": 7},
    {"x": "2", "y": "C", "value": 8},
    {"x": "3", "y": "C", "value": 9},
    {"x": "1", "y": "D", "value": 10},
    {"x": "2", "y": "D", "value": 11},
    {"x": "3", "y": "D", "value": 12},
])

# Create interval selection with initial value
brush = alt.selection_interval(
    encodings=['y'],
    value={'y': ['A', 'B', 'C']}
)

# Create heatmap
chart = alt.Chart(data).mark_rect().encode(
    x=alt.X('x:N'),
    y=alt.Y('y:N'),
    color=alt.Color('value:Q', scale=alt.Scale(scheme='blues')),
    opacity=alt.condition(brush, alt.value(1), alt.value(0.3))
).add_params(
    brush
).properties(
    width=400,
    height=300,
)

chart
```

But it does not work. I get
```
Multiple errors were found. Error 1: 'A' is an invalid value for `0`. Valid values are of type 'boolean', 'number', or 'object'. Error 2: 'B' is an invalid value for `1`. Valid values are of type 'boolean', 'number', or 'object'. Error 3: 'C' is an invalid value for `2`. Valid values are of type 'boolean', 'number', or 'object'.
```

At this point: I am confused for 2 reasons: 
- the error message is not helpful at all to know why it fails (I tried to do it again with number instead of letters, and same error).
- Should the content of the list be the interval (start-end value) or the list of all selected values ? The documentation does not give any hint (https://altair-viz.github.io/user_guide/generated/api/altair.selection_interval.html)

I was able to implement what I wanted in Vega-lite, and it actually works:
```json
{
      "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
      "width": 400,
      "height": 300,
      "data": {
        "values": [
          {"x": "1", "y": "1", "value": 1},
          {"x": "2", "y": "1", "value": 2},
          {"x": "3", "y": "1", "value": 3},
          {"x": "1", "y": "2", "value": 4},
          {"x": "2", "y": "2", "value": 5},
          {"x": "3", "y": "2", "value": 6},
          {"x": "1", "y": "3", "value": 7},
          {"x": "2", "y": "3", "value": 8},
          {"x": "3", "y": "3", "value": 9},
          {"x": "1", "y": "4", "value": 10},
          {"x": "2", "y": "4", "value": 11},
          {"x": "3", "y": "4", "value": 12},
        ]
      },
      "params": [{
        "name": "brush",
        "select": {
          "type": "interval",
          "encodings": ["y"]
        },
        "value": {"y": ["1", "2", "3"]}
      }],
      "mark": "rect",
      "encoding": {
        "x": {"field": "x", "type": "ordinal"},
        "y": {"field": "y", "type": "ordinal"},
        "color": {
          "field": "value",
          "type": "quantitative",
          "scale": {"scheme": "blues"}
        },
        "opacity": {
          "condition": {"param": "brush", "value": 1},
          "value": 0.3
        }
      }
}
```
Output:
Image

Altair should show exactly the same thing, without the weird error.

Note: I don't know if the specification somewhere, but there might be a mismatch between the vega-lite spec and the current way user understand the parameter (see [this post](https://stackoverflow.com/questions/73898496/setting-the-default-altair-selection-interval-domain) as an example)

5.5.0

genrichez:
## Fix: Allow categorical values with >2 items in `selection_interval`

Fixes #3896.

### Problem

`alt.selection_interval(encodings=['y'], value={'y': ['A', 'B', 'C']})` raises a `SchemaValidationError` because the Vega-Lite JSON schema restricts `SelectionInitInterval` to `Vector2` (exactly 2 items), even though the Vega-Lite runtime supports arbitrary-length arrays for categorical selections.

### Root Cause

The `SelectionInitInterval` schema definition only allows `Vector2<boolean>`, `Vector2<number>`, `Vector2<string>`, and `Vector2<DateTime>` — all constrained to exactly 2 items (representing a range). However, Vega-Lite actually supports selecting specific discrete categories (e.g., `['A', 'B', 'C']`) via interval selections, as documented in Altair's own docs and confirmed by the Vega-Lite editor.

### Fix

Added a `_patch_schema_file()` step in `tools/generate_schema_wrapper.py` that extends the `SelectionInitInterval` definition to also accept arrays of any length (`minItems: 1`) for `boolean`, `number`, `string`, and `DateTime` types. The patch is applied after downloading the upstream schema, ensuring idempotent code generation.

This aligns the schema validation with:

- The existing `_SelectionIntervalValueMap` type alias (which already supports `Sequence[str]`, `Sequence[float]`, etc.)
- The actual Vega-Lite runtime behavior
- The Altair documentation examples

### Changes

- `tools/generate_schema_wrapper.py` — added `_patch_schema_file()` to patch the schema post-download
- `altair/vegalite/v6/schema/vega-lite-schema.json` — 28 lines added (4 new `anyOf` variants, generated by the patch)
- `tests/vegalite/v6/test_api.py` — regression test covering string categories, numeric categories, mixed channels, and existing 2-item range behavior

### Testing

- All CI checks pass (Python 3.10–3.14, linting, schema generation, docs build)
- New regression test covers: string categories, numeric categories, mixed channels (range + categorical), and existing 2-item range behavior

### DIFF
--- a/altair/vegalite/v6/api.py
+++ b/altair/vegalite/v6/api.py
@@ -1397,6 +1397,8 @@ def when(
 
 def value(value: Any, **kwargs: Any) -> _Value:
     """Specify a value for use in an encoding."""
+    if isinstance(value, _expr_core.Expression):
+        value = core.ExprRef(expr=repr(value))
     return _Value(value=value, **kwargs)  # type: ignore
 
 
--- a/altair/vegalite/v6/schema/vega-lite-schema.json
+++ b/altair/vegalite/v6/schema/vega-lite-schema.json
@@ -23135,6 +23135,34 @@
         },
         {
           "$ref": "#/definitions/Vector2<DateTime>"
+        },
+        {
+          "type": "array",
+          "items": {
+            "type": "boolean"
+          },
+          "minItems": 1
+        },
+        {
+          "type": "array",
+          "items": {
+            "type": "number"
+          },
+          "minItems": 1
+        },
+        {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "minItems": 1
+        },
+        {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/DateTime"
+          },
+          "minItems": 1
         }
       ]
     },
@@ -32457,4 +32485,3 @@
     }
   }
 }
-
--- a/tests/vegalite/v6/test_api.py
+++ b/tests/vegalite/v6/test_api.py
@@ -1053,6 +1053,42 @@ def test_selection():
     assert sel1.name != sel3.name  # Different specifications get different names
 
 
+def test_selection_interval_categorical_value():
+    """
+    Test that selection_interval accepts categorical values with >2 items.
+
+    Regression test for https://github.com/vega/altair/issues/3896
+    """
+    # String categories (the original failing case)
+    brush = alt.selection_interval(
+        encodings=["y"],
+        value={"y": ["A", "B", "C"]},
+    )
+    d = brush.param.to_dict()
+    assert d["value"] == {"y": ["A", "B", "C"]}
+
+    # Numeric categories
+    brush = alt.selection_interval(
+        encodings=["x"],
+        value={"x": [4, 6, 8]},
+    )
+    d = brush.param.to_dict()
+    assert d["value"] == {"x": [4, 6, 8]}
+
+    # Mixed channels: range + categorical
+    brush = alt.selection_interval(
+        encodings=["x", "y"],
+        value={"x": [1, 10], "y": ["A", "B", "C"]},
+    )
+    d = brush.param.to_dict()
+    assert d["value"] == {"x": [1, 10], "y": ["A", "B", "C"]}
+
+    # Existing 2-item (range) case still works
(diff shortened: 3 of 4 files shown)


joelostblom:
Thanks for contributing this PR! Instead of patching the VL schema from altair, could you make a PR to the VL repo to update the schema there instead?

joelostblom:
I'll go ahead and close this since there hasn't been any follow up, but feel free to reopen.

Issue #3403, opened 2024-04-13
Explore and add examples for additional input elements


joelostblom:
### What is your suggestion?

There are many undocumented [input options](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types) that work with Altair. It would be interesting to explore these more, e.g. maybe the `date` input could be useful to filter time series data which is not that easy with sliders currently. `number` also looks useful and maybe even `file`?

This example can be modified:

```py
search_input = alt.selection_point(
    fields=['Name'],
    empty=False,  # Start with no points selected
    bind=alt.binding(
        input='date',  # Change this to any of the options at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types
        placeholder="Car model",
        name='Search ',
    )
)
alt.Chart(data.cars.url).mark_point(size=60).encode(
    x='Horsepower:Q',
    y='Miles_per_Gallon:Q',
    tooltip='Name:N',
    opacity=alt.condition(  # This condition would also need to be updated accordingly
        search_input,
        alt.value(1),
        alt.value(0.05)
    )
).add_params(
    search_input
)
```

![image](https://github.com/vega/altair/assets/4560057/0734aacf-3b8b-4735-9712-9314c00766bd)

Kato-MT:
Hi, I made a date input example for this issue and opened a draft PR: #4088. It includes both the arguments and methods versions and uses a small local time series. I'd appreciate any feedback.

joelostblom:
I think can be closed since https://github.com/vega/altair/pull/4091 and https://github.com/vega/altair/pull/4088

PR #4088, opened 2026-07-28
docs: add date input gallery example

Kato-MT:
## What I changed

- added a date input example that filters a small time series
- added both the arguments and methods syntax versions
- used `timeParse` so the selected date works the same way across time zones
- kept the default chart width and default axis titles so the example stays focused

The data is kept in the example, so it does not depend on an external URL.

Related to #3403.

## Preview

This is the chart with the default January 5 selection. The gallery version has the date picker above it.

Date input example with the default January 5 selection

## Testing

- `pytest tests/test_examples.py -k interactive_date_input -q -n0 -p no:cacheprovider` (`6 passed`)
- checked both files with Ruff format
- confirmed both versions generate the same Vega-Lite spec
- confirmed the default width and field-name axis titles are used

### DIFF
--- a/tests/examples_arguments_syntax/interactive_date_input.py
+++ b/tests/examples_arguments_syntax/interactive_date_input.py
@@ -1,0 +1,57 @@
+"""
+Filter a Time Series with a Date Input
+======================================
+This example shows how to bind an HTML date input to a parameter and use the
+selected date to filter a time-series chart.
+"""
+
+# :new:
+# category: interactive charts
+import altair as alt
+import pandas as pd
+
+source = pd.DataFrame(
+    {
+        "date": pd.date_range("2024-01-01", periods=14, freq="D"),
+        "visitors": [
+            120,
+            135,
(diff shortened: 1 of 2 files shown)


mattijn:
Thanks for this PR @Kato-MT! Reading from the spec it seems nice, but I prefer to have a screenshot included to see how it looks like. My feeling says that we don't need the `.properties(width=600)`, and capitalization of the axis titles also may not be needed. So we keep the spec minimal and focused, but it is hard to tell without a visual. Again, thanks for this PR especially this `alt.expr.timeParse` is not used a lot yet within examples.

Kato-MT:
Thanks, that makes sense. I removed the fixed width and the custom axis titles from both versions, and I added the updated render to the PR description. The gallery version still has the date picker above the chart. I reran the focused example tests and all 6 passed.

mattijn:
Thank you @Kato-MT!

PR #4089, opened 2026-07-29
fix: report missing anywidget for JupyterChart

### LINKED ISSUE #3801: type object 'JupyterChart' has no attribute 'enable_offline'
### What happened?

I'm trying to follow the documentation [here](https://altair-viz.github.io/user_guide/interactions/jupyter_chart.html#offline-usage) but I'm getting an error.

With this code:

```
import altair as alt
alt.JupyterChart.enable_offline()
```

And the output is:

```
AttributeError                            Traceback (most recent call last)
Cell In[4], line 5
      1 import altair as alt
      3 print(alt.__version__)
----> 5 alt.JupyterChart.enable_offline()

AttributeError: type object 'JupyterChart' has no attribute 'enable_offline'
```

I tried to enable the jupyter renderer with offline set to true:

```
import altair as alt
alt.renderers.enable("jupyter", offline=True)

source = data.cars()

alt.Chart(source).mark_circle(size=60).encode(
    x='Horsepower',
    y='Miles_per_Gallon',
    color='Origin',
    tooltip=['Name', 'Origin', 'Horsepower', 'Miles_per_Gallon']
).interactive()
```

and go this (related?) error:

```
AttributeError                            Traceback (most recent call last)
File [~/.pyenv/versions/3.12.1/lib/python3.12/site-packages/IPython/core/formatters.py:977](http://localhost:8889/~/.pyenv/versions/3.12.1/lib/python3.12/site-packages/IPython/core/formatters.py#line=976), in MimeBundleFormatter.__call__(self, obj, include, exclude)
    974     method = get_real_method(obj, self.print_method)
    976     if method is not None:
--> 977         return method(include=include, exclude=exclude)
    978     return None
    979 else:

File [~/.pyenv/versions/3.12.1/lib/python3.12/site-packages/altair/vegalite/v5/api.py:3682](http://localhost:8889/~/.pyenv/versions/3.12.1/lib/python3.12/site-packages/altair/vegalite/v5/api.py#line=3681), in TopLevelMixin._repr_mimebundle_(self, *args, **kwds)
   3680 else:
   3681     if renderer := renderers.get():
-> 3682         return renderer(dct)

File [~/.pyenv/versions/3.12.1/lib/python3.12/site-packages/altair/vegalite/v5/display.py:100](http://localhost:8889/~/.pyenv/versions/3.12.1/lib/python3.12/site-packages/altair/vegalite/v5/display.py#line=99), in jupyter_renderer(spec, **metadata)
     97 offline = metadata.get("offline", False)
     99 # mypy doesn't see the enable_offline class method for some reason
--> 100 JupyterChart.enable_offline(offline=offline)  # type: ignore[attr-defined]
    102 # propagate embed options
    103 embed_options = metadata.get("embed_options")

AttributeError: type object 'JupyterChart' has no attribute 'enable_offline'
```

Any advice on how to resolve this? Is it a package version issue? Which packages should I check? 

Not have an error ! 

5.5.0

dk3yyyy:
## Summary

- add the missing `enable_offline()` API to the `JupyterChart` fallback used when `anywidget` is unavailable
- detect the optional dependency with `find_spec()` and expose the real class during type checking
- remove obsolete type ignores from the Jupyter renderer
- add regression coverage for the missing-dependency behavior and restore imported module state after the test

Fixes #3801

## Testing

- `uv run task test`
- `uv run pytest tests/test_jupyter_chart.py -q --numprocesses=1`
- `uv run pytest tests/vegalite/v6/test_renderers.py -q --numprocesses=1`
- verified a base-only installation without `anywidget` reports the intended `ImportError` for `alt.JupyterChart`, `enable_offline()`, and the `jupyter` renderer

### DIFF
--- a/altair/jupyter/__init__.py
+++ b/altair/jupyter/__init__.py
@@ -1,21 +1,34 @@
-try:
-    import anywidget  # noqa: F401
-except ImportError:
-    # When anywidget isn't available, create stand-in JupyterChart class
-    # that raises an informative import error on construction. This
+from importlib.util import find_spec
+from typing import TYPE_CHECKING, NoReturn
+
+_ANYWIDGET_IMPORT_ERROR = (
+    "The Altair JupyterChart requires the anywidget \n"
+    "Python package which may be installed using pip with\n"
+    "    pip install anywidget\n"
+    "or using conda with\n"
+    "    conda install -c conda-forge anywidget\n"
+    "Afterwards, you will need to restart your Python kernel."
+)
+
+
+def _raise_anywidget_import_error() -> NoReturn:
+    raise ImportError(_ANYWIDGET_IMPORT_ERROR)
+
+
+if TYPE_CHECKING:
+    from .jupyter_chart import JupyterChart
+elif find_spec("anywidget") is None:
+    # When anywidget isn't available, create a stand-in JupyterChart class
+    # that raises an informative import error when its API is used. This
     # way we can make JupyterChart available in the altair namespace
-    # when anywidget is not installed
+    # when anywidget is not installed.
     class JupyterChart:
-        def __init__(self, *args, **kwargs):
-            msg = (
-                "The Altair JupyterChart requires the anywidget \n"
-                "Python package which may be installed using pip with\n"
-                "    pip install anywidget\n"
-                "or using conda with\n"
-                "    conda install -c conda-forge anywidget\n"
-                "Afterwards, you will need to restart your Python kernel."
-            )
-            raise ImportError(msg)
+        def __init__(self, *args: object, **kwargs: object) -> NoReturn:
+            _raise_anywidget_import_error()
+
+        @classmethod
+        def enable_offline(cls, offline: bool = True) -> NoReturn:
+            _raise_anywidget_import_error()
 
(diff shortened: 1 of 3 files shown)


mattijn:
Thanks for this PR @dk3yyyy! Would it be possible to adapt your contribution to have the test run without a `pytest.monkeypach()`? I think we have a CI run that only does the base installation or at least options to have certain optional deps not included. Thanks again for this PR!

dk3yyyy:
Thanks @mattijn! I’ve updated the test to avoid `pytest.monkeypatch` and module reloading. It now runs directly against `alt.JupyterChart` when `anywidget` is unavailable, using the existing Python 3.10 CI jobs where `anywidget` is explicitly uninstalled.

The test covers both `JupyterChart.enable_offline()` and construction through the public API. I also verified it locally in the equivalent no-`anywidget` setup: 1 passed, with the 5 tests requiring `anywidget` correctly skipped.

mattijn:
Thanks @dk3yyyy! CI is happy with your latest commit. Thanks!

dk3yyyy:
Thanks for the review and merge, @mattijn!

PR #4090, opened 2026-07-29
fix: improve schema validation error relevance

### LINKED ISSUE #3752: `SchemaValidationError` relevance heuristic is flawed
### What happened?

> [!NOTE]
> Originally wrote as comment in #2913, but this will require more work to fix

While debugging some tests for (#3750), I've begun to think the underlying assumption here is flawed:

> ...validated against multiple schemas and its parent is a common anyOf validator.
> The error messages produced from these cases are usually 
> very similar and we just take the shortest one.

<details><summary><code>schemapi._deduplicate_additional_properties_errors</code></summary>
<p>

https://github.com/vega/altair/blob/be5e9ecd61d099c847dca44fb3a4283940c1e3b8/tools/schemapi/schemapi.py#L452-L479

</p>
</details> 

This *assumption* conflicts directly with the channel wrappers, since the `"anyOf"` is how they are differentiated.

https://github.com/vega/altair/blob/be5e9ecd61d099c847dca44fb3a4283940c1e3b8/tools/generate_schema_wrapper.py#L816

# Examples
These are all tested against (https://github.com/vega/altair/pull/3750/commits/8fb2661555ff9f70db381b27506fdb3eb31b5a6d)

```py
import altair as alt

chart = alt.Chart().mark_point()

one_err = alt.value(1, bin=True)
two_err = alt.value(1, bin=True, axis=1)
two_err_2 = alt.value(1, bin=True, aggregate="sum")
three_err = alt.value(1, bin=True, axis=1, aggregate="sum")
```

We're trying to match [`alt.YValue`](https://altair-viz.github.io/user_guide/generated/channels/altair.YValue.html).
The most helpful thing would be identifying **all of the invalid properties**.

Naively picking the *shortest* message is an **unfit heuristic**.
This can be observed by seeing how the final message changes in an *unpredicatble* way, simply by adding **more** invalid properties:

```py
>>> chart.encode(y=one_err)
SchemaValidationError: `YValue` has no parameter named 'bin'
Existing parameter names are:
value

>>> chart.encode(y=two_err)
SchemaValidationError: '1' is an invalid value for `axis`. Valid values are of type `Mapping[str, Any] | None`.

>>> chart.encode(y=two_err_2)
SchemaValidationError: `YValue` has no parameter named 'value'
Existing parameter names are:
value

>>> chart.encode(y=three_err_2)
SchemaValidationError: '1' is an invalid value for `axis`. Valid values are of type `Mapping[str, Any] | None`.
```

## Exploring the error tree
Diving into the error tree for `two_err_2` gives us some more clarity:

### Capturing an error
```py
import altair as alt
from altair.utils.schemapi import SchemaValidationError

two_err_2 = alt.value(1, bin=True, aggregate="sum")
try:
    alt.Chart().mark_point().encode(y=two_err_2).to_dict()
except SchemaValidationError as err:
    some = err
```

### Spec being validated
```py
>>> some.instance
{'value': 1, 'bin': True, 'aggregate': 'sum'}
```

### `altair` transformed message
Since (https://github.com/vega/altair/pull/3750/commits/ea647ebc83e84c0cc0b9731b3acf99b35c4393a4) we now at least get `YValue` included:
```py
>>> some.message
"`YValue` has no parameter named 'value'\n\nExisting parameter names are:\nvalue   \n\nSee the help for `YValue` to read the full description of these parameters"

```

### The first error
Backtracking to what the message originally looked like from `jsonschema`.

I think it is pretty clear this **is not** the error we want:
```py
>>> some._original_message 
"Additional properties are not allowed ('value' was unexpected)"
```

### Parent message is more useful
Going back one level `(.parent`) starts to make more sense:
```py
>>> some.parent.message
"{'value': 1, 'bin': True, 'aggregate': 'sum'} is not valid under any of the given schemas"
```

### Significant backtracking
It is still *possible* to undo all of the work from `validate_jsonschema`

<details><summary><code>schemapi.validate_jsonschema</code></summary>
<p>

https://github.com/vega/altair/blob/be5e9ecd61d099c847dca44fb3a4283940c1e3b8/tools/schemapi/schemapi.py#L135-L170

</p>
</details> 

I wouldn't want to solve the problem this way, but an example of backtracking to a useful context:
```py
tree_nav = {
    err.parent.schema["anyOf"][idx]["$ref"].removeprefix("#/definitions/"): err.message
    for idx, err in enumerate(some.parent.context)
}
>>> tree_nav
{'PositionFieldDef': "Additional properties are not allowed ('value' was unexpected)",
 'PositionDatumDef': "Additional properties are not allowed ('aggregate', 'bin', 'value' were unexpected)",
 'PositionValueDef': "Additional properties are not allowed ('aggregate', 'bin' were unexpected)"}
```

So the hacky way to do this would be replacing the error we got from `'PositionFieldDef'` with the one for `'PositionValueDef'`.
For our `{'value': 1, 'bin': True, 'aggregate': 'sum'}`, we identify the two extra invalid properties `"...('aggregate', 'bin' were unexpected)"`

I think #3750 offers *an* improvement, but really we need to fix this **before raising in `validate_jsonschema`**.
By that stage, we've dropped errors that are better candidates which makes fixing it later much more difficult.

Maybe this would be something to pick up when revisiting #3547, as that PR contains a lot of refactoring/redocumenting in `tools.schemapi.schemapi.py`

5.6.0dev

dk3yyyy:
Fix for #3752

# Problem

Altair's schema error reporting can attribute failures to the wrong part of a Vega-Lite `anyOf` union. In positional channels this reported `value` itself as unexpected instead of the parameters that are invalid for the value branch — and `datum=` was never wrapped in its own channel class, so errors on datum channels were reported against the wrong type.

# Reproduce

```python
import altair as alt

channel = alt.value(1, bin=True, aggregate="sum")
alt.Chart().mark_point().encode(y=channel).to_dict()
```

Current output:

```text
SchemaValidationError: `YValue` has no parameter named 'value'
```

With this PR:

```text
SchemaValidationError: `YValue` has no parameters named 'aggregate', 'bin'
```

The same error-selection issue could hide an array-size error behind a nested item error:

```python
alt.selection_interval(
    resolve="global",
    value={"x": ["Europe"], "y": [10, 20]},
)
```

With this PR, the error points to the actual constraint:

```text
'['Europe']' is an invalid value for `x`. Valid values are of type `Sequence` with at least 2 items.
```

# Proposed solution

When serialization fails, Altair usually already knows the type of the object it is validating. The error tree is now narrowed by re-validating against the live object found at the error path (`_narrow_error_from_live_object`, called when constructing `SchemaValidationError`), instead of reconstructing the intended union branch after the fact.

Two changes make that live-object path cover the reported cases:

- `datum=` is now wrapped in its `*Datum` channel class, just as `value=` is wrapped in its `*Value` class (`_invert_group_channels` / `_wrap_in_channel` in `core.py`). This is the other half of #3752: datum channels were never typed, so their errors fell back to generic union handling.
- Validation errors are grouped by JSON path and narrowed against the object found at that path before the final message is assembled. When the object's own schema is itself a union, or no live object exists, the errors are left untouched and handled by the fallback below.

Where no live object exists — most notably a raw dict passed to `bind=` — a conservative union-branch fallback remains: a branch is preferred only when root-level required properties and their discriminator values reliably identify the instance. A `const`/`enum` mismatch excludes a branch only when the value satisfies another branch's discriminator (it then positively identifies that branch); when the value satisfies no branch's discriminator it is treated as ambiguous and that branch's own error is kept, so a typo like `"ragne"` surfaces instead of a misleading `additionalProperties` message. If multiple branches can accept the same properties, or a reference cannot be resolved, the competing errors are retained rather than guessed.

Local Vega-Lite schema references are resolved against the effective validation root using the same `referencing` registry as validation itself — including Altair's internal `urn:vega-lite-schema` form on current `jsonschema` — with a `RefResolver` fallback on older versions. Chained references resolve through the returned resolver.

Supporting improvements to error selection and messages:

- JSON paths are compared by path segments, not string prefixes (`x2` is a sibling of `x`, not a child).
- `minItems`/`maxItems` errors take precedence over nested item errors and are reported as `of type `Sequence` with at least/at most N items`.
- The "has no parameter named ..." message is computed from the instance against the schema — reporting all unexpected parameters and respecting `patternProperties` — instead of parsing jsonschema's message string.

My main uncertainty remains how aggressively to choose a union branch when the schema is ambiguous. I kept it conservative: only narrow the error tree with positive evidence and otherwise retain the available branch errors, so the fallback generalizes across Vega-Lite's referenced unions rather than special-casing `YValue`.

# Tests

The synthetic union-schema matrix was replaced with regressions that exercise Altair's shipped Vega-Lite schema:

- the `YValue` cases reported in #3752, including multiple invalid parameters;
- the interval-selection `minItems` case reported in #3752;
- Vega-Lite's referenced `Binding` union, including its `const`, `enum`, and general input branches;
- the `bind=` typo case from review, and the const-discriminated-union case where a value satisfying another branch must exclude the contradicted branch;
- `datum=` wrapping, `patternProperties`-aware unexpected-parameter messages, sibling JSON paths, and `maxItems`;
- union references resolved against the validation root (`schema` vs `rootschema`).

Local validation: `uv run task test` passes (1,943 main tests + 2 serial tests) together with Ruff, formatting, MyPy, and Ty; focused resolver coverage also passes with the minimum supported `jsonschema==3.0.2`; `tools/schemapi/schemapi.py` stays synchronized and regeneration is idempotent. CI is green across Python 3.11–3.14 with minimum and latest dependencies.

Closes #3752


mattijn:
Thanks for the PR! Could you instruct your agent to first fix your review environment so the uv toolchain and pytest are available locally?

Also can you update the PR description? Currently it feels more like feedback of your code generator to you than an invitation for a review. Maybe something like this as template?:

***


```bash
<!-- Title: fix: <short imperative> -->

Fix for #<issue>

# Problem
One or two sentences: what breaks, and why the current code does it.

# Reproduce
```python
# minimal runnable Altair snippet

Current output:
<actual error message>

With this PR:
<new error message>

Proposed solution
What changed and why, in prose or a short list. 

Please write in first person, and explain briefly your trail of thought (make your uncertainty visible).
```

***

Also a reminder to try we generalize to the varies reported cases and not have a hotfix for a single usecase.

I'm also a bit confused on the Draft7Validator tests added, these seem more appropriate for jsonschema itself? Please use the vegalite jsonschema itself where necessary.

Thanks for your work!🦾🙌

dk3yyyy:
Thanks, this was helpful. I fixed my local setup first and can now run the full `uv` test, lint, formatting, and type-checking workflow.

I also rewrote the PR description around the actual problem, included runnable examples with the before/after errors, and explained why I kept the branch selection conservative.

You were right about the synthetic tests too. I replaced the broad `Draft7Validator` matrix with regressions using Altair’s Vega-Lite schema, including the reported `YValue` and interval-selection cases and the referenced `Binding` branches. I’ve pushed the updated implementation and tests.

mattijn:
Thanks for the ongoing work. I'm still on mobile, so I can't check the details but looking to the code diff I don't think we need to add another schema-resolution layer for this, my gut feeling says we can use existing functions that are available for this, including usage of `referencing` package.

Please expect some scrutiny, since we have aimed to simplify these routines previously, see for reference https://github.com/vega/altair/pull/2771 and other mentioned issues/PR in there. 

Some crumbs from my LLM assistant: 

Instead of resolving pointers would it be possible to thread the registry/resolver that `_get_errors_from_spec` already builds into `_get_leaves_of_error_tree`? For jsonschema ≥4.18 that’s `registry.resolver().lookup(ref).contents`, on older versions `resolver.resolve(ref)[1]`. Hopefully with this we can reduce the added code like `_decode_json_pointer_*`, `_resolve_json_pointer` and `_resolve_local_schema_reference` all go, along with their tests, also we add the `urn:vega-lite-schema` prefix by ourselves, so it feels off to strip it again too

Thanks again for the ongoing effort. We will get there🤗

dk3yyyy:
Thanks for the guidance — I've reworked the reference resolution to reuse the registry, as suggested.

- `_get_errors_from_spec` now returns the `registry` it builds, and `_get_leaves_of_error_tree` receives it and threads it down; `_union_branch_schemas` resolves branch refs through `registry.resolver()`.
- The pointer helpers (`_decode_json_pointer_*`, `_resolve_json_pointer`, `_resolve_local_schema_reference`) and their unit tests are gone.
- `_resolve_references` now uses `registry.resolver().lookup(...)`, falling back to `jsonschema.RefResolver` on older jsonschema versions. Since we build the registry with the `urn:vega-lite-schema` prefix ourselves, refs are prefixed when missing rather than stripped.
- Compatibility with the minimum supported `jsonschema==3.0.2` is covered by a focused test.

On CI: the current failures are pre-existing network flakes in `tests/test_datasets.py` (external dataset downloads returning `URLError: Connection reset by peer`, plus a polars `BufferError` under xdist) — unrelated to these changes. Locally, `tests/utils/test_schemapi.py` passes in full, and `make lint`/`make typecheck`/`make test` are green.

Happy to iterate further if the branch-selection heuristics need tightening.

altair/utils/schemapi.py
```diff
@@ -324,12 +331,295 @@ def _get_leaves_of_error_tree(
...
+        if not required_names or not required_names.issubset(instance):
+            continue
+        required_values_match = _required_values_match_instance(
+            required, schema, instance
+        )
+        if required_values_match is None:
+            required_values_match = not _branch_has_required_value_error(
+                error.context, branch_index, required_names
+            )
+        if required_values_match and _required_values_identify_branch(
```
joelostblom, line 630: I think that this branch-selection heuristic can hide the actual enum error. A required property is not necessarily an unambiguous discriminator. In particular, a generic fallback branch may accept the property's type while a more specialized branch restricts it with `enum` or `const`. If another property identifies the specialized branch, selecting the generic branch produces a misleading `additionalProperties` error.

Take this example:

```python
import altair as alt

slider = alt.param(
    name="threshold",
    value=5,
    bind={
        "input": "ragne",  # Intentional typo: should be "range"
        "min": 0,
        "max": 10,
    },
)

chart = (
    alt.Chart({"values": [{"x": 1}]})
    .mark_point()
    .encode(x="x:Q")
    .add_params(slider)
)
```

On this PR, the result is:

```text
`VariableParameter` has no parameters named 'max', 'min'
```

That is misleading because `min` and `max` are valid properties of a range binding. They also provide evidence that the intended branch is `BindRange`. The actual error is the misspelled value `"ragne"`.

On `main`, the result correctly identifies the invalid input value (although the error message could still be improved):

```text
'ragne' is an invalid value for `input`. Valid values are one of ['radio', 'select'].

'checkbox' was expected
'range' was expected
```

The same behavior can be demonstrated with a reduced schema:

```python
from altair.utils.schemapi import validate_jsonschema

schema = {
    "anyOf": [
        {
            "type": "object",
            "required": ["kind"],
            "properties": {
                "kind": {"enum": ["allowed"]},
                "payload": {"type": "integer"},
            },
            "additionalProperties": False,
        },
        {
            "type": "object",
            "required": ["kind"],
            "properties": {
                "kind": {"type": "string"},
            },
            "additionalProperties": False,
        },
    ]
}

error = validate_jsonschema(
    {"kind": "typo", "payload": 1},
    schema,
    raise_error=False,
)

assert error is not None
print(error.message)
```

This PR reports:

```text
Additional properties are not allowed ('payload' was unexpected)
```

`main` reports the more actionable error:

```text
'typo' is not one of ['allowed']
```

Could we adjust the heuristic so that an enum/const mismatch does not automatically make a specialized branch irrelevant when other properties identify that branch? Please also add the real binding example as a regression test and mirror the fix in `tools/schemapi/schemapi.py`.

*generated with codex*

joelostblom (changes requested):
Thanks for working on this! It will be great to have improved error heuristics. I found one smaller issue that I think would be good to fix in addition to any additional comments from mattijn.

altair/utils/schemapi.py
dk3yyyy: Thanks for the review. I've addressed all three points: the heuristic now keeps a specialized branch when its only mismatch is a const/enum on a required property (so the BindRange / 'allowed' enum errors surface instead of the misleading additionalProperties / unknown-parameters ones), the real binding example is added as a regression test, and the fix is mirrored in tools/schemapi/schemapi.py. Both examples now produce the expected messages, and the full suite is green. Re-requesting review — happy to iterate further.

joelostblom: Thanks @dk3yyyy, it seems like the fix migh overcorrect by no longer treating const/enum mismatches as branch-exclusion evidence in _branch_excludes_required_value (altair/utils/schemapi.py:556). This can retain a branch that is clearly contradicted by a valid discriminator.

For example, with branches distinguishing kind="integer" from kind="string":

```
from altair.utils.schemapi import SchemaBase, SchemaValidationError

schema = {
    "anyOf": [
        {
            "type": "object",
            "required": ["kind"],
            "properties": {
                "kind": {"const": "integer"},
                "payload": {"type": "integer"},
            },
            "additionalProperties": False,
        },
        {
            "type": "object",
            "required": ["kind"],
            "properties": {
                "kind": {"const": "string"},
                "payload": {"type": "string"},
            },
            "additionalProperties": False,
        },
    ]
}


class Example(SchemaBase):
    _schema = schema


try:
    Example(
        kind="integer",
        payload="bad",  # Invalid: the "integer" branch requires an integer.
    ).to_dict()
except SchemaValidationError as error:
    print(error)
```

At commit c8865185, this prints:

```
Multiple errors were found.

Error 1: 'bad' is an invalid value for `payload`. Valid values are of type `int`.

Error 2: 'integer' is an invalid value for `kind`.

    'string' was expected
```

The second error should not appear. kind="integer" positively identifies the first branch, so the only relevant error is:
'bad' is an invalid value for `payload`. Valid values are of type `int`.
The parent commit reports only that relevant error.

---

Also note that The new regression test fails mypy at tests/utils/test_schemapi.py:1046:

```
Argument "bind" to "param" has incompatible type "dict[str, object]";
expected "Binding | UndefinedType"
```

dk3yyyy: Thanks — I've refined the heuristic so a const/enum mismatch is only treated as ambiguous when the value satisfies no branch's discriminator; when it matches another branch's const/enum, the contradicted branch is still excluded. The overcorrection example now reports only the payload error, the original enum-typo and binding cases are unchanged, and the mypy issue in the test is fixed with a scoped `# type: ignore[arg-type]` (mirroring the existing test pattern). Full suite, ruff, and mypy are green.

mattijn:
While reviewing last evening, I think I found something that may account for half of this issue. In our `_reduce` function (inside `_invert_group_channels`) we currently have this:

https://github.com/vega/altair/blob/1db41564e617d8ad265a00937c045f80e08641a0/altair/utils/core.py#L957-L973

with a docstring saying: *"Never includes `datum`, as it is never utilised in `wrap_in_channel`."* If we actually accept `datum`, it gets wrapped in a `*Datum` channel class, just as `value` is wrapped in a `*Value` class (which we then subsequently can use in the error validation).

The other thing, I think this PR's current approach of reconstructing the relevant branch after the fact is a bit risky, since it has to guess. Altair already has the object's own type available at that point, so we can re-validate against that directly instead.

My branch with applied changes is here: https://github.com/vega/altair/compare/main...fix/3752-validate-against-live-object-schema (or [code.diff](https://github.com/vega/altair/compare/main...fix/3752-validate-against-live-object-schema.diff))

Because I can reuse the existing object the introduced new LOC is much lower.

The only thing I haven't get to is this example from original post (and is working correctly in this PR), is this example

```python
alt.selection_interval(
    resolve="global",
    value={"x": ["Europe"], "y": [10, 20]},
)
```

Technically speaking, while still super relevant to be fixed, I think it is a bit of a different issue than #3752, since its because `x` has two true complaints at once (the list is too short, and its one item is the wrong type), and the current path-picking rule shows the more specific one (about the item) instead of the more useful one (about the length). 

Its also defined in the doctsring of this `_subset_to_most_specific_json_paths` function:

https://github.com/vega/altair/blob/fde4b47a4eb219dd51aeff934e5289faab587a19/altair/utils/schemapi.py#L330-L340

_This is done under the assumption that more specific json paths give more helpful error messages to the user._

All errors can be retrieved from this code:

```python
from altair.utils.schemapi import (
    _get_errors_from_spec,
    _get_leaves_of_error_tree,
    _group_errors_by_json_path,
)
from altair.vegalite.v6.schema.core import SelectionParameter

spec = {
    "name": "__TEMP__",
    "select": {"type": "interval", "resolve": "global"},
    "value": {"x": ["Europe"], "y": [10, 20]},
}

errors = _get_errors_from_spec(spec, SelectionParameter._schema, rootschema=SelectionParameter._rootschema)
leaves = _get_leaves_of_error_tree(errors)
grouped = _group_errors_by_json_path(leaves)

grouped
```

```python
{
 '$.value': [
  <ValidationError: "{'x': ['Europe'], 'y': [10, 20]} is not of type 'number', 'string', 'boolean', 'null'">,
  <ValidationError: "Additional properties are not allowed ('x', 'y' were unexpected)">,
  <ValidationError: "{'x': ['Europe'], 'y': [10, 20]} is not of type 'array'">
],
 '$.value.x[0]': [
  <ValidationError: "'Europe' is not of type 'boolean'">,
  <ValidationError: "'Europe' is not of type 'number'">,
  <ValidationError: "'Europe' is not of type 'object'">
],
 '$.value.x': [
  <ValidationError: "['Europe'] is too short">,
  <ValidationError: "['Europe'] is too short">,
  <ValidationError: "['Europe'] is too short">,
  <ValidationError: "['Europe'] is too short">
]}
```

I'm wondering what you think @dk3yyyy here

dk3yyyy:
@/tmp/mattijn_reply.md

dk3yyyy:
Thanks for the detailed look — you were right that guessing branches is the fragile part. I've switched to your approach: `SchemaValidationError` now re-validates against the live object's own schema first, so it uses the type Altair already knows instead of reconstructing it. I also applied your `_invert_group_channels` observation, so `datum=` now gets wrapped in its `*Datum` class the same way `value=` is — that covers the other half of #3752.

The only thing I kept from the previous approach is a fallback for cases where there is no live object at the error path, like a raw dict passed to `bind=`. With `alt.param(value=1, bind={"input": "range", "min": "bad"})` there's no `Binding` object to re-validate against, so it needs branch selection to get to `'bad' is an invalid value for 'min'` instead of an `additionalProperties` message. It's a much smaller surface than before, but I kept it for exactly that case, and the new tests cover both paths.

On the `selection_interval` example — agreed, that's a separate issue: `x` has two genuine complaints and the path-picking rule intentionally shows the more specific one, so I've left it out of scope here.

CI is green on the current head. Happy to adjust anything else if you have suggestions.

dk3yyyy:
@joelostblom just a heads-up that both points from the last round are addressed on the current head:

- the overcorrection on `{"kind": "integer", "payload": "bad"}` no longer leaks a spurious `'integer' is an invalid value for 'kind'` — a value that satisfies another branch's discriminator now counts as evidence *against* the current branch, and the message only reports the `payload` error
- the mypy failure on the `bind=` dict is fixed (typed with `# type: ignore[arg-type]`, mirroring the existing pattern)

The branch is also rebased on the latest `main` (including the referencing/registry refactor) and CI is green. Would appreciate another look whenever you have a chance.

mattijn:
Heads up: not forgotten this PR and so now and then I'm reading this thread and the code again. But my feeling is that it has become a bit messy, since we are piling fixes.

Maybe good idea if we can get a overview of the different Altair examples discussed and discovered in this thread that we feel can improve in its error message justification.

So it's easier to create a few follow up issues if these don't touch the same heuristic as the original linked issue aimed to be fixed.

dk3yyyy:
Agreed, and good prompt to separate concerns. Here's the overview of the error-justification cases from this thread, split by whether they touch the #3752 heuristic.

**In scope of #3752, fixed in this PR (each has a regression test):**

| Example | Behavior now |
|---|---|
| `alt.value(1, bin=True, aggregate="sum")` | `YValue has no parameters named 'aggregate', 'bin'` instead of `no parameter named 'value'` |
| Multiple invalid value params (`bin`, `axis`, `aggregate`) | all unexpected params reported, `patternProperties` respected |
| `datum=` in encoding channels | wrapped in its `*Datum` class, same as `value=` (core.py) |
| `bind={"input": "ragne", "min": 0, "max": 10}` | `'ragne' is an invalid value for 'input'` instead of `no parameters named 'max', 'min'` |
| `kind="integer", payload="bad"` (const-discriminated union) | only the `payload` error, no spurious `'integer' is an invalid value for 'kind'` |
| Union branch `$ref`s | resolved against the effective validation root via the registry |
| `additionalProperties` message | computed from instance vs schema instead of parsing the message string |

**Filed as follow-up issues, since they don't touch this heuristic:**

1. Raw dicts passed to `bind=` are never wrapped into a `Binding` object, which is the only remaining reason for the branch-selection fallback (#4111)
2. Sequence values with two genuine complaints (too short and wrong item type); the "most specific JSON path" rule drops one (#4112)

Two more minor observations I kept as notes rather than issues: branch narrowing currently only fully engages for top-level unions, and the referencing registry is rebuilt per union branch during error narrowing (error path only). Happy to file those too if you'd prefer.

@joelostblom, both points from the last round are addressed on the current head (the overcorrection leak and the mypy failure). CI is green and the branch is rebased on latest main. Would appreciate another pass whenever you have time.

PR #4091, opened 2026-08-01
docs: add number input to date input example

SR0725:
## What I changed

- added an HTML number input to the existing date-input gallery example
- used the date value to filter the x-axis range and the minimum visitor count to filter y-axis values
- kept matching arguments-syntax and methods-syntax versions
- removed the separate number-input example so the related controls stay on one gallery page

This covers the number-input example suggested in #3403 without adding another gallery entry.

## Testing

- uv run --frozen pytest tests/test_examples.py -k interactive_date_input -q -n0 -p no:cacheprovider (6 passed)
- uv run --frozen task ruff-check
- uv run --frozen task doc-build -- --clean --no-autosummary (build succeeded with one warning in doc/gallery/index.rst, which this PR does not modify)


joelostblom:
Thanks for the PR @SR0725 ! Instead of making a new example, I wonder if we can combine this with the recently added example from https://github.com/vega/altair/pull/4088? Could you add the number input there as well and have it filter the y-axis? Or feel free to come up with a new example that can combine both. We have quite a lot of gallery examples, so it would be nice if related functionality like this can be in the same example.

SR0725:
That makes sense. I folded the number input into interactive_date_input: the date parameter filters the x-axis range, and minimum_visitors filters the y values. I also removed the standalone number-input files. The focused example tests pass 6/6, ruff passes, and the local docs build succeeds. Updated in 10f2e20e.

tests/examples_arguments_syntax/interactive_date_input.py
```diff
@@ -50,7 +56,8 @@
         ],
     )
     .transform_filter(alt.datum.date >= alt.expr.timeParse(start_date, "%Y-%m-%d"))
-    .add_params(start_date)
+    .transform_filter(alt.datum.visitors >= minimum_visitors)
+    .add_params(start_date, minimum_visitors)
 )
```
joelostblom:
```suggestion
) + alt.Chart().mark_rule(strokeDash=(5,3)).encode(alt.Y(datum=minimum_visitors, type='quantitative'))
```

tests/examples_methods_syntax/interactive_date_input.py
```diff
@@ -49,7 +55,8 @@
         ],
     )
     .transform_filter(alt.datum.date >= alt.expr.timeParse(start_date, "%Y-%m-%d"))
-    .add_params(start_date)
+    .transform_filter(alt.datum.visitors >= minimum_visitors)
+    .add_params(start_date, minimum_visitors)
 )
```
joelostblom:
```suggestion
) + alt.Chart().mark_rule(strokeDash=(5,3)).encode(alt.Y(datum=minimum_visitors, type='quantitative'))
```

joelostblom (approved):
Looks great! I will add a line for the threshold and will then merge

joelostblom (approved):
Thanks again!

PR #4094, opened 2026-08-08
docs: document lookup key compatibility

### LINKED ISSUE #2323: Type conversion and matching FIPS codes
Hi - I'm not 100% sure if this is an issue with tutorials, how altair handles lookups, or the vega-lite data referenced in tutorials.

I have been working through the Choropleth examples to build a county-level heatmap of data from the United States. One column of my data is the [5-digit FIPS code](https://transition.fcc.gov/oet/info/maps/census/fips/fips.txt), encoded in the column `id`. This is used to map data to shapes drawn using 

```python
  counties = alt.topo_feature(data.us_10m.url, 'counties')
  alt.Chart(counties).mark_geoshape(
    alt.Color(alt.repeat("row"), type = "quantitative")
  ).transform_lookup(
    lookup="id",
    from_ = alt.LookupData(data=my_data, key="id", fields=["county_rate"])
)
```

However, `id` in dataframe `my_data` is encoded as a string rather than as an integer. This means that running this without converting the column to an integer type means that counties with ids that have zero padding in front of them (so Alabama, California, Arkansas) the lookup fails. 

The problem, from my perspective, is that the lookup *doesn't* fail for non-zero padded codes. Figuring out why the state of California wasn't showing up took a fair amount of guesswork, as it's not obvious for a beginner at this as to how to debug the lookup from the url data sources used in the tutorial. 

I'm filling out this issue for three reasons:

1. So that other people encountering this issue might find it. 

2. As a feature request: to suggest that it might be useful to be able to view a sample of what the alt.Lookup is matching on and that it might be useful to have an error or a "strict" type checking flag for performing lookups.

3. As a suggestion about documentation. 

As for reason 2, I realize there may be good reasons not to not have either however I just want to raise the possibility.

SR0725:
The lookup guide did not explain that lookup values need compatible types and formatting. This is especially easy to miss with geographic data: a zero-padded FIPS string such as `"06001"` does not match the numeric county ID `6001`.

This adds a note next to the `LookupData` explanation and shows how to normalize a pandas FIPS column before creating the chart.

Fixes #2323

## Validation

- `uv run task doc-build -- --clean --no-autosummary --no-gallery`

The build succeeded. Its only warning references the unchanged `case_studies/index.rst` file, and the rendered lookup page contains the new note.

## Risk

Documentation only; runtime behavior is unchanged.

### DIFF
--- a/doc/user_guide/transform/lookup.rst
+++ b/doc/user_guide/transform/lookup.rst
@@ -100,10 +100,11 @@ extract.
 .. note::
 
    Make sure the lookup fields use compatible values and data types. For
-   example, a zero-padded FIPS code such as ``"06001"`` will not match the
-   numeric county ID ``6001``. If a pandas column contains zero-padded FIPS
-   strings while the geographic data uses numeric IDs, convert it before
-   creating the chart, for example with ``df["id"] = df["fips"].astype(int)``.
+   example, a zero-padded spatial ID such as ``"06001"`` will not match the
+   numeric spatial ID ``6001``. If your dataframe contains zero-padded spatial 
+   IDs as strings while the geographic data uses numeric IDs, convert the column
+   to a compatible type before creating the chart. For example, in pandas use
+   ``df["id"] = df["spatial_id"].astype(int)``.
 
 Example: Lookup Transforms for Geographical Visualization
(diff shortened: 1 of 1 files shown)


mattijn:
Thanks @SR0725! Appreciated👍

PR #4095, opened 2026-08-10
fix: preserve existing config in `configure()`

### LINKED ISSUE #3841: Bug: chart `configure_axisBottom()` has no effect if `configure()` is called later

suzunn:
## Description

`configure()` replaced `chart.config` wholesale, so any sub-config set by an earlier `configure_*()` call was silently discarded. The two orderings gave different specs:

```python
import pandas as pd, altair as alt

data = pd.DataFrame(dict(foo=[1.1, 2.2, 3.3], bar=[3, 2, 4]))
base = alt.Chart(data).mark_circle().encode(x="foo", y="bar")

base.configure_axisBottom(orient="top").configure(tooltipFormat={"numberFormat": ".2f"})
# config -> {"view": ..., "tooltipFormat": ...}                  <- axisBottom dropped

base.configure(tooltipFormat={"numberFormat": ".2f"}).configure_axisBottom(orient="top")
# config -> {"view": ..., "axisBottom": ..., "tooltipFormat": ...}
```

The `configure_*()` family already merges into an existing config (`copy.config[prop] = ...`); `configure()` was the outlier, and the failure was silent — no error, just a missing property in the output spec.

Fixes #3841

## Changes

- `CONFIG_METHOD` in `tools/generate_schema_wrapper.py`: `configure()` now copies with `deep=["config"]` and merges the set properties of the newly built `Config` into the existing one, rather than overwriting it. Regenerated `altair/vegalite/v6/schema/mixins.py` accordingly.
- Added regression tests in `tests/vegalite/v6/test_api.py` covering both orderings, accumulation and override across repeated `configure()` calls, and that the source chart is not mutated.

## Notes

- Properties are still last-write-wins: `configure(background="red").configure(background="blue")` yields `"blue"`.
- A **positional** argument (`configure({...})`) keeps the old replacement behaviour, since there are no keywords to merge. That path's behaviour is unchanged from `main`.
- `deep=["config"]` matches what every `configure_*()` method already does, so the merge does not mutate the chart it was called on.

## Testing

- `python tools/generate_schema_wrapper.py` — `mixins.py` is the only regenerated file that changes.
- `pytest tests` passes, except `tests/test_transformed_data.py::test_primitive_chart_examples[annual_weather_heatmap.py]` and `test_compound_chart_examples[layered_chart_with_dual_axis.py]`, which fail identically on unmodified `main` on this machine and are unrelated to this change.
- The two behavioural tests added here fail without the `mixins.py` change.
- `ruff check`, `ruff format --check` and `mypy tools` are clean.

### DIFF
--- a/altair/vegalite/v6/schema/mixins.py
+++ b/altair/vegalite/v6/schema/mixins.py
@@ -1162,6 +1162,11 @@ class ConfigMethodMixin:
     @use_signature(core.Config)
     def configure(self, *args, **kwargs) -> Self:
         copy = self.copy(deep=["config"])  # type: ignore[attr-defined]
+        if not args and not kwargs:
+            # A bare ``configure()`` resets the config, preserving the
+            # behaviour it had before merging was introduced.
+            copy.config = core.Config()
+            return copy
         if args or copy.config is Undefined:
             copy.config = core.Config(*args, **kwargs)
         else:
--- a/altair/vegalite/v6/schema/mixins.py
+++ b/altair/vegalite/v6/schema/mixins.py
@@ -1161,8 +1161,15 @@ class ConfigMethodMixin:
 
     @use_signature(core.Config)
     def configure(self, *args, **kwargs) -> Self:
-        copy = self.copy(deep=False)  # type: ignore[attr-defined]
-        copy.config = core.Config(*args, **kwargs)
+        copy = self.copy(deep=["config"])  # type: ignore[attr-defined]
+        if args or copy.config is Undefined:
+            copy.config = core.Config(*args, **kwargs)
+        else:
+            # Merge into the existing config so that properties set by earlier
+            # ``configure_*()`` calls are kept instead of silently discarded.
+            for prop, val in core.Config(**kwargs)._kwds.items():
+                if val is not Undefined:
+                    copy.config[prop] = val
         return copy
 
     @use_signature(core.RectConfig)
--- a/tests/vegalite/v6/test_api.py
+++ b/tests/vegalite/v6/test_api.py
@@ -1552,6 +1552,51 @@ def test_themes():
         assert "config" not in chart.to_dict()
 
 
+def test_configure_preserves_existing_config():
+    # https://github.com/vega/altair/issues/3841
+    # `configure()` used to replace the whole config object, silently dropping
+    # anything an earlier `configure_*()` call had set.
+    base = alt.Chart("data.txt").mark_point().encode(x="x:Q", y="y:Q")
+
+    axis_first = base.configure_axisBottom(orient="top").configure(
+        tooltipFormat={"numberFormat": ".2f"}
+    )
+    axis_last = base.configure(
+        tooltipFormat={"numberFormat": ".2f"}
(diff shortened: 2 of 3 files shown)


joelostblom:
Thanks @suzunn, could you clarify if this is meant to supersede https://github.com/vega/altair/pull/4035/ and confirm that you have seen my comments there already?

suzunn:
Yes, this supersedes #4035 — apologies, I opened it without noticing that PR was already open for the same issue, which was my mistake. Credit to @puneetdixit200 for getting there first; happy to close this if they would rather pick theirs back up.

I have now read your review on #4035. On your first point, this PR merges via `_kwds` rather than `to_dict()`, so `chart.config.axis` stays an `AxisConfig` instead of becoming a `dict`. I did not apply your test suggestions verbatim, but the cases they were aimed at are covered: `test_configure_preserves_existing_config` asserts the merge is order-independent, and `test_configure_merges_repeated_calls` asserts that a repeated property is overridden by the later call.

Your point about an empty `configure()` was still outstanding here — you were right that it had become a breaking change. I have now pushed the guard you proposed, so a bare `configure()` clears the config exactly as it does on `main`, with a regression test covering it.

suzunn:
Yes — apologies for the slow reply. To be direct about the overlap: #4095 came after I read your review on #4035, and it's built on the two points you raised there. The merge goes through `Config(**kwargs)._kwds.items()` rather than `to_dict()`, so `chart.config.axis` stays an `AxisConfig` instead of becoming a plain `dict`. A bare `configure()` still resets the config, using exactly the `copy.config = core.Config(); return copy` shape you suggested. Tests cover both call orderings, repeated `configure()` calls both accumulating and overriding, and that the source chart isn't mutated.

That said, #4035 got there first and @puneetdixit200 should have the chance to finish it. I'm happy to close #4095 and move the diff over as review suggestions on #4035 instead. If that PR has gone quiet and you'd rather land the fix now, I'll keep #4095 open and also fold in the test suggestions you left there. Your call — either works for me.

PR #4098, opened 2026-08-12
chore: Update ty, ruff and mypy type-checker version

alec-bike:
Update type-checker versions to latest (ty, ruff, mypy) in lock file. 
Fix associated issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 **Vega-Altair** is a declarative statistical visualization library for Python. With Vega-Altair, you can spend more time understanding your data and its meaning. Vega-Altair's
 API is simple, friendly and consistent and built on top of the powerful
 [Vega-Lite](https://github.com/vega/vega-lite) JSON specification. This elegant
-simplicity produces beautiful and effective visualizations with a minimal amount of code. 
+simplicity produces beautiful and effective visualizations with a minimal amount of code.
 
 *Vega-Altair was originally developed by [Jake Vanderplas](https://github.com/jakevdp) and [Brian
 Granger](https://github.com/ellisonbg) in close collaboration with the [UW
@@ -44,10 +44,10 @@ alt.Chart(cars).mark_point().encode(
 
 ![Vega-Altair Visualization](https://raw.githubusercontent.com/altair-viz/altair/main/images/cars.png)
 
-One of the unique features of Vega-Altair, inherited from Vega-Lite, is a declarative grammar of not just visualization, but _interaction_. 
+One of the unique features of Vega-Altair, inherited from Vega-Lite, is a declarative grammar of not just visualization, but _interaction_.
 With a few modifications to the example above we can create a linked histogram that is filtered based on a selection of the scatter plot.
 
-```python 
+```python
 import altair as alt
 from altair.datasets import data
 
@@ -55,22 +55,20 @@ source = data.cars()
 
 brush = alt.selection_interval()
 
-points = (
-    alt.Chart(source)
-    .mark_point()
-    .encode(
-        x="Horsepower",
-        y="Miles_per_Gallon",
-        color=alt.when(brush).then("Origin").otherwise(alt.value("lightgray")),
-    )
-    .add_params(brush)
+points = alt.Chart(source).mark_point().encode(
+    x="Horsepower",
+    y="Miles_per_Gallon",
+    color=alt.when(brush).then("Origin").otherwise(alt.value("lightgray")),
+).add_params(
+    brush
 )
 
-bars = (
-    alt.Chart(source)
-    .mark_bar()
-    .encode(y="Origin", color="Origin", x="count(Origin)")
-    .transform_filter(brush)
+bars = alt.Chart(source).mark_bar().encode(
+    x="count(Origin)",
+    y="Origin",
+    color="Origin",
+).transform_filter(
+    brush
 )
 
 points & bars
@@ -80,24 +78,26 @@ points & bars
 
 ## Features
 
-* Carefully-designed, declarative Python API.
-* Auto-generated internal Python API that guarantees visualizations are type-checked and
+- Carefully-designed, declarative Python API.
+- Auto-generated internal Python API that guarantees visualizations are type-checked and
   in full conformance with the [Vega-Lite](https://github.com/vega/vega-lite)
   specification.
-* Display visualizations in JupyterLab, Jupyter Notebook, Visual Studio Code, on GitHub and
+- Display visualizations in JupyterLab, Jupyter Notebook, Visual Studio Code, on GitHub and
   [nbviewer](https://nbviewer.jupyter.org/), and many more.
-* Export visualizations to various formats such as PNG/SVG images, stand-alone HTML pages and the
-[Online Vega-Lite Editor](https://vega.github.io/editor/#/).
-* Serialize visualizations as JSON files.
+- Export visualizations to various formats such as PNG/SVG images, stand-alone HTML pages and the
+  [Online Vega-Lite Editor](https://vega.github.io/editor/#/).
+- Serialize visualizations as JSON files.
 
 ## Installation
 
 Vega-Altair can be installed with:
+
 ```bash
 pip install altair
 ```
 
 If you are using the conda package manager, the equivalent is:
+
 ```bash
 conda install altair -c conda-forge
 ```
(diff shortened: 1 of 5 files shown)


alec-bike:
@mattijn @joelostblom this PR is passing CI but failing `uv run task test-all` for python 3.10 locally (it also fails on `main`). Any ideas what may be causing this? 

```sh
=================================================================== short test summary info ===================================================================
FAILED tests/test_examples.py::test_render_examples_to_chart
  [-examples_methods_syntax.deviation_ellipses.py] - 
  ImportError: dlopen(/Users/alec/Library/Application Support/hatch/env/virtual/altair/xiLrh2a5/hatch-test.py3.10/lib/python3.10/site-packages/scipy/sparse/...

FAILED tests/test_examples.py::test_render_examples_to_chart
  [-examples_arguments_syntax.deviation_ellipses.py] - 
  ImportError: dlopen(/Users/alec/Library/Application Support/hatch/env/virtual/altair/xiLrh2a5/hatch-test.py3.10/lib/python3.10/site-packages/scipy/sparse/...

FAILED tests/test_examples.py::test_from_and_to_json_roundtrip
  [-examples_methods_syntax.deviation_ellipses.py] - 
  ImportError: dlopen(/Users/alec/Library/Application Support/hatch/env/virtual/altair/xiLrh2a5/hatch-test.py3.10/lib/python3.10/site-packages/scipy/sparse/...

FAILED tests/test_examples.py::test_from_and_to_json_roundtrip
  [-examples_arguments_syntax.deviation_ellipses.py] - 
  ImportError: dlopen(/Users/alec/Library/Application Support/hatch/env/virtual/altair/xiLrh2a5/hatch-test.py3.10/lib/python3.10/site-packages/scipy/sparse/...

FAILED tests/test_examples.py::test_render_examples_to_png
  [-examples_arguments_syntax.deviation_ellipses.py] - 
  ImportError: dlopen(/Users/alec/Library/Application Support/hatch/env/virtual/altair/xiLrh2a5/hatch-test.py3.10/lib/python3.10/site-packages/scipy/sparse/...

FAILED tests/test_examples.py::test_render_examples_to_png
  [-examples_methods_syntax.deviation_ellipses.py] - 
  ImportError: dlopen(/Users/alec/Library/Application Support/hatch/env/virtual/altair/xiLrh2a5/hatch-test.py3.10/lib/python3.10/site-packages/scipy/sparse/...
======================================== 6 failed, 1892 passed, 6 skipped, 6 xfailed, 4 xpassed, 80 warnings in 32.89s ========================================
```

README.md
```diff
@@ -54,20 +55,22 @@ source = data.cars()
...
-    brush
+points = (
+    alt.Chart(source)
+    .mark_point()
+    .encode(
+        x="Horsepower",
+        y="Miles_per_Gallon",
+        color=alt.when(brush).then("Origin").otherwise(alt.value("lightgray")),
+    )
+    .add_params(brush)
```
joelostblom: Was there any specific reason to change the multi-line syntax in the readme? Was that done automatically by ruff after updating it? I remember that we were discussing at some point which of these two should be our canonical multiline syntax; if we want to change it, maybe we should change it everywhere at once? Othwerwise it might be easier to keep what we have?

altair/vegalite/v6/theme.py
```diff
@@ -103,7 +103,7 @@ def __repr__(self) -> str:
 
 themes.register(
```
joelostblom: If we rewrite these two as typed dictionaries, we might not need the ignore comments?


```py
themes.register(
    "default",
    lambda: ThemeConfig(
        config={"view": {"continuousWidth": 300, "continuousHeight": 300}}
    ),
)
themes.register(
    "opaque",
    lambda: ThemeConfig(
        config={
            "background": "white",
            "view": {"continuousWidth": 300, "continuousHeight": 300},
        }
    ),
)
```

joelostblom (commented):
Thanks, the PR looks good in general, just a couple of minor comments.

Regarding your issue, I can't reproduce on Linux, so all I can offer are some suggestions from codex:

> The traceback fails while macOS loads a compiled SciPy extension via dlopen, before the Altair example executes. The most likely cause is a stale or damaged Hatch environment, or a macOS architecture mismatch between Python and the installed SciPy wheel.
> 
> The first recommendation should be to recreate the Hatch environment:
> 
> ```
> uv run hatch env remove hatch-test.py3.10
> uv run hatch test -py 3.10
> ```
> 
> If that still fails, ask for the complete ImportError; the omitted final lines contain the important detail, such as a missing dynamic library or incompatible architecture. These diagnostics would also help:
> 
> ```
> uv run hatch run hatch-test.py3.10:python -c \
> 'import platform, sys; print(sys.executable); print(platform.machine()); print(platform.platform())'
> ```
> ```
> uv run hatch run hatch-test.py3.10:python -c \
> 'import numpy, scipy; print(numpy.__version__, numpy.__file__); print(scipy.__version__, scipy.__file__); import scipy.sparse._sparsetools'
> ```
> 
> On Apple Silicon, platform.machine() should normally be arm64. If it reports x86_64, Python is running under Rosetta and all compiled dependencies need matching x86_64 wheels. Mixing an arm64 Python or NumPy installation with x86_64 SciPy, or the reverse, causes this kind of dlopen failure.

Also note that I have this PR open about dropping py10 support https://github.com/vega/altair/pull/4029

README.md
alec-bike: Since [version 0.16.0](https://github.com/astral-sh/ruff/releases#release-0.16.0) ruff checks markdown files by default, and ruff flagged the two code blocks in `README.md`. These are the changes that `ruff format` proposed. 

It's possible to disable this in `pyproject.toml`, though it doesn't seem to have flagged any other markdown files.

altair/vegalite/v6/theme.py
alec-bike: Thanks for the suggestions, that works.

Interestingly, pyright warns that `themes.register` is deprecated:

```py
The method "register" in class "ThemeRegistry" is deprecated
  Deprecated since `altair=5.5.0`. Use @altair.theme.register instead.
```

alec-bike:
I tried to remove and re-create the 3.10 hatch environment but the test still fails. All other python versions are ok. If it works on your Linux this probably explains why CI is passing. 

Running `uv run task test-all` used to pass (as of June) so this is recent regression for my system. I'll keep investigating.

Edit: may be a `.so` mismatch between scipy 1.15.3 and macOS 27.0 Beta (which I installed recently). For python 3.11 the scipy version is more recent 1.17.1 which does work.

altair/vegalite/v6/theme.py
joelostblom: Right, `ThemeRegistry.register` is deprecated for external callers, but the module still uses it internally to initialize built-in themes. I think we could avoid that warning by calling the non-deprecated base implementation:

```py
super(ThemeRegistry, themes).register(
    "default",
    lambda: ThemeConfig(...),
)
```

The syntax is a bit less clean though, but I don't know of a better solution. What do you prefer?

joelostblom:
> may be a .so mismatch between scipy 1.15.3 and macOS 27.0 Beta (which I installed recently). For python 3.11 the scipy version is more recent 1.17.1 which does work.

That sounds like a plausible cause. If you feel comfortable reviewing #4029, we could deprecate py3.10 support soon instead of you having to work around it locally.

README.md
alec-bike: I propose we disable ruff formatting for `README.md` but adopt some of the `ruff format` recommendations. This is easier to read, in my opinion.

```py
points = (
        x="Horsepower",
        y="Miles_per_Gallon",
        color=alt.when(brush).then("Origin").otherwise(alt.value("lightgray")),
    )
    .add_params(brush)
)

bars = (
        x="count(Origin)",
        y="Origin", 
        color="Origin",
    )
    .transform_filter(brush)
)
```

altair/vegalite/v6/theme.py
alec-bike: That works. I propose the following to make it easier to read:

```py
themes = ThemeRegistry(entry_point_group=ENTRY_POINT_GROUP)
register = super(ThemeRegistry, themes).register

register(
    "default",
    lambda: ThemeConfig(
        config={
            "view": {"continuousWidth": 300, "continuousHeight": 300},
        },
    ),
)
```

alec-bike:
Yes, I'm all for deprecating 3.10. Your PR #4029 looks good to me.

I think this PR could still be merged despite my issue with `uv run task test-min` as the PR runs ok in CI.

altair/vegalite/v6/theme.py
joelostblom: Good suggestion, I agree that makes it easier to read!

README.md
joelostblom: I agree with disabling the auto formatting for the time being and turning it on later if we decide to change the multi-line syntax. Long term, I think the syntax ruff suggests if probably the clearest, but for now my vote would be to keep the readme as is.

PR #4107, opened 2026-08-26
docs: marimo previews

koaning:
## Thanks for contributing to Altair! 🎉

Please follow these guidelines:

### 1. **PR Description**
   - This PR contains a small change to the Altair gallery to allow users to play around with the code in a WASM marimo notebook. 

### 2. **Tests & Docs**
   - In order to make sure that no annoying install pop-ups appear, I also had Claude update all the examples to add the PEP metadata. This ensures that marimo installs the packages needed without requiring any extra user input.

### DIFF
--- a/sphinxext/altairgallery.py
+++ b/sphinxext/altairgallery.py
@@ -200,6 +200,16 @@ EXAMPLE_TEMPLATE = jinja2.Template(
         .. code:: python
 
 {{ code | indent(12) }}
+
+    .. tab-item:: Interactive (marimo)
+
+        .. raw:: html
+
+            <iframe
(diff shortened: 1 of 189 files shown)


joelostblom (changes requested):
Thank you @koaning! I think the addition of the metadata to the gallery files will work well. It's also great to see that marimo is running so well with the examples and that the relatively little code needed to be added to the docs to enable this functionality.

In terms of the styling on the page, I think it is a bit confusing that both the chart and the code is appearing inside the tab, because now there are two charts on the page at on the page and some scrolling is required inside the tab to see everything.

Do you think it is possible to instead add a button that replaces the entire section below the title with the live marimo notebook? Similar to this demo (ideally the chart would also appear above the code block in the marimo notebook):

https://github.com/user-attachments/assets/f4576e5e-6e50-41ea-bfa2-964ad378b8f7

PR #4114, opened 2026-09-10
chore: Expand ty type-checking

alec-bike:
Expand the folders `ty` is able to type-check and exclude fewer folders within `altair`.

This is achieved by:
- replacing `mypy`-specific annotations (eg `type: ignore [return]`) with generic `type: ignore`;
- using `mypy`-specific or `ty`-specific (`ty: ignore`) annotations in some cases; and
- fixing some minor typing issues that surfaced in recent versions of `ty`.

Exclude `tests` for now (which is checked by `mypy` and `pyright`) due to 800+ errors with `ty`.


tools/vega_expr.py
```diff
@@ -384,7 +384,7 @@ def __call__(self, s: str, count: int = 0, /, refresh: bool = False) -> str:
 
     def _compile(self) -> Pattern[str]:
         if not self._mapping:
-            name = self._mapping.__qualname__  # type: ignore[attr-defined]
+            name = self._mapping.__qualname__  # type: ignore
```
joelostblom: Is this an actual bug in the code that we could fix and get rid of the ignore altogether?

Take the following example:

```py
from tools.vega_expr import ReplaceMany

replacer = ReplaceMany()
replacer("hello")
```

It currently raises `AttributeError: 'dict' object has no attribute '__qualname__'`.

I wonder if the fix is to instead use `name = type(self._mapping).__qualname__`, which would then return the qualified name of the dictionary class (rather than the instance), ie "dict"

joelostblom (commented):
Thanks for keeping improving the `ty` support @alec-bike ! I think this looks good in general, although one caution I have is that if we broaden ignores to freely we might end up missing something that we would like the type checkers to fix. I don't immediately see any case of that here, but I didn't notice one place where I wonder if there is a bug in the code that should not have been ignored previously (see inline comment).

tools/vega_expr.py
alec-bike: Thanks Joel, good catch! I implemented your fix and it passed CI.

alec-bike:
Thanks @joelostblom, good point. For new type-check issues raised by `ty` I have generally tried to find a solution rather than relying on blanket `ty: ignore` annotations.

Where an existing `mypy`-specific annotation already existed and is also flagged by `ty`, I have typically only reverted to `type: ignore`. 

I will go through some of these recent changes to see if the original `mypy` ignore can be fixed.

Note that currently I'm blocked by #4134 which causes `uv run task generate-schema-wrapper` to fail (though downgrading to `mistune 3.2.1` is a temporary workaround). _Interestingly, CI does not catch this issue._

alec-bike:
Another issue (warning), raised by both `ty` and `pyright`. The following is deprecated and should be updated:

```sh
❯ uv run task type-check
Success: no issues found in 404 source files
warning[[deprecated](https://ty.dev/rules#deprecated)]: 
   The class `RefResolver` is deprecated
   --> tools/schemapi/schemapi.py:194:24
    |
194 |             jsonschema.RefResolver.from_schema(rootschema)
    |                        ^^^^^^^^^^^ jsonschema.RefResolver is deprecated as of
    v4.18.0, in favor of the https://github.com/python-jsonschema/referencing library, 
    which provides more compliant referencing behavior as well as more flexible APIs
    for customization. A future release will remove RefResolver. Please file a feature
    request (on referencing) if you are missing an API for the kind of customization 
    you need.
```

Also line 572.

pyproject.toml
```diff
@@ -371,17 +372,17 @@ ignore = [
     "./altair/vegalite/v6/schema/mixins.py",    # 1001 warns
     "./altair/jupyter/",                        # Mostly untyped
     "./tests/test_jupyter_chart.py",            # Based on untyped module
-    "../../../**/Lib",                          # stdlib
-    "../../../**/typeshed*"                     # typeshed-fallback
```
joelostblom, line 375: Could you say more about why these two lines were removed? They seem to be unrelated to the `ty` changes?

joelostblom (commented):
Thanks for the update @alec-bike. I think this looks good except one question in my inline comment.

Also, do you want to deal with the deprecation you mentioned in this PR or create a new issue and follow up on it later?

pyproject.toml
alec-bike: These point to a location outside the altair tree, on my macOS it's the `/Users` folder. Are these out-of-tree references a common requirement for pyright?

I use an out-of-tree reference for local debug of `altair` in my personal repo:

```toml
[tool.uv.sources]
altair = { path = "../altair" }
```

but I've not seen out-of-tree references in `pyproject.toml` otherwise.

alec-bike:
> Also, do you want to deal with the deprecation you mentioned in this PR or create a new issue and follow up on it later?

I created issue #4140 to track the deprecation warnings. Neither cause local testing or CI to fail with the current `pyproject.toml` configuration, so they don't interfere with this PR.

pyproject.toml
joelostblom: Makes sense!

joelostblom (approved):
I believe this is ready to merge in 👍

PR #4135, opened 2026-09-12
docs: add color input gallery example

LouisDeconinck:
## What I changed

- added a gallery example that binds HTML `color` inputs to parameters and uses them as the range of a color scale, so readers can pick the color of each series interactively
- added both the arguments-syntax and methods-syntax versions
- used a small inline dataset so the example does not depend on an external URL

The same color-picker pattern is shown in the bindings user guide; this makes it discoverable from the gallery.

Related to #3403.

## Testing

- `uv run pytest tests/test_examples.py -k interactive_color_input -q -n0 -p no:cacheprovider` (6 passed)
- `uv run task ruff-check`
- confirmed both versions generate the same Vega-Lite spec


joelostblom:
Thanks for the example @LouisDeconinck, I believe this is already covered by the second example in this section of the docs https://altair-viz.github.io/user_guide/interactions/bindings_widgets.html#logic-driven-comparisons; is there anything you think is missing there which motivates adding the gallery example?

LouisDeconinck:
You are right — I checked that section and the color-picker pattern is indeed already covered there; a gallery version does not add enough to justify the duplication.

What that section does not cover (and what #3403 explicitly suggested) is the `date` input — e.g. filtering a time series with a native date picker, which sliders handle awkwardly. If you think a gallery example for that would be useful, I am happy to rework this PR around `input='date'` instead. Otherwise I will close it.

joelostblom:
Thanks @LouisDeconinck; we actually added a date input example recently in https://github.com/vega/altair/pull/4088

LouisDeconinck:
Thanks for checking @joelostblom — agreed, #4088 covers this ground. Closing.

PR #4136, opened 2026-09-12
ci: Fix generation script and ci check

### LINKED ISSUE #4134: `generate-schema-wrapper` task fails after mistune update
### What happened?

After the recent dependabot updates, `generate-schema-wrapper` task fails:

```sh
❯ uv run task generate-schema-wrapper
tools/markup.py:151: error: Signature of "process_text" incompatible with supertype
"mistune.inline_parser.InlineParser"  [override]
        def process_text(self, text: str, state: InlineState) -> None:
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
tools/markup.py:151: note:      Superclass:
tools/markup.py:151: note:          def process_text(self, text: str, state: InlineState,
                                      parse_emphasis: bool = ...) -> None
tools/markup.py:151: note:      Subclass:
tools/markup.py:151: note:          def process_text(self, text: str, state: InlineState) -> None
Found 1 error in 1 file (checked 21 source files)
```

6.2.2.post27+gb37a571b

joelostblom:
This PR adds support for the new signature of Mistune 3.3 in markup parser and also adds the tools to the mypy check that runs CI so we don't miss changes like this again.

close https://github.com/vega/altair/issues/4134

@alec-bike Thanks for finding and diagnosing the issue; do you also have time to review this one?


tools/markup.py
```diff
@@ -148,14 +148,18 @@ class InlineParser(_InlineParser):
...
+    ) -> None:
         """
         Removes `liquid`_ templating markup.
 
         .. _liquid:
             https://shopify.github.io/liquid/
         """
-        state.append_token({"type": "text", "raw": _RE_LIQUID_INCLUDE.sub(r"", text)})
+        super().process_text(
+            _RE_LIQUID_INCLUDE.sub(r"", text), state, parse_emphasis=parse_emphasis
```
joelostblom, line 161: My understanding is that Mistune 3.3 added a `parse_emphasis` parameter and additional text-token handling to `InlineParser.process_text`. So rather than updating our override while continuing to call append_token directly, we can now remove the Liquid markup and delegate the remaining work to Mistune’s parent implementation.

Calling `append_token` directly would then duplicate Mistune’s old behavior and bypass its new handling for emphasis, adjacent text-token merging, and word/entity boundaries. Using `super().process_text(...)` preserves Altair’s Liquid-markup removal while allowing Mistune to manage its own tokenization behavior. This also reduces the risk of our override becoming outdated again if Mistune changes its internal processing.

alec-bike: This works for me running the `generate-schema-wrapper` task locally, and passes CI. Your proposed change matches what Copilot suggested when I prompted, but I'm not familiar enough with these tools to judge whether this could have unintended side-effects!

alec-bike (approved):

alec-bike:
@joelostblom I see that with this commit, there is a changed `metadata.parquet` in `altair/datasets/_metadata` after running task `generate-schema-wrapper`.

Should this `metadata.parquet` be updated in the repo?

Issue #4112, opened 2026-09-03
Report both length and item-type errors for sequence values


dk3yyyy:
Follow-up from the review of #4090 (fix for #3752).

## Problem

```python
import altair as alt

alt.selection_interval(
    resolve="global",
    value={"x": ["Europe"], "y": [10, 20]},
)
```

`value.x` has two genuine complaints: the list is too short (`minItems`), and its only item has the wrong type. Altair's error pipeline keeps only the most specific JSON path, so one complaint is always hidden. #4090 makes the length error win over the nested item error:

```text
'['Europe']' is an invalid value for `x`. Valid values are of type `Sequence` with at least 2 items.
```

That is more useful than before, but the item-type complaint is still dropped.

## Context

`_subset_to_most_specific_json_paths` documents the assumption that "more specific json paths give more helpful error messages to the user". Sequence-length errors are the counterexample: `$.x` (too short) is less specific than `$.x[0]` (wrong item type), yet often more actionable.

## Suggested direction

For sequence values that violate both length and item constraints, surface both complaints — e.g. report that the sequence needs at least 2 items *and* that `'Europe'` is not a valid item type. Consider whether the most-specific-path rule should be length-aware generally instead of special-casing `minItems`/`maxItems`.

webdevsamran:
Hi team, I would like to work on improving sequence validation error messages.

When sequence values are validated in Altair schema objects, validation stops at the first mismatch rather than reporting both sequence length and item type constraints. I can update schema validation in `altair/vegalite/v5/schema/core.py` to collect and report both constraint violations, with unit tests in `altair/vegalite/v5/tests/`.

Could you please assign this to me? Thanks!

PR #4137, opened 2026-09-13
fix: use the constructed channel class for validation error messages

### LINKED ISSUE #3752: `SchemaValidationError` relevance heuristic is flawed

mattijn:
Closes #3752.

A channel can be three things: a field, a datum, or a value. Altair checks your chart against all three. If you make more than one mistake, it sometimes picks the wrong one to complain about. Then it names the parameter that was fine.

```python
import altair as alt

alt.Chart().mark_point().encode(y=alt.value(1, bin=True, aggregate="sum")).to_dict()
```

**Before**
```
`YValue` has no parameter named 'value'

Existing parameter names are:
value   

See the help for `YValue` to read the full description of these parameters
```

But `value` is fine. `bin` and `aggregate` are the mistakes.

**After**
```
`YValue` has no parameters named 'aggregate', 'bin'

Existing parameter names are:
value   

See the help for `YValue` to read the full description of these parameters
```

### The fix

Altair already made a `YValue` object for this input. So it knows which of the three it is. No need to guess. The message now uses that object. Sometimes there is no such object, for example when you check a plain dict. In that case Altair still guesses, just like before this PR.

Many thanks to @dk3yyyy for the work in #4090, and for keeping the discussion going there. That is what led to this smaller fix. #4090 can be closed once this lands.


mattijn:
Hmm. Found a regression:

1. `alt.Chart().encode(y=alt.value(1, bin=True, aggregate="sum"))`
    - main: `YValue` has no parameter named 'value'
    - this PR: `YValue` has no parameters named 'aggregate', 'bin'
2. `alt.Chart().encode(alt.YValue(1, bin=True, aggregate="sum"))`
    - main: `YValue` has no parameter named 'value'
    - this PR: `YValue` has no parameters named 'aggregate', 'bin'
3. `alt.Chart().encode(alt.Y(value=1, bin=True, aggregate="sum"))`
    - main: `YValue` has no parameter named 'value'
    - this PR: `Y` has no parameter named 'value'
4. `alt.Chart().encode(alt.Y(datum=1, bin=True))`
    - main: `YDatum` has no parameter named 'bin'
    - this PR: `Y` has no parameter named 'datum'

Case 1 and 2 are better with this PR. Case 3 is equally bad, but for case 4 main is better.

mattijn:
After commit: https://github.com/vega/altair/pull/4137/commits/95c1e6649efb2b04e46467a6c5187ea78c00017a this PR now gives:
- case 3: `YValue` has no parameters named 'aggregate', 'bin'
- case 4: `YDatum` has no parameter named 'bin'

Side-fix: channels with no support for values (latitude, longitude, column, detail, facet, key, row, time) get a better message if it is used:
- case 5: `alt.Chart().encode(latitude=alt.value(1))`
  - this PR: `Latitude` has no parameter named 'value'
  - main was: `KeyError: 'value'`

altair/utils/schemapi.py
```diff
@@ -708,16 +737,24 @@ def _get_additional_properties_error_message(
...
+        live_cls, live_schema = _live_class_schema(self.obj, error)
+        altair_cls = live_cls or self._get_altair_class_for_error(error)
         param_dict_keys = inspect.signature(altair_cls).parameters.keys()
         param_names_table = self._format_params_as_table(param_dict_keys)
 
-        # Error messages for these errors look like this:
-        # "Additional properties are not allowed ('unknown' was unexpected)"
-        # Line below extracts "unknown" from this string
-        parameter_name = error.message.split("('")[-1].split("'")[0]
+        schema = live_schema if live_schema is not None else error.schema
```
joelostblom: Since this uses `error.schema` whenever `_live_class_schema()` rejects a union schema. That schema may represent an arbitrary failing union branch, so comparing all instance keys against its properties produces false positives. Take this as an example:


```py 
alt.Chart(
    data=alt.Data({"values": [], "bogus": 1})
).mark_point().to_dict()
```

The `main` branch correctly outputs this error:

```
SchemaValidationError: `Data` has no parameter named 'bogus'
```

But with this PR, the error message incorrectly also flags `values` as an invalid parameter:

```
SchemaValidationError: `Data` has no parameters named 'bogus', 'values'
```

My agent suggests to consider retaining the existing message parsing when `live_schema` is unavailable, or selecting the matching union branch before computing the key difference.

The raw-dictionary fallback also still exhibits the original problem (it incorrectly reports `value`, rather than `aggregate` and `bin`):

```py
alt.Chart(
    mark="point",
    encoding={"y": {"value": 1, "bin": True, "aggregate": "sum"}},
).to_dict()
```

Both branches also seem to fail to mention the actual valid parameters since they just output: `Existing parameter names are: args`.

joelostblom (changes requested):
This is great! I think improvements in the error messages makes Altair much more friendly to new users =) 

In addition to my inline comment, I noticed that the example in https://github.com/vega/altair/issues/3752#issuecomment-3275852891 still shows the incorrect error message. Is that within scope for this PR or a separate issue?

mattijn:
At the moment out of scope. I think it is covered by https://github.com/vega/altair/issues/4112

altair/utils/schemapi.py
mattijn: Thanks for looking! Great review and good observations.

Regarding first example, `Data` can be six things: a url, inline values, a name, a sequence, a sphere, or a graticule. Altair checked only against on of these (here, the graticule one). That one has no `values` and no `bogus`. So it named both. But only `bogus` is wrong.

Now with latests two commits (7b2b48a & 32898ca) Altair collects the names from all valid options of that object to test against. `values` is in that set, so it drops out. `bogus` is not, so it stays. It also reads those names from the same class it names in the message. So the message and the table can no longer disagree.

```python
alt.Chart(
    data=alt.Data({"values": [], "bogus": 1})
).mark_point().to_dict()
```

> ```cmd
> SchemaValidationError: `Data` has no parameter named 'bogus'
> 
> Existing parameter names are:
> format      sequence   url      
> graticule   sphere     values   
> name                            
> 
> See the help for `Data` to read the full description of these parameters
> ```

That fixes your last point too. The table used to say `args`. A class like `Data` has no real signature, just `(*args, **kwds)`, so there was nothing to show. Now it shows the names from the schema

The plain dict is fixed as well. Similar to the previous one

```python
alt.Chart(
    mark="point",
    encoding={"y": {"value": 1, "bin": True, "aggregate": "sum"}},
).to_dict()
```

> ```cmd
> SchemaValidationError: `YValue` has no parameters named 'aggregate', 'bin'
> 
> Existing parameter names are:
> value   
> 
> See the help for `YValue` to read the full description of these parameters
> ```

Regarding #4112 (distilled from your comment in #3752) that is out of scope, see also second half of this comment where I discussed why I think it is out of scope: https://github.com/vega/altair/pull/4090#issuecomment-5308916769. TLDR, it is about which error to pick of available options, not about not finding the error.

altair/utils/schemapi.py
```diff
@@ -620,6 +620,37 @@ def _maybe_channel(tp: type[Any], spec: Any, /) -> type[Any]:
     return next(_iter_channels(tp, spec), tp) if _is_channel(spec) else tp
 
 
+def _known_properties(schema: Any, rootschema: Any, /) -> set[str]:
```
joelostblom, line 623: If I call data with two parameters that are valid separately but can't be combined, I get a confusing error message:

```py
alt.Data({"values": [], "graticule": True}).to_dict()
```
```
SchemaValidationError: `Data` has no parameter named 'graticule'

Existing parameter names are:
format      sequence   url      
graticule   sphere     values   
name                            

See the help for `Data` to read the full description of these parameters
```

It's confusing in the sense that `graticule` is both said to not exist and listed as an existing parameter. I wonder if it is possible for the error message to output that the issue is the combination of parameters in this case?

joelostblom: I think this example might also stem from the same underlying issue, but adding it as another example to test:

```py
alt.Chart().mark_point().encode(y=alt.Color('test:N').legend()).to_dict()
```
```
SchemaValidationError: `Color` has no parameter named 'legend'

Existing parameter names are:
```

The error message on `main` is more helpful:

```
`Y` has no parameter named 'legend'`
```

altair/utils/schemapi.py
```diff
@@ -708,16 +739,33 @@ def _get_additional_properties_error_message(
...
-        # "Additional properties are not allowed ('unknown' was unexpected)"
-        # Line below extracts "unknown" from this string
-        parameter_name = error.message.split("('")[-1].split("'")[0]
+        live_cls, known = _live_class_schema(self.obj, error)
+        altair_cls = live_cls or self._get_altair_class_for_error(error)
+        if known is None:
+            known = _known_properties(
+                getattr(altair_cls, "_schema", None),
+                getattr(altair_cls, "_rootschema", None),
+            )
```
joelostblom, line 748: This is minor and we might not need to fix it since it is not a very common use case, but I wanted to mention it in case you see it as something worth dealing with because it could lead to more robustness in general and not just for this rare case.

Since the fallback gets a class from the property name and uses that class’s schema, the property name does not necessarily identify the contextual schema for nested raw dictionaries:

```py
alt.Chart(
    mark="point",
    config={
        "legend": {
            "gradientDirection": "horizontal",
            "bogus": 1,
        }
    },
).to_dict()
```

```
`Legend` has no parameters named 'bogus', 'gradientDirection'
```

gradientDirection is valid for LegendConfig. `main` correctly reports only `bogus`. Maybe a fix could be if the fallback could retain contextual information from error.schema, or infer the generated class from that schema rather than solely from the path name?

The problem occurs specifically with the raw dictionary: `config={"legend": {...}}`, where the fallback sees the path component legend and guesses `alt.Legend`, rather than the contextually correct `alt.LegendConfig`.

The much more common use case with `configure_legend` works correctly:

```py
alt.Chart(mark="point").configure_legend(
    gradientDirection="horizontal",
    bogus=1,
).to_dict()
```

```
`LegendConfig` has no parameter named 'bogus'
```

Other affected LegendConfig-only properties include:

```
disable
gradientHorizontalMaxLength
gradientVerticalMaxLength
layout
strokeDash
symbolDirection
unselectedOpacity
```

joelostblom (changes requested):
Great updates! Just a couple of more things in the inline comments. Maybe we could also add some of the recent examples from our comments as regression tests?

altair/utils/schemapi.py
mattijn: Both fixed, with tests.

**On the combination.** Yes, that is possible, and it is what it does now. `values` and `graticule` are both fine as `Data` parameters, just not together. One belongs to inline data, the other to a graticule. "Has no parameter named X" is the wrong sentence there, whichever name you pick.

So now, if every name exists somewhere in the class and the value is still invalid, the names cannot be combined. Altair says so, and points at the form each name belongs to. Only a union ends up here, and a union has no single set of parameters to list.

```python
alt.Data({"values": [], "graticule": True}).to_dict()
```

> **on `main`**
> ```cmd
> SchemaValidationError: `Data` has no parameter named 'graticule'
> 
> Existing parameter names are:
> args   
> ```
> **now**
> ```cmd
> SchemaValidationError: `Data` does not accept 'graticule' and 'values' together
> 
> 'graticule' belongs to `GraticuleGenerator`
> 'values' belongs to `InlineData`
> 
> See the help for `Data` to read the full description of these parameters
> ```

A name that exists in no form at all still gets the normal message, since there the sentence is true:

```python
alt.Data({"values": [], "bogus": 1}).to_dict()
```
```cmd
SchemaValidationError: `Data` has no parameter named 'bogus'

Existing parameter names are:
format      sequence   url      
graticule   sphere     values   
name                            
```

**On your second example.** Same root cause, you were right.

```python
alt.Chart().mark_point().encode(y=alt.Color("test:N").legend()).to_dict()
```
```cmd
SchemaValidationError: `Y` has no parameter named 'legend'
```
Same as `main`.

`Color` was passed to `y`. Altair builds that object, so the message trusted it and named `Color`. But `Color` does have a `legend` parameter, and there it was in the table below. Now, if the built object knows every name in the value, it is not what the error is about, so Altair uses the schema that did produce the error.

altair/utils/schemapi.py
mattijn: Fixed, with a test.

```python
alt.Chart(
    mark="point",
    config={"legend": {"gradientDirection": "horizontal", "bogus": 1}},
).to_dict()
```
```cmd
SchemaValidationError: `Legend` has no parameter named 'bogus'
```
Same as `main`.

You had it right. There is no built object here, so Altair guessed the class from the path name: `legend` becomes `alt.Legend`. But in this spot it is a `LegendConfig`, and that one does have `gradientDirection`.

The class only counts as proof when the value itself says which variant it is, through a `datum` or `value` key. Without that it is just a guess from the path name, and the schema that produced the error is the better source. So that is what it uses now.

altair/utils/schemapi.py
```diff
@@ -708,21 +750,74 @@ def _get_additional_properties_error_message(
...
+        )
+        unexpected = sorted(n for n in instance if n not in known) if known else []
+
+        if not unexpected:
+            # There is no such object, or the one Altair constructed knows every
+            # name here and so is not what this error is about.
+            altair_cls = self._get_altair_class_for_error(error)
+            known = _known_properties(altair_cls._schema, altair_cls._rootschema)
+            if known and all(n in known for n in instance):
+                return self._combined_names_message(altair_cls, instance)
```
joelostblom, line 768: I wonder if we could change this to only report incompatible combinations when the contextual schema establishes a union conflict and find every property in the guessed class triggers `_combined_names_message()` before consulting the contextual `error.schema`.

```python
alt.Chart(
    mark="point",
    config={"legend": {"direction": "horizontal", "format": ".2f"}},
).to_dict()
```

The PR reports:

```text
`Legend` does not accept 'direction' and 'format' together
```

However, `Legend` accepts those properties together. The actual issue is that `format` is not accepted in **`LegendConfig`**. On `main`, the error identifies `format` as the invalid parameter.

Could we consult the contextual schema before deciding that this is a combination error? The combination message should only be used when the relevant schema is a union and the properties genuinely span incompatible branches. A path-derived class knowing all the names is not sufficient evidence.

Please also add the example above as a regression test.

altair/utils/schemapi.py
```diff
@@ -708,21 +750,74 @@ def _get_additional_properties_error_message(
...
-        message = f"""\
-`{altair_cls.__name__}` has no parameter named '{parameter_name}'
+        instance = error.instance if isinstance(error.instance, dict) else {}
+        altair_cls = _live_class(self.obj, error)
+        known = (
+            _known_properties(altair_cls._schema, altair_cls._rootschema)
+            if altair_cls
+            else set()
+        )
+        unexpected = sorted(n for n in instance if n not in known) if known else []
```
joelostblom, line 760: I wonder if we could check that the live channel class matches its encoding location before trusting it. Now the constructed class is still trusted whenever it identifies any unexpected property. This means the mismatched-channel fix fails with a small variation on the previous example:

```python
    y=alt.Color("test:N", axis=None).legend()
).to_dict()
```

The PR reports:

```text
`Color` has no parameter named 'axis'
```

On `main`, it reports:

```text
`Y` has no parameter named 'legend'
```

`axis` is valid in the actual `y` context; `legend` is the problem. Removing `.legend()` makes this chart serialize successfully, confirming that the updated message points at the wrong parameter.

Could we check that the live channel class matches its encoding location before using its schema? For a `Color` object placed under `y`, the contextual `Y` schema should determine which parameters are invalid, even when the live object also contains parameters that `Color` does not recognize.

Please add this variation as a regression test too.

joelostblom (changes requested):
Thanks for updating this! (a)I found a couple of more cases where the error message could be improved.

PR #4143, opened 2026-09-15
ci: Separate publishing of docs and pypi package

joelostblom:
To make it easier to re-run the docs workflow if something goes awry (e.g. an expired token) or if we want to publish some minor tweaks to the docs independent of a new package release.


.github/workflows/release-draft.yml
```diff
@@ -244,6 +245,6 @@ jobs:
             gh release create "$TAG" \
               --draft \
               --verify-tag \
-              --title "$TAG" \
+              --title "Version $VERSION" \
```
joelostblom, line 248: Consistent with how we have been naming past releases

.github/workflows/code-checks.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     - name: "Set up Python"
       uses: actions/setup-python@v7
       with:
-        python-version: "3.12"
+        python-version: "3.14"
```
joelostblom, line 17: Use latest stable python in all workflows

alec-bike:
@joelostblom should we also bump the python venv version in CONTRIBUTING.md to `3.14`? Right now it's still pegged at `3.12` for local development.

PR #4145, opened 2026-09-16
chore: Add class typing to generated mixins

### LINKED ISSUE #4144: Fix `type: ignore` in generated mixins
Opening this issue to look into fixing type issues that have been suppressed by `type: ignore`.

Addresses this comment by @joelostblom in https://github.com/vega/altair/pull/4114#pullrequestreview-5178353788:

> I think this looks good in general, although one caution I have is that if we broaden ignores too freely we might end up missing something that we would like the type checkers to fix. I don't immediately see any case of that here, but I did notice one place where I wonder if there is a bug in the code that should not have been ignored previously (see inline comment).

alec-bike:
Generated mixins require multiple `type: ignore` annotations, as several classes are untyped. 

This PR adds typing to the following:
- method mixins: `MarkMethodMixin` and `ConfigMethodMixin`
- channel mixins: `ValueChannelMixin`, `DatumChannelMixin` and `_EncodingMixin`

Closes #4144.

To test:

```sh
uv run task generate-schema-wrapper
```

Note: this work was done with help from Copilot.


alec-bike:
Adding typing to the channels mixins (`ValueChannelMixin` and `DatumChannelMixin`) alters the altair API (`doc/user_guide/api.rst`). `_EncodingMixin` is excluded as it's already excluded.

Fix: add additional exclusion to `generate_api_docs.py` for all `ChannelMixin`.

alec-bike:
@mattijn, @joelostblom, any feedback or concerns about this approach? 

It allows for the removal of all `type: ignore` from `schema/mixins.py` and a good number from `schema/channels.py`. It does need to explicitly exclude all `ChannelMixin` classes from the public API, but the existing `TopLevelMixin` is unaffected.

mattijn:
From my side no concerns, I checked the code diff, and the changes seems reasonable by just reading them

PR #4148, opened 2026-09-17
fix: Replace deprecated jsonschema RefResolver with referencing

### LINKED ISSUE #4140: Fix deprecation warnings in pytest and jsonschema
With the latest dependabot and type-checker version updates, several deprecation warnings have surfaced. Originally these were detailed in the following places:
- https://github.com/vega/altair/pull/4114#issuecomment-5642188798
- https://github.com/vega/altair/issues/4134#issuecomment-5640729505. 

The deprecation warnings are reproduced below.

### `parametrize-iterators` deprecation in `pytest`

```sh
> uv run task generate-schema-wrapper
Success: no issues found in 21 source files
...
.venv/lib/python3.12/site-packages/_pytest/python.py:124
  /Users/alec/Developer/altair/.venv/lib/python3.12/site-packages/_pytest/python.py:124: 
    PytestRemovedIn10Warning: Passing a non-Collection iterable to parametrize is deprecated.
  `Test: tests/expr/test_expr.py::test_expr_methods, argvalues type: generator
  Please convert to a list or tuple.`
  See https://docs.pytest.org/en/stable/deprecations.html#parametrize-iterators
    metafunc.parametrize(*marker.args, **marker.kwargs, _param_mark=marker)
...
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

Also occurs for:
- `tests/expr/test_expr.py::test_expr_consts`
- `tests/expr/test_expr.py::test_expr_consts_immutable`

### `RefResolver` deprecation in `jsonschema`

Raised by both `ty` and `pyright`:

```sh
❯ uv run task type-check
...
warning[[deprecated](https://ty.dev/rules#deprecated)]: 
   The class `RefResolver` is deprecated
   --> tools/schemapi/schemapi.py:194:24
    |
194 |             jsonschema.RefResolver.from_schema(rootschema)
    |                        ^^^^^^^^^^^ jsonschema.RefResolver is deprecated as of
    v4.18.0, in favor of the https://github.com/python-jsonschema/referencing library, 
    which provides more compliant referencing behavior as well as more flexible APIs
    for customization. A future release will remove RefResolver. Please file a feature
    request (on referencing) if you are missing an API for the kind of customization 
    you need.
```

Also occurs on line 572.

00200200:
### 1. **PR Description**

Fixes #4140.

Pytest 10 warns if you pass a generator to `@pytest.mark.parametrize`. Those expr test argvalues are now tuples.

`jsonschema.RefResolver` is deprecated as of 4.18.0. Hiding the old fallback from type checkers is not enough, so this PR finishes the move to `referencing`:

- Raise the minimum `jsonschema` version from `>=3.0` to `>=4.18.0` (the first release with the registry API). The minimum CI job now pins `jsonschema==4.18.0`.
- Always resolve `$ref` through `referencing.Registry`. The jsonschema 3.x / RefResolver path is gone.
- Keep pytest `filterwarnings` so a RefResolver deprecation would fail the suite if it came back.

This is a dependency floor bump. Environments still on jsonschema 3.x will need to upgrade.

### 2. **Tests & Docs**

`uv run pytest tests/expr/test_expr.py tests/utils/test_schemapi.py --numprocesses=0`

Also ran the same tests against `jsonschema==4.18.0`.

### 3. **Commit Message**

`fix: Require jsonschema>=4.18 and always resolve refs with referencing`


alec-bike:
Thanks @00200200, the addition of `tuple()` fixes the pytest warning.

However, I'm not sure that just hiding the `RefResolver` warning is the best approach. From the deprecation message:

> jsonschema.RefResolver is deprecated as of v4.18.0, in favor of the [referencing](https://github.com/python-jsonschema/referencing) library, which provides more compliant referencing behavior as well as more flexible APIs for customization. **A future release will remove RefResolver.**

Do you know what is required to upgrade to the `referencing` library?

00200200:
Yeah — the helper is just a temporary compromise so we can keep the jsonschema==3.0 CI job. Moving fully to `referencing` would mean bumping the min jsonschema version and rewriting the RefResolver / from_schema / in_scope bits in schemapi.py, which is more than this PR is trying to do.

00200200:
Went ahead and did the referencing switch you asked about — schemapi is on referencing now, jsonschema min is 4.18, and I dropped the 3.0 job because it was blocking.

mattijn:
I'm fine with bumping jsonschema to version 4.18 from 3.0.0👍

pyproject.toml
alec-bike: You can remove `deprecated = "ignore"` as this only applied to `RefResolver`.

pyproject.toml
```diff
@@ -329,6 +331,11 @@ markers = [
     "no_xdist: Unsafe to run in parallel",
     "geospatial: Tests that require geopandas (deselect with '-m \"not geospatial\"')"
 ]
+filterwarnings = [
+    "error::pytest.PytestRemovedIn10Warning",
+    "error:jsonschema.RefResolver is deprecated:DeprecationWarning",
+    "error:The class `RefResolver` is deprecated:DeprecationWarning",
+]
```
alec-bike, line 336: I can get the first entry `PytestRemovedIn10Warning` to flag the warnings but cannot find a way to trigger the other two. Are they needed?

alec-bike (changes requested):
Looks good @00200200, just a couple of notes/questions in `pyproject.toml`.

pyproject.toml
00200200: Removed in the latest push.

pyproject.toml
00200200: Dropped the two RefResolver filters — they were only there as a tripwire, and after the migration they do not fire. Left the PytestRemovedIn10Warning entry.

alec-bike (approved):

PR #4151, opened 2026-09-22
ci: Collect altair and tools doctests in tests.yml again after #3972

glaziermag:
Since #3972 (d61b42af, 2026-03-04) the Tests workflow has run only the tests under `tests/`. That PR moved `"tests"`, `"altair"`, `"tools"` from pytest `addopts` into `testpaths`, and pytest consults `testpaths` only when no path is given on the command line (https://docs.pytest.org/en/stable/reference/reference.html#confval-testpaths, the link in #3972). tests.yml:84 still passes `tests`, so the `altair/` and `tools/` doctests that `--doctest-modules` is there for are never collected in CI.

Today's `main` run https://github.com/vega/altair/actions/runs/35669625991 (4c4493e7) reports `1894 passed, 12 skipped, 6 xfailed, 4 xpassed` on the 3.12–3.14 jobs and `1692 passed, 220 skipped, 4 xfailed` on the two 3.11 jobs. With `tests altair tools` on the same command line at the same commit, the same jobs report `1922 passed, 16 skipped, 6 xfailed, 4 xpassed` and `1720 passed, 225 skipped, 4 xfailed`, identical on two attempts: https://github.com/glaziermag/altair/actions/runs/35677652073. The 32 extra items are doctests from 12 modules (8 in `altair/datasets/_data.py`, 6 in `altair/vegalite/v6/api.py`, 5 in `altair/utils/_transformed_data.py`, the rest in `altair/utils`, `altair/expr` and `tools/`): 28 pass and 4 consist only of `+SKIP` examples. On 3.11 the 33rd item is `altair/jupyter/jupyter_chart.py`, skipped by `--doctest-ignore-import-errors` because anywidget is uninstalled there.

`altair` doctests were added to the test command on purpose in #3175 ("Fix doctest and run doctests in altair module"), after #3174: "docstring tests are only run in the `tests` folder. We should probably change the pytest command to include the altair package". #3972 was about letting a single-test invocation stop pulling in everything, and did not mention CI.

This PR names the three paths on the pytest command line in tests.yml, as `addopts` used to. Dropping `tests` from the command so that `testpaths` applies would also work; I kept the explicit form because the fork run above verifies this exact command.

Generated with [Claude Code](https://claude.com/claude-code): the counts above were read from the job logs, and the fork run is the verification.


mattijn:
Good research! Well spotted @glaziermag

